PR #21, opened 2023-01-26
Fix/dependabot alerts

vscaiceanu-1a:


kpanot:
This is already part of #22

PR #32, opened 2023-01-27
feat: add cascading to the repository

pginoux-1A:
Migration from the Azure dev Ops cascading task.
API calls replaced using octokit js


.github/workflows/main.yml
```diff
@@ -9,6 +9,7 @@ on:
     branches:
       - main
       - 'release/*'
+  workflow_dispatch: #Necessary to allow API based trigger of the workflow
```
kpanot, line 12: Global question regarding the mechanism of this plugin and the fact that it's trigger this workflow manually:
Would it not be safer to actually create the pull request (with automerge and a dedicated label) for each cascading?
Actually the `github.context` object prorived by github is different according to the trigger of the pipeline. When we multiply the list of different triggers we risk to have issue because missing information in the context.

packages/@o3r/dev-tools/src/utils/cascading/cascading.ts
```diff
@@ -0,0 +1,286 @@
...
+ */
+export class Cascading {
+  /** Name of the repository */
+  public repositoryName: string;
+
+  /** Repository owner login */
+  public ownerName: string;
+
+  constructor(public githubClient: Octokit, public githubContext: Context.Context, public options: CascadingOptions) {
+    // this.options.logger.verbose(`Cascading plugin options: ${JSON.stringify(options)}`);
```
kpanot, line 58: oops

packages/@o3r/dev-tools/src/utils/cascading/helpers.ts
```diff
@@ -0,0 +1,81 @@
+import { Logger } from 'winston';
```
kpanot, line 1: could be `import type`

packages/@o3r/dev-tools/src/utils/index.ts
```diff
@@ -5,3 +5,4 @@ export * from './new-version';
 export * from './check-peer-deps';
 export * from './package-manager';
 export * from './is-package-installed';
+export * from './cascading';
```
kpanot, line 8: please provide file name: `from './cascading/index';`

.github/workflows/main.yml
```diff
@@ -70,3 +71,19 @@ jobs:
...
+  cascading:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Setup
+        uses: ./.github/actions/setup
+        env:
+          AZ_AR_PASSWORD: ${{ secrets.AZ_AR_PASSWORD }}
```
kpanot, line 83: I don't think it's needed as you do not install the `jdk`

.github/actions/cascading/cascading.mjs
```diff
@@ -0,0 +1,80 @@
+import * as github from '@actions/github';
+import * as core from '@actions/core';
+import * as winston from 'winston';
```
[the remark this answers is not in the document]
pginoux-1A: Indeed, took the commit from the internal repo forgot about this

.github/actions/cascading/cascading.mjs
```diff
@@ -0,0 +1,80 @@
...
+import {inspect, promisify} from 'util';
+import {exec} from 'child_process';
+
+const promisifiedExec = promisify(exec);
+
+const run = async () => {
+  try {
+    // Processing options from inputs
+    const noFf = core.getInput('noFf');
+    const cascadingPattern = core.getInput('cascadingPattern');
```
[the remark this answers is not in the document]
pginoux-1A: Currently passed as a default option from the action.yml, isn't it enough ?
image

.github/workflows/main.yml
pginoux-1A: Automerge but still required approvals ?

.github/workflows/main.yml
```diff
@@ -70,3 +71,19 @@ jobs:
...
+      - name: Cascading
+        uses: ./.github/actions/cascading
+        with:
+          defaultBranch: main
+          logLevel: debug
```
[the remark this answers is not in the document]
pginoux-1A: I can switch, but what added value do you see ? we are talking about adding an env:logLevel: debug instead of the with right ?

.github/actions/cascading/cascading.mjs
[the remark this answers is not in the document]
kpanot: nop, I am talking about `getInput('cascadingPattern', { required: true })`

.github/workflows/main.yml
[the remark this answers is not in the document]
kpanot: about: `logLevel: ${{ ACTIONS_RUNNER_DEBUG && 'debug' | 'info' }}`

.github/workflows/main.yml
```diff
@@ -70,3 +71,19 @@ jobs:
     with:
       version: ${{ needs.version.outputs.nextVersionTag }}
       prerelease: ${{ github.event_name == 'pull_request' }}
+  cascading:
```
matthieu-crouzet, line 74: shouldn't we wait for[build, test, lint] to be done before cascading

.github/workflows/main.yml
kpanot: It can be bypass actually but would it be an issue to require approvals?

.github/actions/cascading/cascading.mjs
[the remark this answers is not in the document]
pginoux-1A: Since we provide a default value, won't it always be defined anyway ? I'd consider required : true only for parameters that are mandatory and without a default value.

.github/workflows/main.yml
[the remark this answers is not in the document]
pginoux-1A: I will consider it if not given as input, defaulting on verbose : 
image

.github/workflows/main.yml
pginoux-1A: Need to test 1/ if the bypass works 2/ if the build on the destination branch is trigger

.github/workflows/main.yml
pginoux-1A: To be discussed, are we waiting for publish or not ?

.github/actions/cascading/cascading.mjs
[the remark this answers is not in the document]
kpanot: Let's keep it simple:
You expect a type `string | undefined`, you use `getInput('cascadingPattern')`.
You expect a type `string`, you use `getInput('cascadingPattern', { required: true })`.

According to `CascadingOptions`, you expect `string` then you are in the second case.
This is what ever is specified in you YAML file, especially that `inputs` can be provided via environment variables.

.github/actions/cascading/action.yml
```diff
@@ -0,0 +1,34 @@
+name: 'Cascading'
+description: 'Cascade the commit(s) after a merge on the next branches based on semver and default branch'
+inputs:
+  noFf:
```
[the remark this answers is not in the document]
pginoux-1A: Yes, it's the flag --no-ff from the git merge command

PR #33, opened 2023-01-27
feat: update nx 15.6.3 and Angular 15.1

fpaul-1A:


matthieu-crouzet (approved):

matthieu-crouzet (approved):

apps/vscode-extension/project.json
```diff
@@ -25,7 +25,8 @@
             "input": ".",
             "output": "."
           }
-        ]
+        ],
+        "updateBuildableProjectDepsInPackageJson": false
```
vscaiceanu-1a, line 29: why is this necessary?

vscaiceanu-1a (dismissed):

kpanot:
- [ ] should clean peerDeps
image

apps/vscode-extension/project.json
fpaul-1A: `@nrwl/js:tsc` executor is failing with error `Cannot find dist/package.json` if I don't disable it, I suspect the package.json are not copied to the dist yet when running this

kpanot (dismissed):

kpanot (approved):

matthieu-crouzet (approved):

PR #34, opened 2023-01-27
fix: remove unused artifact-retriever

vscaiceanu-1a:
Removing the script for downloading the `swagger-codegen-cli.jar` because it is already versioned here https://github.com/AmadeusITGroup/otter/blob/main/packages/%40ama-sdk/generator-sdk/src/generators/resources/swagger-codegen-cli.jar


package.json
```diff
@@ -20,9 +20,8 @@
     "test": "yarn nx run-many --target=test --parallel 2 --cacheDirectory=$(yarn get:current-dir)/.cache/jest",
     "test:affected": "yarn nx affected --target=test --parallel 2 --cacheDirectory=$(yarn get:current-dir)/.cache/jest",
     "test:int": "yarn nx run-many --target=test:int --parallel 2",
-    "postinstall": "husky install && yarn build:tools && yarn version-harmonize && yarn in-source-dep-check && yarn setup:generators && yarn yarn:harmonization",
+    "postinstall": "husky install && yarn build:tools && yarn version-harmonize && yarn in-source-dep-check && yarn yarn:harmonization",
     "yarn:harmonization": "node scripts/yarn-harmonize.js",
-    "setup:generators": "yarn artifact-retriever --repository-manager \"Azure Artifacts\" --organization AmadeusDigitalAirline --project DES-SDKs --feed des-sdks -u \"AmadeusDigitalAirline\" -g \"io.swagger\" -a \"swagger-codegen-cli\" -v \"2.4.0-AMADEUS\" --out ./packages/@ama-sdk/generator-sdk/src/generators/resources/swagger-codegen-cli.jar || true",
```
kpanot, line 25: Why removing it? Maybe just removing the `|| true` would do the job

vscaiceanu-1a: as discussed on Teams, we remove the `artifact-retriever` because 1) the jar is versioned in the repo and 2) we never update it

matthieu-crouzet (approved):

kpanot (approved):

PR #39, opened 2023-01-31
fix: sdk generator cache

matthieu-crouzet:


packages/@ama-sdk/generator-sdk/src/generators/core/templates/swagger-codegen-typescript/src/main/resources/typescriptFetch/api/api.mustache
```diff
@@ -1,7 +1,7 @@
 {{#noUnusedImport}}{{#imports}}import { {{import}} } from '../../models/base/{{#kebabCase}}{{import}}{{/kebabCase}}/index';
 import { revive{{import}} } from '../../models/base/{{#kebabCase}}{{import}}{{/kebabCase}}/{{#kebabCase}}{{import}}{{/kebabCase}}.reviver';
 {{/imports}}
-import { Api, ApiClient, ApiTypes, computePiiParameterTokens, RequestBody, RequestMetadata } from '@ama-sdk/core';
+import { Api, ApiClient, ApiTypes, computePiiParameterTokens, RequestBody, RequestMetadata, utils } from '@ama-sdk/core';
```
vscaiceanu-1a, line 4: I don't think `utils` is used.

matthieu-crouzet: Yes it is :)
we just don't see it in the file but the type utils.DateTime could be generated

cpaulve-1A (approved):

fpaul-1A (approved):

cpaulve-1A (approved):

PR #46, opened 2023-02-01
fix: wrong package dependencies

cpaulve-1A:


packages/@o3r/localization/package.json
```diff
@@ -38,8 +40,6 @@
     }
   },
   "dependencies": {
-    "@o3r/dev-tools": "workspace:^",
-    "@o3r/schematics": "workspace:^",
```
matthieu-crouzet, line 42: sure it's not needed to run ng add ?
@mrednic-1A could you clarify please ?

mrednic-1A: it is needed for ng add indeed.
But we adopted a strategy to be able to keep the `@o3r/schematics` as peer dep.
Actually we run the `ng add` and if it's failing when trying to use `o3r/schematics` we display a custom message to the user saying that he needs to install schematics first. 

There is a second usecase when the package has `o3r/core` as peer dependency. Then we will display in the custom message that the user needs to install `o3r/core` directly because this one will install `o3r/schematics` because it has it as a direct dependency. 

Here is the example of the ng add for localization which uses that technique.
packages/@o3r/localization/schematics/ng-add/index.ts

apps/chrome-devtools/package.json
```diff
@@ -86,6 +86,7 @@
     "@o3r/core": "workspace:^",
     "@o3r/localization": "workspace:^",
     "@o3r/rules-engine": "workspace:^",
+    "@o3r/schematics": "workspace:^",
```
kpanot, line 89: why is it required?

packages/@o3r/extractors/package.json
```diff
@@ -65,6 +64,7 @@
     "rxjs": "^7.4.0",
     "ts-jest": "^28.0.5",
     "type-fest": "^2.13.1",
+    "typedoc": "~0.23.5",
```
kpanot, line 67: This is incorrect, it is actually a real `depedency`.
It is used with `import type` in the code but it is part of exposed types (via the `public_api.ts` file) so the user will need the dependency to resolve the typings, it is not used only for build.
(It would be fine only if the types would not have been exposed, such as schematics usage)

packages/@o3r/storybook/package.json
```diff
@@ -52,7 +50,9 @@
     "@o3r/components": "workspace:^",
     "@o3r/configuration": "workspace:^",
     "@o3r/core": "workspace:^",
+    "@o3r/dev-tools": "workspace:^",
     "@o3r/localization": "workspace:^",
+    "@o3r/schematics": "workspace:^",
```
kpanot, line 55: missing devDep

kpanot (changes requested):

Issue #35, opened 2023-01-27
Change publish job to rely on build job's artifacts


kpanot:
**Context**:
Currently the Workflow job `publish-packages/publish` is (re)building the packages before publishing the artifacts.
This is duplicating the build process through the main workflow.

**Request**:
As a PR creator, I would like the `publish` step to use the artifact generated in the `build` step of the main workflow.
When the publish step is triggered on the creation of a release, the build step should still be run.

kpanot:
We discovered that using the artifact mechanism from Github may not be the solution as it is taking 8 minutes (4 + 4) to upload and download the artifact (which is 4 time longer that the whole pipeline without it).
Should we continue to investigate on it?

jbourgeois-1A:
> (which is 4 time longer that the whole pipeline without it)
This run: https://github.com/AmadeusITGroup/otter/actions/runs/4082635162 took 29 minutes.
So let's rather use an "average" run time as comparison instead of the best case scenario.

Besides, the current implementation uploads more than 4000 files individually.
Maybe we should try to archive them first before blaming build artifacts.
Or see how much time we'd really "loose" if we executed the build, tests and publish in the same job to benefit from the actual cache without having to use network resources to share it.

Regardless whether build artifacts are sufficient or not, we've already discussed that Git Cache seems not appropriate for that usage.
As can be seen in the current workflows runs, build times are made highly unpredictable because of caches hit or miss.
Since caches are not overridden at the end of a workflow run, in an ideal world it would never be replaced and become more and more irrelevant with time the more commits are added to the PR/branch.
On top of that, reaching the 10GB limit will clear some caches and allow to magically refresh them.

This behavior is erratic and I would strongly advise against relying on it, **especially for release branches**.

PR #47, opened 2023-02-01
feat: #35: rely on build artifact for publishing pr

matthieu-crouzet:


.github/workflows/main.yml
```diff
@@ -36,6 +36,13 @@ jobs:
           AZ_AR_PASSWORD: ${{ secrets.AZ_AR_PASSWORD }}
       - name: Build
         run: yarn build
+      - name: Store dist folders
+        uses: actions/upload-artifact@v3
+        if: github.event_name == 'pull_request'
+        with:
+          path: |
```
kpanot, line 43: missing `apps` folder

.github/workflows/publish.yml
```diff
@@ -60,7 +60,14 @@ jobs:
         run: yarn build:swagger-gen
         env:
           AZ_AR_PASSWORD: ${{ secrets.AZ_AR_PASSWORD }}
+      - name: Download dist folders
+        uses: actions/download-artifact@v3
+        if: github.event_name == 'pull_request'
```
kpanot, line 65: Actually the check is not correct, it should be according to the trigger.
Maybe something like `github.event.release != null` or equivalent to see if you are triggered by `release` event or not

matthieu-crouzet: if the event_name is pull_request then it couldn't be a release event 
else event_name would have been release no ?

.github/workflows/main.yml
```diff
@@ -36,6 +36,15 @@ jobs:
           AZ_AR_PASSWORD: ${{ secrets.AZ_AR_PASSWORD }}
       - name: Build
         run: yarn build
+      - name: Store dist folders
+        uses: actions/upload-artifact@v3
+        if: github.event.release == null
```
kpanot, line 41: not needed, there is no `release` trigger ;)

kpanot:
Duplicate of #81

PR #54, opened 2023-02-02
feat: pass version to install to ng add

mrednic-1A:


packages/@o3r/schematics/src/utility/package-version.ts
```diff
@@ -0,0 +1,12 @@
+import * as fs from 'node:fs';
+import { PackageJson } from 'type-fest';
```
kpanot, line 2: `import type`

kpanot (dismissed):

packages/@o3r/schematics/src/utility/package-version.ts
mrednic-1A: indeed :)

vscaiceanu-1a (dismissed):

kpanot (approved):

matthieu-crouzet (dismissed):

kpanot (approved):

matthieu-crouzet (approved):

PR #19, opened 2023-01-26
feat: ng add o3r core

mrednic-1A:

PR #57, opened 2023-02-02
feat: add ng add for ama-sdk-core

matthieu-crouzet:


mrednic-1A (dismissed):

vscaiceanu-1a (approved):

packages/@ama-sdk/core/schematics/ng-add/index.ts
```diff
@@ -7,5 +9,27 @@ import type { Rule } from '@angular-devkit/schematics';
  */
 export function ngAdd(): Rule {
   /* ng add rules */
-  return () => {};
+  return (tree: Tree) => {
```
mrednic-1A, line 12: Here it would be good to use the technique of dynamic import to be able to display a custom message if the '@o3r/schematics' is not yet installed.
here is an example of how we did for [eslint-config-otter package](https://github.com/AmadeusITGroup/otter/pull/19/files#diff-5a4d821d2e217ee51ec518db673d86d72831b1c743116f5f7b00f10bac755fff)

packages/@ama-sdk/core/package.json
```diff
@@ -41,6 +41,7 @@
     "uuid": "^8.3.2"
   },
   "peerDependencies": {
+    "@o3r/schematics": "workspace:^",
```
kpanot, line 44: Should be in `dep` in this case because we don't want to require `@o3r/schematics` as dep in the SDKs

packages/@ama-sdk/generator-sdk/src/generators/shell/templates/base/package.json
```diff
@@ -124,5 +124,5 @@
     "npm": ">=6.0.0",
     "yarn": ">=3.1.0"
   },
-  "packageManager": "yarn@3.3.1"
+  "packageManager": "yarn@3.4.1"
 }
```
kpanot, line 128: missing EOF as empty line

mrednic-1A (approved):

cpaulve-1A (approved):

PR #69, opened 2023-02-03
fix: several deprecated code

kpanot:


packages/@o3r/dev-tools/src/cli/artifact-cleaner.ts
```diff
@@ -9,8 +9,8 @@ import * as winston from 'winston';
 
 program
   .description('Clean old artifacts from artifactory repositories')
-  .option('--artifactory-url <artifactoryUrl>', 'Artifactory URL', /.*/, 'https://digitalforairlines.jfrog.io/digitalforairlines/')
-  .option('-d, --duration-kept <durationKept>', 'All artifacts which have not been downloaded and are older than this value(ms) will be deleted. Default to 10080000ms (7 days)', /.*/, '604800000')
+  .option('--artifactory-url <artifactoryUrl>', 'Artifactory URL', 'https://digitalforairlines.jfrog.io/digitalforairlines/')
```
cpaulve-1A, line 12: does it make sense to keep this default value?

kpanot: I did not think much than removing deprecated function call.
If so maybe a clean up of several CLIs options can be done in separated task

guilhermevrs:
Still no tests - I am not surprised @kpanot

kpanot:
> Still no tests - I am not surprised @kpanot 

This is because this PR is not breaking test that there is no test in the diff ;)

matthieu-crouzet (approved):

mrednic-1A (approved):

PR #75, opened 2023-02-07
fix: small fixes ng add for env files

mrednic-1A:


packages/@o3r/core/schematics/rule-factories/differential-loading/index.ts
```diff
@@ -60,7 +60,7 @@ export function updateDifferentialLoading(options?: { projectName: string | null
       }
 
     } else {
-      context.logger.warn('The tsconfig.prod.json file has not been created');
+      context.logger.warn('Diferential loading setup skipped. The tsconfig.prod.json file has not been created');
```
jbourgeois-1A, line 63: is differential loading still relevant now that we don't support IE11 anymore?

packages/@o3r/core/schematics/rule-factories/additional-modules/index.ts
```diff
@@ -170,8 +184,18 @@ export function updateAdditionalModules(options: { projectName: string | null },
...
+    const envProdFilePathNg14 = path.join(path.dirname(workspaceProject.architect!.build.options.main), 'environments', 'environment.prod.ts');
+
+    let envProdFilePath;
+    if (tree.exists(envProdFilePathNg14)) {
+      // we are in ng 14, environment prod file name is: environment.prod.ts
+      envProdFilePath = envProdFilePathNg14;
+    } else {
+      // we are in ng 14, environment prod file name is: environment.ts
+      envProdFilePath = path.join(path.dirname(workspaceProject.architect!.build.options.main), 'environments', 'environment.ts');
+    }
```
jbourgeois-1A, line 196:
```suggestion
    // ng15 prod environment file name
    let envProdFilePath = path.join(path.dirname(workspaceProject.architect!.build.options.main), 'environments', 'environment.ts');
    if (!tree.exists(envProdFilePath )) {
      // use the ng14 file name
      envProdFilePath = path.join(path.dirname(workspaceProject.architect!.build.options.main), 'environments', 'environment.prod.ts');
    }
```

packages/@o3r/core/schematics/rule-factories/additional-modules/index.ts
```diff
@@ -119,8 +119,17 @@ export function updateAdditionalModules(options: { projectName: string | null },
     }
 
     const workspaceProject = getProjectFromTree(tree);
-    const envDevFilePath = path.join(path.dirname(workspaceProject.architect!.build.options.main), 'environments', 'environment.development.ts');
-    if (!envDevFilePath) {
+    const envProdFilePathNg14 = path.join(path.dirname(workspaceProject.architect!.build.options.main), 'environments', 'environment.prod.ts');
```
jbourgeois-1A, line 122: same comment as below

packages/@o3r/core/schematics/rule-factories/differential-loading/index.ts
mrednic-1A: Nice catch :)
it's not relevant. I will remove it from the ng add.

packages/@o3r/core/schematics/rule-factories/additional-modules/index.ts
fpaul-1A: `environment.ts` file exists in ng14, but it contains the development config (instead of prod config in ng15)
this is why we check `environment.prod.ts` to know if we are in ng14 or ng15

jbourgeois-1A: oh OK, then we can do it the other way (first try the ng14 name).
My point was that we could simplify the code by not using a temporary variable and the else clause

mrednic-1A: I could update here to use directly the prod name from ng 14 if exists. like then we simplify removing the else clause and the temp var.

packages/@o3r/core/schematics/rule-factories/additional-modules/index.ts
mrednic-1A: updated. we'll base on env dev file name

jbourgeois-1A (approved):

kpanot (approved):

PR #80, opened 2023-02-08
fix: node resolution to support exports fields

kpanot:


kpanot:
I close this PR because the process is a bit more complex.
Today NG does not really support node16 module resolution but it would be required by the TS Server to read the `exports` field of the dependency.
Meaning that, in the current structure, each package tsconfig.json file (used only for the TS Server) should override the module resolution to node16.
As this is not mandatory in this monorepo (thanks to `paths` redirection) lets keep it as it is.

PR #81, opened 2023-02-08
feat: publish artifact to reuse the build output

fpaul-1A:


.github/workflows/main.yml
```diff
@@ -32,6 +32,13 @@ jobs:
         run: yarn build:swagger-gen
       - name: Build
         run: yarn build
+      - name: zip dist content
+        run: zip -q -r dist.zip . -i "*/dist/*" -x "*/node_modules/*" ".cache/*"
```
kpanot, line 36: Could be more specific with the regexp form `workspaces` field: `apps/*/dist`, `packages/*/*/dist` and `packages/@o3r/amaterasu/*/dist`

.github/workflows/publish.yml
```diff
@@ -51,13 +51,14 @@ jobs:
...
         uses: ./.github/actions/setup
+      - name: Download dist folders
+        uses: actions/download-artifact@v3
+        if: github.event.release == null
         with:
-          enable-build-cache: 'true'
-          install-jdk: 'true'
-      - name: Build Swagger CodeGen
-        run: yarn build:swagger-gen
-      - name: Build
```
kpanot, line 59: you still need to build in case of `release` trigger

.github/workflows/publish.yml
```diff
@@ -51,13 +51,14 @@ jobs:
...
-        run: yarn build
+          name: dist
+          path: '.'
+      - name: Unzip dist
```
kpanot, line 60: Should not be run when `release` trigger

kpanot:
I think you should move out the `release` trigger to another pipeline dedicated to it.
Both event triggers seems to diverge quite a significantly

fpaul-1A:
> I think you should move out the `release` trigger to another pipeline dedicated to it. Both event triggers seems to diverge quite a significantly

done

.github/workflows/main.yml
```diff
@@ -17,21 +17,15 @@ on:
...
-        uses: ./.github/actions/setup
+      - uses: ./.github/actions/setup
         with:
           enable-build-cache: 'true'
           install-jdk: 'true'
-      - name: Build Swagger CodeGen
-        run: yarn build:swagger-gen
-      - name: Build
-        run: yarn build
+      - run: yarn build:swagger-gen
```
kpanot, line 26: why did you remove the task names?

.github/workflows/publish.yml
```diff
@@ -49,17 +47,9 @@ jobs:
...
-          install-jdk: 'true'
-      - name: Build Swagger CodeGen
-        run: yarn build:swagger-gen
-      - name: Build
-        run: yarn build
-      - name: Set new version
-        run: yarn set:version ${{ steps.get-version.outputs.version }}
+      - uses: ./.github/actions/setup
+      - uses: ./.github/actions/download-build-output
+      - run: yarn set:version ${{ steps.get-version.outputs.version }}
```
kpanot, line 52: you can change the `is-prerelease` (in following lines) to `is-prerelease: ${{ inputs.prerelease }}`

.github/workflows/release.yml
```diff
@@ -0,0 +1,72 @@
+name: Release Packages
+
+on:
+  workflow_call:
```
kpanot, line 4: this trigger is not needed on this workflow

.github/workflows/release.yml
```diff
@@ -0,0 +1,72 @@
...
+          enable-build-cache: 'false'
+          install-jdk: 'true'
+      - run: yarn build:swagger-gen
+      - run: yarn build
+      - run: yarn set:version ${{ inputs.version }}
+      - name: Get tag name
+        id: get-npm-tag
+        uses: ./.github/actions/get-npm-tag
+        with:
+          is-prerelease: ${{ inputs.prerelease || (github.event.release && github.event.release.prerelease) }}
```
kpanot, line 54: or `is-prerelease: ${{ github.event.release && github.event.release.prerelease }}`

.github/workflows/release.yml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Publish
+        run: yarn run publish --tag=${{ steps.get-npm-tag.outputs.tag }} --always-auth=true
+        env:
+          GITHUB_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+          AZURE_AUTH_TOKEN: ${{ secrets.AZURE_AUTH_TOKEN }}
+          AZURE_USERNAME: ${{ secrets.AZURE_USERNAME }}
+          AZURE_EMAIL: ${{ secrets.AZURE_EMAIL }}
+      - name: Publish Extensions
+        if: '!inputs.prerelease'
```
kpanot, line 65: `if: github.event.release && !github.event.release.prerelease`

.github/workflows/main.yml
fpaul-1A: I removed it because I think that `Build` is not necessarily a better name than the default `Run yarn build`

jbourgeois-1A (approved):
That's a lot less than 4mn when we upload only one file instead of 4000 😄

.github/workflows/release.yml
fpaul-1A: done

.github/workflows/release.yml
fpaul-1A: done

.github/workflows/release.yml
fpaul-1A: removed

.github/workflows/publish.yml
fpaul-1A: done

kpanot (dismissed):

.github/actions/download-build-output/action.yml
```diff
@@ -0,0 +1,15 @@
+name: Download Build Output
+description: Download and restore the `dist` folders from github artifact (check ../upload-build-output)
+
+runs:
+  using: 'composite'
+  steps:
+    - uses: actions/download-artifact@v3
+      if: github.event.release == null
```
kpanot, line 8: should not be part of the action (which is not used in case of `release` trigger)

fpaul-1A: removed

jbourgeois-1A (approved):

cpaulve-1A (approved):

PR #90, opened 2023-02-10
Bugfix/extension prerelease publish

kpanot:


kpanot:
I close this PR because the pre-release feature is available for VsCode extension but does not support our version number

PR #92, opened 2023-02-10
fix: extension publishing with Node18 and catch chrome extension publish Api issue

kpanot:


.github/workflows/publish.yml
```diff
@@ -59,7 +61,7 @@ jobs:
           AZURE_EMAIL: ${{ secrets.AZURE_EMAIL }}
       - name: Publish Extensions
         if: '!inputs.prerelease'
-        run: yarn run publish:extensions
+        run: yarn run publish:extensions:affected --base=HEAD~1
```
fpaul-1A, line 64: if I understand correctly, this will run the publish only if the latest commit has changes on `vscode-extension` project
I assume the last commit will most of the times be the merge-commit of a PR
with the `fetch-depth` set to 2, I suppose you will only get the HEAD commit plus the merge-commit but not the actual commit(s) with the changes ? or am I missing something ?
I have a bad feeling about this for some reason

fpaul-1A (approved):

.github/workflows/publish.yml
kpanot: The purpose of the `fetch-depth: 2` is actually to get the 2 latest commits of the branch (which will actually be a merge commits if it is resulting of a PR merge) to allow Nx to do its diff between head the head~1.
Do you think it can be an issue here regarding the fact that they are merge commits?
(cf: https://github.com/actions/checkout#checkout-head)

kpanot: Did a fix to always publish on new minor/major but run `affected` for the updates on release branch

matthieu-crouzet (approved):

PR #107, opened 2023-02-16
fix: optional peer deps

mrednic-1A:


packages/@o3r/rules-engine/package.json
```diff
@@ -37,6 +37,7 @@
     "globby": "^11.1.0",
     "jsonpath-plus": "^7.0.0",
     "rxjs": "^7.4.0",
+    "typescript": "~4.9.4",
```
mrednic-1A, line 40: __OPEN QUESTION__ For the dependencies used inside the schematics/builders of a package, should we turn them as optional ( we can talk here about @o3r/schematics which is used only by the schematics scripts on all packages, or typescript (api) which is used in builders/schematic scripts on the majority of the packages ?

packages/@ama-sdk/core/package.json
```diff
@@ -43,11 +43,15 @@
   },
   "peerDependencies": {
     "@angular-devkit/schematics": "~15.1.3",
-    "isomorphic-fetch": "^2.2.1"
+    "isomorphic-fetch": "^2.2.1",
```
kpanot, line 46: maybe this one can be optional as well?

packages/@o3r/rules-engine/package.json
kpanot: I would say that if the dependency is required only by builder and/or schematics, this should be `Optional PeerDependency` (or `dependencies` in case of `core` package).

kpanot:
As bugfix, should this PR target release/8.0 branch instead?

matthieu-crouzet (approved):

vscaiceanu-1a (approved):

PR #120, opened 2023-02-21
feat: move actions to dedicated package, move to typescript, packagin…

pginoux-1A:
This pull request contains the cascading action as well as the move of all the other github actions to a dedicated folder. It aims to have the project white listed so all actions can be reused in other projects.
The current recommanded strategy to expose typescript actions is to use ncc to create a standalone index.js, and version it inside the repository so it can be checked out. Even if the solution is not perfect (we need to be careful in future changes to run the package command and add the changes).


kpanot:
Why targeting release/8.0?

pginoux-1A:
> Why targeting release/8.0? As feature it should target `main` branch

pginoux-1A:
> Why targeting release/8.0? As feature it should target `main` branch

In the end this is more a feature for us than for our users, since it improves our pipeline. If we consider that the fixes that will be merged on 8.0 should be cascade on main, I think we should consider merging this one on 8.0

kpanot:
> > Why targeting release/8.0? As feature it should target `main` branch
> 
> In the end this is more a feature for us than for our users, since it improves our pipeline. If we consider that the fixes that will be merged on 8.0 should be cascade on main, I think we should consider merging this one on 8.0

The purpose of this PR is to expose it outside of this repo (even if today it is only for our private repo only), it will be exposed anyway.
Moreover, let's respect our own guideline and, as a new feature, integrate it to the alpha version.
I do not really see reason to be applied as patch to a released version. Do we have an emergency on it that I don't see?

pginoux-1A:
> > > Why targeting release/8.0? As feature it should target `main` branch
> > 
> > 
> > In the end this is more a feature for us than for our users, since it improves our pipeline. If we consider that the fixes that will be merged on 8.0 should be cascade on main, I think we should consider merging this one on 8.0
> 
> The purpose of this PR is to expose it outside of this repo (even if today it is only for our private repo only), it will be exposed anyway. Moreover, let's respect our own guideline and, as a new feature, integrate it to the alpha version. I do not really see reason to be applied as patch to a released version. Do we have an emergency on it that I don't see?

I'm not sure I get your point, today the cascading is not in the pipeline at all, this PR is not only about exposing it, but as well using it in the repository itself (+ being able to target it from internal repository). It only contains enhancements for the pipeline, nothing that is part of our published packages, and today fixes merged on release/8.0 needs to be cherry picked on main as well.

tools/github-actions/get-npm-tag/src/main.ts
```diff
@@ -0,0 +1,45 @@
...
+        const tag = isLatest ? 'latest' : `v${major(version)}-updated`;
+        core.info(`Version "${version}" is a released version, tag "${tag}" will be used`);
+        core.setOutput('tag', tag);
+      }
+      return;
+    }
+    core.info(`Tag "${preRelease[0]}" will be used for version "${version}"`);
+    core.setOutput('tag', preRelease[0]);
+  } catch (error) {
+    if (error instanceof Error) {core.setFailed(error.message);}
```
kpanot, line 41: would be cleaner with string handling:
```typescript
const errorMessage = err instanceof Error ? err.message : typeof err === 'string' ? err : 'Caught an error during NPM tag caculation';
core.setFailed(errorMessage);
```

tools/github-actions/new-version/src/new-version.ts
```diff
@@ -1,17 +1,18 @@
...
-import {exec} from 'child_process';
-import {promisify} from 'util';
+import {exec} from 'node:child_process';
+import {promisify} from 'node:util';
 import * as semver from 'semver';
 import {formatGitTagsOutput} from './helpers';
 
 const promisifiedExec = promisify(exec);
 
+export type BaseLogger = { debug: (log: string) => void; info: (log: string) => void; warning: (log: string) => void; error: (log: string) => void };
```
kpanot, line 8: missing comment

tools/github-actions/audit/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "version": "0.0.0-placeholder",
+  "private": true,
+  "description": "Yarn audit GitHub action provided by Otter",
+  "main": "tmp/main.js",
```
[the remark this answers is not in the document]
pginoux-1A: It's created by the build, and used as an input for ncc to generate the index.js

tools/github-actions/audit/src/main.ts
```diff
@@ -0,0 +1,107 @@
+import * as core from '@actions/core';
+import {getExecOutput} from '@actions/exec';
+import * as os from 'node:os';
+import {GitHubAdvisoryId, NPMAuditReportV1, Severity} from 'audit-types';
+import Audit = NPMAuditReportV1.Audit;
```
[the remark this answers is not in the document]
pginoux-1A: it's declared as namespace : 
image

tools/github-actions/get-npm-tag/src/main.ts
pginoux-1A: Agreed, I will update other main as well

kpanot:
You should add these new packages in the list in `.github\ISSUE_TEMPLATE\1-O3R-ISSUE-FORM.yaml`

tools/github-actions/upload-build-output/readme.md
```diff
@@ -0,0 +1,7 @@
+# Upload build output action
+
+## Overview
+Upload a GitHub artifact containing all `dist` outputs using upload-artifact GitHub action
```
jbourgeois-1A, line 4: could specify it's a single zip file that contains all the dist folders inside the current working dir

tools/github-actions/audit/readme.md
```diff
@@ -0,0 +1,23 @@
+# Yarn Audit action
+## Overview
+This GitHub action runs the yarn audit with the given command parameters, and generates a markdown report out of the json result.
```
jbourgeois-1A, line 3: it runs `yarn npm audit`

we can specify that the action fails if it finds vulnerabilities of at least the specified severity.

tools/github-actions/cascading/readme.md
```diff
@@ -0,0 +1,29 @@
+# Cascading action
+## Overview
+This plugin is meant to replace the Bitbucket cascading mechanism on Azure Dev Ops.
```
jbourgeois-1A, line 3: "is meant to implement the Bitbucket cascading mechanism on Github"

tools/github-actions/cascading/readme.md
```diff
@@ -0,0 +1,29 @@
...
+For the moment, the branching model supported is the following one : release/Maj.min[.0-alpha|beta|next|rc] and a default branch from your choice.
+It has been restrained as much as possible for the first version, to avoid any mistakes.
```
jbourgeois-1A, line 5: It's not very clear what this means

tools/github-actions/cascading/readme.md
```diff
@@ -0,0 +1,29 @@
...
+## Task options
+See [Action specifications](action.yml) directly for more information about the supported parameters.
+
+
+## What it does
+The plugin will look for branches matching release/* in your repository.
+Then the plugin will sort the branches, and based on the current branch you gave as input will figure out if the cascading is needed and on which branch.
+If the current branch is the default branch given as input OR if the current branch is the last one matching the pattern the plugin will do nothing.
+
+If a branch candidate for cascading is found, a git merge attempt will be performed :
```
jbourgeois-1A, line 16: If the plugin finds a branch candidate for cascading, it will try a git merge and:
(always prefer active voice for documentation)

tools/github-actions/cascading/src/cascading.ts
```diff
@@ -0,0 +1,284 @@
...
+  /**
+   * Launch the cascading process
+   */
+  public async execute() {
+    this.options.logger.info('Cascading plugin execution');
+    await handlePromisifiedExecLog(promisifiedExec('git config --global user.email "auto-cascading@amadeus.com"'), this.options.logger);
+    await handlePromisifiedExecLog(promisifiedExec('git config --global user.name "Auto cascading"'), this.options.logger);
+    await handlePromisifiedExecLog(promisifiedExec('git status'), this.options.logger);
+    const branchOutput = await handlePromisifiedExecLog(promisifiedExec(`git branch -al */${this.options.cascadingPattern}`), this.options.logger);
+    let sortedBranches = formatGitBranchOutput(branchOutput.stdout);
```
jbourgeois-1A, line 72: it's the `verifyBranches` that seem to do the sorting.
Function and variable names aren't well chosen

tools/github-actions/cascading/src/helpers.ts
```diff
@@ -0,0 +1,81 @@
+import {BaseLogger} from './cascading';
+
+/**
+ * Ensure that the sorting based on git have been done correctly, update the list if there is an issue
+ *
+ * @param branches
+ */
+export function verifyBranches(branches: string[]) {
+  return branches.sort((branchA, branchB) => {
+    const aExtract = branchA.match(/release\/([0-9]+)\.([0-9]+)/);
```
jbourgeois-1A, line 10: this implementation looks inefficient.

a sort implementation is usually O(n*log(n)) which means you will check the same branch name multiple times.
you could extract the version information used to determine the natural order in a separate variable in a `.map()`

tools/github-actions/cascading/src/helpers.ts
```diff
@@ -0,0 +1,81 @@
+import {BaseLogger} from './cascading';
+
+/**
+ * Ensure that the sorting based on git have been done correctly, update the list if there is an issue
```
jbourgeois-1A, line 4: it's a very convoluted way to simply say "Sort the branches according to semantic versioning"

tools/github-actions/cascading/src/helpers.ts
```diff
@@ -0,0 +1,81 @@
...
+    const aMin = aExtract[2];
+    const aMaj = aExtract[1];
+    const bMin = bExtract[2];
+    const bMaj = bExtract[1];
+    if (parseInt(aMaj, 10) > parseInt(bMaj, 10)) {
+      return 1;
+    } else if (parseInt(aMaj, 10) < parseInt(bMaj, 10)) {
+      return -1;
+    }
+    return parseInt(aMin, 10) - parseInt(bMin, 10);
```
jbourgeois-1A, line 24:
```suggestion
    const aMaj = parseInt(aExtract[1], 10);
    const bMaj = parseInt(bExtract[1], 10);
    if (aMaj  > bMaj) {
      return 1;
    } else if (aMaj  < bMaj) {
      return -1;
    }
    return parseInt(aExtract[2], 10) - parseInt(bExtract[2], 10);
```

tools/github-actions/cascading/src/cascading.ts
```diff
@@ -0,0 +1,284 @@
...
+            nonIgnoringChanges.forEach((fileChange) => {
+              this.options.logger.info(JSON.stringify(fileChange, null, 2));
+            });
+            await this.createFallbackPullRequest(branchToCascade);
+            return;
+          } else {
+            this.options.logger.info('Conflicts can be ignored, discarding those changes');
+            const listOfDiscardingFiles = [...changes.packageChanges.map((file) => file.file), changes.yarnLockFile];
+            for (const file of listOfDiscardingFiles) {
+              if (file) {
```
jbourgeois-1A, line 176: this check is only necessary because of line 174, it could be done so we only check the presence of `changes.yarnLockFile`

tools/github-actions/cascading/src/cascading.ts
```diff
@@ -0,0 +1,284 @@
...
+                this.options.logger.info(`git checkout --ours ${file}`);
+                await handlePromisifiedExecLog(promisifiedExec(`git checkout --ours ${file}`), this.options.logger);
```
jbourgeois-1A, line 178: we could `join` the file names and do only one exec, might be more efficient

tools/github-actions/cascading/src/helpers.ts
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Extract just the branches matching the release/min.maj[.0-alpha|beta|next] format,
+ *
+ * @param gitOutput
+ */
+export function formatGitBranchOutput(gitOutput: string) {
+  return gitOutput.split(/\r?\n\s*/)
+    .map((val) => val.replace(/ /g, ''))
+    .map((val) => val.replace('remotes/origin/', ''))
```
jbourgeois-1A, line 36: could be a single `.map()`

tools/github-actions/cascading/readme.md
```diff
@@ -0,0 +1,29 @@
...
+  * if conflictsIgnoredPackages input is not empty, there are only conflicts in package.json and yarn.lock files and the conflict is only related to those packages, a pull request will be automatically created resolving the conflict by discarding those changes
+  * else a pull request will be automatically created (409 error will be thrown if the pull request already exists). If assignCommitter is true, the first person with a "Merge" commit in the list will be assigned as required approval
+
+The cascading task is logging a lot of information, don't hesitate to have a look to have more precisions about what exactly is executed.
+
+## Usage example
+```yaml
+- name: Cascading 
+  id: cascading
+  uses: ./tools/github-actions/cascading
```
jbourgeois-1A, line 28: this is an example from our repo right? what would it look like from another repo?

tools/github-actions/new-version/src/main.ts
```diff
@@ -0,0 +1,74 @@
...
+    core.debug(JSON.stringify(process.env, null, 2));
+
+    // Storing ENV variables
+    const buildSourceBranch = github.context.ref;
+    const buildReason = github.context.eventName;
+    const authenticatedGitUrl = github.context.payload.repository!.html_url!;
+    const buildId = `${github.context.runId}`;
+
+    // Compute organisationName
+    const organisationName = github.context.actor;
```
jbourgeois-1A, line 24: This was needed in Azure but doesn't seem required here

tools/github-actions/new-version/readme.md
```diff
@@ -0,0 +1,19 @@
+# New version action
```
jbourgeois-1A, line 1: could reuse the doc of the Azure extension: https://dev.azure.com/AmadeusDigitalAirline/AzureTools/_git/new-version-plugin?path=/README.md

tools/github-actions/cascading/src/helpers.ts
pginoux-1A: I will just fix it with the other one : /**
 * For the branches according to semantic versioning
 *
 * @param branches
 */
export function verifyBranches(branches: string[]) {
  return branches.map(branch => {
    const extract = branch.match(/release\/([0-9]+)\.([0-9]+)/);
    if (!extract) {
      throw new Error(`Format of branch doesnt match the release pattern ${branches.join(',')}`);
    }
    return {
      maj: parseInt(extract[1], 10),
      min: parseInt(extract[2], 10),
      branch
    };
  }).sort((branchObjectA, branchObjectB) => {
    if (branchObjectA.maj > branchObjectB.maj) {
      return 1;
    } else if (branchObjectA.maj < branchObjectB.maj) {
      return -1;
    }
    return branchObjectA.min - branchObjectB.min;
  }).map(branchObject => branchObject.branch);
}

tools/github-actions/get-npm-tag/src/main.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    const releaseResponse = await getOctokit(core.getInput('token', {required: true})).rest.repos.listReleases({
+      owner,
+      repo
+    });
+    const latestVersion = releaseResponse.data
+      .filter((release) => !release.draft && !release.prerelease)
+      .map((release) => release.tag_name.replace(/^v/i, ''))
+      .sort(compare)
+      .reverse()[0];
```
jbourgeois-1A, line 24: this was copied, but I'm not very fan of doing a full iteration with a `reverse()` when we could simply access the last element with `length-1` or `.pop()`

tools/github-actions/download-build-output/readme.md
```diff
@@ -0,0 +1,7 @@
+# Download build output action
+
+## Overview
+Download and restore the `dist` folders from GitHub artifact (check ../upload-build-output)
```
jbourgeois-1A, line 4: could be a proper link to `upload-build-output`'s readme

(and upload could also link the donwload's readme)

.github/workflows/main.yml
```diff
@@ -61,3 +61,10 @@ jobs:
     with:
       version: ${{ needs.version.outputs.nextVersionTag }}
       prerelease: ${{ needs.version.outputs.isPreRelease == 'true' }}
+  cascading:
+    runs-on: ubuntu-latest
+    needs: [version, build, checks]
```
matthieu-crouzet, line 66: in fact when it's merged on the release branch should we wait that it build again to cascade the change ?
or should we try to cascade directly ?

tools/github-actions/audit/src/main.ts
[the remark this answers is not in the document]
matthieu-crouzet: why not `type Audit = NPMAuditReportV1.Audit` ?

tools/github-actions/cascading/src/cascading.ts
```diff
@@ -0,0 +1,284 @@
...
+    await handlePromisifiedExecLog(promisifiedExec('git status'), this.options.logger);
+    const branchOutput = await handlePromisifiedExecLog(promisifiedExec(`git branch -al */${this.options.cascadingPattern}`), this.options.logger);
+    let sortedBranches = formatGitBranchOutput(branchOutput.stdout);
+    this.options.logger.info(`Sorted branches computed : ${JSON.stringify(sortedBranches)}`);
+    sortedBranches = verifyBranches(sortedBranches);
+    this.options.logger.info(`Sorted branches after update : ${JSON.stringify(sortedBranches)}`);
+    const currentBranchIndex = sortedBranches.indexOf(this.options.baseBranch);
+    this.options.logger.info(`Current branch index computed : ${currentBranchIndex}`);
+    if (currentBranchIndex === -1) {
+      this.options.logger.info('Current branch doesnt match the pattern, please check inputs given to the cascading extension, exiting cascading...');
```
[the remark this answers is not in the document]
pginoux-1A: this.options.logger.info(`Current branch doesnt match cascadingPattern: ${this.options.cascadingPattern}, no cascading will be performed.`);

.github/workflows/main.yml
pginoux-1A: As discussed, in case there is a cascading already done from the previous branch, better ensure that all the checks are green before cascading again

tools/github-actions/cascading/src/cascading.ts
```diff
@@ -0,0 +1,284 @@
...
+      return;
+    }
+    let branchToCascade: string = '';
+    if (currentBranchIndex + 1 === sortedBranches.length) {
+      if (this.options.defaultBranch) {
+        this.options.logger.info('Current branch is the last branch that matches the pattern given, cascading on default branch instead...');
```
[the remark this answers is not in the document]
pginoux-1A: New option added: ignoredPattern

tools/github-actions/new-version/readme.md
```diff
@@ -1,18 +1,41 @@
...
 ## Overview
-This GitHub action retrieves the npm tag to apply for a release
+
+This task computes new version candidates based on the repository tags as well as the branch on which it is executed.
+It supports in terms of branching model:
+* Release branches named ``release/Major.Minor[.0-alpha|beta|next|rc]``
+* One default branch that can be parametrised (``master``, ``develop``, ...)
+
+The plug-in will end with an Azure status as ``SucceededWithIssues`` in case it is run on a branch that it is not compatible with.  
+Meaning it is made to not block a pipeline if ran on an unsupported branch.
```
jbourgeois-1A, line 11: this is not exactly the case anymore (we just add a warning from what I see)

jbourgeois-1A (dismissed):

tools/github-actions/cascading/src/helpers.ts
```diff
@@ -0,0 +1,143 @@
+import {BaseLogger} from './cascading';
```
kpanot, line 1: it is `import type` but not important for a tool

tools/github-actions/setup/readme.md
```diff
@@ -0,0 +1,20 @@
...
+See [Action specifications](tools/github-actions/setup/action.yml) directly for more information about the supported parameters.
+
+## Usage example
+```yaml
+- name: Setup
+  id: setup
+  uses: AmadeusITGroup/otter/tools/github-actions/setup@v8.0.15
+  with:
+    enable-build-cache: 'true'
+    install-jdk: 'true'
```
kpanot, line 19: This part has changed since the `setup-java` has been moved out

kpanot:
As we need to stock JS resulting of a build in git, we should check that they are still up-to-date with the TS source code associated.
I think you should add a check (in code-check workflow?) that rebuild the packages and compare the result, when one of the github action packages has been touched, to guaranty the consistency

pginoux-1A:
> As we need to stock JS resulting of a build in git, we should check that they are still up-to-date with the TS source code associated. I think you should add a check (in code-check workflow?) that rebuild the packages and compare the result, when one of the github action packages has been touched, to guaranty the consistency

The current packaging strategy (recommanded by github) won't ensure a 0 diff on the packaged output : 
image
image
image

We will have a diff at each new generation, so we will need to post process it to try to determine whether it comes from the packaging or not, looks a bit complex and unstable, for a low gain.

tools/github-actions/cascading/action.yml
```diff
@@ -0,0 +1,37 @@
...
+  defaultBranch:
+    description: 'The default branch if you have one (ex: master, development), if no candidate found with the given pattern this branch will be the last one where the code will be cascaded'
+    required: false
+    default: 'main'
+  assignCommitter:
+    description: 'Assigns the first committer to the cascading failure PR'
+    required: false
+    default: false
+  token:
+    description: 'Github token to use to access to the APIs'
```
kpanot, line 24: I think you need to specify the the permission(s) mandatory to this token (here or in the readme)

tools/github-actions/cascading/src/helpers.ts
```diff
@@ -0,0 +1,143 @@
+import type {BaseLogger} from './cascading';
+
+/**
+ * Sort branches according to semantic versioning
+ *
+ * @param branches
+ */
+export function sortBranches(branches: string[]) {
+  return branches.map(branch => {
+    const extract = branch.match(/release\/([0-9]+)\.([0-9]+)/);
```
kpanot, line 10: This should not be based on `cascadingPattern` input?
If my config is `cascadingPattern: publish/*` this function will not work very well (same as the one below) :S

tools/github-actions/cascading/tsconfig.build.json
```diff
@@ -0,0 +1,14 @@
+{
```
kpanot, line 1: should extend the root `tsconfig.build.json` as the others

tools/github-actions/new-version/readme.md
```diff
@@ -0,0 +1,42 @@
+# New version action
+
+## Overview
+
+This task computes new version candidates based on the repository tags as well as the branch on which it is executed.
+It supports in terms of branching model:
+* Release branches named ``release/Major.Minor[.0-alpha|beta|next|rc]``
```
kpanot, line 7: This is actually based on the `releaseBranchRegExp` input and support more than this pattern (but this is the default pattern)

tools/github-actions/new-version/src/new-version.spec.ts
```diff
@@ -1,16 +1,23 @@
-import {NewVersion, NewVersionOptions} from './new-version';
-import * as winston from 'winston';
+import {BaseLogger, NewVersion, NewVersionOptions} from './new-version';
+
+
+const mockLogger: BaseLogger = {
```
kpanot, line 4: Why not defaulting on `console`?

tools/github-actions/setup/readme.md
```diff
@@ -0,0 +1,19 @@
+# Build setup action
+
+## Overview
+Action to install needed dependencies and handles caching on demand for Otter project.
+Including :
+* nodejs 16
+* JDK 17
```
kpanot, line 7: Not anymore with the split of the setup actions

tools/github-actions/cascading/src/helpers.ts
pginoux-1A: We had a discussion on this and we chose to have  a very strict pattern to avoid any issue (unwanted cascading on other repository branches) as mentionned here : 
`This plugin is meant to implement the Bitbucket cascading mechanism on GitHub.
For the moment, the branching model supported is the following one : release/Maj.min[.0-alpha|beta|next|rc] and a default branch from your choice.
This pattern is restrictive on purpose, to prevent any issue where some non-release branches could match as well`
And the cascading Pattern option applies a pre filter only:
`The cascading pattern used as a pre-filter on repositiory branches considered for cascading`.

The idea was that we would create a new feature in case projects want an enhancement on this.

kpanot: As discuss, a compromise would be to:
- Remove the input to not have inconsistency with this code
- Create a task to enhance the plugin with the flexibility bring by the input

kpanot (approved):

tools/github-actions/cascading/action.yml
```diff
@@ -0,0 +1,33 @@
...
+  defaultBranch:
+    description: 'The default branch if you have one (ex: master, development), if no candidate found with the given pattern this branch will be the last one where the code will be cascaded'
+    required: false
+    default: 'main'
+  assignCommitter:
+    description: 'Assigns the first committer to the cascading failure PR'
+    required: false
+    default: false
+  token:
+    description: 'Github token to use to access to the APIs (need rights to create a pull request and write/trigger_build on release branches'
```
kpanot, line 20: Missing `)`

tools/github-actions/cascading/readme.md
```diff
@@ -0,0 +1,32 @@
...
+Then it will sort the branches, and based on the current branch you gave as input will figure out if the cascading is needed and on which branch.
+If the current branch is the default branch given as input OR if the current branch is the last one matching the pattern the plugin will do nothing.
+
+If the plugin finds a branch candidate for cascading, it will try a git merge and:
+* If there is no conflict, the result will be pushed to the repository, triggering a build on the next branch than will run the cascading again
+* If there is a conflict:
+  * if conflictsIgnoredPackages input is not empty, there are only conflicts in package.json and yarn.lock files and the conflict is only related to those packages, a pull request will be automatically created resolving the conflict by discarding those changes
+  * else a pull request will be automatically created (409 error will be thrown if the pull request already exists). If assignCommitter is true, the first person with a "Merge" commit in the list will be assigned as required approval
+
+The token provided needs to have the permission to create a pull request, write and trigger builds on release branches.
```
kpanot, line 21: It is actually not only the provided token, as the default value of the token is `github.token`, this will work only if your job has the `contents: write`.
Maybe you can have something like:
```suggestion
> ***Warning**: The action requires to be run in a job with `contents: write` permission or the provided token needs to have the permission to create a pull request, write and trigger builds on release branches.
```

kpanot (approved):

PR #121, opened 2023-02-21
feat: programmatic checkbox for playwright

matthieu-crouzet:


packages/@o3r/testing/src/core/elements/programmatic-checkbox-element.ts
```diff
@@ -0,0 +1,36 @@
+import { TranspilationPurposeOnlyError } from '../../errors/index';
+
+import { O3rElement } from '../element';
+import { CheckboxElementProfile } from './checkbox-element';
+
+/**
+ * Interface to describe the Checkbox elements that are used inside a fixture.
+ * As for ComponentFixtureProfile, this abstracts the testing framework that is used by choosing the right
+ * implementation at runtime.
```
kpanot, line 9: Maybe you should specify that this fixture is provided only for the case of "complex" checkbox label

packages/@o3r/testing/src/core/playwright/elements/programmatic-checkbox-element.ts
```diff
@@ -0,0 +1,61 @@
+import {CheckboxElementProfile} from '../../elements';
+import {O3rElement, PlaywrightSourceElement} from '../element';
+
+/**
+ * Implementation dedicated to Playwright.
+ */
+export class O3rProgrammaticCheckboxElement extends O3rElement implements CheckboxElementProfile {
```
mrednic-1A, line 7: Can't you extend `O3rCheckboxElement` class and override only `check` method ?

mrednic-1A (approved):

PR #124, opened 2023-02-22
fix: checkbox playwright

matthieu-crouzet:


mrednic-1A (approved):

packages/@o3r/testing/src/core/playwright/elements/checkbox-element.ts
```diff
@@ -12,8 +12,10 @@ export class O3rCheckboxElement extends O3rElement implements CheckboxElementPro
   private async getInputElement() {
     try {
       const subInputElement = this.sourceElement.element.locator('input[type="checkbox"]');
-      await subInputElement.waitFor({state: 'attached'});
-      return subInputElement;
+      if (await subInputElement.count()) {
```
kpanot, line 15: I have a question: are we sure this change does not change the behavior a bit?
What append this the DOM is not ready when calling the `.count()` do we will retrieve a `0` or would if way for the DOM to count?
In the fist case we may have issue if the action is required to early. Would it be of the responsibility of caller to make sure the DOM is ready? Is the caller able to do it?

kpanot (approved):

cpaulve-1A (approved):

PR #126, opened 2023-02-22
feat: #1286 Improve logs in o3r

cpaulve-1A:


packages/@ama-sdk/generator-sdk/src/generators/core/templates/swagger-codegen-typescript/src/main/resources/typescriptFetch/model/reviver.mustache
```diff
@@ -14,18 +14,17 @@ import { {{.}} } from '../{{#kebabCase}}{{.}}{{/kebabCase}}';
...
-export function revive{{classname}}(data: any, dictionaries?: any): {{classname}} {{#discriminator}}{{#vendorExtensions}}{{#x-discriminator-subtypes}}| {{.}}{{/x-discriminator-subtypes}}{{/vendorExtensions}}{{/discriminator}} | undefined;
-export function revive{{classname}}<T extends {{classname}}>(data: T, dictionaries?: any): T {{#discriminator}}{{#vendorExtensions}}{{#x-discriminator-subtypes}}| {{.}}{{/x-discriminator-subtypes}}{{/vendorExtensions}}{{/discriminator}};
-export function revive{{classname}}<T extends {{classname}}>(data: any, dictionaries?: any): T {{#discriminator}}{{#vendorExtensions}}{{#x-discriminator-subtypes}}| {{.}}{{/x-discriminator-subtypes}}{{/vendorExtensions}}{{/discriminator}} | undefined;
-export function revive{{classname}}<T extends {{classname}} = {{classname}}>(data: any, dictionaries?: any): T {{#discriminator}}{{#vendorExtensions}}{{#x-discriminator-subtypes}}| {{.}}{{/x-discriminator-subtypes}}{{/vendorExtensions}}{{/discriminator}} | undefined {
+export function revive{{classname}}<T extends {{classname}} = {{classname}}>(data: undefined, dictionaries?: any, logger = console): undefined;
```
kpanot, line 17: don't you need to explicit the `logger` type to `Logger`?
Otherwise it will take the type from `console` which is more complex

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -248,9 +249,8 @@ export class RulesetExecutor {
               if (this.rulesEngine.debugMode) {
                 output.evaluation = handleRuleEvaluationDebug(rule, this.ruleset.name, output.actions, output.error, runtimeFactValues, factValues, oldFactValues);
               } else if (output.error) {
-                // eslint-disable-next-line no-console
-                console.error(output.error);
-                console.warn(`Skipping rule ${rule.id}, and the associated ruleset`);
+                (this.rulesEngine.logger || console).error(output.error);
```
kpanot, line 252: Seeing this, maybe it would make sense to default the rules engine logger to `console`. What do you think?

cpaulve-1A: I know, 
What I am doing here is not the best because the logger is a LoggerService which is not the same type as the console (so I can't default on the console). 
I can't default to a new service with the consoleLoggerService client as the service will be injected in the app module.
I could eventually change the type of the logger to reflect something similar to what I have done in the ama-sdk. 

What do you think?

packages/@o3r/rules-engine/src/engine/engine.ts
```diff
@@ -7,6 +7,7 @@ import { filterRulesetsEventStream } from './helpers/filter-ruleset-event.operat
 import { Operator, operatorList, UnaryOperator } from './operator/index';
 import { RulesetExecutor } from './ruleset-executor';
 import { ActionBlock, Ruleset } from './structure';
+import {Logger} from '@o3r/logger';
```
kpanot, line 10: Could be `import type`

packages/@o3r/rules-engine/src/engine/engine.interface.ts
```diff
@@ -1,3 +1,4 @@
+import {Logger} from '@o3r/core';
```
kpanot, line 1: Could be `import type`

kpanot (dismissed):

mrednic-1A (approved):

kpanot (approved):

PR #141, opened 2023-03-01
Cascading v8 to main

matthieu-crouzet:


mrednic-1A (dismissed):

mrednic-1A (approved):

kpanot (approved):

packages/@ama-sdk/generator-sdk/src/generators/java-client-core/templates/swagger-codegen-java-client/src/main/resources/javaClient/settings.mustache
```diff
@@ -3,17 +3,17 @@
...
+      <mirrorOf>des-sdks</mirrorOf>
     </mirror>
   </mirrors>
   <servers>
     <server>
-      <id>public-artifactory</id>
-      <username>${env.AR_USER}</username>
-      <password>${env.AR_PASSWORD}</password>
+      <id>des-sdks</id>
+      <username>AmadeusDigitalAirline</username>
```
kpanot, line 15: Why using static string instead of variable?

packages/@ama-sdk/generator-sdk/project.json
```diff
@@ -40,7 +40,9 @@
           "packages/@ama-sdk/generator-sdk/src/**/swagger-codegen-cli.jar",
           "packages/@ama-sdk/generator-sdk/src/**/config/*.json",
           "packages/@ama-sdk/generator-sdk/src/generators/**/target/*.jar",
-          "packages/@ama-sdk/generator-sdk/src/generators/*/templates/**/*",
+          "packages/@ama-sdk/generator-sdk/src/generators/shell/templates/**/*",
```
kpanot, line 43: Why this change, the `*` was not doing the job?

packages/@o3r/stylelint-plugin/src/public_api.ts
```diff
@@ -0,0 +1,3 @@
+import rulesPlugins from './index';
+// Rules plugins have to be exported as default
+export default rulesPlugins;
```
kpanot, line 3: missing additional empty line

kpanot (approved):

packages/@ama-sdk/generator-sdk/project.json
mrednic-1A: The `generators/*/templates` allowed in the packaging all the files inside `templates` folder for all the generators, not only the templates files for schematics generator. 
This was a problem for the packaging of java part of generators, because along with the `.jar` files in `target` folder we got also the `.class` files, which break the generators at runtime because it is trying to execute the them.
Ex: `generator-sdk/src/generators/java-client-core/templates/target` has matched with the `/generators/*/templates/**/*`, and that's why I specified exactly which are the `schematics` tempaltes to package.

packages/@ama-sdk/generator-sdk/src/generators/java-client-core/templates/swagger-codegen-java-client/src/main/resources/javaClient/settings.mustache
matthieu-crouzet: It's just the cascading 
maybe we need to fix it on v8 if you are not agree with what @mrednic-1A has done

packages/@o3r/stylelint-plugin/src/public_api.ts
matthieu-crouzet: in fact there is one 
else we would see the red minus sign in a round

kpanot (approved):

packages/@ama-sdk/generator-sdk/src/generators/java-client-core/templates/swagger-codegen-java-client/src/main/resources/javaClient/settings.mustache
mrednic-1A: It's coming from a fix done when passing from jfrog to azure. we can redo it  8.0 if needed

mrednic-1A (approved):

PR #144, opened 2023-03-01
fix: prepare e2e visual testing

matthieu-crouzet:


packages/@o3r/application/src/devkit/application-devkit.interface.ts
```diff
@@ -1,6 +1,8 @@
 import type { ConnectContentMessage, DevtoolsCommonOptions, MessageDataTypes, OtterMessageContent, RequestMessagesContentMessage } from '@o3r/core';
 
 export interface ApplicationDevtoolsServiceOptions extends DevtoolsCommonOptions {
+  /** Class for e2e visual testing */
```
kpanot, line 4: The comment is a bit unclear and seems to mean the opposite of what this field is for

cpaulve-1A (approved):

kpanot (approved):

PR #148, opened 2023-03-02
feat: #1291 Support localization key in placeholders

cpaulve-1A:


packages/@o3r/localization/src/tools/localization.service.ts
```diff
@@ -106,7 +106,7 @@ export class LocalizationService {
    * @param interpolateParams Object to use in translation binding
    * @returns A stream of the translated key
    */
-  private getTranslationStream(translationKey: string, interpolateParams?: object) {
```
kpanot, line 109: why do you need to expose this field instead of using `translate` ?

packages/@o3r/rules-engine/src/services/rules-engine.effect.ts
```diff
@@ -61,35 +62,52 @@ export class PlaceholderTemplateResponseEffect {
...
 
   /**
-   * Renders the html template, replacing facts and urls
+   * Renders the html template, replacing facts and urls and localizationKeys
    *
    * @param template
    * @param vars
    * @param facts
    */
-  public async renderHTML(template?: string, vars?: Record<string, PlaceholderVariable>, facts?: { name: string; factValue: any }[]) {
```
kpanot, line 74: this is actually a breaking change, maybe you can reprovide the `renderHTML` as `depdecated` with a `latestValueFrom(this.getRenderedHTML$)`

packages/@o3r/localization/src/tools/localization.service.ts
cpaulve-1A: I did not caught translate emitted on language changes, I will revert it

packages/@o3r/rules-engine/src/services/rules-engine.effect.ts
cpaulve-1A: That is what I first intended - hence the new name - but we said it was internal logic and was ok to clean up. I can leave it for safety reasons, it does not matter much anyway

packages/@o3r/rules-engine/src/services/rules-engine.effect.ts
```diff
@@ -100,9 +118,13 @@ export class PlaceholderTemplateResponseEffect {
         }
       }
     }
-    return {renderedTemplate: template, unknownTypeFound};
+    return replacements$.length > 0 && !!template ?
+      combineLatest(replacements$).pipe(
+        map((replacements: { ejsVar: RegExp; value: string }[]) =>
+          replacements.reduce((acc, replacement) => ({
+            renderedTemplate: acc.renderedTemplate.replace(replacement.ejsVar, replacement.value),
+            unknownTypeFound: acc.unknownTypeFound
```
jbourgeois-1A, line 126: this property is never touched during the iteration, the `reduce` could return the replaced string

packages/@o3r/rules-engine/src/services/rules-engine.effect.ts
jbourgeois-1A: I'm not sure it made sense to be public to be honest.
Maybe the new one should be private

kpanot (approved):

kpanot (approved):

PR #152, opened 2023-03-02
fix: add waitFor timeout not to rely on global test timeout

cpaulve-1A:


jbourgeois-1A (commented):
I think you need to modify the functions signatures in the `core` folder since it's the one used by the IDE
(you may also have to add the optional param to every implementation otherwise typescript may fail when using the fixture for UTs, to be checked)

matthieu-crouzet (dismissed):

packages/@o3r/testing/src/core/playwright/angular-materials/autocomplete-material.ts
```diff
@@ -12,12 +12,12 @@ export class MatAutocomplete extends O3rElement implements MatAutocompleteProfil
   }
 
   /** @inheritdoc */
-  public async selectByValue(value: string) {
+  public async selectByValue(value: string, timeout = 5000) {
```
kpanot, line 15: Why setting a static default value instead of relying on the default one from Playwright?
(Same question for the other impl)

cpaulve-1A: because they set it at global level pretty high for other tests (due to calls etc.). I could only find a value that impacts all the timeout on the app or all the one linked to the navigation.
In the prior version we already had this kind of timeout. It is something we lost with the migration to the Locator.

matthieu-crouzet (approved):

kpanot (approved):

PR #160, opened 2023-03-06
feat: provide rule iteration number

kpanot:
image


kpanot:
Canceled, information already provided to through the execution counter

PR #162, opened 2023-03-06
fix: multiple request param service

matthieu-crouzet:


packages/@o3r/routing/src/services/request-parameters/request-parameters.service.ts
```diff
@@ -16,6 +16,8 @@ export interface PartialRequestParametersConfig extends Partial<RequestParameter
 /**
  * Service used to store the request parameters of your requests so that subsequent calls or refresh the page will preserve
  * them.
+ *
+ * @deprecated use RequestParametersService from @o3r/dynamic-content instead, will be removed in v10
```
mrednic-1A, line 20: If there is a module file, @deprecated should be added there too.

packages/@o3r/dynamic-content/src/fixtures/jasmine/request-parameters.fixture.jasmine.ts
```diff
@@ -0,0 +1,41 @@
+import type { InterfaceOf } from '@o3r/core';
+import type { RequestParametersConfig, RequestParametersService } from '@o3r/dynamic-content';
+import { StorageStrategy } from '@o3r/dynamic-content';
```
kpanot, line 3: `import type` was not possible for it?

packages/@o3r/dynamic-content/tsconfig.fixture.jasmine.json
```diff
@@ -0,0 +1,15 @@
+{
```
kpanot, line 1: missing registration into the `tsconfig.json`'s references

packages/@o3r/dynamic-content/src/fixtures/jasmine/request-parameters.fixture.jasmine.ts
matthieu-crouzet: no because it's an enum

packages/@o3r/dynamic-content/tsconfig.fixture.jasmine.json
matthieu-crouzet: not done for all others fixture 
"Referenced project '/Users/crouzetm/git_clones/public-otter/packages/@o3r/dynamic-content/tsconfig.fixture.jasmine.json' must have setting "composite": true."
should I add composite true to all of them and reference them in their respectiver tsconfig ?

mrednic-1A (approved):

kpanot (approved):

PR #176, opened 2023-03-13
fix: #1391 Fix export issue on module with subentries

cpaulve-1A:


tsconfig.base.json
```diff
@@ -66,7 +66,6 @@
       "@o3r/stylelint-plugin": ["packages/@o3r/stylelint-plugin/src"],
       "@o3r/styling": ["packages/@o3r/styling/src/public_api"],
       "@o3r/testing": ["packages/@o3r/testing/src/public_api"],
-      "@o3r/testing/*": ["packages/@o3r/testing/src/*"],
```
kpanot, line 69: Not sure you should remove it.
Without this rules, the deep imports will be rerouted to src/ folder in this monorepo

packages/@ama-sdk/swagger-builder/package.json
```diff
@@ -10,14 +10,14 @@
     "nx": "nx",
     "ng": "yarn nx",
     "build": "yarn nx build ama-sdk-swagger-builder",
-    "postbuild": "patch-package-json-main",
+    "postbuild": "patch-package-json-main && cpy 'src/schemas/*.json' 'dist/schemas'",
```
kpanot, line 13: is `src/schemas/*.json` already included in Nx task outputs?

packages/@ama-sdk/core/package.json
```diff
@@ -2,27 +2,13 @@
...
+  "module": "dist/public_api.js",
   "esm2015": "dist/esm2015/public_api.js",
-  "esm2020": "dist/src/public_api.js",
-  "typings": "dist/src/public_api.d.ts",
+  "esm2020": "dist/public_api.js",
+  "typings": "dist/public_api.d.ts",
+  "main": "dist/public_api.js",
+  "types": "dist/public_api.d.ts",
   "sideEffects": false,
-  "exports": {
```
kpanot, line 10: Not sure you should remove these exports.
The purpose of it is to be able to use this package (and then the SDK) in a esm context on node.
Why did you remove it here? Is it causing issue somewhere?

packages/@ama-sdk/swagger-builder/package.json
cpaulve-1A: I want to remove the src folder from the dist. Hence, the one from nx will be removed.

packages/@ama-sdk/generator-sdk/tsconfig.build.json
```diff
@@ -7,6 +7,7 @@
     "moduleResolution": "node",
     "noEmitHelpers": false,
     "outDir": "dist/",
+     "rootDir": "src",
```
impy88, line 10: space :)

nx.json
```diff
@@ -152,8 +152,8 @@
         "jest-fixtures"
       ],
       "outputs": [
-        "{projectRoot}/dist/**/fixtures/jasmine",
-        "{projectRoot}/dist/**/fixtures/jest"
+        "{projectRoot}/dist/fixtures/jasmine",
```
impy88, line 155: Do you keep build fixtures, so jest and jasmine seem mixed because of that...

cpaulve-1A: I misunderstood your remark, we still build both but they are exported via different entrypoints. 
We should not have reference of jest in jasmine and vice-versa

packages/@o3r/analytics/project.json
```diff
@@ -22,11 +22,27 @@
...
       "options": {
-        "project": "packages/@o3r/analytics/ng-package.json",
+        "project": "packages/@o3r/analytics/src/ng-package.json",
         "tsConfig": "packages/@o3r/analytics/tsconfig.build.json"
       }
     },
+    "prepare": {
+      "executor": "nx:run-script",
+      "outputs": [
+        "{projectRoot}/src/package.json"
```
kpanot, line 37: maybe you can also specify input as it is only 1 file

packages/@o3r/testing/package.json
```diff
@@ -6,85 +6,87 @@
     "testing",
     "otter"
   ],
+  "main": "dist/public_api.js",
+  "typings": "dist/public_api.d.ts",
+  "ng-update": {
+    "migrations": "./src/migration.json"
```
kpanot, line 12: should not be in `./migration.json`?

cpaulve-1A: updated

packages/@ama-sdk/generator-sdk/package.json
```diff
@@ -78,5 +86,7 @@
     "node": ">=14.0.0",
     "npm": ">=4",
     "yarn": "<4.0.0"
-  }
+  },
+  "main": "./dist/generators/sdk-generator.js",
```
mrednic-1A, line 90: I think it's without `./dist` and I am not sure it is needed

packages/@o3r/testing/package.json
```diff
@@ -6,85 +6,87 @@
     "testing",
     "otter"
   ],
+  "main": "dist/public_api.js",
```
mrednic-1A, line 9: `dist` should not be needed as in the published package we dont have a dist folder.

cpaulve-1A: It is updated during the postbuild task which goes over all the dist reference and remove them

packages/@ama-sdk/generator-sdk/package.json
cpaulve-1A: Yeah it can probably be removed as this is not the way the generator shall be imported

packages/@o3r/testing/package.json
cpaulve-1A: Well it does not change anything after tests, I will leave it this way :)

kpanot (dismissed):

kpanot (approved):

matthieu-crouzet (approved):

mrednic-1A (approved):

kpanot (approved):

PR #190, opened 2023-03-17
fix: #1403 eslint rules for Jest activated by default (jasmine rules missing)

cpaulve-1A:


kpanot (dismissed):

packages/@o3r/eslint-config-otter/jasmine.js
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+/* eslint-disable quote-props */
+module.exports = {
```
kpanot, line 3: Should we really provide this?
- we do not use Jasmine
- we do not recommend Jasmine
- this is just extending the recommended rule that can be done by the user directly
- this will require our generator to be adapted to include or not this file

I would actually not provide it.
The split index.js/base.js would be enough for me.

packages/@o3r/eslint-config-otter/jasmine.js
```diff
@@ -0,0 +1,11 @@
...
+  'extends': [
+    './base',
+    'plugin:eslint-plugin-jasmine/recommended'
```
kpanot, line 6: Is it already part of the optional peer deps?

packages/@o3r/eslint-config-otter/index.js
```diff
@@ -1,567 +1,15 @@
...
   'extends': [
-    'eslint:recommended',
-    'plugin:jsdoc/recommended',
-    'plugin:@angular-eslint/recommended',
-    'plugin:@typescript-eslint/recommended',
-    'plugin:@typescript-eslint/eslint-recommended',
-    'plugin:@typescript-eslint/recommended-requiring-type-checking',
-    'plugin:@o3r/recommended',
+    './base',
     'plugin:jest/recommended'
```
kpanot, line 6: Then this should be part of the optional peer deps to not be install by people using /base

packages/@o3r/eslint-config-otter/package.json
```diff
@@ -10,7 +10,7 @@
     "linter"
   ],
   "scripts": {
-    "build": "yarn cpy '{index,template,fast}.js' package.json dist && yarn postbuild",
+    "build": "yarn cpy '{index,base,template,fast,jasmine,jasmine-fast}.js' package.json dist && yarn postbuild",
```
kpanot, line 13: Should we take the opportunity of this PR to add `exports` for the non-default js files?

packages/@o3r/eslint-config-otter/jasmine.js
jbourgeois-1A: yet we still provide services fixtures for Jasmine.
it doesn't cost much to export it in this package that doesn't change very often

> this will require our generator to be adapted to include or not this file

same applies to the `index/base` split that you suggested.

kpanot: Then fine for me as long as it is optional peerDeps 🙂

kpanot (dismissed):

kpanot (approved):

matthieu-crouzet (approved):

PR #193, opened 2023-03-20
feat: support multiple templates in placeholders

pginoux-1A:
__Previous Behavior__:
Today the current placeholder mechanism only allow to fetch a single URL to retrieve and display the template.
A common usage is to set an action as a result of a rule, and the RulesEngineService will trigger an action on the placeholder store.
The store contains a 1-1 mapping between placeholderId and url, and the fetch of the url + the rendering of the template is handled by the PlaceholderTemplateResponseEffect. The rendered template is set in the store, and is retrieved by the PlaceholderComponent that watch for a specific placeholderId.

__Issue__:
The issue with the current behavior is that as soon as we want to display a second message in the same placeholder, it forces to create a new template containing the 2 messages, and a a third one containing only the second messages, and have it driven by rules.
This doesn't scale much.

__Proposal__:
This pull request decorelates the mapping between placeholder and urls from the url resolve and render.
We have now the PlaceholderTemplate store that stores a mapping between placeholderId and associated Urls (as strings).
The new PlaceholderRequest store will handle the call to the Url, and with the effect the computation and rendering of the response template.
There is now a single selector from PlaceholderTemplate store that allow to get the list of rendered tempates for a given placeholderId : selectPlaceholderRenderedTemplates.
The result will be pending as long as one template is still pending (to prevent UI consecutive refresh adding a new part in the placeholder), but the ones in failure will just be filtered out of the result list.

### DIFF
--- a/docs/rules-engine/how-to-use/placeholders.md
+++ b/docs/rules-engine/how-to-use/placeholders.md
@@ -99,7 +99,23 @@ And then in the angular.json :
 * Html limited to angular sanitizer supported behavior
 * Urls (relative ones will be processed to add the dynamic-media-path)
 * Facts references
-* Localization : for each change a new url will be fetched
+
+### Static localization
+The first choice you have when you want to localize your template is the static localization.
+You need to create a localized template for each locale and provide the template URL with [LANGUAGE] (ex: assets/placeholders/[LANGUAGE]/myPlaceholder.json)
+The rules engine service will handle the replacement of [LANGUAGE] for you, and when you change language a new call will be performed to the new 'translated' URL.
+
+Note that the URL caching mechanism is based on the url NOT 'translated', meaning that if you change from en-GB to fr-FR then you decide to switch back and all the calls will be made again.
+This behavior is based on the fact that a real user rarely goes back and forth with the language update.
+
+### Dynamic localization
+
+### Multiple templates in same placeholder
+You can use placeholder actions to target the same placeholderId with different template URLs.
+It groups the rendered templates in the same placeholder, and you can choose the order by using the `priority` attribute in the action.
+If not specified, the priority defaults to 0. Then the higher the number, the higher the priority. The final results are displayed in descending order of priority.
+The placeholder component waits for all the calls to be resolved (not pending) to display the content.
+The placeholder component ignores a template if the application failed to retrieve it.
 
 ## Investigate issues
 If the placeholder is not rendered properly, you can perform several check to find out the root cause, simply looking at the store state.
--- a/packages/@o3r/components/src/tools/placeholder/placeholder.component.ts
+++ b/packages/@o3r/components/src/tools/placeholder/placeholder.component.ts
@@ -1,8 +1,16 @@
-import { ChangeDetectionStrategy, ChangeDetectorRef, Component, Input, OnDestroy, OnInit, ViewEncapsulation } from '@angular/core';
-import { select, Store } from '@ngrx/store';
-import { ReplaySubject, Subscription } from 'rxjs';
-import { distinctUntilChanged, switchMap } from 'rxjs/operators';
-import { PlaceholderTemplateStore, selectPlaceholderTemplateEntity } from '../../stores/placeholder-template';
+import {
+  ChangeDetectionStrategy,
+  ChangeDetectorRef,
+  Component,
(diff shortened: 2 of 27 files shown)


packages/@o3r/components/src/tools/placeholder/placeholder.component.ts
```diff
@@ -44,11 +52,20 @@ export class PlaceholderComponent implements OnInit, OnDestroy {
...
+        switchMap((id) => this.store.select(selectPlaceholderRenderedTemplates(id)))
+      ).subscribe((templates) => {
+        if (templates) {
+          this.isPending = templates.isPending;
+          const orderedRenderedTemplates = templates.orderedRenderedTemplates;
+          if (!orderedRenderedTemplates || !orderedRenderedTemplates.length) {
+            this.template = undefined;
+          } else if (orderedRenderedTemplates.length === 1) {
+            // Behavior before the multi url in placeholder support
+            this.template = orderedRenderedTemplates[0];
```
cpaulve-1A, line 64: Isn't it the same as line 67 ? Why do you need to handle it separately ?

cpaulve-1A (dismissed):

fpaul-1A (approved):

docs/rules-engine/how-to-use/placeholders.md
```diff
@@ -99,7 +99,23 @@ And then in the angular.json :
...
+You need to create a localized template for each locale and provide the template Url with [LANGUAGE] (ex: assets/placeholders/[LANGUAGE]/myPlaceholder.json)
+The rules engine service will handle the replacement of [LANGUAGE] for you, and when you change lang a new call will be performed to the new 'translated' Url.
+
+Note that the URL caching mechanism is based on the url NOT 'translated', meaning that if you change from en-GB to fr-FR then you decide to switch back all the calls will be done again.
+This behavior is based on the fact that a real user rarely go back and force with the language update.
+
+### Dynamic localization
+
+### Multiple templates in same placeholder
+You can use placeholder actions to target the same placeholderId with different templates URLs.
```
sdo-1A, line 114:
```suggestion
You can use placeholder actions to target the same placeholderId with different template URLs.
```

docs/rules-engine/how-to-use/placeholders.md
```diff
@@ -99,7 +99,23 @@ And then in the angular.json :
...
 * Urls (relative ones will be processed to add the dynamic-media-path)
 * Facts references
-* Localization : for each change a new url will be fetched
+
+### Static localization
+The first choice you have when you want to localize your template is the static localization.
+You need to create a localized template for each locale and provide the template Url with [LANGUAGE] (ex: assets/placeholders/[LANGUAGE]/myPlaceholder.json)
+The rules engine service will handle the replacement of [LANGUAGE] for you, and when you change lang a new call will be performed to the new 'translated' Url.
+
+Note that the URL caching mechanism is based on the url NOT 'translated', meaning that if you change from en-GB to fr-FR then you decide to switch back all the calls will be done again.
```
sdo-1A, line 108:
```suggestion
Note that the URL caching mechanism is based on the url NOT 'translated', meaning that if you change from en-GB to fr-FR then you decide to switch back and all the calls will be made again.
```

docs/rules-engine/how-to-use/placeholders.md
```diff
@@ -99,7 +99,23 @@ And then in the angular.json :
 * Html limited to angular sanitizer supported behavior
 * Urls (relative ones will be processed to add the dynamic-media-path)
 * Facts references
-* Localization : for each change a new url will be fetched
+
+### Static localization
+The first choice you have when you want to localize your template is the static localization.
+You need to create a localized template for each locale and provide the template Url with [LANGUAGE] (ex: assets/placeholders/[LANGUAGE]/myPlaceholder.json)
+The rules engine service will handle the replacement of [LANGUAGE] for you, and when you change lang a new call will be performed to the new 'translated' Url.
```
sdo-1A, line 106:
```suggestion
The rules engine service will handle the replacement of [LANGUAGE] for you, and when you change language a new call will be performed to the new 'translated' URL.
```

cpaulve-1A (approved):

fpaul-1A (approved):

PR #194, opened 2023-03-20
feat: add cache for yarn for test-app on it tests

fpaul-1A:
Persist the cache generated by yarn on the install of the test-app generated by the it tests
The cache should be created once per month and the same instance should be reused by all the build

### DIFF
--- a/.github/workflows/test-publish.yml
+++ b/.github/workflows/test-publish.yml
@@ -21,6 +21,18 @@ jobs:
       - uses: ./.github/actions/setup-java
         with:
           install-jdk: 'true'
+      - name: Get current month to build the cache key
+        run: echo "currentMonth=$(date +'%Y-%m')" >> $GITHUB_ENV
+        shell: bash
+      - name: Cache test-app yarn
+        uses: actions/cache@v3
+        with:
+          path: |
+            .cache/test-app
+            !.cache/test-app/v6/npm-@ama-sdk*
+            !.cache/test-app/v6/npm-@ama-terasu*
+            !.cache/test-app/v6/npm-@o3r*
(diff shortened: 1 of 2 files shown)


.github/actions/setup/action.yml
```diff
@@ -48,6 +52,15 @@ runs:
           ${{ runner.os }}-build-${{ env.hash }}-
           ${{ runner.os }}-build-
           ${{ runner.os }}-
+    - name: Cache e2e
+      if: inputs.enable-e2e-cache == 'true'
+      uses: actions/cache@v3
+      with:
+        path: |
+          .cache/e2e
+        key: ${{ runner.os }}-e2e-${{ hashFiles('yarn.lock') }}
```
kpanot, line 61: It will be reset only in case of `yarn.lock` update.
Maybe you can add `hashFiles('**/*.it.spec.ts')` as well

fpaul-1A: as discussed, since we don't have a good way to generate a hash key to identify the dependencies of the test-app we decided to use a self-invalidating key that will change every month

pginoux-1A (dismissed):

tools/github-actions/setup/action.yml
```diff
@@ -6,6 +6,10 @@ inputs:
     description: 'Enable Cache'
     required: false
     default: 'false'
+  enable-test-app-cache:
```
kpanot, line 9: Why putting this cache option in the setup action when it is actually used only by 1 workflow?
Do you need to share this task? If so maybe the cache name should be part of the parameter as well.

kpanot (approved):

kpanot (approved):

PR #200, opened 2023-03-23
fix: otter 8 requires nodejs16 for migration scripts

pginoux-1A:
Migration scripts use ||= that is not supported by nodejs14. 
The security maintenance of nodejs 14 ends next month, time to upgrade our minimal requirement.


matthieu-crouzet (dismissed):

kpanot (approved):

packages/@ama-terasu/schematics/src/create-module/templates/package.json.template
```diff
@@ -22,7 +22,7 @@
     "@ama-terasu/core": "^<%= version %>"
   },
   "engines": {
-    "node": ">=14.0.0",
+     "node": ">=16.0.0",
```
mrednic-1A, line 25: unfortunately there is an extra space added :D

mrednic-1A (approved):

packages/@ama-terasu/schematics/src/create-module/templates/package.json.template
pginoux-1A: sad :(

mrednic-1A (approved):

PR #219, opened 2023-03-29
feat: support jsonPath in placeholders

pginoux-1A:
Today the placeholder supports facts variables : 
```
{
  "template": "<p>My fact : <%= myFact %>",
  "vars": {
    "myFact ": {
      "value": "myFact ",
      "type": "fact"
    }
  }
}
```
The goal is to add support to an optional field, path, that will be used as a json path to extract a slice of a complex object:
```
{
  "template": "<p>My cart data : <%= cart %>",
  "vars": {
    "cart": {
      "value": "cart",
      "type": "fact",
      "path": "$.airOffers[*].offerItems[*].air.bounds[*].destinationLocation.countryCode"
    }
  }
}
```

### DIFF
--- a/packages/@o3r/rules-engine/src/services/rules-engine.effect.ts
+++ b/packages/@o3r/rules-engine/src/services/rules-engine.effect.ts
@@ -14,6 +14,7 @@ import {LocalizationService} from '@o3r/localization';
 import {combineLatest, EMPTY, firstValueFrom, Observable, of} from 'rxjs';
 import {map, switchMap, take} from 'rxjs/operators';
 import {RulesEngineService} from './rules-engine.service';
+import {JSONPath} from 'jsonpath-plus';
 
 /**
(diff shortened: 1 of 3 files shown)


kpanot (dismissed):

packages/@o3r/rules-engine/src/services/rules-engine.effect.ts
```diff
@@ -102,7 +103,10 @@ export class PlaceholderTemplateResponseEffect {
               break;
             }
             case 'fact': {
```
mrednic-1A, line 105: it might worth to add a unit test

mrednic-1A (approved):

kpanot (dismissed):

kpanot (approved):

mrednic-1A (approved):

cpaulve-1A (approved):

PR #242, opened 2023-04-05
feat: expose json schema for functional content

matthieu-crouzet:


packages/@o3r/application/schemas/functional-content.metadat.schema.json
```diff
@@ -0,0 +1,306 @@
...
+                  "$ref": "#/$defs/widget"
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  },
+  "$defs": {
```
kpanot, line 145: I don't think it is part of  `http://json-schema.org/schema`, usually we use `definitions` instead

packages/@o3r/application/schemas/functional-content.metadat.schema.json
```diff
@@ -0,0 +1,306 @@
...
+            "pattern": ".*\\.json$"
+          },
+          "rootReference": {
+            "type": "string",
+            "description": "Name of the object definition to use as the root of the model described by this section"
+          },
+          "definitions": {
+            "type": "array",
+            "description": "Object definitions for this section",
+            "minItems": 1,
```
kpanot, line 68: why this `minItems: 1` to a mandatory list?

packages/@o3r/application/schemas/functional-content.metadat.schema.json
```diff
@@ -0,0 +1,306 @@
...
+      "properties": {
+        "type": {
+          "type": "string"
+        },
+        "parameters": {
+          "type": "object",
+          "description": "map of widget parameters. Each parameter can be a string, number or boolean or an array of those",
+          "additionalProperties": false,
+          "patternProperties": {
+            "^[a-zA-Z0-9-_:.]*$": {
```
kpanot, line 286: maybe `+` instead of `*`

packages/@o3r/application/schemas/functional-content.metadat.schema.json
guillaume-goulet: The objective here is to have a mandatory list and it should not be empty.
Without setting the minItems, "sections":[] should not be allowed

kpanot: Not having definitions is not possible?

matthieu-crouzet:
PR raised by @guillaume-goulet https://github.com/AmadeusITGroup/otter/pull/245

PR #248, opened 2023-04-05
fix: selectByIndex resolved before mat-option is displayed

cpaulve-1A:
Wait for first option to be displayed before attempting to click on the element


packages/@o3r/testing/src/core/playwright/angular-materials/select-material.ts
```diff
@@ -11,9 +11,10 @@ export class MatSelect extends O3rElement implements MatSelectProfile {
   }
 
   /** @inheritdoc */
-  public async selectByIndex(index: number) {
+  public async selectByIndex(index: number, timeout = 5000) {
```
kpanot, line 14: Should not be also added in the `SelectElementProfile` interface as well?

kpanot (approved):

PR #276, opened 2023-04-17
fix: generators templates

cpaulve-1A:
## Proposed change

Apply subentries fix on generated module. 
Move ng-package into the source folder to take the src folder as the root of the subentries instead of the package folder.
Include the scripts to target the src ng-package.json

- :bug: Fixes #(issue)
## Type of change

- [x] :bug: Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/packages/@o3r/core/schematics/module/templates/base/ng-package.json.template
+++ b/packages/@o3r/core/schematics/module/templates/base/ng-package.json.template
@@ -1,10 +1,10 @@
 
 {
   "$schema": "https://raw.githubusercontent.com/ng-packagr/ng-packagr/master/src/ng-package.schema.json",
-  "dest": "../dist",
+  "dest": "./dist",
   "deleteDestPath": false,
 
   "lib": {
-    "entryFile": "public_api.ts"
+    "entryFile": "src/public_api.ts"
   }
 }


matthieu-crouzet (approved):

packages/@o3r/core/schematics/module/templates/nx/project.json.template
```diff
@@ -48,6 +49,21 @@
           "<%= path %>/<%= name %>/src/**/*.ts"
         ]
       }
+    },
+     "prepare": {
+       "executor": "nx:run-script",
+       "outputs": [
+         "{projectRoot}/src/package.json"
```
kpanot, line 56: So that mean that this file should also be nxignored (or removed from `source` inputs of the other tasks)

packages/@o3r/core/schematics/module/templates/base/package.json.template
```diff
@@ -11,6 +11,7 @@
     "ng": "<%= runner %> nx",
     "build": "<%= runner %> nx build apis-manager",
     "prepare:build:builders": "<%= runner %> cpy 'collection.json' dist && <%= runner %> cpy 'schematics/**/*.json' dist/schematics",
+    "prepare:compile": "cp-package-json",
```
kpanot, line 14: Not sure to understand why doing this process as it is needed only for package with subentries (which may be asked by the generator but not considered as default behavior :S)

kpanot: But it is adding more that boilerplate, it is also adding complexity to Nx rules.
I would really advice to do it only when there is no other choice as workaround for sub entries packages (which should not be the often anymore)

cpaulve-1A: that's fine by me, I will keep the PR to update the ng-package.json that is incorrect anyway

matthieu-crouzet (approved):

PR #294, opened 2023-04-25
fix: throwOnUndefined to work with playwrigh

kpanot:
fix: throwOnUndefined to work with playwrigh
feat: add promise support to throwOnUndefined

## Proposed change

Add support of Promise to `throwOnUndefined` to avoid breaking change of the fix

## Related issues

- :bug: Fixes 1441

## Remains To Do:

- [x] Ng Update for migration

### DIFF
--- a/packages/@o3r/core/schematics/fixture/helpers.ts
+++ b/packages/@o3r/core/schematics/fixture/helpers.ts
@@ -93,38 +93,40 @@ export const getImplementation = (
   switch (methodType) {
     case 'clickOnButton': {
       return `{
-    const elt = await this.query(this.${classPropSelector});
-    return this.throwOnUndefined(elt).click();
+    const elt = await this.throwOnUndefined(this.query(this.${classPropSelector}));
+    return elt.click();
   }`;
     }
     case 'getText': {
       return `{
-    const elt = await this.query(this.${classPropSelector});
-    return this.throwOnUndefined(elt).getText();
+    const elt = await this.throwOnUndefined(this.query(this.${classPropSelector}));
+    return elt.getText();
   }`;
     }
     case 'getInputValue': {
       return `{
-    const elt = await this.query(this.${classPropSelector});
-    return this.throwOnUndefined(elt).getValue();
+    const elt = await this.throwOnUndefined(this.query(this.${classPropSelector}));
+    return elt.getValue();
   }`;
     }
     case 'setInputValue': {
       return `{
-    const elt = await this.query(this.${classPropSelector});
-    return this.throwOnUndefined(elt).setValue(value);
+    const elt = await this.throwOnUndefined(this.query(this.${classPropSelector}));
+    return elt.setValue(value);
   }`;
     }
     case 'getTextInList': {
       return `{
-    const elements = await this.queryAll(this.${classPropSelector});
-    return this.throwOnUndefined(elements[index]).getText();
+    const elements = this.queryAll(this.${classPropSelector});
+    const elt = await this.throwOnUndefinedElement(items[index]);
+    return elf.getText();
   }`;
     }
     case 'clickButtonInList': {
       return `{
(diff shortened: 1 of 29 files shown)


matthieu-crouzet (dismissed):

packages/@o3r/testing/src/core/playwright/component-fixture.ts
```diff
@@ -26,11 +26,31 @@ export class O3rComponentFixture<V extends O3rElement = O3rElement> implements C
...
 
-  protected throwOnUndefined<T extends O3rElement>(elemnt?: T) {
-    if (!elemnt) {
+  /** @deprecate not supported for Playwright */
+  protected throwOnUndefined<T extends O3rElement>(element?: T): T;
+  /** @inheritdoc */
+  protected async throwOnUndefined<T extends O3rElement>(element: Promise<T | undefined>): Promise<T>;
+  /**
+   * @param element
+   * @deprecate not supported for Playwright
```
cpaulve-1A, line 35: you provide an implementation but the way you phrase the deprecated it does not seem possible to implement it which is a bit confusing.
Do you mean that we deprecate it because it goes against playwright's locator design?

kpanot: If I put something like `@deprecated not valid in case of Playwright due to the Locator implementation`, what do you think>?

packages/@o3r/testing/src/core/protractor/component-fixture.ts
```diff
@@ -28,11 +28,25 @@ export class O3rComponentFixture<V extends O3rElement = O3rElement> implements C
...
 
-  protected throwOnUndefined<T extends O3rElement>(elemnt?: T) {
-    if (!elemnt) {
+  /** @inheritdoc */
+  protected throwOnUndefined<T extends O3rElement>(elemnt?: T): T;
+  /** @inheritdoc */
+  protected async throwOnUndefined<T extends O3rElement>(elemnt: Promise<T | undefined>): Promise<T>;
+  /** @inheritdoc */
+  protected throwOnUndefined<T extends O3rElement>(elemnt?: Promise<T | undefined> | T): Promise<T> | T {
+    if (!element) {
```
guillaumerochelle, line 37: Should be `elemnt` ?

packages/@o3r/core/schematics/fixture/helpers.ts
```diff
@@ -93,38 +93,40 @@ export const getImplementation = (
   switch (methodType) {
     case 'clickOnButton': {
       return `{
-    const elt = await this.query(this.${classPropSelector});
-    return this.throwOnUndefined(elt).click();
+    const eltPromise = await this.throwOnUndefined(this.query(this.${classPropSelector}));
```
fpaul-1A, line 96:
```suggestion
    const elt = await this.throwOnUndefined(this.query(this.${classPropSelector}));
```
?

matthieu-crouzet (approved):

packages/@o3r/core/schematics/fixture/helpers.ts
```diff
@@ -93,38 +93,40 @@ export const getImplementation = (
   switch (methodType) {
     case 'clickOnButton': {
       return `{
-    const elt = await this.query(this.${classPropSelector});
-    return this.throwOnUndefined(elt).click();
+    const elt = await this.throwOnUndefined(this.query(this.${classPropSelector}));
```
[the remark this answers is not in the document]
kpanot: Both `clickOnButton` and `this.throwOnUndefined` are async functions.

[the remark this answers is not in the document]
vscaiceanu-1a: but `getImplementation` isn't ... ?

vscaiceanu-1a (approved):

PR #298, opened 2023-04-28
feat: update FC metadata json schema to be more future proof

guillaume-goulet:
## Proposed change

Update FC metadata JSON schema to be more future proof:
- Remove additionalProperties everywhere except for the widget properties

## Related issues

- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/application/schemas/functional-content.metadata.schema.json
+++ b/packages/@o3r/application/schemas/functional-content.metadata.schema.json
@@ -10,7 +10,6 @@
     "description",
     "sections"
   ],
-  "additionalProperties": false,
   "properties": {
     "$schema": {
       "type": "string",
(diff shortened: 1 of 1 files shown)


packages/@o3r/application/schemas/functional-content.metadata.schema.json
```diff
@@ -10,7 +10,6 @@
     "description",
     "sections"
   ],
-  "additionalProperties": false,
```
kpanot, line 13: Why removing strong typing on JSON check?

guillaume-goulet: I am going to sound crazy and seem afraid of you but do not get too excited :p.
After some discussions internally, we decided to cancel this PR.

FYI- the original reason:
This story has been raised after some discussions as the FC were not correctly used by some internal apps (That is why we now have the new property outputObject).
So the goal of this story was to let them work/test then add back the additionalProperties to false later.

PR #299, opened 2023-04-28
fix: mock helper getPath - adding an HTTP method constraint when several matches

### LINKED ISSUE #296: [Bug]: Mock helper - several pathObjects matching for a requestUrl and wrong match returned
### Package name

core

### Package version

8.1.5

### Reproduction steps

Issue is localized in the `core/fwk/mocks/helper.ts` > `getPath` method.

Context :

> Considering these 2 endpoints :
> - `river/otter` - methods : ['POST']
> - `river/{otter-id}` - methods : ['GET', 'DELETE']
>
> `{otter-id}` parameter do not have any constraint and can be any string
> 

Calling the `getPath` method with a `requestUrl = "river/otter"` returns the PathObject with baseUrl `river/{otter-id}` instead of the PathObject with baseUrl `river/otter`. Indeed, the method is not able to separate them correctly, as "otter" URI element could be interpreted as an '{otter-id}' (as '{otter-id}' does not have any constraint). 

In this context, the only thing on which we could take a correct decision is basing on the HTTP method.  Adding a constraint on the HTTP method when several matches are found would have filtered the wrong result and returned the good one.

### Current result

The first PathObject match is returned, as the method didn't have enough infos to choose the good one.

### Expected result

The correct PathObject is returned (adding a constraint on the HTTP method)

### Additional comments

_No response_

alexandre-mazurier-ama:
## Proposed change

Adding a last way to filter out several matches in the `packages/@ama-sdk/core/src/fwk/mocks/helper.ts` > `getPath` method, based on the HTTP method of the matching results.

## Related issues

- :bug: Fixes #296

### DIFF
--- a/packages/@ama-sdk/core/src/fwk/mocks/helpers.ts
+++ b/packages/@ama-sdk/core/src/fwk/mocks/helpers.ts
@@ -22,8 +22,9 @@ export function getOperationId(pathObject: PathObject, method: string): string |
  *
  * @param requestUrl the URL string
  * @param pathObjects the list of available path objects
+ * @param method the optional HTTP method used in case of several matches
  */
-export function getPath(requestUrl: string, pathObjects: PathObject[]): PathObject | undefined {
+export function getPath(requestUrl: string, pathObjects: PathObject[], method?: string): PathObject | undefined {
   const pathName = new URL(requestUrl, requestUrl.startsWith('/') ? 'http://example.com' : undefined).pathname;
 
   if (!pathName) {
@@ -31,11 +32,12 @@ export function getPath(requestUrl: string, pathObjects: PathObject[]): PathObje
   }
 
   // Find all matching paths
-  let matches = pathObjects.reduce<{index: number; segments: string[]}[]>((newMatches, pathObject, index) => {
+  let matches = pathObjects.reduce<{index: number; segments: string[]; methods: string[]}[]>((newMatches, pathObject, index) => {
     if (pathObject.regexp.test(pathName)) {
       newMatches.push({
         index,
-        segments: pathObject.path.split('/')
+        segments: pathObject.path.split('/'),
+        methods: pathObject.operations.map((operation) => operation.method)
       });
     }
     return newMatches;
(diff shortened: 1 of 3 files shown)


packages/@ama-sdk/core/src/fwk/mocks/helpers.ts
```diff
@@ -22,20 +22,22 @@ export function getOperationId(pathObject: PathObject, method: string): string |
...
 
   // Find all matching paths
-  let matches = pathObjects.reduce<{index: number; segments: string[]}[]>((newMatches, pathObject, index) => {
+  let matches = pathObjects.reduce<{index: number; segments: string[]; methods: string[]}[]>((newMatches, pathObject, index) => {
     if (pathObject.regexp.test(pathName)) {
       newMatches.push({
         index,
-        segments: pathObject.path.split('/')
+        segments: pathObject.path.split('/'),
+        methods: pathObject.operations.map((operation) => operation.method)
```
kpanot, line 40: Or `({ method }) => method)`

packages/@ama-sdk/core/src/fwk/mocks/helpers.ts
```diff
@@ -22,20 +22,22 @@ export function getOperationId(pathObject: PathObject, method: string): string |
...
+export function getPath(requestUrl: string, pathObjects: PathObject[], method?: string): PathObject | undefined {
   const pathName = new URL(requestUrl, requestUrl.startsWith('/') ? 'http://example.com' : undefined).pathname;
 
   if (!pathName) {
     throw new Error(`Couldn't parse url ${requestUrl}`);
   }
 
   // Find all matching paths
-  let matches = pathObjects.reduce<{index: number; segments: string[]}[]>((newMatches, pathObject, index) => {
+  let matches = pathObjects.reduce<{index: number; segments: string[]; methods: string[]}[]>((newMatches, pathObject, index) => {
```
kpanot, line 35: Instead of `string` you can use literal union type to avoid the following `toLowerCase`

alexandre-mazurier-ama: The initial idea was to restrict the method type to HTTP methods `get`, `post`, etc, But as discussed with @kpanot, each objects in this context are typed with a `string` type (both `PatchObject.operation.method` and `EncodedApiRequest.method` are of type `string`) => Too much changes for this contribution. The `toLowerCase` will be kept, as it is already used by others methods of the helper.

kpanot (approved):

vscaiceanu-1a (approved):

PR #301, opened 2023-05-01
feat: add code analysis

kpanot:
## Proposed change

Add code analysis (remaining Github security checks)

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,58 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,57 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches:
+    - main
+    - 'release/*'
+  pull_request:
+    branches:
+    - main
```
vscaiceanu-1a, line 21: why not PRs targeting `release/*` branches?

kpanot: Just took the default one provided by Github.
But I asked myself the same question so I think I will add it

cpaulve-1A (approved):

PR #306, opened 2023-05-03
fix: mock helper getPath - adding an HTTP method constraint when several matches

### LINKED ISSUE #296: [Bug]: Mock helper - several pathObjects matching for a requestUrl and wrong match returned

alexandre-mazurier-ama:
## Proposed change

Adding a last way to filter out several matches in the `packages/@ama-sdk/core/src/fwk/mocks/helper.ts` > `getPath` method, based on the HTTP method of the matching results.

Fix already merged on `main` branch - PR #299 

## Related issues

- :bug: Fixes #296

### DIFF
--- a/packages/@ama-sdk/core/src/fwk/mocks/alf-mock-adapter.ts
+++ b/packages/@ama-sdk/core/src/fwk/mocks/alf-mock-adapter.ts
@@ -65,7 +65,7 @@ function getRequest(log: string, corrId: string, operationAdapter: PathObject[])
     // so we remove the SAP information at the beginning and URL parameters at the end if any
     const requestUrl = match[2].substring(match[2].indexOf('/'), match[2].indexOf('?') > -1 ? match[2].indexOf('?') : match[2].length);
     const method = match[1];
-    const pathObject = getPath(requestUrl, operationAdapter);
+    const pathObject = getPath(requestUrl, operationAdapter, method);
     if (pathObject) {
       const operationId = getOperationId(pathObject, method);
       return {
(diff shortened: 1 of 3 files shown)


kpanot:
For next time: a bug fix can be done directly to the branch you need it to be applied to, without duplicating your PR.
Please avoid as much a possible the PR duplication (#299).

kpanot (approved):

vscaiceanu-1a (approved):

PR #313, opened 2023-05-05
fix: reduce logging noise on configuration

cpaulve-1A:
The override of a component configuration can be done several time a flow with the same value

This happens if extension is done at component initialization. This is actually a common use case

## Related issues

- :bug: Fixes #(issue)


kpanot (dismissed):

kpanot (approved):

kpanot:
I understand that this fix is mainly to reduce the number of emitted messags in case of network logger reporter but I still think that the issue is more on the reporter side as it is quite standard to have verbose debug messages.

matthieu-crouzet (approved):

PR #321, opened 2023-05-09
fix: register complex client at runtime

cpaulve-1A:
## Proposed change

Allow the dynamic registeration of LoggerClient to avoid issues with complex loggers (potential circular dependencies).

### DIFF
--- a/packages/@o3r/logger/src/services/logger/logger.service.ts
+++ b/packages/@o3r/logger/src/services/logger/logger.service.ts
@@ -16,6 +16,11 @@ export class LoggerService implements Logger {
   /** Loggers */
   private readonly clients: LoggerClient[];
 
+  /**
+   * Record the recording status to make sure that new clients recording status will be consistent with the service
+   */
+  private recordingState: 'default' | 'resumed' | 'stopped' = 'default';
+
   constructor(@Optional() @Inject(LOGGER_CLIENT_TOKEN) clients?: LoggerClient | LoggerClient[]) {
     this.clients = clients ? (Array.isArray(clients) ? clients : [clients]) : [new ConsoleLogger()];
   }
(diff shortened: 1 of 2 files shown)


packages/@o3r/logger/src/services/logger/logger.service.ts
```diff
@@ -16,6 +16,13 @@ export class LoggerService implements Logger {
   /** Loggers */
   private readonly clients: LoggerClient[];
 
+  /**
+   * Record the recording status to make sure that new clients recording status will be consistent with the service
+   *
+   * @private
```
kpanot, line 22: not needed

packages/@o3r/logger/src/services/logger/logger.service.ts
```diff
@@ -50,17 +57,37 @@ export class LoggerService implements Logger {
     return sessionUrls.length <= 1 ? sessionUrls[0] : sessionUrls;
   }
 
+  /**
+   * Register a new client to the logger service
+   *
+   * @param client
+   */
+  public registerClient(client: LoggerClient) {
```
kpanot, line 65: maybe you should expose the a getter of the client list as well?

packages/@o3r/logger/src/services/logger/logger.service.ts
cpaulve-1A: auto-generated

packages/@o3r/logger/src/services/logger/logger.service.ts
cpaulve-1A: what for?

packages/@o3r/logger/src/services/logger/logger.service.ts
kpanot: By your IDE? Maybe you should review a bit its configu

packages/@o3r/logger/src/services/logger/logger.service.ts
kpanot: List the registered clients to not register several time the same and wait for the warning saying it has been registered several time?

kpanot (approved):

PR #341, opened 2023-05-12
fix: parsing of the cascading app config

kpanot:


vscaiceanu-1a (dismissed):

apps/github-cascading-app/src/cascading/cascading-probot.ts
```diff
@@ -51,19 +51,21 @@ export class CascadingProbot extends Cascading {
       }).catch(() => null)));
 
     const configFileValidResponses = configFileResponses
-      .filter((res) => res && Array.isArray(res.data) && res.data.length > 0);
+      .map((res) => res && !Array.isArray(res.data) && res.data || null)
```
pginoux-1A, line 54: Just to be sure I got it right: octokit.repos.getContent returns an array if the path given is a directory, so the goal here is to filter those outs ?

apps/github-cascading-app/src/cascading/cascading-probot.ts
```diff
@@ -51,19 +51,21 @@ export class CascadingProbot extends Cascading {
...
 
     if (configFileValidResponses.length > 1) {
-      this.logger.warn(`Several configuration files have been found (${configFileValidResponses.map((c) => c && Array.isArray(c.data) && c.data[0].path).join(', ')}). The files will be ignored.`);
+      this.logger.warn(`Several configuration files have been found (${configFileValidResponses.map((c) => c?.path).join(', ')}). The files will be ignored.`);
     } else if (!configFileResponse) {
       this.logger.warn('No remote Configuration found, the default configuration will be used');
     } else {
       try {
-        const parsedConfig = Array.isArray(configFileResponse.data) && configFileResponse.data[0].content && JSON.parse(configFileResponse.data[0].content);
+        const configFileResponseWithContent: { content?: string; encoding?: 'utf8' | 'utf-8' | 'base64' } = configFileResponse as unknown as any;
```
pginoux-1A, line 65: What's the difference here between as unknown as any and just as any ?

apps/github-cascading-app/src/cascading/cascading-probot.ts
kpanot: yes, we keep only files

apps/github-cascading-app/src/cascading/cascading-probot.ts
kpanot: `as any` was rising an exception at transpile time so I added the `as unknown` to make it passed (not sure why the `as any` was rising an exception here :S)

apps/github-cascading-app/src/cascading/cascading-probot.ts
pginoux-1A: Should never happen since it's driven by CONFIGURATION_FILES right ?

kpanot: Could happen if someone use named a folder with one of the file names listed :D.
And it is also a way to ensure typing without casting

matthieu-crouzet (approved):

.husky/commit-msg
cpaulve-1A, line 1: why?

cpaulve-1A (approved):

.husky/commit-msg
kpanot: Give execution right for Unix/Linux based OS

PR #343, opened 2023-05-15
fix: remove the dependency of schematics to dev-tools

### LINKED ISSUE #332: [Bug]: ng add @o3r/core doesn't work
### Package name

core

### Package version

8.1.6

### Reproduction steps

- create a new Angular app with `ng new`
- run `ng add @o3r/core`  in the app folder

### Current result

```shell
$ ng add @o3r/core
ℹ Using package manager: npm
✔ Found compatible package version: @o3r/core@8.1.6.
✔ Package information loaded.

The package @o3r/core@8.1.6 will be installed and executed.
Would you like to proceed? Yes
✔ Packages successfully installed.
require() of ES Module /Users/vscaiceanu/workspace/otter/otter-app/node_modules/globby/index.js from /Users/vscaiceanu/workspace/otter/otter-app/node_modules/@o3r/dev-tools/src/utils/jest.helpers.js not supported.
index.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /Users/vscaiceanu/workspace/otter/otter-app/node_modules/globby/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

### Expected result

```shell
? Activate prefetch builder to generate prefetcher JavaScript file ? (Y/n) 
```

### Additional comments

As a workaround you can run the command again and the schematics will be executed.

```shell
$ ng version

     _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/
    

Angular CLI: 15.2.8
Node: 18.14.1
Package Manager: npm 9.5.0
OS: darwin arm64

Angular: 15.2.9
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1502.8
@angular-devkit/build-angular   15.2.8
@angular-devkit/core            15.2.8
@angular-devkit/schematics      15.2.8
@angular/cli                    15.2.8
@schematics/angular             15.2.8
rxjs                            7.8.1
typescript                      4.9.5
```

kpanot:
## Proposed change

Removing dependency of schematics to the dev-tools to avoid ESM / CommonJs clash

## Related issues

- :bug: Fixes #332

### DIFF
--- a/packages/@o3r/dev-tools/src/utils/check-peer-deps.ts
+++ b/packages/@o3r/dev-tools/src/utils/check-peer-deps.ts
@@ -67,6 +67,7 @@ export function getPackagesToInstallOrUpdate(packageName: string) {
 /**
  * Log an instruction with the packages to install or update to match a package peer dependencies
  *
+ * @deprecated use `checkPackagesRule` functions exposed by `@o3r/schematics` instead. Will be removed in Otter v10
  * @param packageName
  * @param angularJsonString
  */
--- a/packages/@o3r/schematics/package.json
+++ b/packages/@o3r/schematics/package.json
@@ -28,7 +28,6 @@
     }
   },
   "dependencies": {
-    "@o3r/dev-tools": "workspace:^",
     "comment-json": "^4.1.0",
     "globby": "^11.1.0",
     "minimatch": "^6.1.6",
@@ -55,6 +54,7 @@
     "@nrwl/js": "~15.9.0",
     "@nrwl/linter": "~15.9.0",
     "@o3r/build-helpers": "workspace:^",
+    "@o3r/dev-tools": "workspace:^",
     "@o3r/eslint-plugin": "workspace:^",
     "@schematics/angular": "~15.2.0",
     "@types/jest": "~28.1.2",
--- a/packages/@o3r/schematics/src/rule-factories/check-packages-peers/index.ts
+++ b/packages/@o3r/schematics/src/rule-factories/check-packages-peers/index.ts
@@ -1,5 +1,114 @@
 import { SchematicContext, Tree } from '@angular-devkit/schematics';
-import { checkPackagesToInstallOrUpdate } from '@o3r/dev-tools';
+import type { PackageJson } from 'type-fest';
+import * as path from 'node:path';
+import * as commentJson from 'comment-json';
+import { satisfies } from 'semver';
+import type { LoggerApi } from '@angular-devkit/core/src/logger';
+
+/** Interface containing a npm package name, needed version and optionally found version */
+interface PackageVersion {
+  /** Npm package name */
+  packageName: string;
+  /** Npm package needed version */
+  version: string;
+  /** Npm package installed version found */
+  foundVersion?: string;
+}
+
+/**
+ * Get package manager used in runs
+ * Defaults to the package manager setup in process.env if no package manager set in angular.json
+ *
+ * @param angularJsonString Content of angular.json file
+ */
+function getPackageManager(angularJsonString?: string | null) {
+  let packageManager = process.env && process.env.npm_execpath && process.env.npm_execpath.indexOf('yarn') === -1 ? 'npm' : 'yarn';
+  if (angularJsonString) {
+    const angularJsonObj = commentJson.parse(angularJsonString) as any;
+    if (angularJsonObj?.cli?.packageManager) {
+      packageManager = angularJsonObj.cli.packageManager;
+    }
+  }
+  return packageManager;
+}
+
+
+/**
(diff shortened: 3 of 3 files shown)


matthieu-crouzet (approved):

packages/@o3r/dev-tools/src/utils/check-peer-deps.ts
```diff
@@ -67,6 +67,7 @@ export function getPackagesToInstallOrUpdate(packageName: string) {
 /**
  * Log an instruction with the packages to install or update to match a package peer dependencies
  *
+ * @deprecated use `checkPackagesRule` or `checkPackagesToInstallOrUpdate` functions exposed by `@o3r/schematics` instead. Will be removed in Otter v10
```
vscaiceanu-1a, line 70: I think this method is only used in `packages/@o3r/schematics/src/rule-factories/check-packages-peers/index.ts` where you replaced it with the new one. In this case, why not removing it completely from here?

packages/@o3r/schematics/package.json
```diff
@@ -55,6 +54,7 @@
     "@nrwl/js": "~15.9.0",
     "@nrwl/linter": "~15.9.0",
     "@o3r/build-helpers": "workspace:^",
+    "@o3r/dev-tools": "workspace:^",
```
vscaiceanu-1a, line 57: shouldn't this PR remove the dependency?

packages/@o3r/schematics/src/rule-factories/check-packages-peers/index.ts
```diff
@@ -1,15 +1,124 @@
...
+}
+
+/**
+ * Log an instruction with the packages to install or update to match a package peer dependencies
+ *
+ * @param packageName
+ * @param angularJsonString
+ * @param logger
+ */
+export function checkPackagesToInstallOrUpdate(packageName: string, logger: LoggerApi, angularJsonString?: string | null) {
```
vscaiceanu-1a, line 84: if not used outside this file, is it worth exporting it?

packages/@o3r/schematics/src/rule-factories/check-packages-peers/index.ts
```diff
@@ -1,15 +1,124 @@
...
+  });
+  return { packagesToInstall, packagesWrongVersion };
+}
+
+/**
+ * Log an instruction with the packages to install or update to match a package peer dependencies
+ *
+ * @param packageName
+ * @param angularJsonString
+ * @param logger
```
vscaiceanu-1a, line 82: wrong params order

packages/@o3r/dev-tools/src/utils/check-peer-deps.ts
kpanot: Because it is exposed by the package, so removing it would be a breaking change (even if on our side we you it only once :S)

packages/@o3r/schematics/package.json
kpanot: It is now a devDep, used only in test purpose

packages/@o3r/schematics/src/rule-factories/check-packages-peers/index.ts
kpanot: There is no order in JsDoc parameters, this is the reason why it is named :)

packages/@o3r/schematics/src/rule-factories/check-packages-peers/index.ts
kpanot: I hesitated by I finally kept the export because it was exported before and because it was not a big deal to do so

vscaiceanu-1a: Well, this way, next time we want to remove it, we will have to deprecate it because we won't know if anyone else is using it. Let's keep it as a private method and, if we ever need it outside, we can export it?

vscaiceanu-1a (approved):

matthieu-crouzet (approved):

matthieu-crouzet (approved):

PR #344, opened 2023-05-15
fix: adapt command runner to package manager in schematics

cpaulve-1A:
## Proposed change

Do not force yarn on schematics as we both support npm and yarn

### DIFF
--- a/packages/@o3r/core/schematics/ng-add-create/index.ts
+++ b/packages/@o3r/core/schematics/ng-add-create/index.ts
@@ -30,9 +30,10 @@ export function updateTemplates(options: NgGenerateUpdateSchematicsSchema): Rule
       // prepare needed deps for schematics
       const angularVersion = packageJson.devDependencies?.['@angular/cli'] || packageJson.devDependencies?.['@angular/core'];
       const otterVersion = o3rCorePackageJson.dependencies!['@o3r/schematics'];
-      const command = getPackageManagerRunner();
+      const packageManagerRunner = getPackageManagerRunner();
       packageJson.scripts ||= {};
-      packageJson.scripts['build:schematics'] = `tsc -b tsconfig.builders.json --pretty && ${command} cpy 'schematics/**/*.json' dist/schematics && ${command} cpy 'collection.json' dist`;
+      packageJson.scripts['build:schematics'] =
+        `tsc -b tsconfig.builders.json --pretty && ${packageManagerRunner} cpy 'schematics/**/*.json' dist/schematics && ${packageManagerRunner} cpy 'collection.json' dist`;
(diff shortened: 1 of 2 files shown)


vscaiceanu-1a (dismissed):

packages/@o3r/core/schematics/rule-factories/cms-adapter/index.ts
```diff
@@ -97,8 +98,9 @@ export function updateCmsAdapter(options: { projectName: string | null }, rootPa
   const addExtractersScripts = (tree: Tree, _context: SchematicContext) => {
     const workspaceProject = getProjectFromTree(tree);
     const packageJson = readPackageJson(tree, workspaceProject);
+    const command = getPackageManagerRunner();
```
pginoux-1A, line 101: Command can be renamed to packageManagerRunner for more clarity

packages/@o3r/core/schematics/ng-add-create/index.ts
```diff
@@ -1,4 +1,5 @@
 import { chain, move, Rule, SchematicContext, Tree } from '@angular-devkit/schematics';
+import { getPackageManagerRunner } from '@o3r/dev-tools';
```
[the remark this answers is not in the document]
cpaulve-1A: in core or schematics?

pginoux-1A (approved):

packages/@o3r/core/schematics/ng-add-create/index.ts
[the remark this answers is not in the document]
kpanot: Yes, maybe schematics would make more sense in the end

pginoux-1A (approved):

packages/@o3r/schematics/schematics/ng-add/index.ts
```diff
@@ -16,15 +15,15 @@ import { NodePackageName } from '@angular-devkit/schematics/tasks/package-manage
 class DevInstall extends NodePackageInstallTask {
   /** @inheritDoc */
   public toConfiguration() {
-    const installOptions = getPackageManager();
+    const installOptions = process.env && process.env.npm_execpath && process.env.npm_execpath.indexOf('yarn') === -1 ? 'npm' : 'yarn';
```
kpanot, line 18: why not importing it from the `src`?

cpaulve-1A: could do now that it is in the schematics

cpaulve-1A: Nope, you can't refer to the source until the peerdep are resolved, hence I cannot use the one from src.

kpanot (approved):

PR #345, opened 2023-05-16
feat: add useless custom Otter logo

kpanot:


.attachments/logos/otter-straw-hat.png
mrednic-1A, line 1: A `pool party` otter would be a nice to have :D

matthieu-crouzet (approved):

PR #357, opened 2023-05-18
fix: schematics project discovery to get first project based on type

kpanot:


packages/@o3r/core/schematics/rule-factories/cms-adapter/index.ts
```diff
@@ -93,9 +99,16 @@ export function updateCmsAdapter(options: { projectName: string | null }, rootPa
...
    * @param _context
+   * @param context
    */
-  const addExtractersScripts = (tree: Tree, _context: SchematicContext) => {
-    const workspaceProject = getProjectFromTree(tree);
+  const addExtractersScripts = (tree: Tree, context: SchematicContext) => {
+    const workspaceProject = getProjectFromTree(tree, null, 'application');
+
+    if (!workspaceProject) {
+      context.logger.error('No application detected, the extractor scripts will not be added');
```
matthieu-crouzet, line 108: Are you sure we don't want to extract metadata in a library ?
If a project has a library and several applications they would like to extract metadata of the library to be able to specify this library in the field `libraries` in the extractors of their app, no ?

packages/@o3r/core/schematics/rule-factories/otter-environment/index.ts
```diff
@@ -35,11 +35,16 @@ export function updateOtterEnvironmentAdapter(
...
+   * @param context
    */
-  const editAngularJson = (tree: Tree, _context: SchematicContext) => {
+  const editAngularJson = (tree: Tree, context: SchematicContext) => {
     const workspace = readAngularJson(tree);
-    const projectName = options.projectName || workspace.defaultProject || Object.keys(workspace.projects)[0];
-    const workspaceProject = getProjectFromTree(tree);
+    const workspaceProject = getProjectFromTree(tree, null, 'application');
+    if (!workspaceProject) {
+      context.logger.error('No application detected, the extractor will not be added');
```
matthieu-crouzet, line 44: Not sure about the message here 
the following code is not adding extractor

cpaulve-1A (approved):

Issue #363, opened 2023-05-22
[Feature]: fromApiEffectSwitchMap support for undefined callbacks


OxyFlax:
### Description

In the fromApiEffectSwitchMap operator for store, we currently cannot leave any of the callbacks undefined. 
This is a good thing in general to ensure everything is handled for the default effect but it is a limitation for side effects we could create.
Let's say if we want to create effect for particular scenarios such as particular success or error case and we don't want to affect the default effect, we could still use this operator and define only a success or error handlers without affecting the rest.

PR #365, opened 2023-05-23
feat: add the capability to not provide callback to effect helpers

kpanot:
## Proposed change

Add the capability to not provide callback to effect helpers

## Related issues

- :rocket: Feature #363

### DIFF
--- a/packages/@o3r/core/src/store/async/async.operators.ts
+++ b/packages/@o3r/core/src/store/async/async.operators.ts
@@ -14,10 +14,15 @@ import { AsyncRequest, ExtractFromApiActionPayloadType, FromApiActionPayload } f
  * @param cancelRequestActionFactory function that returns the action to emit in case the FromApi action is 'cancelled' because a new action was received by the switchMap
  */
 // eslint-disable-next-line @typescript-eslint/ban-types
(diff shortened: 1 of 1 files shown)


packages/@o3r/core/src/store/async/async.operators.ts
```diff
@@ -14,10 +14,10 @@ import { AsyncRequest, ExtractFromApiActionPayloadType, FromApiActionPayload } f
  * @param cancelRequestActionFactory function that returns the action to emit in case the FromApi action is 'cancelled' because a new action was received by the switchMap
  */
 // eslint-disable-next-line @typescript-eslint/ban-types
-export function fromApiEffectSwitchMap<T extends FromApiActionPayload<any>, S extends ExtractFromApiActionPayloadType<T>, U extends Action, V extends Action, W extends Action>(
+export function fromApiEffectSwitchMap<T extends FromApiActionPayload<any>, S extends ExtractFromApiActionPayloadType<T>, U extends Action, V extends Action | never, W extends Action | never>(
   successHandler: (result: S, action: T) => U | Observable<U>,
```
OxyFlax, line 18: We could also have an optional successHandler, in case we want to do an effect reacting to only errors for eg

kpanot: Unfortunately, this is not possible because the success callback is required to determine the operator return type

OxyFlax (approved):

matthieu-crouzet (approved):

mrednic-1A (approved):

PR #369, opened 2023-05-25
fix: playwright integration

matthieu-crouzet:
## Proposed change

`playwright` is only a devDep for `@o3r/testing`
When install after `@playwright/test` it cause the following issue when running `playwright` command.

```shell
Please install @playwright/test package to use Playwright Test.
  npm install -D @playwright/test
```

## Related issues

- :bug: Fixes

### DIFF
--- a/packages/@o3r/testing/src/core/playwright/element.ts
+++ b/packages/@o3r/testing/src/core/playwright/element.ts
@@ -1,4 +1,4 @@
-import {Locator, Page} from 'playwright';
+import type {Locator, Page} from 'playwright';
 import {ElementProfile} from '../element';
 
 export {ElementProfile} from '../element';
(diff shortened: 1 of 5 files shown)


sdo-1A (approved):

packages/@o3r/testing/src/core/playwright/element.ts
```diff
@@ -1,4 +1,4 @@
-import {Locator, Page} from 'playwright';
+import type {Locator, Page} from 'playwright';
 import {ElementProfile} from '../element';
```
mrednic-1A, line 2: As we export a new type using these 2 types from playwright package ([few lines below](https://github.com/AmadeusITGroup/otter/blob/2c573973f167588597da5a4dc3710e22a9dfef91/packages/%40o3r/testing/src/core/playwright/element.ts#LL13C1-L16C3)) I am not sure we can keep `playwright` only as a devDep.
cc: @kpanot

kpanot: Indeed, it would need to be peerDep as well

PR #385, opened 2023-06-02
feat: improve error logs format

mrednic-1A:
## Proposed change

Improve error log messages

### DIFF
--- a/packages/@ama-sdk/core/src/fwk/errors.ts
+++ b/packages/@ama-sdk/core/src/fwk/errors.ts
@@ -8,6 +8,8 @@ interface GenericErrorContext {
   operationId?: string;
   /** Base URL */
   url?: string;
+  /** Origin domain initiating the call */
+  origin?: string | null;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


matthieu-crouzet (approved):

packages/@ama-sdk/core/src/fwk/errors.ts
```diff
@@ -16,10 +18,10 @@ interface GenericErrorContext {
 export class GenericError extends Error {
 
   constructor(message: string, context?: GenericErrorContext) {
```
kpanot, line 20: Maybe you should turn the `context` field as `public` to have it in the error serialization

packages/@ama-sdk/core/src/fwk/api.helpers.ts
```diff
@@ -115,3 +115,24 @@ export function tokenizeRequestOptions(tokenizedUrl: string, queryParameters: {
...
 }
+
+/**
+ * Remove http/https from the given url
+ * Extract origin part from the given url
+ * If no url return 'unknown' string;
+ *
+ * @param url The string url
+ */
+export function extractInfoFromUrl(url?: string | null) {
```
kpanot, line 126: Would not work to use the ES6 URL object?

```typescript
if (url) {
  const parsedUrl = new URL(url);
  return parsedURL.hostname + parsedURL.pathname;
}

return 'unknown';

File: packages/@ama-sdk/core/src/fwk/api.helpers.spec.ts [hunk 3]
```diff
@@ -0,0 +1,33 @@
+import { extractInfoFromUrl } from './api.helpers';
+
+describe('ExtractInfoFromUrl', () => {
```
kpanot, line 3: your forgot to test your supported Basic Auth URL

packages/@ama-sdk/core/src/fwk/errors.ts
mrednic-1A: I have done some changes from the previous implementation, and it does not seem to be needed to be serialized.

packages/@ama-sdk/core/src/fwk/api.helpers.spec.ts
mrednic-1A: I removed the factory function and the test too. I found another way to get the origin from the `headers` directly

matthieu-crouzet (approved):

kpanot (approved):

pginoux-1A (approved):

PR #400, opened 2023-06-12
feat: add schematic to add a module

kpanot:
## Proposed change

Add a schematics to facilitate the add of a module to an application/library

### Features:

- #401 
The PR add the support of private default registry defined in NPM configuration only

### DIFF
--- a/packages/@o3r/core/README.md
+++ b/packages/@o3r/core/README.md
@@ -81,26 +81,28 @@ Then uncomment the following lines in the `src/styles.scss` file to apply the Ot
 
 Otter library provides a set of code generators based on [angular schematics](https://angular.io/guide/schematics).
 
-| Schematics          | Description                                                                   | How to use                                  |
-| ------------------- | ----------------------------------------------------------------------------- | ------------------------------------------- |
-| add                 | Include Otter in a library / application.                                     | `ng add @o3r/core`                          |
-| page                | Create a new Page in your application.                                        | `ng generate @o3r/core:page`                |
-| service             | Create a new Otter Service in your library / application.                     | `ng generate @o3r/core:service`             |
-| fixture             | Adding functions to an Otter fixture based on a selector and default methods. | `ng generate @o3r/core:fixture`             |
-| component           | Create a new Otter component in your library / application.                   | `ng generate @o3r/core:component`           |
-| component-container | Create a new Otter container component in your library / application.         | `ng generate @o3r/core:component-container` |
-| component-presenter | Create a new Otter presenter component in your library / application.         | `ng generate @o3r/core:component-presenter` |
-| playwright-scenario | Create a new Playwright scenario in your application.                         | `ng generate @o3r/core:playwright-scenario` |
-| store               | Create a new store in your library / application.                             | `ng generate @o3r/core:store`               |
-| store-entity-async  | Create an entity async new store in your library / application.               | `ng generate @o3r/core:store-entity-async`  |
-| store-entity-sync   | Create an entity sync new store in your library / application.                | `ng generate @o3r/core:store-entity-sync`   |
-| store-simple-async  | Create a simple async new store in your library / application.                | `ng generate @o3r/core:store-simple-async`  |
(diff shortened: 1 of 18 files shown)


packages/@o3r/core/README.md
```diff
@@ -81,26 +81,28 @@ Then uncomment the following lines in the `src/styles.scss` file to apply the Ot
...
-| renovate-bot        | Create a basic Renovate Bot                                                   | `ng generate @o3r/core:renovate-bot`        |
-| module              | Create a new Otter compatible module in your monorepo                         | `ng generate @o3r/core:module`              |
-| ng-add-create       | Include `ng add` schematic into your project                                  | `ng generate @o3r/core:ng-add-create`       |
+| Schematics          | Description                                                                   | How to use                 |
+| ------------------- | ----------------------------------------------------------------------------- | -------------------------- |
+| add                 | Include Otter in a library / application.                                     | `ng add @o3r/core`         |
+| page                | Create a new Page in your application.                                        | `ng g page`                |
+| service             | Create a new Otter Service in your library / application.                     | `ng g service`             |
```
fpaul-1A, line 88: is there no risk of conflict with the default Angular generators ?

fpaul-1A (dismissed):

mrednic-1A (dismissed):

packages/@o3r/core/README.md
kpanot: Yes and this is the goal.
Actually `ngCli` is using (and displaying with `ng help`) the generators of the collections listed in angular.json 's [schematicCollections](https://angular.io/guide/workspace-config#angular-cli-configuration-options) config.

The priority is given to the first on of the list in case of conflict and the default Angular generators are provided by de page `@schematics/angular`.
This means that if you do not want to override the default Angular generators you can configure the `angular.json` as following:
```json
{
  "schematicCollections": [
    "@schematics/angular",
    "@o3r/core"
  ]
}
```

In our case we **do want** to override the default Angular schematics with our so we *unshift* the `@o3r/core` collection within this array.

> **note**: The default overriden generators can still be used with it full name: `ng g @schematics/angular:service`

packages/@o3r/core/schematics/add-modules/schema.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "http://json-schema.org/schema",
+  "$id": "NgAddModules",
+  "title": "Add available Otter modules",
+  "description": "ng generate Renovate bot",
```
matthieu-crouzet, line 5: copy paste ?

matthieu-crouzet (dismissed):

packages/@o3r/core/schematics/add-modules/index.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Select the available modules to add to the project
+ *
+ * @param _options
+ */
+export function ngAddModules(_options: NgAddModulesSchematicsSchema): Rule {
+  return async (tree, context) => {
+    if (!context.interactive) {
+      context.logger.error('This command is available only for interactive shell');
+      return;
```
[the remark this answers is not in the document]
cpaulve-1A: so that means we will never chain it with another rule? As far as I understood, you need to return the tree if it has been updated by another rule before

mrednic-1A (approved):

packages/@o3r/schematics/src/modules/modules.constants.ts
```diff
@@ -0,0 +1,35 @@
+import type { PackageJson } from 'type-fest';
+
+/** Package as return by the NPM Registry */
+export type NpmRegistryPackage = Pick<PackageJson, 'name' | 'description' | 'version' | 'keywords'> &
```
matthieu-crouzet, line 4: Why do you pack `name` to override it ?

matthieu-crouzet (approved):

packages/@o3r/schematics/src/modules/modules.constants.ts
kpanot: To enforce the support of the type of packageJson's name type (actually `string | undefined` :D).
It is not exactly overridden, it is extended (to remove the `undefined` part)

PR #407, opened 2023-06-14
fix: changed cascading origin branch

kpanot:
## Proposed change

Changed cascading origin branch to run target branch CI

### DIFF
--- a/apps/github-cascading-app/src/cascading/cascading.ts
+++ b/apps/github-cascading-app/src/cascading/cascading.ts
@@ -364,7 +364,14 @@ ${MESSAGE_NEW_UPDATE} ${baseBranch}
         return;
       }
     } else {
-      await this.createBranch(cascadingBranch, currentBranch.branch);
+      try {
+        await this.createBranch(cascadingBranch, targetBranch.branch);
+        await this.merge(currentBranch.branch, cascadingBranch);
+      } catch {
+        this.logger.warn(`Fail to merge ${currentBranch.branch} into ${cascadingBranch} before creating the PR, will switch to conflict mode (and create the PR from ${currentBranch.branch})`);
+        await this.deleteBranch(cascadingBranch);
+        await this.createBranch(cascadingBranch, currentBranch.branch);
+      }
     }
     await this.createPullRequestWithMessage(cascadingBranch, currentBranch.branch, targetBranch.branch, config);
   }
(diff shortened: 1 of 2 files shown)


apps/github-cascading-app/src/cascading/cascading.ts
```diff
@@ -364,7 +364,14 @@ ${MESSAGE_NEW_UPDATE} ${baseBranch}
         return;
       }
     } else {
-      await this.createBranch(cascadingBranch, currentBranch.branch);
+      try {
+        await this.createBranch(cascadingBranch, targetBranch.branch);
+        await this.merge(currentBranch.branch, cascadingBranch);
+      } catch {
+        this.logger.warn(`Fail to merge ${currentBranch.branch} into ${cascadingBranch} to create the PR, will which to conflict mode`);
```
mrednic-1A, line 371: The message is a bit unclear.
Fail to merge or Fail to create the PR ?

apps/github-cascading-app/src/cascading/cascading.ts
```diff
@@ -364,7 +364,14 @@ ${MESSAGE_NEW_UPDATE} ${baseBranch}
         return;
       }
     } else {
-      await this.createBranch(cascadingBranch, currentBranch.branch);
+      try {
+        await this.createBranch(cascadingBranch, targetBranch.branch);
+        await this.merge(currentBranch.branch, cascadingBranch);
```
pginoux-1A, line 369: image
In case of a 403, delete / create won't make a difference, same for 404 even if it should not happen. In case of 422 it looks just like a retry.We are left with a 409, shouldn't we keep the merge conflict so it's shown in the branch ? Do we want to delete/create in this case ? If I remember correctly we had an issue while create the PR with the branch after the create ref, that the build for the PR wasn't considering the pipeline changes from the target branch.

kpanot: Actually the try/catch is mainly for the 409. The problem is that if this merge fail we will do a PR without any diff, not a config.
The way to be able to create a conflicting PR (I put in place in this PR) is to create the PR from the previous release branch (without merge then). It is creating a conflicting PR that can not run the CI (because conflicting).
The drawback it that according the way the conflict it fix it may run the CI from the previous release branch indeed.
If you think about a better way I will take it :)

pginoux-1A: After discussion with the limits running it on Azure functions, I agree with the changes

PR #414, opened 2023-06-19
fix: add missing package.json in schematics build builder's output

kpanot:
## Proposed change

Add missing package.json in schematics build builder's output

### DIFF
--- a/packages/@o3r/schematics/project.json
+++ b/packages/@o3r/schematics/project.json
@@ -29,7 +29,17 @@
       "executor": "nx:run-script",
       "options": {
         "script": "build:builders"
-      }
+      },
+      "outputs": [
+        "{projectRoot}/dist/package.json",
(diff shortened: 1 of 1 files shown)


packages/@o3r/schematics/project.json
```diff
@@ -29,7 +29,17 @@
       "executor": "nx:run-script",
       "options": {
         "script": "build:builders"
-      }
+      },
+      "outputs": [
+        "{projectRoot}/dist/package.json",
```
matthieu-crouzet, line 34: Shouldn't we have this change for all build:builders ?
because I think we have the same issue for @ama-sdk/schematics

kpanot: Actually no, they are not really touching the package.json usually

matthieu-crouzet (approved):

mrednic-1A (approved):

PR #402, opened 2023-06-12
feat: angular 16

vscaiceanu-1a:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

- yarn audit error: wait for https://github.com/yarnpkg/berry/pull/5501 to be released
- Jest warnings: https://github.com/jestjs/jest/issues/13576#issuecomment-1526477142

PR #425, opened 2023-06-22
feat: base the module generation on Angular cli library generation

kpanot:
Requires https://github.com/AmadeusITGroup/otter/pull/402 to get `@nx/js` as dependency

- [x] Require to be merged #402


kpanot:
Duplicate of #473

PR #427, opened 2023-06-22
fix: correctly cache o3r/schematics schematics

cpaulve-1A:
## Proposed change
Remove incorrect cache output.

Task build-builders is in charge of compiling the typescripts and should only cache javascript files
Task prepare-build-builders does not impact the package.json file

Thus both task can rely on nx.json output

## Related issues

- :bug: Missing files in published @o3r/schematics

### DIFF
--- a/packages/@o3r/schematics/project.json
+++ b/packages/@o3r/schematics/project.json
@@ -29,17 +29,7 @@
       "executor": "nx:run-script",
       "options": {
         "script": "build:builders"
-      },
-      "outputs": [
-        "{projectRoot}/dist/package.json",
-        "{projectRoot}/dist/builders.json",
-        "{projectRoot}/dist/migration.json",
-        "{projectRoot}/dist/collection.json",
-        "{projectRoot}/dist/builders/**/*.json",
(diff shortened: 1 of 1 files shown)


packages/@o3r/schematics/project.json
```diff
@@ -29,17 +29,7 @@
       "executor": "nx:run-script",
       "options": {
         "script": "build:builders"
-      },
-      "outputs": [
```
matthieu-crouzet, line 33: ` "{projectRoot}/dist/package.json",`
 should be added in nx.json for `prepare-build-builders`

kpanot (approved):

packages/@o3r/schematics/project.json
cpaulve-1A: Not really
As mentionned by @kpanot , this is quite unusual the package.json is copied into the dist by the ama-sdk task.
This is an output that is specific by the way of this package is build, the prepare:build-builders do not touch the package.json at all.

matthieu-crouzet (approved):

PR #444, opened 2023-06-27
ci: publish verdaccio state in the PR in case of failure

fpaul-1A:
In case of failure on the it-tests, we'll now publish an artifact containing the storage of verdaccio, so we can download it and run the docker image locally with the same state as the CI

### DIFF
--- a/.verdaccio/README.md
+++ b/.verdaccio/README.md
@@ -1,0 +1,27 @@
+# Verdaccio
+
+Verdaccio is a tool used to publish npm packages in a local registry (https://verdaccio.org/)
+
+You can run the following commands inside this folder
+
+## How to start
+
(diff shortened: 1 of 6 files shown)


.verdaccio/README.md
```diff
@@ -0,0 +1,27 @@
+# Verdaccio
```
kpanot, line 1: This file should be linked in the main readme.md or developer.md to be more easily be found (Cf. Comment from payment user)

fpaul-1A: this README is mostly for when you download the artifact from the PR but I can add a link

matthieu-crouzet (approved):

kpanot (approved):

PR #463, opened 2023-06-30
fix: enforce style to SCSS in create command

kpanot:


fpaul-1A (dismissed):

matthieu-crouzet (approved):

packages/@o3r/create/src/index.ts
```diff
@@ -3,15 +3,15 @@
 import { spawnSync } from 'node:child_process';
 import { join, resolve } from 'node:path';
 import * as minimist from 'minimist';
+import type { PackageJson } from 'type-fest';
+
+const { version } = require(resolve(__dirname, 'package.json')) as PackageJson;
```
vscaiceanu-1a, line 8: are we sure we have a meaningful version in all cases (not 0.0.0)?

kpanot: Not sure to understand the point :S.
- In "normal" case (if `npm create @o3r`) we got the latest version of the `@o3r/create` which is the version we want for the `@o3r/core`
- In "advance" case (if `npm create @o3r@latest`) we got the latest version of the `@o3r/create` which is the version we want for the `@o3r/core`
- In "advance version" case (if `npm create @o3r@~8.3.0`) we got the **8.3.x** version of the `@o3r/create` which is the version we want for the `@o3r/core`
- In "local" case (if `npm run ./dist/index.js` or `npm create file:....`) we got **0.0.0-placeholder** which is the version of the local @o3r/core.

Is there a case I did not think you see that can lead to issue?

matthieu-crouzet (dismissed):

packages/@o3r/core/schematics/ng-add/index.ts
```diff
@@ -19,7 +19,7 @@ class DevInstall extends NodePackageInstallTask {
 
   /** @inheritdoc */
   public toConfiguration() {
-    const installOptions = process.env && process.env.npm_execpath && process.env.npm_execpath.indexOf('yarn') === -1 ? 'npm' : 'yarn';
+    const installOptions = !process.env || !process.env.npm_execpath || process.env.npm_execpath.indexOf('yarn') === -1 ? 'npm' : 'yarn';
```
fpaul-1A, line 22: can't you just:
```suggestion
    const installOptions = process.env?.npm_execpath?.includes('yarn')  ? 'yarn' : 'npm';
```
?

fpaul-1A (dismissed):

fpaul-1A (approved):

cpaulve-1A (approved):

PR #467, opened 2023-07-03
fix: allow renovate post run commands

kpanot:
## Related issues

- :bug: Fixes
  - #465
  - #466

### DIFF
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -10,6 +10,11 @@
     "group:linters",
     "schedule:nonOfficeHours"
   ],
+  "allowedPostUpgradeCommands": [
+    "^yarn install",
+    "^yarn sdks",
+    "^yarn run harmonize:version"
+  ],
   "ignorePaths": [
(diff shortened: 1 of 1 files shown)


matthieu-crouzet (approved):

.renovaterc.json
```diff
@@ -10,6 +10,11 @@
     "group:linters",
     "schedule:nonOfficeHours"
   ],
+  "allowedPostUpgradeCommands": [
```
fpaul-1A, line 13: is this working for github ? 
I thought this was only for the config of the self hosted renovate bot

also, don't you need to add `yarn ng update` and `yarn dlx` as well ?

kpanot: I did not check the doc and has been misled by the linked PR messages.
Actually you are right, this options is only for self-hosted Renovate (cf: https://docs.renovatebot.com/self-hosted-configuration/#allowpostupgradecommandtemplating).
Then I do not know how to fix the current issue :S

PR #473, opened 2023-07-04
feat: base the module generation on Angular cli library generation

kpanot:


packages/@o3r/core/schematics/library/index.spec.ts
```diff
@@ -23,26 +23,39 @@ describe('New module generator', () => {
...
     expect(tree.exists('/packages-test/@my/new-module/collection.json')).toBe(true);
+    expect(tree.exists('/packages-test/@my/new-module/schematics/ng-add/schema.json')).toBe(true);
     expect(tree.exists('/packages-test/@my/new-module/project.json')).toBe(false);
     expect(JSON.parse(tree.readContent('/tsconfig.base.json')).compilerOptions.paths['@my/new-module']).toContain('packages-test/@my/new-module/src/public_api');
     expect(JSON.parse(tree.readContent('/tsconfig.build.json')).compilerOptions.paths['@my/new-module'][0]).toBe('packages-test/@my/new-module/dist');
     expect(tree.files.length).toBeGreaterThan(16);
   });
 
-  describe('in NX monorepo', () => {
+  // Skip relative to nx generator issue
```
matthieu-crouzet, line 46: Maybe we can put a TODO with the reference to the issue 
to not forget to remove the skip when it's implemented

packages/@o3r/core/schematics/library/index.ts
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
matthieu-crouzet, line 1: why do you have to disable it for the whole file ?

packages/@o3r/core/schematics/library/rules/rules.ng.ts
```diff
@@ -0,0 +1,153 @@
...
+        eslintRcPath: findConfigFileRelativePath(tree, ['.eslintrc.json', '.eslintrc.js'], options.targetPath)
+      }),
+      renameTemplateFiles(),
+      move(options.targetPath)
+    ]);
+
+    return chain([
+      mergeWith(templateNg, MergeStrategy.Overwrite),
+      (t, _c) => {
+        /* eslint-disable @typescript-eslint/naming-convention */
```
matthieu-crouzet, line 43: in what this line doesnt follow the namingConvention ?

packages/@o3r/core/schematics/library/index.ts
kpanot: No idea why it is here, it is even not needed :S

packages/@o3r/core/schematics/library/rules/rules.ng.ts
```diff
@@ -0,0 +1,153 @@
...
+   */
+  const updateNgTemplate: Rule = (tree, context) => {
+    const o3rCorePackageJsonPath = path.resolve(__dirname, '..', '..', '..', 'package.json');
+    const o3rCorePackageJson: PackageJson & { generatorDependencies?: Record<string, string> } = JSON.parse(readFileSync(o3rCorePackageJsonPath)!.toString());
+    const otterVersion = o3rCorePackageJson.dependencies!['@o3r/schematics'];
+
+    const templateNg = apply(url('./templates/ng'), [
+      template({
+        ...options,
+        runner: process.env.npm_execpath && /[\\/][^\\/]yarn[^\\/]js$/.test(process.env.npm_execpath) ? 'yarn run' : 'npm run',
```
matthieu-crouzet, line 29: again a new method to decided between `yarn` and `npm`

packages/@o3r/core/schematics/library/rules/rules.ng.ts
```diff
@@ -0,0 +1,153 @@
...
+    ])(tree, context);
+  };
+
+  /**
+   * Update the root tsconfig files mappings
+   *
+   * @param tree File tree
+   * @param context Context of the schematics
+   */
+  const updateTsonfigFiles: Rule = (tree, context) => {
```
matthieu-crouzet, line 111:
```suggestion
  const updateTsConfigFiles: Rule = (tree, context) => {
```

packages/@o3r/core/schematics/library/rules/rules.nx.ts
```diff
@@ -0,0 +1,183 @@
...
+        runner: process.env?.npm_execpath?.includes('yarn') ? 'yarn run' : 'npm run'
+      }),
+      renameTemplateFiles(),
+      move(targetPath)
+    ]);
+    rules.push(mergeWith(templateNx, MergeStrategy.Overwrite));
+
+    return chain([
+      ...rules,
+      (t, _c) => {
```
matthieu-crouzet, line 104: can't we share this rule with the one for angular ?
it seems to be the same code

packages/@o3r/core/schematics/library/rules/rules.nx.ts
```diff
@@ -0,0 +1,183 @@
...
+          'jest-preset-angular': o3rCorePackageJson.generatorDependencies!['jest-preset-angular'],
+          'rxjs': o3rCorePackageJson.peerDependencies!.rxjs,
+          'typescript': o3rCorePackageJson.peerDependencies!.typescript,
+          'zone.js': o3rCorePackageJson.generatorDependencies!['zone.js']
+        };
+        /* eslint-enable @typescript-eslint/naming-convention */
+        t.overwrite(path.posix.join(targetPath, 'package.json'), JSON.stringify(packageJson, null, 2));
+        return t;
+      },
+      (t, _c) => {
```
matthieu-crouzet, line 158: same for this one

packages/@o3r/core/schematics/library/schema.ts
```diff
@@ -8,7 +8,7 @@ export interface NgGenerateModuleSchema extends JsonObject {
   path: string;
 
   /** Description of the new module */
-  description: string;
+  description: string | null;
```
matthieu-crouzet, line 11: could it be null, as default is `""`

packages/@o3r/core/schematics/library/index.spec.ts
```diff
@@ -23,26 +23,40 @@ describe('New module generator', () => {
   });
 
   it('should generate the minimum mandatory files', async () => {
+    initialTree.create('angular.json', '{"version": 1, "projects": {} }');
+    initialTree.create('package.json', '{ "version": "0.0.0-test" }');
     const runner = new SchematicTestRunner('schematics', collectionPath);
-    const tree = await runner.runExternalSchematic('schematics', 'module', {
+    const angularPackageJson = require.resolve('@schematics/angular/package.json');
+    runner.registerCollection('@schematics/angular', path.resolve(path.dirname(angularPackageJson), require(angularPackageJson).schematics));
```
fpaul-1A, line 30: Shouldn't we mock the collection from `@schematics/angular` in the unit tests instead of calling the actual thing ?

packages/@o3r/core/schematics/library/rules/shared.ts
```diff
@@ -0,0 +1,87 @@
...
+ */
+export function updatePackageDependenciesFactory(
+    targetPath: string,
+    otterVersion: string,
+    o3rCorePackageJson: PackageJson & { generatorDependencies?: Record<string, string>},
+    options: NgGenerateModuleSchema): Rule {
+  return (tree) => {
+    /* eslint-disable @typescript-eslint/naming-convention */
+    const packageJson = tree.readJson(path.posix.join(targetPath, 'package.json')) as PackageJson;
+    const runner = process.env.npm_execpath && /[\\/][^\\/]yarn[^\\/]js$/.test(process.env.npm_execpath) ? 'yarn run' : 'npm run';
```
fpaul-1A, line 22: `process.env?.npm_execpath?.includes('yarn')` was not enough for this one ?

packages/@o3r/core/schematics/library/index.spec.ts
kpanot: I think about it at first and I realized that if we do not actually test that the update is done on top of the code generated by the generator in dep, this can break on any dependency update without breaking the UTs.
Maybe it is not really UT anymore but I think it make more sense in the end 😄

fpaul-1A: also, shouldn't we read this information from `angular.json` / `nx.json` ?

packages/@o3r/core/schematics/library/rules/shared.ts
fpaul-1A: also, shouldn't we read this information from `angular.json` / `nx.json` ?

packages/@o3r/core/schematics/library/index.spec.ts
fpaul-1A: this should break the IT tests in that case ?
the UT should only break if we change the code of the thing being tested I think

kpanot: That's true for pure UT, but I don't think it make sense for generators.
Let imagine we mock here:
- we will need to redo the mocks for every minor upgrade of the angular generator dep (because the code generated can change on minor versions) and find a mechanism to keep them up to date.
- It will only guaranty that there is no change in the generator input options but do not guaranty the output is correct regarding the version of the generator is based on.

As I said, I agree this is not pure UT but it I am not convinced it make sense to have some in that case.
If we want pure UTs not relying on used generator, I even think it would be better to not provide UT at all (regarding the benefice/maintenance cost) and rely on e2e only.

packages/@o3r/core/README.md
```diff
@@ -76,7 +76,7 @@ Otter framework provides a set of code generators based on [angular schematics](
 | store-simple-sync          | Create a simple sync new store in your library / application.                 | `ng g store-simple-sync`          |
 | store-action               | Create an action into an existing store.                                      | `ng g store-action`               |
 | renovate-bot               | Create a basic Renovate Bot                                                   | `ng g renovate-bot`               |
-| module                     | Create a new Otter compatible module in your monorepo                         | `ng g module`                     |
+| library                    | Add a new Otter Module to the current project                                 | `ng g library`                    |
```
mrednic-1A, line 79: IMO this PR should target `9.0`. Otherwise we will be with `module` in 9.0 and `library` in 9.1.
And it's more a fix then a feature.

packages/@o3r/core/schematics/add-modules/schema.json
```diff
@@ -4,5 +4,15 @@
   "title": "Add available Otter modules",
   "description": "ng generate Add module",
   "properties": {
+    "preset": {
+      "type": "string",
+      "description": "Preset of module list to automatically install",
+      "default": "",
+      "enum": [
+        "",
```
matthieu-crouzet, line 12: could be none instead of empty string to be clearer ?

packages/@o3r/core/schematics/library/index.spec.ts
```diff
@@ -23,26 +23,40 @@ describe('New module generator', () => {
...
+    expect(tree.exists('/packages-test/@my/new-module/schematics/ng-add/schema.json')).toBe(true);
     expect(tree.exists('/packages-test/@my/new-module/project.json')).toBe(false);
     expect(JSON.parse(tree.readContent('/tsconfig.base.json')).compilerOptions.paths['@my/new-module']).toContain('packages-test/@my/new-module/src/public_api');
     expect(JSON.parse(tree.readContent('/tsconfig.build.json')).compilerOptions.paths['@my/new-module'][0]).toBe('packages-test/@my/new-module/dist');
     expect(tree.files.length).toBeGreaterThan(16);
   });
 
-  describe('in NX monorepo', () => {
+  // Skip relative to nx generator issue
+  // TODO: re-enable when Nx Alias conflict is fixed
```
vscaiceanu-1a, line 47: we can mention https://github.com/AmadeusITGroup/otter/issues/491 here

packages/@o3r/core/schematics/library/index.ts
```diff
@@ -0,0 +1,38 @@
...
+import { NgGenerateModuleSchema } from './schema';
+import { NodePackageInstallTask } from '@angular-devkit/schematics/tasks';
+import { nxGenerateModule } from './rules/rules.nx';
+import { ngGenerateModule } from './rules/rules.ng';
+
+/**
+ * Add an Otter compatible module to a monorepo
+ *
+ * @param options
+ * @param rootPath
```
vscaiceanu-1a, line 13: there is no `rootPath` param

packages/@o3r/core/schematics/library/rules/rules.ng.ts
```diff
@@ -0,0 +1,96 @@
...
+import { apply, chain, externalSchematic, MergeStrategy, mergeWith, move, renameTemplateFiles, Rule, template, url } from '@angular-devkit/schematics';
+import * as path from 'node:path';
+import { readFileSync } from 'node:fs';
+import { updateNgPackagrFactory, updatePackageDependenciesFactory } from './shared';
+
+/**
+ * generate the rules to adapt the library generated by ng cli
+ *
+ * @param options
+ * @param rootPath
```
vscaiceanu-1a, line 13: `options.targetPath`

packages/@o3r/core/schematics/library/rules/shared.ts
```diff
@@ -0,0 +1,88 @@
...
+      '@angular/cli': packageJson.peerDependencies['@angular/common'],
+      '@angular/common': packageJson.peerDependencies['@angular/common'],
+      '@angular/compiler': packageJson.peerDependencies['@angular/common'],
+      '@angular/compiler-cli': packageJson.peerDependencies['@angular/common'],
+      '@angular/core': packageJson.peerDependencies['@angular/common'],
+      '@angular/platform-browser': packageJson.peerDependencies['@angular/common'],
+      '@angular/platform-browser-dynamic': packageJson.peerDependencies['@angular/common'],
+      '@o3r/dev-tools': otterVersion,
+      '@o3r/eslint-plugin': otterVersion,
+      '@schematics/angular': o3rCorePackageJson.peerDependencies!['@angular-devkit/core'],
```
vscaiceanu-1a, line 51:
```suggestion
      '@schematics/angular': o3rCorePackageJson.peerDependencies!['@schematics/angular'],
```

or we don't have this peerdep? Anyway, I am not sure `@schematics` and `@angular-devkit` are aligned.

packages/@o3r/core/schematics/ng-add/schema.json
```diff
@@ -1,9 +1,26 @@
...
+      "enum": [
+        "basic",
+        "cms"
+      ],
+      "x-prompt": {
+        "message": "Which preset to use to start your application?",
+        "type": "list",
+        "items": [
+          { "value": "basic", "label": "BASIC - Minimum plugin list to install for a basic Otter application (no additional modules)" },
+          { "value": "cms", "label": "CMS - Add the Otter modules allowing the full administration of the application via the CMS" }
```
vscaiceanu-1a, line 20: should we link a MD from our readme for further details about what the CMS stack contains?

packages/@o3r/core/schematics/ng-add/schema.ts
```diff
@@ -1,6 +1,11 @@
 import type { JsonObject } from '@angular-devkit/core';
 
+export type PresetNames = 'basic' | 'cms';
+
 export interface NgAddSchematicsSchema extends JsonObject {
+  /** reset of module list to automatically install */
```
vscaiceanu-1a, line 6:
```suggestion
  /** Preset of module list to automatically install */
```

packages/@o3r/core/schematics/add-modules/schema.json
kpanot: Not sure, it is convenient in the way to have a nullish default value

packages/@o3r/core/schematics/ng-add/schema.json
kpanot: I will do it but not sure the links in select are working in all consoles

packages/@o3r/core/schematics/library/index.spec.ts
kpanot: Not related to this issue.
The problem is that instead of calling the `library` schematic of `@anguler/schamtics`, Nx is calling its alias `lib`.
`lib` is an alias a different schematics which result to an error.
I did not put links to issue here because I did not find any issue open on there side.

packages/@o3r/core/README.md
```diff
@@ -86,3 +86,24 @@ Otter framework provides a set of code generators based on [angular schematics](
...
+
+Will add all the modules allowing the allowing the full managing of an application by the CMS.
+It includes the following modules:
+
+* @o3r/localization
+* @o3r/styling
+* @o3r/components
+* @o3r/configuration
+* @o3r/dynamic-content
+* @o3r/rules-engin
```
mrednic-1A, line 109: rules-engin**e**

fpaul-1A (approved):

packages/@o3r/schematics/src/utility/package-manager-runner.ts
```diff
@@ -1,39 +1,64 @@
...
+}
+
 /**
  * Get the Package Manager
  *
  * @param enforcedNpmManager package manager to enforce
  */
-export function getPackageManagerName(enforcedNpmManager?: SupportedPackageManagers | '' | undefined | null): SupportedPackageManagers {
-  return enforcedNpmManager || (process.env?.npm_execpath?.includes('yarn') ? 'yarn' : 'npm');
+function getPackageManagerName(enforcedNpmManager?: SupportedPackageManagers): SupportedPackageManagers {
```
mrednic-1A, line 21: `export` the function to stay not breaking

packages/@o3r/schematics/src/utility/package-manager-runner.ts
```diff
@@ -1,39 +1,64 @@
...
- * @param angularJsonString Content of angular.json file
+ * @param options.workspaceConfig Workspace configuration
+ * @param options.enforcedNpmManager  package manager to enforce, will be used if supported
+ * @param options
  */
-export function getPackageManager(angularJsonString?: string | null) {
-  let packageManager = process.env && process.env.npm_execpath && process.env.npm_execpath.indexOf('yarn') === -1 ? 'npm' : 'yarn';
-  if (angularJsonString) {
-    const angularJsonObj = commentJson.parse(angularJsonString) as any;
+export function getPackageManager(options?: {workspaceConfig?: WorkspaceSchema | string | null; enforcedNpmManager?: string | null}) {
```
mrednic-1A, line 35: this is a bit breaking. Maybe rename it and deprecate the other.
`getUsedPackageManager` maybe ?

kpanot: Not breaking if committed before 9.0.0

packages/@o3r/schematics/src/utility/package-manager-runner.ts
kpanot: Not breaking if committed before 9.0.0

packages/@o3r/schematics/src/utility/package-manager-runner.ts
mrednic-1A: ok :) I did not check when was added :)

mrednic-1A (approved):

fpaul-1A (approved):

PR #477, opened 2023-07-04
fix: build of generation missign files and update doc

pginoux-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


packages/@ama-sdk/generator-sdk/src/generators/shell/index.ts
```diff
@@ -36,13 +36,13 @@ export default class extends SdkGenerator {
     this.desc(`${deprecatedMessage}\nGenerate the shell of an SDK package`);
 
     this.option('projectName', {
-      description: 'Project name (NPM package scope)',
+      description: 'Project name (NPM package scope, package.json name will be @{projectName}/{packageName})',
```
cpaulve-1A, line 39: this package will be deprecated

pginoux-1A: Indeed, but updating the description to have something more relevant doesn't cost much :D

PR #485, opened 2023-07-05
fix: update ng add @ama-sdk/schematics to update swagger-ignore file

cpaulve-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


kpanot (dismissed):

mrednic-1A (dismissed):

packages/@ama-sdk/schematics/package.json
```diff
@@ -43,6 +43,7 @@
     "@angular-devkit/schematics": "~16.0.5",
     "@angular-devkit/schematics-cli": "~16.0.5",
     "@angular/cli": "~16.0.5",
+    "@o3r/schematics": "workspace:^",
```
kpanot, line 46: Should be optional.
It is used only for `ng-add` and it can make sense in a Otter project context but a bit less in an SDK.

matthieu-crouzet (approved):

mrednic-1A (approved):

PR #495, opened 2023-07-06
feat: o3r linter on his own ng add

mrednic-1A:
## Proposed change

o3r linter packages handled by ng add o3r/eslint-config-otter and not form ng add o3r/core anymore


packages/@o3r/core/schematics/ng-add/utils/linter/index.ts
```diff
@@ -0,0 +1,25 @@
...
+  let useOtterLinter = false;
+  try {
+    require.resolve(`${linterPackageName}/package.json`);
+    if (context.interactive) {
+      useOtterLinter = await askConfirmation(`You already have eslint installed. Would you like to add otter config rules for eslint ?
+Otherwise, you can add them later via this command: ng add @o3r/eslint-config-otter`, true);
+      if (!useOtterLinter) {
+        context.logger.info(`
+          You can add otter linter config rules later to the project via this command:
+          ng add @o3r/eslint-config-otter
```
matthieu-crouzet, line 15: we already say it just above 
do we really need to say it again ?

packages/@o3r/core/schematics/ng-add/utils/linter/index.ts
```diff
@@ -0,0 +1,25 @@
...
+        `);
+      }
+    }
+
+  } catch {
+    context.logger.debug('eslint package not installed. Skipping otter linter phase !');
```
matthieu-crouzet, line 21: if there is no `eslint` we can propose them to install it with `@o3r/eslint-config-otter` no ?

mrednic-1A: This is only a debug log. I did not want to display something to the normal user.

packages/@o3r/core/schematics/ng-add/utils/linter/index.ts
mrednic-1A: Indeed I can remove the check and this log. It will be simpler

cpaulve-1A (approved):

PR #503, opened 2023-07-10
refactor: breaking-change: modification of the schema for styling extractor

matthieu-crouzet:
## Proposed change
⚠️ Breaking change ⚠️ 
Modification of the schema for the styling extractor

### DIFF
--- a/packages/@o3r/styling/schemas/style.metadata.schema.json
+++ b/packages/@o3r/styling/schemas/style.metadata.schema.json
@@ -3,10 +3,14 @@
   "$id": "StyleMetadataSchema",
   "description": "Schema of Application style",
   "type": "object",
-  "additionalProperties": {
-    "$ref": "#/definitions/StyleMetadataValueSchema"
+  "properties": {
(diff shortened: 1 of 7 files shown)


mrednic-1A (approved):

pginoux-1A (approved):

packages/@o3r/styling/schemas/style.metadata.schema.json
```diff
@@ -3,10 +3,14 @@
   "$id": "StyleMetadataSchema",
   "description": "Schema of Application style",
   "type": "object",
-  "additionalProperties": {
```
kpanot, line 6: Lets turn it to `false` then :)

kpanot (approved):

packages/@o3r/styling/schemas/style.metadata.schema.json
matthieu-crouzet: As discussed will be done later

PR #518, opened 2023-07-13
fix: add Nx In Angular upgrade PR

kpanot:


.renovaterc.json
```diff
@@ -106,10 +106,10 @@
         "^@schematics",
         "angular",
         "^@ngrx",
-        "^zone.js"
-      ],
-      "excludePackagePatterns": [
-        "^@nrwl"
+        "^zone.js",
+        "^@nrwl",
+        "^@nX",
```
fpaul-1A, line 111: is it case sensitive ? the scope seems to be `@nx` in our package.json, same for `nx`

kpanot: oops, it is not case sensitive but it's a typo :'(

mrednic-1A (approved):

fpaul-1A (approved):

PR #519, opened 2023-07-13
fix: Use tagname for mat-option in fixtures

### LINKED ISSUE #520: [Bug]: Autocomplete fixtures do not work with MDC
### Package name

testing

### Package version

8.0.x

### Reproduction steps

Try to test an autocomplete in fixtures

### Current result

You get a timeout as the selector used is incorrect for mdc

### Expected result

The selector should be correct

### Additional comments

The tagname allows to target both MDC and legacy

rglearns:
## Proposed change

Using mat option tagname instead of class as the class changed with MDC 

## Related issues

- :bug: Fixes #520 
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/testing/src/core/playwright/angular-materials/autocomplete-material.ts
+++ b/packages/@o3r/testing/src/core/playwright/angular-materials/autocomplete-material.ts
@@ -16,7 +16,7 @@ export class MatAutocomplete extends O3rElement implements MatAutocompleteProfil
     await this.setValue(value);
     await this.click();
     const element = this.sourceElement.page;
-    const matOptions = element.locator('.mat-option');
+    const matOptions = element.locator('.mat-option, mat-option');
     await matOptions.first().waitFor({state: 'attached', timeout});
     const matOptionsCount = await matOptions.count();
     const options: (string | undefined)[] = [];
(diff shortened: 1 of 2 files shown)


kpanot (approved):

packages/@o3r/testing/src/core/playwright/angular-materials/autocomplete-material.ts
```diff
@@ -16,7 +16,7 @@ export class MatAutocomplete extends O3rElement implements MatAutocompleteProfil
     await this.setValue(value);
     await this.click();
     const element = this.sourceElement.page;
-    const matOptions = element.locator('.mat-option');
+    const matOptions = element.locator('mat-option');
```
matthieu-crouzet, line 19: Shouldn't we put `.mat-option, mat-option` to not break the previous behaviour ?

rglearns: The tag would select both mdc and legacy and imo is more future proof as classes can change with each angular release

matthieu-crouzet: as we are on a released version

rglearns: done

kpanot (approved):

fpaul-1A (approved):

PR #521, opened 2023-07-13
ci: run it-tests on affected files only

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/.nxignore
+++ b/.nxignore
@@ -2,3 +2,4 @@
 packages/@*/*/{builders,schematics}/**/templates/**/*.ts
 packages/@*/*/{builders,schematics}/**/mocks/**/*.ts
 packages/@*/*/build/**
+!dist/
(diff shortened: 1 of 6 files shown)


mrednic-1A (dismissed):

.nxignore
```diff
@@ -2,3 +2,4 @@
 packages/@*/*/{builders,schematics}/**/templates/**/*.ts
 packages/@*/*/{builders,schematics}/**/mocks/**/*.ts
 packages/@*/*/build/**
+!dist/
```
kpanot, line 5: :'(

matthieu-crouzet:
if we do a modification in `@o3r/localization` for example, `@o3r/core` will not be part of affected but in fact we have one it-test in `@o3r/core` which rely on it

fpaul-1A:
I am closing this for now
To be checked when https://github.com/AmadeusITGroup/otter/pull/766 will be done, if the split makes it easier to identify the actual dependencies of the it-tests

PR #529, opened 2023-07-17
fix: sdk generator - date support and doc

pginoux-1A:
## Proposed change

* Fix a bug preventing to disable utils.Date usage using the vendor extension
* Add documentation associated

### DIFF
--- a/packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/java/com/amadeus/codegen/ts/AbstractTypeScriptClientCodegen.java
+++ b/packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/java/com/amadeus/codegen/ts/AbstractTypeScriptClientCodegen.java
@@ -523,15 +523,11 @@ public abstract class AbstractTypeScriptClientCodegen extends DefaultCodegen imp
         + "list container.");
     }
 
-    // If the x-date-timezone is present in the timestamp, we use Date instead of utils.Date
+    // If the x-date-timezone is present in the specification, replace Date with utils.Date
     if (property.vendorExtensions.containsKey("x-date-timezone")) {
-      for (CodegenProperty modelVariable : model.allVars) {
-        if (modelVariable.baseName != null && modelVariable.baseName.toString().equals("timestamp")) {
-          modelVariable.dataType = "Date";
-          modelVariable.datatypeWithEnum = "Date";
-          modelVariable.baseType = "Date";
-        }
-      }
+      property.dataType = "Date";
+      property.datatypeWithEnum = "Date";
+      property.baseType = "Date";
     }
     if (model.name.indexOf("_allOf") > -1 && model.discriminator != null) {
       property.vendorExtensions.put("x-exposed-classname", model.classname.replace("AllOf", ""));
--- a/packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/readme.md
+++ b/packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/readme.md
@@ -109,11 +109,95 @@ You can build and run UT with:
 yarn test
 ```
 
-## Vendor extensions
+## Manage dates
 
-These are some of the available configurations using vendor extensions in the swagger spec
+### The timezone issue
+Managing dates with timezones has always been a bit painful in front end applications.
+Let's give a concrete example to understand the problem:
+An API returns the date and hour of your flight in the timezone of the airport location. In our use case, let's say the departure airport is on GTM+7 : 2023-07-10T00:37:00.000+07:00.
+The timezone sent is the one from the airport, here GMT+7. If you just use the Date(), the computer browser will convert this in its own timezone.
+For example, if the user is in GMT+2 you will end up displaying the following: 2023-07-09T19:37:00.000+02:00.
+This is not what you want. You want the exact date time of the flight at the airport timezone, not the one of your user's computer.
+However, there might be cases where you might still need the timezone information.
+For example, you want to be able to display that the flight is in X hours.
+You will need to compute this information with the two timezones -- the airport's and the user's.
 
-### Manage timezone in date
+### Solution proposed to remove the timezone: utils.DateTime
+We have introduced the utils.Date object to replace the Date implementation and ignore the timezone.
+As we need to get rid of the timezones more often than not, this will be our default behavior.
+By default, the Date models are replaced with utils.Date.
+
+When we need to keep the timezone information, we create a new field directly in the SDK.
+As this field does not exist in the specification, it will not be part of the base model but of the core model instead.
+
+Simple example:
+```yaml
+  Flight:
+    type: "object"
+    required:
+      - departureDateTime
+    properties:
+      departureDateTime:
+        type: string
+        format: date-time
+```
+Base model generated
+```typescript
+// flight.ts generated in base models
+export interface Flight {
+  /** @see utils.DateTime */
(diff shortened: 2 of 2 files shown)


packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/readme.md
```diff
@@ -109,11 +109,88 @@ You can build and run UT with:
 yarn test
 ```
 
-## Vendor extensions
+## Manage dates
+
+### The timezone issue
+Managing dates with timezones have always been a bit painful in front end applications.
```
cpaulve-1A, line 115:
```suggestion
Managing dates with timezones has always been a bit painful in front end applications.
```

packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/readme.md
```diff
@@ -109,11 +109,88 @@ You can build and run UT with:
...
+Let's give a concrete example to understand the problem:
+Let's say you have an API that returns you the date and hour of your flight : 2023-07-10T00:37:00.000+07:00
+The timezone sent is the one from the airport, here GMT+7. If you just use the Date(), the computer browser will convert this in its own timezone.
+You will end up having for example if the user is in GMT+2 : 2023-07-09T19:37:00.000+02:00.
+You don't want that, you want the exact date time of the flight, at the airport timezone, not the user one.
+However, sometimes you might need this timezone information, for example to be able to display to the user that the flight is in X hours (need both airport and user timezone).
+
+### Solution proposed: utils.DateTime
```
kpanot, line 123: The title may not be matching the content of the doc.
Actually the main purpose of the document under this section is to provide an example to get the date considering timezone.
Maybe it would make sense to move this example in a new section or to change the title of this one?

packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/java/com/amadeus/codegen/ts/AbstractTypeScriptClientCodegen.java
```diff
@@ -525,13 +525,9 @@ public void postProcessModelProperty(CodegenModel model, CodegenProperty propert
 
     // If the x-date-timezone is present in the timestamp, we use Date instead of utils.Date
```
matthieu-crouzet, line 526: Maybe we could change the comment as it's not based for timestamp only now

packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/readme.md
pginoux-1A: The purpose of this section is to get rid of the timezone, keeping the timezone if the next one, I'll update the titles to be more accurate

packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/readme.md
```diff
@@ -109,11 +109,88 @@ You can build and run UT with:
 yarn test
 ```
 
-## Vendor extensions
+## Manage dates
+
+### The timezone issue
+Managing dates with timezones has always been a bit painful in front end applications.
+Let's give a concrete example to understand the problem:
+In the case of an API returning the date and hour of your flight : 2023-07-10T00:37:00.000+07:00
```
cpaulve-1A, line 117:
```suggestion
An API returns the date and hour of your flight in the timezone of the airport location. In our use case, let's say the departure airport is on GTM+7 : 2023-07-10T00:37:00.000+07:00.
```

packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/readme.md
```diff
@@ -109,11 +109,88 @@ You can build and run UT with:
...
+The timezone sent is the one from the airport, here GMT+7. If you just use the Date(), the computer browser will convert this in its own timezone.
+You will end up having for example if the user is in GMT+2 : 2023-07-09T19:37:00.000+02:00.
+You don't want that, you want the exact date time of the flight, at the airport timezone, not the user one.
```
cpaulve-1A, line 120:
```suggestion
You don't want that, you want the exact date time of the flight at the airport timezone, not the time of your user's computer when the flight will depart.
```

packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/readme.md
```diff
@@ -109,11 +109,88 @@ You can build and run UT with:
...
+However, sometimes you might need this timezone information, for example to be able to display to the user that the flight is in X hours (need both airport and user timezone).
```
cpaulve-1A, line 121:
```suggestion
However, there might be cases where you might still need this timezone information. 
For example, you want to be able to display that the flight is in X hours. You will need to compute this information with the two timezones -- the airport's and the user's.
```

sdo-1A, line 121:
```suggestion
However, you may sometimes need this timezone information, for example to be able to indicate to the user that the flight is in X hours (both the airport and the user timezones are needed).
```

packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/readme.md
```diff
@@ -109,11 +109,88 @@ You can build and run UT with:
...
+
+### Solution proposed to remove the timezone: utils.DateTime
+That's why the default Date behavior is to get rid of the timezone, by replacing Date by utils.Date.
```
cpaulve-1A, line 124:
```suggestion
As we need to get rid the timezones more often than not, this will be our default behaviour. This is why, we introduced the utils.Date object to replace Date in those use cases.
```

sdo-1A: Date or DateTime ?

cpaulve-1A (dismissed):

packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/readme.md
```diff
@@ -109,11 +109,88 @@ You can build and run UT with:
...
+```typescript
+export * from './flight/index';
+```
 
-These are some of the available configurations using vendor extensions in the swagger spec
+You can now use departureDateTimeConsideringTimezone to access the timezone information.
+See [utils.Date](https://github.com/AmadeusITGroup/otter/blob/main/packages/%40ama-sdk/core/src/fwk/date.ts) for more information.
 
-### Manage timezone in date
+### How to keep the timezone (not use utils.Datetime)
```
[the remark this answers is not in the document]
sdo-1A: utils.Datetime or utils.Date ?

packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/readme.md
kpanot, line 1: This documentation is provided only to new SDK, it means that the solution documented is not provided/accessible to previously generated owner.
Maybe it should also be provided to the Otter repository documentation (in `docs/` or on the `@ama-sdk/core` readme)

pginoux-1A: For the sdk-core doc, they are a few things that needs to be updated, I will tackle it a dedicated PR. But yes that's a good point.

cpaulve-1A (approved):

Issue #534, opened 2023-07-18
[Bug]: state not syncing with storage for feature modules


Abhirocks889:
### Package name

@o3r

### Package version

9.0.0-rc

### Reproduction steps

When attempting to sync the state of a store inside a feature module using the `syncStorage` method provided by `o3r`, it fails to sync the state. It works perfectly fine when performing the same in the app module using `StoreModule.forRoot`.

```ts
export function storageSyncReducer(reducer: ActionReducer<any>): ActionReducer<any> {
  return syncStorage ({keys: ['todos'], rehydrate: true})(reducer);
}
const metaReducers: Array<MetaReducer<any, any>> = [storageSyncReducer];

@NgModule({
  imports: [
    BrowserModule,
    StoreModule.forFeature(
        'feature',
        featureReducer,
        {metaReducers}
    )
  ]
})
export class MyFeatureModule {}
```

### Current result

The state is not synced with the browser storage.

### Expected result

The state should sync with the browser storage and rehydrate whenever the browser tab is refreshed.

### Additional comments

There should be a way of specifying that the store sync is for a feature store. Adding a new flag like `forFeature` and then doing the following should sync it correctly:

```ts
export function storageSyncReducer(reducer: ActionReducer<any>): ActionReducer<any> {
  return syncStorage ({keys: ['todos'], rehydrate: true, forFeature: true})(reducer);
}
const metaReducers: Array<MetaReducer<any, any>> = [storageSyncReducer];

@NgModule({
  imports: [
    BrowserModule,
    StoreModule.forFeature(
        'feature',
        featureReducer,
        {metaReducers}
    )
  ]
})
export cl

PR #537, opened 2023-07-18
fix: state not syncing with storage for feature modules

Abhirocks889:
## Proposed change

This PR introduces a fix for the issue: state not syncing with storage for feature modules

## Related issues

- :bug: Fixes #(https://github.com/AmadeusITGroup/otter/issues/534)

## Description

The basic difference between syncing a state in the root app module via StoreModule.forRoot versus in a feature through StoreModule.forFeature is the fact that in the former, the syncStateUpdate method receives the entire State where each feature is described by key-value pairs (eg: {cart: {items: []} checkout: {methods: []}}), whereas in the latter, the syncStateUpdate method receives the respective feature slice state (eg: {items: []} for cart and {methods: []} for checkout).

PFA the screenshots describing the same:

**Syncing in the app module**

image

**Syncing in the feature module**

image

Therefore, the latter scenario was not handled as it always looked for the key specified in the config and fetch the corresponding state for the key.

The fix introduces a new boolean property to the `StorageConfig` called `forFeature` which when enabled, handles the state syncing for feature stores.

> Note: PFA a stackblitz that illustrates the behavior after the fix: https://stackblitz.com/edit/angular-ivy-bdu1sx?file=README.md


matthieu-crouzet (approved):

kpanot:
I think you failed your rebase :S
image

Abhirocks889:
> I think you failed your rebase :S image

Yes massively :(

kpanot:
Duplicate of #576

PR #538, opened 2023-07-18
fix: remove eslint installed by default

mrednic-1A:


packages/@o3r/core/schematics/ng-add/index.ts
```diff
@@ -33,6 +33,24 @@ class DevInstall extends NodePackageInstallTask {
...
+ *
+ * @param tree
+ * @returns package manager as string or undefined
+ */
+function getPackageManagerFromTree(tree: Tree) {
+  let packageManager;
+  const angularJsonString = tree.read('/angular.json')?.toString();
+  if (angularJsonString) {
+    const angularJsonObj = JSON.parse(angularJsonString);
+    if (angularJsonObj?.cli?.packageManager) {
```
fpaul-1A, line 47: we could do the same check on `nx.json` or do we want to handle Nx support later ?

packages/@o3r/core/schematics/ng-add/index.ts
```diff
@@ -33,6 +33,24 @@ class DevInstall extends NodePackageInstallTask {
   }
 }
 
+/**
+ * Retrive the package manager from the angular.json
+ *
+ * @param tree
+ * @returns package manager as string or undefined
+ */
+function getPackageManagerFromTree(tree: Tree) {
```
fpaul-1A, line 42: this is very similar to the code here https://github.com/AmadeusITGroup/otter/blob/main/packages/%40o3r/schematics/src/utility/package-manager-runner.ts#L21, I suppose you don't reuse because it's needed to install `@o3r/schematics` ?

packages/@o3r/core/schematics/ng-add/index.ts
mrednic-1A: from what I have seen, if we run in nx context, we have an `angular.json` in the tree built from nx.json and project.json.
I have checked how the package manager is declared in nx.json and it is still under cli.package.manager same as in angular.json. So that's why this code should be compatible with nx too.
@kpanot any thoughts ?

packages/@o3r/core/schematics/ng-add/index.ts
mrednic-1A: Indeed. It's the case for the class above.

packages/@o3r/core/schematics/ng-add/index.ts
```diff
@@ -19,7 +19,7 @@ class DevInstall extends NodePackageInstallTask {
 
   /** @inheritdoc */
   public toConfiguration() {
-    const installOptions = process.env?.npm_execpath?.includes('yarn')  ? 'yarn' : 'npm';
+    const installOptions = this.packageManager || process.env?.npm_execpath?.includes('yarn') ? 'yarn' : 'npm';
```
kpanot, line 22: Can use the getPackageManaager() from `@o3r/schematics`

packages/@o3r/core/schematics/ng-add/index.ts
```diff
@@ -43,8 +61,10 @@ export function ngAdd(options: NgAddSchematicsSchema): Rule {
     const corePackageJsonContent = JSON.parse(fs.readFileSync(path.resolve(__dirname, '..', '..', 'package.json'), {encoding: 'utf-8'})) as PackageJson;
     const o3rCoreVersion = corePackageJsonContent.version ? `@${corePackageJsonContent.version}` : '';
     const schematicsDependencies = ['@o3r/dev-tools', '@o3r/schematics'];
+    const packageManager = getPackageManagerFromTree(tree);
     for (const dependency of schematicsDependencies) {
       context.addTask(new DevInstall({
+        packageManager,
```
kpanot, line 67: You should use the one exposed by `@o3r/schematics` that determine the packageManager automatically

mrednic-1A: At this moment 'o3r/schematics' is not installed. We call the `DevInstall` constructor to install schematics and dev-tools. 
I find it a bit strange because `o3r/schematics` is a dep for `o3r/core` and 'o3r/dev-tools' is a dep for 'o3r/schematics'. I think it is for 'pnp' that needs to have them explicitly in `package.json`.
@cpaulve-1A Not sure if this sounds familiar to you :)

kpanot: `o3r/schematics` is in prod dep of `o3r/core`, it should be installed :S

mrednic-1A: Indeed. I gave it a try.

fpaul-1A (approved):

matthieu-crouzet (approved):

packages/@o3r/core/schematics/ng-add/index.ts
mrednic-1A: finally done in https://github.com/AmadeusITGroup/otter/pull/473

packages/@o3r/core/schematics/ng-add/index.ts
mrednic-1A: finally done in https://github.com/AmadeusITGroup/otter/pull/473

packages/@o3r/core/schematics/ng-add/index.ts
mrednic-1A: finally done in https://github.com/AmadeusITGroup/otter/pull/473

packages/@o3r/core/schematics/ng-add/index.ts
mrednic-1A: finally done in https://github.com/AmadeusITGroup/otter/pull/473

PR #543, opened 2023-07-18
fix: sdk generator - fix build without spec or with partial spec

pginoux-1A:
## Proposed change
* Add default exports in files to ensure that the sdk will build just after the shell generation and with a small spec


packages/@ama-sdk/create/src/index.it.spec.ts
```diff
@@ -26,7 +26,7 @@ const execAppOptions: ExecSyncOptions = {
 };
 const registry = 'http://localhost:4873';
 
-const packageManager = process.env.TEST_PACKAGE_MANAGER || 'npm';
```
pginoux-1A, line 29: I've put yarn as default here for the moment, waiting for the PR that fixes shell for npm

mrednic-1A (approved):

packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/src/models/base/index.ts
```diff
@@ -0,0 +1 @@
+export type { }
```
cpaulve-1A, line 1: missing semi-colon

sdo-1A (approved):

cpaulve-1A (approved):

PR #560, opened 2023-07-20
feat: use angular schematic to generate page

matthieu-crouzet:
## Proposed change
Use angular schematic to generate page


fpaul-1A (dismissed):

packages/@o3r/core/tsconfig.builders.json
```diff
@@ -11,7 +11,7 @@
   "include": [
     "builders/**/*.ts",
     "schematics/**/*.ts"
-  ],
+, "schematics/page/templates/index.ts.template"  ],
```
mrednic-1A, line 14: is this needed/working ? :D

matthieu-crouzet: oups automatic change
I removed it

fpaul-1A (approved):

PR #562, opened 2023-07-20
o3r create workspace

mrednic-1A:
## Proposed change
- o3r/create will build a monorepo skeleton

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/core/schematics/ng-add/project-setup/index.ts
+++ b/packages/@o3r/core/schematics/ng-add/project-setup/index.ts
@@ -1,9 +1,10 @@
-import { chain, noop, SchematicContext, Tree } from '@angular-devkit/schematics';
+import { chain, noop, Rule, SchematicContext, Tree } from '@angular-devkit/schematics';
 import { NodeDependencyType } from '@schematics/angular/utility/dependencies';
 import * as fs from 'node:fs';
 import * as path from 'node:path';
 import { PackageJson } from 'type-fest';
 import {
+  addWorkspacesToProject,
   createAzurePipeline,
   generateRenovateConfig,
   o3rBasicUpdates,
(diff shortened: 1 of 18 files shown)


packages/@o3r/core/schematics/rule-factories/o3r-basic-updates/index.ts
```diff
@@ -9,7 +9,7 @@ import { ngAddPackages, readAngularJson, readPackageJson } from '@o3r/schematics
  * @param o3rCoreVersion
  * @param projectType
  */
-export function o3rBasicUpdates(pName: string | null, o3rCoreVersion?: string, projectType?: 'application' | 'library'): Rule {
+export function o3rBasicUpdates(pName: string | null, o3rCoreVersion?: string, projectType?: 'application' | 'library' | 'monorepo'): Rule {
```
kpanot, line 12: Why not considering `undefined` as monorepo case, as the type is mandatory otherwise?

packages/@o3r/core/schematics/rule-factories/monorepo/index.ts
```diff
@@ -0,0 +1,33 @@
+import { Rule, SchematicContext, SchematicsException, Tree } from '@angular-devkit/schematics';
+import * as commentJson from 'comment-json';
+import { PackageJson } from 'type-fest';
```
kpanot, line 3: `import type`

packages/@o3r/core/schematics/rule-factories/monorepo/index.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Update root package.json to include workspaces
+ *
+ * @param patterns workspaces patterns to include in package.json
+ */
+export function addWorkspacesToProject(patterns = ['apps/*', 'libs/*']): Rule {
```
kpanot, line 10: Maybe they should be set in constants (as you need to add them in `application`, `sdk` and `library` config in the angular.json)

packages/@o3r/core/schematics/rule-factories/monorepo/index.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Update root package.json to include workspaces
+ *
+ * @param patterns workspaces patterns to include in package.json
+ */
+// eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+export function addWorkspacesToProject(patterns = [`${APPLICATIONS_FOLDER_NAME}/*`, `${LIBRARIES_FOLDER_NAME}/*`]): Rule {
+
+  return (tree: Tree, _context: SchematicContext) => {
+
+    // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
```
matthieu-crouzet, line 17: there is no template expressions in the next line

packages/@o3r/core/schematics/rule-factories/monorepo/index.ts
```diff
@@ -0,0 +1,36 @@
+import type { Rule, SchematicContext, Tree } from '@angular-devkit/schematics';
+import { SchematicsException } from '@angular-devkit/schematics';
+import * as commentJson from 'comment-json';
```
matthieu-crouzet, line 3: we should try to avoid using comment-json in schematics @kpanot has open a PR to remove it

packages/@o3r/core/schematics/ng-add/project-setup/index.ts
```diff
@@ -40,54 +41,69 @@ export const prepareProject = (options: NgAddSchematicsSchema) => async (tree: T
...
   } = await import('@o3r/schematics');
   const installOtterLinter = await shouldOtterLinterBeInstalled(context);
-  const workspaceProject = tree.exists('angular.json') ? getProjectFromTree(tree) : undefined;
-  const projectType = workspaceProject?.projectType || 'application';
+  const workspaceProject = tree.exists('angular.json') ? getProjectFromTree(tree, options.projectName) : undefined;
+  const projectType = workspaceProject?.projectType;
   const internalPackagesToInstallWithNgAdd = Array.from(new Set([
     ...(projectType === 'application' ? ['@o3r/application'] : []),
-    ...(options.enableStyling ? ['@o3r/styling'] : []),
+    ...(options.enableApisManager && !!projectType ? ['@o3r/apis-manager'] : []),
```
kpanot, line 48: All these options could have been removed actually

packages/@o3r/core/schematics/rule-factories/monorepo/index.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Update root package.json to include workspaces
+ *
+ * @param patterns workspaces patterns to include in package.json
+ */
+// eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+export function addWorkspacesToProject(patterns = [`${APPLICATIONS_FOLDER_NAME}/*`, `${LIBRARIES_FOLDER_NAME}/*`]): Rule {
+
+  return (tree: Tree, _context: SchematicContext) => {
+
+    const rootPackageJson = tree.read('/package.json');
```
kpanot, line 16: can use `tree.readJson('/package.json') as PackageJson ` instead

packages/@o3r/core/schematics/rule-factories/monorepo/index.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (!rootPackageJson) {
```
kpanot, line 17: `tree.exists` would be cleaner

packages/@o3r/core/schematics/rule-factories/monorepo/index.ts
```diff
@@ -0,0 +1,34 @@
...
+      throw new SchematicsException('Root package.json does not exist');
+    }
+
+    const rootPackageJsonObject = JSON.parse(rootPackageJson.toString()) as PackageJson;
+
+    rootPackageJsonObject.workspaces ||= [];
+    patterns.forEach(pattern => {
```
kpanot, line 24: or
```typescript
rootPackageJsonObject.workspaces = [...new Set(patterns.concat(rootPackageJsonObject.workspaces))];
```

packages/@o3r/core/schematics/rule-factories/o3r-basic-updates/index.ts
```diff
@@ -11,37 +12,50 @@ import { ngAddPackages, readAngularJson, readPackageJson } from '@o3r/schematics
...
+
+    tree.overwrite(packageJsonPath, JSON.stringify(packageJsonObj, null, 2));
+  };
+
   const updatePackageJson = (tree: Tree, _context: SchematicContext) => {
     const workspace = readAngularJson(tree);
 
+    if (!projectType) { // at the root of the monorepo
+      const rootPackageJsonPath = '/package.json';
+      const rootPackageJsonString = tree.read(rootPackageJsonPath)?.toString();
```
kpanot, line 30: can use `tree.readJson`

packages/@o3r/core/schematics/rule-factories/o3r-basic-updates/index.ts
```diff
@@ -11,37 +12,50 @@ import { ngAddPackages, readAngularJson, readPackageJson } from '@o3r/schematics
...
+        ngUpdateScript(tree, packageJson, `${workspaceProject.root}/package.json`);
       });
     return tree;
   };
 
   return async (tree: Tree, context: SchematicContext) => {
     // eslint-disable-next-line @typescript-eslint/naming-convention
     const { NodeDependencyType } = await import('@schematics/angular/utility/dependencies');
-
+    const packagesToAddByProjectType = projectType ? ['@ama-sdk/core'] : ['@ama-sdk/schematics'];
```
kpanot, line 49: Not sure you need to add `@ama-sdk/core` in case of package

packages/@o3r/core/schematics/rule-factories/store/index.ts
```diff
@@ -29,25 +28,27 @@ const ngrxRouterStoreDevToolDep = '@ngrx/store-devtools';
  * @param options @see RuleFactory.options
  * @param rootPath @see RuleFactory.rootPath
  * @param options.projectName
- * @param _rootPath
+ * @param projectType
  */
-export function updateStore(options: { projectName: string | null | undefined }, _rootPath: string): Rule {
+export function updateStore(options: { projectName: string | undefined}, projectType?: 'application' | 'library'): Rule {
```
kpanot, line 33: the type `'application' | 'library'` can be replace by `WorkspaceProject['projectType']`

packages/@o3r/schematics/src/utility/monorepo.ts
```diff
@@ -25,3 +25,9 @@ export function findConfigFileRelativePath(tree: Tree, files: string[], originPa
 export function isNxContext(tree: Tree) {
   return tree.exists('/nx.json');
 }
+
+/** Name of the folder where libraries will be generated inside monorepo */
+export const LIBRARIES_FOLDER_NAME = 'libs' as const;
```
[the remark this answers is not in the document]
mrednic-1A: ok

packages/@o3r/core/schematics/rule-factories/o3r-basic-updates/index.ts
mrednic-1A: we use it for utils.datetime so I think we can add it

packages/@o3r/core/schematics/ng-add/project-setup/index.ts
mrednic-1A: don't you want to wait and remove them only when we implement the dynamic presets ?

packages/@o3r/schematics/src/utility/monorepo.ts
```diff
@@ -25,3 +25,9 @@ export function findConfigFileRelativePath(tree: Tree, files: string[], originPa
 export function isNxContext(tree: Tree) {
   return tree.exists('/nx.json');
 }
+
+/** Default name of the folder where libraries will be generated inside monorepo */
+export const LIBRARIES_FOLDER_NAME = 'libs';
+/** Default name of the folder where applications will be generated inside monorepo */
+export const APPLICATIONS_FOLDER_NAME = 'apps';
+
```
kpanot, line 33: would be valuable to provide a function to retrieve the folder name as well.
Something like:

```typescript
export const DEFAULT_ROOT_FOLDERS: WorkspaceLayout = {
  libsDir: LIBRARIES_FOLDER_NAME,
  appsDir: APPLICATIONS_FOLDER_NAME
};
const BASE_ROOT_FOLDERS_MAP: Record<WorkspaceProject['projectType'], keyof WorkspaceLayout> = {
  library: 'libsDir',
  application: 'appsDir'
};

// not in this file :)
export function getSchematicOptions<T extends SchematicOptions['*:*'] = SchematicOptions['*:*']>(config: WorkspaceConfig, schematicName?: string): T | undefined {
  if (!schematicName) {
    return config.schematics?.['*:*'];
  }
  const options = config.schematics && Object.entries(config.schematics)
    .filter(([name]) => new RegExp(name.replace(/\*/g, '.*')).test(schematicName))
    .sort(([a], [b]) => ((a.match(/\*/g)?.length || 0) - (b.match(/\*/g)?.length || 0)))
    .reduce((acc, [, opts]) => ({ ...acc, ...opts }), {} as any);
  return options && Object.keys(options).length ? options : undefined;
}

export function getPackagesBaseRootFolder(config: WorkspaceConfig, tree: Tree, projectType?: WorkspaceProject['projectType'], schematicName?: string) {
  const configName = projectType && BASE_ROOT_FOLDERS_MAP[projectType];
  const nxExplicitDir = configName && isNxContext(tree) && tree.readJson('/nx.json').workspaceLoyout?.[configName];
  const schematicConfigDir = configName && getSchematicConfig(config, schematicName)?.[configName];
  return '/' + (schematicConfigDir || nxExplicitDir || (projectType && DEFAULT_ROOT_FOLDERS[projectType]) || '');
}
```

mrednic-1A: updated

packages/@o3r/core/schematics/ng-add/project-setup/index.ts
```diff
@@ -40,54 +41,71 @@ export const prepareProject = (options: NgAddSchematicsSchema) => async (tree: T
...
+      updateOtterEnvironmentAdapter(options, coreSchematicsFolder),
+      updateStore(options, projectType),
+      options.enableCustomization && projectType === 'application' ? updateCustomizationEnvironment(coreSchematicsFolder, o3rCoreVersion, options, false) : noop,
+      options.enablePlaywright && projectType === 'application' ? updatePlaywright(coreSchematicsFolder, options) : noop,
+      projectType === 'application' ? updateAdditionalModules(options, coreSchematicsFolder) : noop,
+      removePackages(packagesToRemove)
+    ];
+  } else {
+    monorepoRules = [
+      generateRenovateConfig(coreSchematicsFolder),
```
cpaulve-1A, line 87: no renovate on the applications?

packages/@o3r/core/schematics/rule-factories/o3r-basic-updates/index.ts
```diff
@@ -9,39 +11,51 @@ import { ngAddPackages, readAngularJson, readPackageJson } from '@o3r/schematics
...
+export function o3rBasicUpdates(pName: string | null | undefined, o3rCoreVersion?: string, projectType?: WorkspaceProject['projectType']): Rule {
+
+  const ngUpdateScript = (tree: Tree, packageJsonObj: PackageJson, packageJsonPath = '/package.json') => {
+    if (!packageJsonObj.scripts) {
+      packageJsonObj.scripts = {};
+    }
+
+    packageJsonObj.scripts['update:otter'] = 'ng update @o3r/core';
+
+    tree.overwrite(packageJsonPath, JSON.stringify(packageJsonObj, null, 2));
```
cpaulve-1A, line 23: you don't need to return the tree?

packages/@o3r/core/schematics/ng-add/project-setup/index.ts
```diff
@@ -40,54 +41,71 @@ export const prepareProject = (options: NgAddSchematicsSchema) => async (tree: T
...
+    const angularMajorVersion = angularVersion?.match(simplifiedSemVerRegexp)?.[1];
+    const ngxPrefetchVersion = angularMajorVersion ? `^${angularMajorVersion}.0.0` : undefined;
+    externalPackagesToInstallWithNgAdd['@o3r/ngx-prefetch'] = ngxPrefetchVersion;
+  }
+  const type = projectType === 'library' ? NodeDependencyType.Peer : NodeDependencyType.Default;
+  let appLibRules: Rule[] = [];
+  let monorepoRules: Rule[] = [];
+  if (projectType) {// this means we run the ng add in a lib/app inside the workspace
+    appLibRules = [
+      updateImports(mapImportV7toV8, renamedPackagesV7toV8) as any,
```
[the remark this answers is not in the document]
mrednic-1A: I had the same question to the team and we decided to keep this until we drop the support of v7, because we can have users migrating from v7 to v9+.

packages/@o3r/core/schematics/ng-add/project-setup/index.ts
mrednic-1A: This is tricky good point. The reco what we'll put in place will be to have a always  monorepo even if you have only one app. This will prepare the things in case you want to add a new app or a new lib (the structure will be ready). For this we will setup renovate at monorepo level.
But if we run the ng add on an existing repository which is only an app for instance, we should add it there too. 
So I have to identify somehow that I am in app but not in a monorepo :S.
cc: @kpanot

packages/@o3r/schematics/src/utility/monorepo.ts
```diff
@@ -25,3 +29,49 @@ export function findConfigFileRelativePath(tree: Tree, files: string[], originPa
 export function isNxContext(tree: Tree) {
   return tree.exists('/nx.json');
 }
+
+/**
+ * Determine if we are in an monorepo context
+ *
+ * @param tree
+ */
+export function isMonorepoContext(tree: Tree) {
```
kpanot, line 38: This is a bit unclear.
Does "monorepo" means "workspace" in this context? Then you should check `nx.json` or `angular.json`.
Does "monorepo" means "multi packages"? then checking the `workspaces` field make sense but not checking the `nx.json`

mrednic-1A: My intention was to detect multipackages. 
It's what I need if I want to apply changes from `ng add o3r core` at project root level. 
So the root rules will be applied when I run the `o3r create`, but also when I run the `ng add o3r core` in a project which I detect that it is not a multipackage.
So I'll stay with checking `workspaces` filed in package.json and rename the helper function.

kpanot: then you can remove the `isNxContext(tree) || `

mrednic-1A: it's what I have done indeed :)

kpanot (approved):

matthieu-crouzet (approved):

Issue #491, opened 2023-07-06
[Bug]: 


matthieu-crouzet:
### Package name

analytics, configuration, core, localization, rules-engine, testing

### Package version

9.0.0-alpha

### Reproduction steps

nx generate does not support schematic with nullable boolean option

### Current result

throw an error if we run the schematic with a null value

### Expected result

should execute correctly the schematics

kpanot:
Actually Nx does support only nullable option for boolean, but not for other types (including `string`).
I did following for:
- turn non-mandatory fields to `| null` to ensure null case validation
- removed non mandatory default value (kept `"default": ""` to keep `string` type when mandatory by the schematic implementation, mainly for `path` options)

PR #567, opened 2023-07-21
fix: support nx for schematics

matthieu-crouzet:
## Proposed change
Remove nullable and default null value
When no default value provided we receive undefined instead of null but json type tell it's null
so instead of comparing with null I check that `typeof myNullableOption !== 'boolean'`

## Related issues

- :bug: Fixes https://github.com/AmadeusITGroup/otter/issues/491


packages/@o3r/schematics/src/rule-factories/ng-add/index.ts
```diff
@@ -152,7 +151,7 @@ export function registerPackageCollectionSchematics(packageJson: PackageJson, an
  * @param schematicsDefaultParams default params to setup by schematic
  * @param angularJsonFile Path to the Angular.json file. Will use the workspace root's angular.json if not specified
  */
-export function setupSchematicsDefaultParams(schematicsDefaultParams: Record<string, Record<string, JsonValue>>, angularJsonFile?: string): Rule {
+export function setupSchematicsDefaultParams(schematicsDefaultParams: Record<string, Record<string, any>>, angularJsonFile?: string): Rule {
```
kpanot, line 154: Cleaner to be `JsonValue | undefined`

kpanot:
As discuss, maybe good to provide an interface (in `@o3r/schematics`) for all `schema.ts`.

packages/@o3r/schematics/src/rule-factories/ng-add/index.ts
```diff
@@ -7,6 +6,7 @@ import { lastValueFrom } from 'rxjs';
 import type { PackageJson } from 'type-fest';
 import type { NgAddPackageOptions } from '../../tasks/index';
 import { getExternalDependenciesVersionRange, getNodeDependencyList, getPackageManager, readAngularJson, registerCollectionSchematics, writeAngularJson } from '../../utility/index';
+import { SchematicOptionObject } from '../../interfaces';
```
sdo-1A, line 9: import order

sdo-1A (approved):

PR #569, opened 2023-07-21
fix: rollback to initial cascading strategy to avoid wrong conflict r…

kpanot:
fix: rollback to initial cascading strategy to avoid wrong conflict resolution
feat: add custom branch prefix name


kpanot:
As discussed, this PR is redirected to 9.1 via the PR #577

PR #570, opened 2023-07-21
fix: export AssetPathOverrideStoreModule to allow path override in ap…

cpaulve-1A:
…plications

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


packages/@o3r/dynamic-content/src/stores/index.ts
```diff
@@ -3,5 +3,6 @@ export * from './asset-path-override/asset-path-override.module';
 export * from './asset-path-override/asset-path-override.reducer';
```
kpanot, line 3: Should not export `./asset-path-overrive/index` instead? (And export the files from this barrel)?

kpanot (approved):

matthieu-crouzet (approved):

PR #576, opened 2023-07-24
fix: state not syncing with storage for feature modules

Abhirocks889:
## Proposed change

This PR introduces a fix for the issue: state not syncing with storage for feature modules

## Related issues

- :bug: Fixes #(https://github.com/AmadeusITGroup/otter/issues/534)

## Description

The basic difference between syncing a state in the root app module via StoreModule.forRoot versus in a feature through StoreModule.forFeature is the fact that in the former, the syncStateUpdate method receives the entire State where each feature is described by key-value pairs (eg: {cart: {items: []} checkout: {methods: []}}), whereas in the latter, the syncStateUpdate method receives the respective feature slice state (eg: {items: []} for cart and {methods: []} for checkout).

PFA the screenshots describing the same:

**Syncing in the app module**

image

**Syncing in the feature module**

image

Therefore, the latter scenario was not handled as it always looked for the key specified in the config and fetch the corresponding state for the key.

The fix introduces a new boolean property to the `StorageConfig` called `forFeature` which when enabled, handles the state syncing for feature stores.

> Note: PFA a stackblitz that illustrates the behavior after the fix: https://stackblitz.com/edit/angular-ivy-bdu1sx?file=README.md

### DIFF
--- a/packages/@o3r/store-sync/src/sync-storage/interfaces.ts
+++ b/packages/@o3r/store-sync/src/sync-storage/interfaces.ts
@@ -16,6 +16,8 @@ export interface SyncStorageSyncOptions {
   filter?: string[];
   /** Spacing for serializer */
   space?: string | number;
+  /** Defines whether the store is loaded as forFeature */
+  syncForFeature?: boolean;
 }
 
 /**
--- a/packages/@o3r/store-sync/src/sync-storage/storage-sync.ts
+++ b/packages/@o3r/store-sync/src/sync-storage/storage-sync.ts
@@ -74,9 +74,12 @@ export const rehydrateApplicationState = (
     let reviver = restoreDates ? dateReviver : dummyReviver;
     let deserialize: ((arg0: string) => any) | undefined;
     let decrypt: ((arg0: string) => string) | undefined;
+    let syncForFeature = false;
 
     if (typeof key === 'object') {
       key = Object.keys(key)[0];
+      syncForFeature = !!(curr?.[key] as SyncStorageSyncOptions)?.syncForFeature;
+
       // use the custom reviver function
       if (typeof curr[key] === 'function') {
         reviver = curr[key];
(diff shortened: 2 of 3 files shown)


packages/@o3r/store-sync/src/sync-storage/storage-sync.ts
```diff
@@ -74,9 +74,12 @@ export const rehydrateApplicationState = (
     let reviver = restoreDates ? dateReviver : dummyReviver;
     let deserialize: ((arg0: string) => any) | undefined;
     let decrypt: ((arg0: string) => string) | undefined;
+    let syncForFeature = false;
```
kpanot, line 77: You removed your previous `forFeature` option at `StorageConfig` and replace it by a static default value `false`.
Is it because having a default value different from `false` does not make sense?

packages/@o3r/store-sync/src/sync-storage/interfaces.ts
```diff
@@ -16,6 +16,8 @@ export interface SyncStorageSyncOptions {
   filter?: string[];
   /** Spacing for serializer */
   space?: string | number;
+  /** Defines whether the state is be synced at feature level*/
```
kpanot, line 19: I think the comment is a bit unclear regarding the purpose of this configuration.
I would suggest something like `/** Defines whether the store is loaded as forFeature */`

Abhirocks889: Changed the comment as suggested.

packages/@o3r/store-sync/src/sync-storage/storage-sync.ts
Abhirocks889: My understanding is that at the moment, every DES fork including RefX is syncing the stores at the root level and if I keep the default as true and they start using this method to sync the store instead of the ngrx-store-localstorage, they will encounter issues. So, basically to maintain backward compatibility, I prefer to keep the default as false.

Abhirocks889: So, essentially the practice has been to register the meta reducers at the root level in the app module. So, keeping the default as true would not work. So, to make it backwards compatible I believe we can keep the default as false.

kpanot (dismissed):

matthieu-crouzet (approved):

kpanot (approved):

Issue #583, opened 2023-07-26
[Feature]: Improve configuration extraction


matthieu-crouzet:
### Context

Currently when a default value of a property cannot be extracted the whole configuration is excluded.
Extraction is not failing, only a warning is logged.

### Proposal

Exclude only the property without default value.
Add a extractor option to throw an error if default value cannot be extracted.

PR #584, opened 2023-07-26
feat: #583: exclude only incompatible properties and not the full config

matthieu-crouzet:
## Proposed change

exclude only incompatible properties and not the full config

## Related issues
- :rocket: Feature #583

### DIFF
--- a/packages/@o3r/components/builders/component-extractor/helpers/component/component.extractor.ts
+++ b/packages/@o3r/components/builders/component-extractor/helpers/component/component.extractor.ts
@@ -4,7 +4,7 @@ import type {
   ComponentConfigOutput,
   ComponentModuleOutput,
   ComponentOutput,
-  ComponentStructure, PlaceholdersMetadata
+  ComponentStructure, ConfigProperty, PlaceholdersMetadata
 } from '@o3r/components';
 import { CmsMedataData, getLibraryCmsMetadata } from '@o3r/extractors';
 import * as fs from 'node:fs';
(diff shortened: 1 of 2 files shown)


kpanot (dismissed):

packages/@o3r/components/builders/component-extractor/helpers/component/component.extractor.ts
```diff
@@ -263,20 +264,44 @@ export class ComponentExtractor {
...
-        return false;
+        return acc;
       }
 
-      const propertiesWithoutDefaultValue = config.properties.filter((property) => property.values === undefined && property.value === undefined);
+      const { propertiesWithDefaultValue, propertiesWithoutDefaultValue } = config.properties.reduce((properties: {
+        propertiesWithDefaultValue: ConfigProperty[];
+        propertiesWithoutDefaultValue: ConfigProperty[];
+      }, property) => {
+        if (property.values === undefined && property.value === undefined) {
```
fpaul-1A, line 277: I assume that `undefined` cannot be a legitimate default value ?

fpaul-1A (approved):

packages/@o3r/components/builders/component-extractor/helpers/component/component.extractor.ts
matthieu-crouzet: Yes configuration value cannot be `undefined`

sdo-1A (approved):

PR #605, opened 2023-07-31
feat(schematics): generate service fixtures only for available frameworks

vscaiceanu-1a:
## Proposed change

The service generator should generate fixtures for Jest or Jasmine only if the test framework is used in the application / library.


packages/@o3r/core/schematics/service/index.ts
```diff
@@ -54,26 +54,45 @@ export function ngGenerateService(options: NgGenerateServiceSchematicsSchema): R
...
+      template(templateData),
       renameTemplateFiles(),
       move(destination)
     ]);
+
+    const rules = [mergeWith(baseTemplateSource, MergeStrategy.Overwrite)];
+
+    ['jest', 'jasmine-core'].forEach(testFramework => {
+      try {
+        require.resolve(`${testFramework}/package.json`);
```
matthieu-crouzet, line 78: Could be good to add test case 
even if it's not testable for now because of [this jest issue](https://github.com/jestjs/jest/issues/9543)

packages/@o3r/core/schematics/service/index.ts
```diff
@@ -54,26 +54,45 @@ export function ngGenerateService(options: NgGenerateServiceSchematicsSchema): R
...
+
+    const baseTemplateSource = apply(url('./templates/base'), [
+      template(templateData),
       renameTemplateFiles(),
       move(destination)
     ]);
+
+    const rules = [mergeWith(baseTemplateSource, MergeStrategy.Overwrite)];
+
+    ['jest', 'jasmine-core'].forEach(testFramework => {
```
kpanot, line 76: it is the second time we use this way to determine the test system, maybe we can provide an helper for it in `@o3r/schematics`:

```typescript
export type AvailableTestFrameworsks = 'jest' | 'jasmine';

const testPackageMaps: Record<AvailableTestFrameworsks, string[]> = {
  'jest': ['jest'],
  'jasmine': ['jasmine-core']
}

export function getTestFramework(workspaceOptions?: WorkspaceOptions): AvailableTestFrameworsks | undefined {
  if (workspaceOptions?.testFramework) {
    return workspaceOptions.testFramework;
  }
  return Object.entries(testPackageMaps)
    .find(([_, deps]) => {
      deps.some((dep) => {
        try {
          return !!require.resolve(`${dep}/package.json`);
        } catch {
          return false;
        }
      });
    })?.[0];
}

Comment (vscaiceanu-1a) [reply packages/@o3r/core/schematics/service/index.ts:76]: Could you please point where we are using it the first time? I couldn't find it anywhere.

Comment (kpanot) [reply packages/@o3r/core/schematics/service/index.ts:76]: I guess in `library` and `application` generators (and it will be needed in `component`, `services` and `store` :))

Comment (vscaiceanu-1a) [reply packages/@o3r/core/schematics/service/index.ts:76]: I am pretty sure this is not used anywhere else yet.
On top of that:
1. your solution returns only one occurrence. As we discussed, to remain backward compatible with our private library, we should generate fixtures for both frameworks if the 2 are installed in the project.
2. `testFramework` does not exist on `WorkspaceLayout`

**Review by fpaul-1A — APPROVED** [review]

File: packages/@o3r/core/schematics/service/index.ts (hunk 2 above)
Comment (vscaiceanu-1a) [reply packages/@o3r/core/schematics/service/index.ts:76]: the helper function will come with https://github.com/AmadeusITGroup/otter/pull/613

File: packages/@o3r/core/schematics/service/index.spec.ts [hunk 3]
```diff
@@ -38,6 +38,21 @@ describe('Service generator', () => {
...
+    const runner = new SchematicTestRunner('schematics', collectionPath);
+    const tree = await runner.runExternalSchematic('schematics', 'service', {
+      projectName: 'test-project',
+      name: 'test-service',
+      featureName: 'test-base',
+      path: 'src/services'
+    }, initialTree);
+
     expect(tree.files.filter((file) => /test-service/.test(file)).length).toEqual(12);
     expect(tree.files.some((file) => /^[\\/]?src[\\/]services[\\/]test-service[\\/]test-base[\\/]test-service\.test-base\.module\.ts$/i.test(file))).toBeTruthy();
```
matthieu-crouzet, line 57: expectation is not the good one I think

cpaulve-1A (approved):

kpanot (approved):

packages/@o3r/core/schematics/service/index.spec.ts
vscaiceanu-1a: when the test will be re-enabled it will be the proof that it works correctly (we will have to adapt the no of files)

PR #612, opened 2023-08-01
ci: increase node memory size for unit tests

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/.github/workflows/code-check.yml
+++ b/.github/workflows/code-check.yml
@@ -46,6 +46,7 @@ jobs:
       - name: Test
         env:
           testCmd: ${{ inputs.affected && format('test:affected --base=remotes/origin/{0}', github.base_ref || github.ref_name) || 'test'}}
+          NODE_OPTIONS: ${{ runner.os == 'Windows' && '--max_old_space_size=4096' || '' }}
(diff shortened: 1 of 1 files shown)


matthieu-crouzet (dismissed):

.github/workflows/code-check.yml
```diff
@@ -21,6 +21,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     env:
       NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
+      NODE_OPTIONS: "--max_old_space_size=4096"
```
kpanot, line 24: Is it needed for `yarn test` command only?
(if so, maybe it should be set only for the step `Test`)

Is it for both OS?
(if so, the set of this EnvVar can be conditional)

fpaul-1A: updated for only `yarn test`, only Windows

kpanot (approved):

matthieu-crouzet (approved):

PR #620, opened 2023-08-02
fix: disable nx cloud for release branches

pginoux-1A:
## Proposed change
disable nx cache on release branches, to prevent any issues on published packaged linked to input/output of nx configuration (force the failure instead)
It's sad but it seems that there is no clean way to reuse env variable defined in the main workflow in reusable ones (cf: https://github.com/orgs/community/discussions/26671#discussioncomment-3931817)

### DIFF
--- a/.github/workflows/code-check.yml
+++ b/.github/workflows/code-check.yml
@@ -8,10 +8,19 @@ on:
         required: false
         default: false
         description: Run test only on affective packages if true
+      skipNxCache:
+        type: boolean
+        default: false
+        required: false
+        description: Skip the nx cache
     secrets:
       NX_CLOUD_ACCESS_TOKEN:
         required: false
         description: Token to use Nx Cloud token
+
+env:
+  NX_SKIP_NX_CACHE: ${{ inputs.skipNxCache }}
+
 jobs:
   test:
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,6 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
+      NX_SKIP_NX_CACHE: ${{ github.event_name == 'push' && startsWith(github.ref, 'release') }}
     steps:
       - uses: actions/checkout@v3
       - uses: ./tools/github-actions/setup
@@ -59,12 +60,15 @@ jobs:
       NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
     with:
       affected: ${{ github.event_name == 'pull_request' }}
+      skipNxCache: ${{ github.event_name == 'push' && startsWith(github.ref, 'release') }}
 
   test-publish:
     uses: ./.github/workflows/test-publish.yml
     secrets:
       NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
     needs: [build]
+    with:
+      skipNxCache: ${{ github.event_name == 'push' && startsWith(github.ref, 'release') }}
 
   publish-packages:
     uses: ./.github/workflows/publish.yml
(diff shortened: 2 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       contents: read
     runs-on: ubuntu-latest
     env:
-      NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
+      NX_CLOUD_ACCESS_TOKEN: ${{ env.NX_CLOUD_ACCESS_TOKEN }}
```
matthieu-crouzet, line 15:
```suggestion
      NX_CLOUD_ACCESS_TOKEN: ${{ github.event_name == 'push' && github.ref != 'main' && '' || secrets.NX_CLOUD_ACCESS_TOKEN }}
```
Because this workflow is not a `workflow_call`

matthieu-crouzet: maybe for this one we don't need to provide the `NX_CLOUD_ACCESS_TOKEN`
and it will be enough to disable nx cloud for releases as we are in the `release` workflow

.github/workflows/main.yml
```diff
@@ -16,11 +16,13 @@ on:
   merge_group:
     types: [checks_requested]
 
+env:
+  NX_CLOUD_ACCESS_TOKEN: ${{ github.event_name == 'push' && github.ref != 'main' && '' || secrets.NX_CLOUD_ACCESS_TOKEN }}
```
matthieu-crouzet, line 20: this will always return `secrets.NX_CLOUD_ACCESS_TOKEN`
because `''` is falsy

pginoux-1A: indeed

.github/workflows/release.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       contents: read
     runs-on: ubuntu-latest
     env:
-      NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
+      NX_CLOUD_ACCESS_TOKEN: ${{ !(github.event_name == 'push' && github.ref != 'main') && secrets.NX_CLOUD_ACCESS_TOKEN || '' }}
```
matthieu-crouzet, line 15: it will always be `''` because `github.event_name` can only be `release` if think 
as the workflow is run `on` that specific event (see line 4)

.github/workflows/main.yml
```diff
@@ -16,11 +16,13 @@ on:
   merge_group:
     types: [checks_requested]
 
+env:
+  NX_CLOUD_ACCESS_TOKEN: ${{ !(github.event_name == 'push' && github.ref != 'main') && secrets.NX_CLOUD_ACCESS_TOKEN || '' }}
```
kpanot, line 20: The PR is not really following its description: instead of "disabling Nx Cache" it disables Nx Cloud.
Nx Cloud is not only for caching but divers metrics and developer local build optimization.
The best solution to disable the cache would be to provide the following envvar:
`NX_SKIP_NX_CACHE: ${{ github.event_name != 'push' && startsWith(github.ref, 'release') }}`

.github/workflows/code-check.yml
```diff
@@ -16,7 +16,7 @@ jobs:
   test:
     runs-on: ubuntu-latest
     env:
-      NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
+      NX_CLOUD_ACCESS_TOKEN: ${{ env.NX_CLOUD_ACCESS_TOKEN }}
```
kpanot, line 19: why using `env` instead of `secret`?

kpanot (approved):

.github/workflows/main.yml
kpanot: Actually when you checkout the release branch in local, the `nx build` will take around 30s thanks to the Nx Cloud remote cache (if you have set the NX_CLOUD_ACCES_TOKEN locally with your personal read token).
If the the Nx Cloud is enabled on the CI, even if the NX_SKIP_NX_CAHE is enabled, the CI will send the cache to Nx Cloud which will be used for local build.

.github/workflows/pr-labels.yml
```diff
@@ -6,7 +6,7 @@ jobs:
   setup-label:
     runs-on: ubuntu-latest
     env:
-      NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
+      NX_CLOUD_ACCESS_TOKEN: ${{ env.NX_CLOUD_ACCESS_TOKEN }}
```
kpanot, line 9: This env does exist in this pipeline? As it is not triggered by the main pipeline

kpanot (approved):

PR #622, opened 2023-08-03
fix: remove useless release workflow

pginoux-1A:


.github/workflows/release.yml
fpaul-1A, line 1: couldn't this be done on `main` ? it's not needed on release/9.0

pginoux-1A: I don't mind moving it to main

fpaul-1A (approved):

vscaiceanu-1a (approved):

kpanot (approved):

PR #624, opened 2023-08-03
ci: fix some ut instability

fpaul-1A:
## Proposed change

Relying on real timers for tests makes them slow and indeterministic
There is an option on jest to use fake timers that we can use to mock all the calls to `setTimeout`, `setInterval`, ...

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@ama-sdk/generator-sdk/jest.config.js
+++ b/packages/@ama-sdk/generator-sdk/jest.config.js
@@ -15,6 +15,10 @@ module.exports = {
     ['jest-junit', {outputDirectory: resolve(__dirname, 'dist-test'), outputName: 'ut-report.xml'}],
     'github-actions'
   ],
+  fakeTimers: {
+    // TODO enable fake timers
+    // enableGlobally: true
+  },
   transform: {
     // eslint-disable-next-line @typescript-eslint/naming-convention
     '^.+\\.tsx?$': [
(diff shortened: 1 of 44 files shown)


kpanot (approved):

packages/@ama-sdk/generator-sdk/jest.config.js
```diff
@@ -15,6 +15,10 @@ module.exports = {
     ['jest-junit', {outputDirectory: resolve(__dirname, 'dist-test'), outputName: 'ut-report.xml'}],
     'github-actions'
   ],
+  fakeTimers: {
+    // TODO enable fake timers
+    // enableGlobally: true
```
kpanot, line 20: Why it is not enabled in this PR?

packages/@o3r/core/src/store/async/async.operator.spec.ts
```diff
@@ -26,65 +27,49 @@ function createAction(
 }
 
 describe('AsyncOperator', () => {
-  it('fromApiEffectSwitchMapById', () => {
+  it('fromApiEffectSwitchMapById', async () => {
     const actionsStream = new Subject<any>();
-    const outputs: any[] = [];
-    actionsStream
+    const allResolved = lastValueFrom(actionsStream
```
kpanot, line 32: Actually `lastValueFrom(....)` + `take(1)` => `firstValueFrom(...)`

kpanot (dismissed):

matthieu-crouzet (dismissed):

packages/@ama-sdk/generator-sdk/jest.config.js
fpaul-1A: it didn't seem that easy to make it work with `yeoman-test`

matthieu-crouzet (approved):

kpanot (approved):

PR #613, opened 2023-08-01
feat: ng add @o3r/testing independent from @o3rcore

vscaiceanu-1a:
## Related issues

- :rocket: Feature #602

PR #636, opened 2023-08-07
ci: set worker-idle-memory-limit for ut tests

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/jest.config.ut.js
+++ b/jest.config.ut.js
@@ -1,0 +1,60 @@
+const { getJestModuleNameMapper } = require('@o3r/dev-tools');
+
+globalThis.ngJest = {
+  skipNgcc: true
+};
+
+/**
+ * @param rootDir {string}
+ * @param isAngularSetup {boolean}
+ * @returns {import('ts-jest/dist/types').JestConfigWithTsJest}
+ */
+module.exports.getJestConfig = (rootDir, isAngularSetup) => ({
+  preset: 'ts-jest',
+  setupFilesAfterEnv: ['<rootDir>/testing/setup-jest.ts'],
+  rootDir: '.',
+  moduleNameMapper: getJestModuleNameMapper(rootDir),
+  modulePathIgnorePatterns: [
+    '<rootDir>/dist'
+  ],
+  testPathIgnorePatterns: [
+    '<rootDir>/.*/templates/.*'
+  ],
+  reporters: [
+    'default',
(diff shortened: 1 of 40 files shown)


packages/@ama-sdk/core/jest.config.js
```diff
@@ -1,21 +1,13 @@
 const { getJestModuleNameMapper } = require('@o3r/dev-tools');
-const {resolve} = require('node:path');
+const defaultConfig = require('../../../jest.config.ut');
 
 /** @type {import('ts-jest/dist/types').JestConfigWithTsJest} */
 module.exports = {
```
kpanot, line 5: the `transform` field should not be something like:
```typescript
transform: {
  ...defaultConfig.transform,
  '^.+\\.tsx?$': [/*...*/]
}
```
to be future proof?

jest.config.ut.js
```diff
@@ -0,0 +1,40 @@
+globalThis.ngJest = {
+  skipNgcc: true
+};
+
+/** @type {import('ts-jest/dist/types').JestConfigWithTsJest} */
+module.exports = {
```
kpanot, line 6: could be a factory to provide `moduleNameMapper` as well:
```typescript
/**
 * @params rootDir {string}
 * @returns {import(https://github.com/kulshekhar/ts-jest).JestConfigWithTsJest}
 */
module.exports.getJestConfig = (rootDir) => ({
   moduleNameMapper: getJestModuleNameMapper(rootDir)
  // ...
})
```

kpanot (dismissed):

jest.config.ut.js
kpanot, line 1: This file may need to be added to the Otter Workspace by the `ng add @o3r/core` in workspace level.
Otherwise the `ng add @o3r/testing` should handle the 2 cases (with or without) the dep to it.

matthieu-crouzet (dismissed):

jest.config.ut.js
fpaul-1A: much better

packages/@ama-sdk/core/jest.config.js
fpaul-1A: no longer needed with the factory

jest.config.ut.js
fpaul-1A: this can be done in a separate pull request (possibly this one https://github.com/AmadeusITGroup/otter/pull/613)

kpanot (approved):

PR #648, opened 2023-08-09
fix: null pointer on this in storage sync

cpaulve-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/store-sync/src/core/storage-sync.ts
+++ b/packages/@o3r/store-sync/src/core/storage-sync.ts
@@ -52,7 +52,7 @@ export class StorageSync {
   /**
    * Returns a meta reducer that handles storage sync
    */
-  public localStorageSync() {
+  public localStorageSync = () => {
     return (reducer: any) => {
       if (isLocalStorageConfig(this.options)) {
         return syncStorage(this.options)(reducer);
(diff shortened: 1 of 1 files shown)


pginoux-1A (approved):

packages/@o3r/core/schematics/rule-factories/store/index.ts
```diff
@@ -131,7 +131,7 @@ const rootReducers = {
   ${isApplicationThatUsesRouterModule(tree) ? 'router: routerReducer' : ''}
 };
 
-const metaReducers = [storageSync.localStorageSync()];
+const metaReducers = [storageSync.localStorageSync];
```
fpaul-1A, line 134: You lose the context of the class by doing this

```suggestion
const metaReducers = [(reducer) => storageSync.localStorageSync(reducer)];
```
or
```suggestion
const metaReducers = [storageSync.localStorageSync.bind(storageSync)];
```

cpaulve-1A: first one does not work as well, only the bind solution does work. I will go back with my original fix which should not impact the applications

matthieu-crouzet (dismissed):

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #658, opened 2023-08-10
feat: do not synchronize untouched state with the localstorage [perf]

cpaulve-1A:
## Proposed change
Only synchronize with the localstorage states that do not match the version already stored.
Keep an internal representation of the state to avoid extra read operations that would slow the application

### DIFF
--- a/packages/@o3r/store-sync/src/core/interfaces.ts
+++ b/packages/@o3r/store-sync/src/core/interfaces.ts
@@ -36,5 +36,11 @@ export interface AsyncStorageSyncOptions extends Omit<SyncStorageConfig, 'storag
   storage?: AsyncStorage;
 }
 
-/** Options for storage sync */
+/** Options for storage sync method */
 export type StorageSyncOptions = SyncStorageConfig | AsyncStorageSyncOptions;
+
+/**
+ * Options that can be set on the Storage sync constructor
+ * postProcess and syncKeyCondition will be handled by the StorageSync class if smart sync is activated
+ */
+export type StorageSyncConstructorOptions = Omit<StorageSyncOptions, 'postProcess' | 'syncKeyCondition'>;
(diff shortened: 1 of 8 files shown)


pginoux-1A (approved):

packages/@o3r/store-sync/src/core/interfaces.ts
```diff
@@ -38,3 +38,5 @@ export interface AsyncStorageSyncOptions extends Omit<SyncStorageConfig, 'storag
 
 /** Options for storage sync */
 export type StorageSyncOptions = SyncStorageConfig | AsyncStorageSyncOptions;
+
+export type StorageSyncConstructorOptions = Omit<StorageSyncOptions, 'postProcess' | 'syncKeyCondition'>;
```
kpanot, line 42: Missing comment

pginoux-1A (dismissed):

kpanot (dismissed):

packages/@o3r/store-sync/src/core/storage-sync.ts
```diff
@@ -1,22 +1,35 @@
...
+  private storeImage: { [key: string]: any } = {};
 
   public options: StorageSyncOptions;
 
-  constructor(options?: Partial<StorageSyncOptions>) {
+  constructor(options?: StorageSyncConstructorOptions, extraOptions?: {disableSmartSync: boolean}) {
     this.options = {
       keys: [],
+      ...(extraOptions?.disableSmartSync ? {} : {
+        syncKeyCondition: (key, state) => !equal(this.storeImage[key], state[key]),
```
fpaul-1A, line 23: I don't know how easy it would be to implement but it could be nice to have some kind of `throttle` (with configurable delay maybe?) to generally avoid any unnecessary CPU overhead

fpaul-1A (approved):

fpaul-1A (approved):

pginoux-1A (approved):

Issue #646, opened 2023-08-09
Migrate VSCode extension configuration to auto-detection


kpanot:
### Context

Several configuration of schematics defined in the workspace (`angular.json` or `nx.json`) are duplicated at VScode extension configuration level.
The purpose it to remove the this duplicated configuration.

### Proposal

- The configuration should be read from the workspace configuration.
- The NPM client should be determined automatically (workspace config + env)

PR #659, opened 2023-08-10
feat(vscode ext): use schematic options from angular.json or nx.json

matthieu-crouzet:
## Proposed change

Use schematic options from angular.json or nx.json

## Related issues

- :rocket: Feature https://github.com/AmadeusITGroup/otter/issues/646


apps/vscode-extension/src/commands/helpers.ts
```diff
@@ -29,7 +29,50 @@ export const wrapCommandWhenExplorerContext = (
 /**
  * Get the runner for NPM scripts
  */
-export const getPackageScriptRunner = () => {
-  const packageManager = vscode.workspace.getConfiguration('otter').get<string>('packageManager', 'npm');
+export const getPackageScriptRunner = async (): Promise<string> => {
+  const [workspaceJsonUri] = await vscode.workspace.findFiles('{angular,nx}.json');
```
vscaiceanu-1a, line 33: we should make sure nx has priority if both are present. you can have angular.json file present for edge cases like workaround for working with ng update, templates, mock files etc

vscaiceanu-1a (dismissed):

apps/vscode-extension/src/commands/helpers.ts
matthieu-crouzet: I would have say the opposite 
first `angular.json` then `nx.json`
as we said we should support first Angular and then Nx if possible

matthieu-crouzet: I have created a method to sort everytime the same way 
let's discuss the order all together

matthieu-crouzet: As discussed we keep `angular.json` first for the moment but it could change later if we face some issue about it

apps/vscode-extension/src/commands/module/add-module.command.ts
```diff
@@ -86,11 +86,12 @@ export function generateModuleAddCommand(_context: ExtensionContext) {
...
-      .forEach(({label}) => {
+    await Promise.all(
+      moduleToAdd.map(async ({label}) => {
         const terminal = vscode.window.createTerminal(`Add the module ${label}`);
-        terminal.sendText(`${getPackageScriptRunner()} ng add ${label} --defaults`, true);
+        terminal.sendText(`${await getPackageScriptRunner()} ng add ${label} --defaults`, true);
         terminal.show();
-      });
+      })
+    );
```
fpaul-1A, line 95: You could avoid this promise thing by just doing 
```typescript
const packageScriptRunner = await getPackageScriptRunner();
``` 
before the forEach

apps/vscode-extension/src/commands/helpers.ts
```diff
@@ -26,10 +26,75 @@ export const wrapCommandWhenExplorerContext = (
...
+export const getSchematicDefaultOptions = async (schematicName: string) => {
+  const getSchematicDefaultOptionsFromWorkspaceUri = async (workspaceJsonUri: vscode.Uri) => {
+    const workspaceJsonDocument = await vscode.workspace.openTextDocument(workspaceJsonUri);
+    const workspaceJson = JSON.parse(workspaceJsonDocument.getText());
+    return basename(workspaceJsonUri.path, '.json') === 'angular'
+      ? workspaceJson.schematics?.[schematicName]
+      : workspaceJson.generators?.[schematicName];
+  };
+  const packageManager = await getInfoFromWorkspaceJsonUris(getSchematicDefaultOptionsFromWorkspaceUri);
+  return packageManager || {};
```
fpaul-1A, line 99:
```suggestion
  return await getInfoFromWorkspaceJsonUris(getSchematicDefaultOptionsFromWorkspaceUri) ?? {};
```

apps/vscode-extension/src/commands/helpers.ts
```diff
@@ -26,10 +26,74 @@ export const wrapCommandWhenExplorerContext = (
...
+
+const getInfoFromWorkspaceJsonUris = async <T>(
+  getInfo: (workspaceJsonUri: vscode.Uri) => Promise<T | undefined>
+) => {
+  const workspaceJsonUris = await vscode.workspace.findFiles('{angular,nx}.json');
+  const sortedWorkspaceJsonUris = workspaceJsonUris.sort(sortWorkspaceUris);
+  const packageManagers = await Promise.all(
+    sortedWorkspaceJsonUris.map(getInfo)
+  );
+  return packageManagers.find((packageManager) => !!packageManager);
```
fpaul-1A, line 49: You only need to read the files until you find one with the info, no need to read all of them from the start
```suggestion
  for workspaceJsonUri of sortedWorkspaceJsonUris) {
    const infoFromWorkspace = await getInfo(workspaceJsonUri);
    if (infoFromWorkspace) {
      return infoWorkspace;
    }
  }
  return {};
```

fpaul-1A (approved):

vscaiceanu-1a (approved):

PR #662, opened 2023-08-11
fix: fallback on missing status code from API response

cpaulve-1A:
## Proposed change

If the API returns a success status code not described in the specification, fallback to the first defined success response code
## Related issues

- :bug:

### DIFF
--- a/packages/@ama-sdk/core/src/fwk/api.helpers.ts
+++ b/packages/@ama-sdk/core/src/fwk/api.helpers.ts
@@ -1,4 +1,5 @@
-import { TokenizedOptions } from '../plugins/core/request-plugin';
+import {TokenizedOptions} from '../plugins/core/request-plugin';
+import type {ReviverType} from './Reviver';
 
 /**
  * prepares the url to be called
@@ -30,7 +31,7 @@ export function extractQueryParams<T extends { [key: string]: any }>(data: T, na
       const prop = data[name];
       acc[name] = (typeof prop.toJSON === 'function') ? prop.toJSON() : prop.toString();
       return acc;
-    }, {} as {[p in keyof T]: string});
+    }, {} as { [p in keyof T]: string });
 }
 
 /**
@@ -39,10 +40,10 @@ export function extractQueryParams<T extends { [key: string]: any }>(data: T, na
  * @param object JSON object to filter
  * @returns an object without undefined values
  */
-export function filterUndefinedValues(object: {[key: string]: string | undefined}): {[key: string]: string} {
+export function filterUndefinedValues(object: { [key: string]: string | undefined }): { [key: string]: string } {
   return Object.keys(object)
     .filter((objectKey) => typeof object[objectKey] !== 'undefined')
-    .reduce<{[key: string]: string}>((acc, objectKey) => {
+    .reduce<{ [key: string]: string }>((acc, objectKey) => {
       acc[objectKey] = object[objectKey] as string;
       return acc;
     }, {});
@@ -105,7 +106,7 @@ export function computePiiParameterTokens(piiParameterNames: string[]): { [key: 
  */
 export function tokenizeRequestOptions(tokenizedUrl: string, queryParameters: { [key: string]: string }, piiParamTokens: { [key: string]: string }, data: any): TokenizedOptions {
   const values: Record<string, string> = {};
-  const tokenizedQueryParams = { ...queryParameters };
+  const tokenizedQueryParams = {...queryParameters};
   Object.entries(piiParamTokens).filter(([parameterName, _token]) => data[parameterName] !== undefined).forEach(([parameterName, token]) => {
     if (tokenizedQueryParams[parameterName]) {
       tokenizedQueryParams[parameterName] = token;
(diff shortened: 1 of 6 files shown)


packages/@ama-sdk/core/src/fwk/clients/api-fetch-client.ts
```diff
@@ -152,11 +152,24 @@ export class ApiFetchClient implements ApiClient {
     } catch (e: any) {
       exception = new ResponseJSONParseError(e.message || 'Fail to parse response body', response && response.status || 0, body, {apiName, operationId, url, origin});
     }
+    let statusCodeReviver: ReviverType<T> | undefined = typeof reviver === 'function' || typeof reviver === 'undefined' ?
+      reviver : typeof response?.status !== 'undefined' && reviver[response.status] || undefined;
+    if (!!response?.status && response.status !== 204 && typeof reviver === 'object' && !Object.hasOwn(reviver, response?.status)) {
+      const fallback = Object.entries(reviver)?.reduce((acc: {statusCode: number; reviver: ReviverType<T> | undefined}, [key, value]) => {
+        if (+key < acc.statusCode && +key !== 204) {
```
fpaul-1A, line 159: should we check that we fallback to a statusCode in the same category ? (2xx, 3xx, 4xx, ...) ?

cpaulve-1A: this is only for the success status code, in the other case, we handle them as error (well not sure for 3xx but we would not revive it anyways)

matthieu-crouzet (dismissed):

packages/@ama-sdk/core/src/fwk/clients/api-fetch-client.ts
```diff
@@ -152,11 +152,25 @@ export class ApiFetchClient implements ApiClient {
...
+    if (isSuccessResponseCode && typeof reviver === 'object' && !Object.prototype.hasOwnProperty.call(reviver, response!.status)) {
+      const fallback = Object.entries(reviver)?.reduce((acc: {statusCode: number; reviver: ReviverType<T> | undefined}, [key, value]) => {
+        if (+key < acc.statusCode && +key !== 204) {
+          return {statusCode: +key, reviver: value};
+        }
+        return acc;
+      }, {statusCode: Number.MAX_SAFE_INTEGER, reviver: undefined});
+      statusCodeReviver = fallback?.reviver;
+      // eslint-disable-next-line no-console
+      console.error(`Unknown status code ${response?.status || 'undefined'} for ${operationId || 'unknown endpoint'} - fallback to status code ${fallback?.statusCode}`);
```
mrednic-1A, line 167: It's not more a 'warning' or 'info' than an 'error' ? 
Same as other logs related to APIs missing fields :)

fpaul-1A (approved):

mrednic-1A (approved):

packages/@ama-sdk/core/src/fwk/clients/api-fetch-client.ts
```diff
@@ -152,11 +152,25 @@ export class ApiFetchClient implements ApiClient {
     } catch (e: any) {
       exception = new ResponseJSONParseError(e.message || 'Fail to parse response body', response && response.status || 0, body, {apiName, operationId, url, origin});
     }
+    let statusCodeReviver: ReviverType<T> | undefined = typeof reviver === 'function' || typeof reviver === 'undefined' ?
+      reviver : typeof response?.status !== 'undefined' && reviver[response.status] || undefined;
+    const isSuccessResponseCode = !!response?.status && response.status >= 200 && response.status < 300 && response.status !== 204;
+    if (isSuccessResponseCode && typeof reviver === 'object' && !Object.prototype.hasOwnProperty.call(reviver, response!.status)) {
+      const fallback = Object.entries(reviver)?.reduce((acc: {statusCode: number; reviver: ReviverType<T> | undefined}, [key, value]) => {
```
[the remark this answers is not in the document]
cpaulve-1A: If I reassign the acc instead of creating a new object, it should be better no?

packages/@ama-sdk/core/src/fwk/api.helpers.ts
```diff
@@ -115,3 +116,31 @@ export function tokenizeRequestOptions(tokenizedUrl: string, queryParameters: {
...
+/**
+ * Compute the reviver to use in case the success response code returned by the API does not match the API specifications.
+ * Does not try to match non-successful responses as error are handled separately
+ *
+ * @param revivers
+ * @param responseCode
+ */
+export function fallbackToFirstEligibleResponseReviver<T>(revivers: {[statusCode: number]: ReviverType<T> | undefined}, responseCode: number):
+  {reviver: ReviverType<T> | undefined; log: string | undefined} {
+  if (responseCode < 200 && responseCode >= 300) {
```
pginoux-1A, line 129: Not sure about the condition

packages/@ama-sdk/core/src/fwk/api.helpers.ts
```diff
@@ -115,3 +116,45 @@ export function tokenizeRequestOptions(tokenizedUrl: string, queryParameters: {
...
+ * Compute the reviver to use in case the success response code returned by the API does not match the API specifications.
+ * Does not try to match non-successful responses as error are handled separately
+ *
+ * @param revivers
+ * @param response
+ * @param disableFallback
+ */
+export function getResponseReviver<T>(revivers: {[statusCode: number]: ReviverType<T> | undefined} | undefined | ReviverType<T>,
+                                                          response: {status?: number; ok?: boolean} | undefined, disableFallback?: boolean):
+  {reviver: ReviverType<T> | undefined; logMessage: string | undefined} {
```
kpanot, line 130: Having a optional logger in input would be simpler and would not require the creation of an object in the function return.
But why not :)

vscaiceanu-1a (approved):

pginoux-1A (dismissed):

packages/@ama-sdk/core/src/fwk/api.helpers.ts
```diff
@@ -115,3 +116,53 @@ export function tokenizeRequestOptions(tokenizedUrl: string, queryParameters: {
...
+ *
+ * @param revivers
+ * @param endpoint
+ * @param response
+ * @param disableFallback
+ */
+export function getResponseReviver<T>(revivers: {[statusCode: number]: ReviverType<T> | undefined} | undefined | ReviverType<T>, response: {status?: number; ok?: boolean} | undefined,
+  endpoint?: string, options?: {disableFallback?: boolean}): ReviverType<T> | undefined {
+  const logPrefix = `API status code error for ${endpoint || 'unknown'} endpoint`;
+  if (!response?.ok) {
```
kpanot, line 133: Why `?` as the `response` is mandatory anyway.

Moreover, the fact that `ok` is optional is leading to issue:
if `ok === undefined` it does not mean that the response is "not ok", it just means that the `ok` property is not provided

So there are 2 possible fixes:

- `ok` is mandatory:

```typescript
if (!response.ok) {
  return undefined;
}
```

- `ok` is optional:

```typescript
if (response.ok === false || (!response.ok && !!response.status && (response.status < 200 || response.status > 299))) {
  return undefined;
}
```

vscaiceanu-1a (approved):

PR #679, opened 2023-08-23
feat: enhance dynamic content module

pginoux-1A:
## Proposed change
* Make media folder name configurable
* Update documentation

### DIFF
--- a/packages/@o3r/dynamic-content/README.md
+++ b/packages/@o3r/dynamic-content/README.md
@@ -38,19 +38,21 @@ A pipe to be used in your component templates:
 
 ```html
 <img src="{{'assets-otter/imgs/logo.png' | dynamicContent}}" /> or
-<img [src]="'assets-otter/imgs/logo.png' | dynamicContent" />
+<img ngSrc="{{'assets-otter/imgs/logo.png' | dynamicContent}}"/>
 ```
 
 and a service to be used in your component classes, for example:
 
 ```typescript
-
+import { firstValueFrom } from 'rxjs';
 @Component({
   /* ... */
 })
 export class MyComponent {
   constructor(private service: DynamicContentService) {
-    const imgSrc = this.service.getMediaPath('assets/assets-otter/imgs/logo.png');
+  }
+  ngOnInit() {
+    const imgSrc = await firstValueFrom(this.service.getMediaPathStream('assets/assets-otter/imgs/logo.png'));
   }
 }
 ```
--- a/packages/@o3r/dynamic-content/src/services/dynamic-content/dynamic-content.service.ts
+++ b/packages/@o3r/dynamic-content/src/services/dynamic-content/dynamic-content.service.ts
@@ -3,9 +3,7 @@ import { select, Store } from '@ngrx/store';
 import { of } from 'rxjs';
 import { distinctUntilChanged, map, shareReplay } from 'rxjs/operators';
 import { AssetPathOverrideStore, selectAssetPathOverride } from '../../stores/index';
-import { CMS_ASSETS_PATH_TOKEN, DYNAMIC_CONTENT_BASE_PATH_TOKEN } from './dynamic-content.token';
-
-const MEDIA_FOLDER_NAME = 'assets';
+import {CMS_ASSETS_PATH_TOKEN, DYNAMIC_CONTENT_BASE_PATH_TOKEN, MEDIA_FOLDER_NAME_TOKEN} from './dynamic-content.token';
 
 /**
  * Service for getting dynamic content path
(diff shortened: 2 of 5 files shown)


packages/@o3r/dynamic-content/README.md
```diff
@@ -38,19 +38,19 @@ A pipe to be used in your component templates:
 
 ```html
 <img src="{{'assets-otter/imgs/logo.png' | dynamicContent}}" /> or
-<img [src]="'assets-otter/imgs/logo.png' | dynamicContent" />
+<img ngSrc="{{'assets-otter/imgs/logo.png' | dynamicContent}}" fill/>
```
matthieu-crouzet, line 41: why using `ngSrc` ?
`fill` is useful for background image but not recommend for others images as it will cause layout shift
So I would not recommend we put that in our recommendation

packages/@o3r/dynamic-content/README.md
```diff
@@ -38,19 +38,19 @@ A pipe to be used in your component templates:
...
 ```typescript
-
+import { firstValueFrom } from 'rxjs';
 @Component({
   /* ... */
 })
 export class MyComponent {
   constructor(private service: DynamicContentService) {
-    const imgSrc = this.service.getMediaPath('assets/assets-otter/imgs/logo.png');
+    const imgSrc = await firstValueFrom(this.service.getMediaPathStream('assets/assets-otter/imgs/logo.png'));
```
matthieu-crouzet, line 53: constructor could not be asynchronous

matthieu-crouzet: why not using the `async` pipe in the template ?

packages/@o3r/dynamic-content/src/services/dynamic-content/dynamic-content.service.ts
```diff
@@ -46,7 +44,7 @@ export class DynamicContentService {
     if (this.cmsOnlyAssetsPath && assetPath) {
       return assetPath.startsWith('/') ? assetPath : `${this.cmsOnlyAssetsPath.replace(/\/$/, '')}/${assetPath}`;
     }
-    return this.getContentPath(this.mediaFolder ? `${this.mediaFolder}/${this.normalizePath(assetPath)}` : assetPath);
+    return this.getContentPath(`${this.mediaFolderName}/${this.normalizePath(assetPath)}`);
```
matthieu-crouzet, line 47:
```suggestion
    return this.getContentPath(`${
      this.mediaFolderName ? this.mediaFolderName + '/' : ''
    }${this.normalizePath(assetPath)}`);
```
prefixing by `/` could maybe cause trouble if the normalized path starts with `..`

packages/@o3r/dynamic-content/README.md
pginoux-1A: It was more to store the imgSrc in the typescript (to have the same example than in the previous version of the doc) but I can definitely move it in the onInit

packages/@o3r/dynamic-content/README.md
pginoux-1A: After discussion, ngSrc is recommended over [src] for performances: https://angular.io/guide/image-directive

packages/@o3r/dynamic-content/src/services/dynamic-content/dynamic-content.service.ts
pginoux-1A: The normalize path called by the getContentPath prevents that
image

packages/@o3r/dynamic-content/README.md
kpanot: I a bit agree with Matthieu, this it not the purpose of the example, but we should not encourage to convert it to promise which would block the rules engine assets override action.
I am a bit scare that people will take this example to use it in their code and get issue because of it :S

fpaul-1A: I believe the `ngOnInit` shouldn't be asynchronous either
also the `await` won't compile if not inside an `async` function

pginoux-1A: We already have an example for the async pipe, this is meant to get the value in the typescript, maybe I can put it in a dedicated async function:
`async getImgSrc() {
  return await firstValueFrom(this.service.getMediaPathStream('assets/assets-otter/imgs/logo.png'));
}`
Would it be a good tradeoff ?

kpanot: Unfortunately, for me it is still not supporting the Rules Engine and should not be an example that we should provide :S

pginoux-1A: I'll update to put the same one than the proposal you did on my other PR :)

PR #682, opened 2023-08-24
feat: add angular client to SDK

kpanot:


packages/@ama-sdk/core/src/clients/api-angular-client.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    let response: HttpResponse<any> | undefined;
+    let root: any | undefined;
+    let exception: Error | undefined;
+
+    const origin = options.headers.get('Origin');
+
+    // Execute call
+    try {
+      const headers = Array.from(options.headers.entries())
```
fpaul-1A, line 103: can't we use directly `Object.fromEntries` ?

packages/@ama-sdk/core/src/clients/api-angular-client.ts
```diff
@@ -0,0 +1,160 @@
...
+    const replyPlugins = this.options.replyPlugins ?
+      this.options.replyPlugins.map((plugin) => plugin.load<T>({
+        dictionaries: root && root.dictionaries,
+        response: response && {
+          ...response,
+          headers: new Headers(
+            Array.from(response.headers.keys()
+              .map((key) => ([key, response!.headers.get(key)!] as [string, string]))
+            )
+          )
```
fpaul-1A, line 138: isn't it equivalent to:
```suggestion
          headers: new Headers(Array.from(response.headers.entries()))
```
?

fpaul-1A (dismissed):

packages/@ama-sdk/core/src/clients/api-angular-client.ts
kpanot: Actually no because `response.headers.entries()` does not exist :S

fpaul-1A (approved):

matthieu-crouzet (approved):

packages/@ama-sdk/core/src/plugins/url-rewrite/url-rewrite.spec.ts
```diff
@@ -12,6 +12,7 @@ describe('URL Rewrite Request Plugin', () => {
 
   beforeEach(() => {
     options = {
+      method: 'get',
```
mrednic-1A, line 15: was this mandatory, or you added it to have the tests more explicit ?

kpanot: It is mandatory. I turned it to mandatory to avoid defaulting to different method according to the request shooter mechanism.
No impact other than test as we always explicit the method

mrednic-1A: I see. The `method` is always passed in the request options in the `prepareOptions` method of the clients we have.

mrednic-1A (approved):

fpaul-1A (approved):

fpaul-1A (approved):

fpaul-1A (approved):

mrednic-1A (approved):

PR #684, opened 2023-08-25
fix: remove override content path and update doc

pginoux-1A:
## Issue

Today, getContentPathStream and getMediaPathStream are plugged on the override store, to be able to take into account any override coming for example from the rules engine.
In reality, only the getMediaPathStream works as intended, because the getContentPathStream compares the value stored with only the file name with the dynamicContentPath+filename on the other side.
The getContentPathStream is meant to access the content outside of the media folder, that should be ONLY driven by the CMS, to prevent unstabilities (for example if you can override in a rule the ruleset.json, or some functional metadata files the behavior will be different if you refresh your app).

## Proposed changes

Remove the store override access in getContentPathStream (not working anyway today)


packages/@o3r/dynamic-content/src/services/dynamic-content/dynamic-content.service.ts
```diff
@@ -50,27 +40,25 @@ export class DynamicContentService {
   }
 
   /**
-   * Gets the asset path with the content root path
+   * Gets the full path of a content relative to the root
+   * Content path doesn't consider any override, you will always get the same file
    *
-   * @param assetPath asset location (e.g /assets-otter/imgs/my-image.png)
+   * Example: getMediaPath('assets/imgs/my-image.png') will give you the basePath + 'assets/imgs/my-image.png'
```
kpanot, line 46:
```suggestion
   * @example getMediaPath('assets/imgs/my-image.png') will give you the basePath + 'assets/imgs/my-image.png'
```

packages/@o3r/dynamic-content/src/services/dynamic-content/dynamic-content.service.ts
```diff
@@ -27,21 +27,11 @@ export class DynamicContentService {
     return !assetPath ? '' : assetPath.replace(/^\//, '');
   }
 
-  /**
```
kpanot, line 30: it was not mandatory but why removing it?

kpanot (dismissed):

packages/@o3r/dynamic-content/src/services/dynamic-content/dynamic-content.service.ts
pginoux-1A: It's a copy paste that is not accurate at all since it doesn't explain the difference content VS media, I prefer adding an accurate one on both public methods, and delete those ones

kpanot (dismissed):

packages/@o3r/dynamic-content/README.md
```diff
@@ -102,9 +102,33 @@ export function myContentPath() {
...
 ```
 
-## getContentPath
+## getContentPathStream
 
-For non-media resources (ex: localization, configuration) one should refer to the ``getContentPath`` method of the ``DynamicContentService``. Doing so will ensure the correct resource path is computed in an ACS-enabled environment (e.g. when the ``data-dynamiccontentpath`` tag is present in the body).
+For non-media resources (ex: localization, configuration) one should refer to the ``getContentPath`` method of the ``DynamicContentService``. Doing so will ensure the correct resource path is computed in any environment (e.g. when the ``data-dynamiccontentpath`` tag is present in the body).
+
+The content path is always related to the root of the application.
+It also ignores any path overrides from the AssetPathOverrideStore store, meaning that you will always get the same file.
```
kpanot, line 110:
```suggestion
It also ignores any path overrides from the `AssetPathOverrideStore` store, meaning that you will always get the same file.
```

packages/@o3r/dynamic-content/README.md
```diff
@@ -102,9 +102,33 @@ export function myContentPath() {
...
+  /** */
+})
+export class MyComponent {
+  constructor(private service: DynamicContentService) {}
+
+  async getDynamicConfig() {
+    const filePath = await firstValueFrom(this.service.getContentPathStream('global.config.post.json'));
+    const fileContent = await fetch(filePath);
+    //  ...
+  }
```
kpanot, line 123: I would have the same argument as the PR #679 :
I understand that you wanted to provide a simple example but it is not a recommended way to use the service (to support the rules engine).
To be more aligned with the guideline, it would be:

```suggestion
  public dynamicConfig$ = this.service.getContentPathStream('global.config.post.json').pipe(
    switchMap((filePath) => from(fetch(filePath))
  );
```

packages/@o3r/dynamic-content/README.md
```diff
@@ -114,8 +138,18 @@ export class MyComponent {
...
   async getDynamicConfig() {
-    const result = await fetch(this.dynamicContentService.getContentPath('global.config.post.json'));
-    ...
+    const filePath = await firstValueFrom(this.service.getMediaPathStream('imgs/my-image.png'));
+    const fileContent = await fetch(filePath);
+    //  ...
   }
 }
 ```
+It also looks for overrides in the AssetPathOverrideStore, so it will return the new value instead if an override is found.
```
vscaiceanu-1a, line 147:
```suggestion
It also looks for overrides in the `AssetPathOverrideStore`, so it will return the new value instead if an override is found.
```

vscaiceanu-1a (approved):

kpanot (approved):

PR #710, opened 2023-08-31
ci: draft of it test for monorepo

fpaul-1A:
## Proposed change

Add an IT test with monorepo for o3r/analytics and @o3r/core

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/test-helpers/src/utilities/package-manager.ts
+++ b/packages/@o3r/test-helpers/src/utilities/package-manager.ts
@@ -1,4 +1,7 @@
 import { execFileSync, ExecSyncOptions } from 'node:child_process';
+import { existsSync, rmSync } from 'node:fs';
+import { join } from 'node:path';
+import { performance } from 'node:perf_hooks';
 
 declare global {
   namespace NodeJS {
@@ -15,14 +18,18 @@ const PACKAGE_MANAGERS_CMD = {
     create: 'npm create',
     exec: 'npm exec',
     install: 'npm install',
-    run: 'npm run'
+    run: 'npm run',
+    workspaceExec: 'npm exec -w',
+    workspaceRun: 'npm run -w'
   },
   yarn: {
     add: 'yarn add',
     create: 'yarn create',
     exec: 'yarn exec',
     install: 'yarn install',
-    run: 'yarn run'
+    run: 'yarn run',
+    workspaceExec: 'yarn workspace',
+    workspaceRun: 'yarn workspace'
   }
 };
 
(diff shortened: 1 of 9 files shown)


packages/@o3r/test-helpers/src/utilities/package-manager.ts
```diff
@@ -37,13 +43,14 @@ export function getPackageManager() {
...
   try {
+    const d = Date.now();
     const output = execFileSync(
       cmd.split(' ')[0],
       [...cmd.split(' ').slice(1), ...args.split(' ')].filter((arg) => !!arg),
       {...options, shell: process.platform === 'win32', stdio: 'pipe', encoding: 'utf8'}
     );
     // eslint-disable-next-line no-console
-    console.log(`${cmd} ${args}\n${output}`);
+    console.log(`${cmd} ${args} [${Date.now() - d}ms]\n${output}`);
```
kpanot, line 53: Using performance API would be cleaner :)

packages/@o3r/test-helpers/src/utilities/package-manager.ts
```diff
@@ -85,6 +92,19 @@ export function packageManagerExec(script: string, options: ExecSyncOptions) {
...
 }
 
+/**
+ * Execute a binary command (npx / yarn exec) for a specific workspace
+ * @param workspace
+ * @param script
+ * @param options
+ */
+export function packageManagerWorkspaceExec(workspace: string, script: string, options: ExecSyncOptions) {
+  if (getPackageManager() === 'npm') {
```
kpanot, line 102: Could be moved out as function (as it is used twice)

packages/@o3r/test-helpers/src/utilities/package-manager.ts
```diff
@@ -15,14 +15,18 @@ const PACKAGE_MANAGERS_CMD = {
     create: 'npm create',
     exec: 'npm exec',
     install: 'npm install',
-    run: 'npm run'
+    run: 'npm run',
+    workspaceExec: 'npm exec -w ',
```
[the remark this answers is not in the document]
fpaul-1A: my mistake

packages/@o3r/test-helpers/src/utilities/package-manager.ts
fpaul-1A: done

packages/@o3r/test-helpers/src/utilities/package-manager.ts
fpaul-1A: done

kpanot (approved):

mrednic-1A (approved):

Issue #603, opened 2023-07-31
Add application to the current project (monorepo)


mrednic-1A:
### Context

When we are in a multi-project context at some point, we would like to create a new application inside this monorepo.
The idea will be to provide a generator to create an otter application inside a monorepo.

### Proposal

The idea is to extend the already existing angular app generator and add the **o3r core** on top of it, plus some additional steps (listed below).

The command to run will be:
_npx ng g application <name>_ which behind will call the angular generator for an application inside a monorepo:
- _npx ng g @schematics/angular:application <name>_
followed by:
- _npx ng add @o3r/core --project <name>_ 

On top of the modification done today by the `ng add @o3r/core` this command will:
- adapt application to Otter recommendations ( add a package.json to the generated app - this is not added by the angular generator: Hint: prepare a template ([here ](https://github.com/angular/schematics-angular-builds/blob/main/workspace/files/package.json.template)is an example of the template from ng-cli which we can use) with the needed deps for a brand new app generated and get the versions of the packages from the root package.json);
- add otter schematics at the new created app level (modify the angular.json by adding the schematics field to the new application workspace)

mrednic-1A:
Requires #633

PR #714, opened 2023-08-31
feat(schematics): add otter application to monorepo

vscaiceanu-1a:
## Related issues

- :rocket: Feature #603

### DIFF
--- a/packages/@o3r/core/collection.json
+++ b/packages/@o3r/core/collection.json
@@ -93,6 +93,14 @@
         "libs"
       ]
     },
+    "application": {
+      "description": "Generate a new Otter Application",
+      "factory": "./schematics/application/index#generateApplication",
+      "schema": "./schematics/application/schema.json",
+      "aliases": [
+        "new-application", "app"
+      ]
+    },
     "sdk": {
       "description": "Generate a new SDK package in an Otter project",
       "factory": "./schematics/sdk/index#generateSdk",
--- a/packages/@o3r/core/schematics/application/index.ts
+++ b/packages/@o3r/core/schematics/application/index.ts
@@ -1,0 +1,79 @@
+import { apply, chain, externalSchematic, MergeStrategy, mergeWith, move, renameTemplateFiles, Rule, Schematic, SchematicContext, strings, template, Tree, url } from '@angular-devkit/schematics';
+import { getPackagesBaseRootFolder, getWorkspaceConfig, isNxContext } from '@o3r/schematics';
+import * as path from 'node:path';
+import type { NgGenerateApplicationSchema } from './schema';
+import type { PackageJson } from 'type-fest';
+import { RunSchematicTask } from '@angular-devkit/schematics/tasks';
+import { NgAddSchematicsSchema } from '../ng-add/schema';
+import { type Schema as ApplicationOptions } from '@schematics/angular/application/schema';
+import { Style } from '@schematics/angular/application/schema';
+
+/**
+ * Add an Otter application to a monorepo
+ *
+ * @param options Schematic options
+ */
+export function generateApplication(options: NgGenerateApplicationSchema): Rule {
+  const cleanName = strings.dasherize(options.name);
+
+  const addProjectSpecificFiles = (targetPath: string, rootDependencies: Record<string, string | undefined>) => {
+
+    return mergeWith(apply(url('./templates'), [
+      template({
+        ...options,
+        name: cleanName,
+        rootDependencies
+      }),
+      move(targetPath),
+      renameTemplateFiles()
+    ]), MergeStrategy.Overwrite);
+  };
+
+  const ngCliUpdate: Rule = (tree, context) => {
+
+    if (options.style && options.style !== 'scss') {
+      context.logger.warn('The provided "style" option will be ignored. "SCSS" will be used.');
+    }
+
+    const config = getWorkspaceConfig(tree);
+    if (!options.path && !config) {
+      throw new Error('The `path` option is not provided and no workspace configuration file found to define it.');
+    }
+    if (isNxContext(tree)) {
+      throw new Error('NX is not currently supported. Feature tracked via https://github.com/AmadeusITGroup/otter/issues/720');
+    }
+
+    const defaultRoot = getPackagesBaseRootFolder(tree, context, config!, 'application');
+
+    /** Path to the folder where to generate the new application */
+    const targetPath = path.posix.join('/', options.path || defaultRoot, cleanName);
+    const extendedOptions = { ...options, targetPath, name: cleanName, projectName: cleanName };
+
+    const rootPackageJson = tree.readJson('/package.json') as PackageJson;
+
+    const rootDependencies = { ...rootPackageJson.dependencies, ...rootPackageJson.devDependencies };
+
+    const angularAppSchema = context.engine.createCollection('@schematics/angular').createSchematic('application');
+
+    const getOptions = (schema: Schematic<any, any>) => {
+      const schemaOptions = schema.description.schemaJson?.properties || {};
+      return Object.keys(schemaOptions);
+    };
+    const angularOptions = getOptions(angularAppSchema);
+
+    return chain([
+      externalSchematic<Partial<ApplicationOptions>>('@schematics/angular', 'application', {
+        ...Object.entries(extendedOptions).reduce((acc, [key, value]) => (angularOptions.includes(key) ? {...acc, [key]: value} : acc), {}),
+        name: extendedOptions.name,
+        projectRoot: extendedOptions.targetPath,
+        style: Style.Scss}),
+      addProjectSpecificFiles(targetPath, rootDependencies),
+      (_host: Tree, c: SchematicContext) => {
+        c.addTask(new RunSchematicTask <Partial<NgAddSchematicsSchema>>('ng-add', extendedOptions));
+      }
+    ])(tree, context);
+  };
+
+  return ngCliUpdate;
+
+}
--- a/packages/@o3r/core/schematics/application/schema.json
+++ b/packages/@o3r/core/schematics/application/schema.json
@@ -1,0 +1,38 @@
+{
+  "$schema": "http://json-schema.org/schema",
+  "$id": "ngGenerateApplicationSchematicsSchema",
+  "title": "Generate a new Otter application",
+  "description": "Generate a new Otter application",
+  "properties": {
+    "name": {
+      "type": "string",
+      "description": "Application name",
+      "x-prompt": "Name of the new application?",
+      "$default": {
+        "$source": "argv",
+        "index": 0
+      }
+    },
+    "description": {
+      "type": "string",
+      "description": "Description of the new application",
+      "default": ""
+    },
+    "path": {
+      "type": "string",
+      "description": "Target directory to generate the application"
+    },
+    "skipLinter": {
+      "type": "boolean",
+      "description": "Skip the linter process",
+      "default": true
+    },
+    "skipInstall": {
+      "description": "Do not install dependency packages.",
+      "type": "boolean",
+      "default": false
+    }
+  },
+  "required": ["name"],
+  "additionalProperties": true
+}
--- a/packages/@o3r/core/schematics/application/templates/package.json.template
+++ b/packages/@o3r/core/schematics/application/templates/package.json.template
@@ -1,0 +1,29 @@
+{
+  "name": "<%= name %>",
+  "version": "0.0.0-placeholder",
+  "scripts": {
+    "ng": "ng",
+    "start": "ng serve",
+    "build": "ng build",
+    "watch": "ng build --watch --configuration development"
+  },
+  "private": true,
+  "dependencies": {
+    "@angular/animations": "<%= rootDependencies['@angular/animations'] || rootDependencies['@angular/core'] || 'latest' %>",
+    "@angular/common": "<%= rootDependencies['@angular/common'] || rootDependencies['@angular/core'] || 'latest' %>",
+    "@angular/compiler": "<%= rootDependencies['@angular/compiler'] || rootDependencies['@angular/core'] || 'latest' %>",
+    "@angular/core": "<%= rootDependencies['@angular/core'] || 'latest' %>",
+    "@angular/forms": "<%= rootDependencies['@angular/forms'] || rootDependencies['@angular/core'] || 'latest' %>",
+    "@angular/platform-browser": "<%= rootDependencies['@angular/platform-browser'] || rootDependencies['@angular/core'] || 'latest' %>",
+    "@angular/platform-browser-dynamic": "<%= rootDependencies['@angular/platform-browser-dynamic'] || rootDependencies['@angular/core'] || 'latest' %>",
+    "@angular/router": "<%= rootDependencies['@angular/router'] || rootDependencies['@angular/core'] || 'latest' %>",
+    "rxjs": "<%= rootDependencies['rxjs'] || 'latest' %>",
+    "tslib": "<%= rootDependencies['tslib'] || 'latest' %>",
+    "zone.js": "<%= rootDependencies['zone.js'] || 'latest' %>"
+  },
+  "devDependencies": {
+    "@angular/cli": "<%= rootDependencies['@angular/cli'] || 'latest' %>",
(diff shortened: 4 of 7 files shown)


packages/@o3r/core/schematics/application/index.ts
```diff
@@ -0,0 +1,56 @@
+import { apply, chain, externalSchematic, MergeStrategy, mergeWith, move, renameTemplateFiles, Rule, schematic, strings, template, url } from '@angular-devkit/schematics';
+import { getPackagesBaseRootFolder, getWorkspaceConfig } from '@o3r/schematics';
+import * as path from 'node:path';
+import { NgGenerateApplicationSchema } from './schema';
```
kpanot, line 4:
```suggestion
import type { NgGenerateApplicationSchema } from './schema';
```

packages/@o3r/core/schematics/application/index.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Add an Otter application to a monorepo
+ *
+ * @param options Schematic options
+ */
+export function generateApplication(options: NgGenerateApplicationSchema): Rule {
+  const cleanName = strings.dasherize(options.name);
+
+  const createAppPackageJson = (targetPath: string, rootDependencies: Record<string, string | undefined>) => {
```
kpanot, line 15: Actually the name is not really matching the implementation.
Today the `templates/` folder does contain only the `package.json` file but this function will take all the files of this folder, not only a `package.json`.
Maybe cleaner to rename it to a more generic manner (or target only the `package.json` file explicitly)

packages/@o3r/core/schematics/application/schema.json
```diff
@@ -0,0 +1,51 @@
+{
+  "$schema": "http://json-schema.org/schema",
+  "$id": "ngGenerateApplicationSchematicsSchema",
+  "title": "Generate a new Otter application",
+  "description": "Generate a new Otter application",
+  "properties": {
+    "name": {
+      "type": "string",
+      "description": "Application name",
+      "x-prompt": "Name of the new application?",
```
kpanot, line 10:
```suggestion
      "x-prompt": "Name the the new application?",
```

packages/@o3r/core/schematics/application/schema.json
```diff
@@ -0,0 +1,51 @@
...
+      "type": "boolean",
+      "description": "Skip the linter process",
+      "default": true
+    },
+    "skipInstall": {
+      "description": "Do not install dependency packages.",
+      "type": "boolean",
+      "default": false
+    },
+    "preset": {
```
kpanot, line 39: not used

packages/@o3r/core/schematics/application/index.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+    const defaultRoot = getPackagesBaseRootFolder(tree, context, config, 'application');
+
+    /** Path to the folder where to generate the new application */
+    const targetPath = path.posix.resolve('/', options.path || defaultRoot, cleanName);
+    const extendedOptions = { ...options, targetPath, name: cleanName, projectName: cleanName };
+
+    const rootPackageJson = tree.readJson('./package.json') as PackageJson;
+
+    const rootDependencies = { ...rootPackageJson.dependencies, ...rootPackageJson.devDependencies };
```
kpanot, line 41: can add `peerDependencies` as well (especially that there is no `devDependency` in the `package.json` generated)

packages/@o3r/core/schematics/application/index.ts
vscaiceanu-1a: how about `addProjectSpecificFiles`?

packages/@o3r/core/schematics/application/schema.json
vscaiceanu-1a: are you sure 😄 ?

packages/@o3r/core/schematics/application/index.ts
vscaiceanu-1a: I don't think we have peerDeps at root level, do we?

packages/@o3r/core/schematics/application/schema.json
kpanot: Just to introduce a typo :D

packages/@o3r/core/collection.json
```diff
@@ -93,6 +93,14 @@
         "libs"
       ]
     },
+    "application": {
+      "description": "Generate a new Otter Application",
+      "factory": "./schematics/application/index#generateApplication",
+      "schema": "./schematics/application/schema.json",
+      "aliases": [
+        "new-application"
```
matthieu-crouzet, line 101:
```suggestion
        "new-application", "app"
```

packages/@o3r/core/schematics/application/templates/package.json.template
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "<%= name %>",
+  "version": "0.0.0",
+  "scripts": {
+    "ng": "ng",
```
matthieu-crouzet, line 5: we should detect if we use nx to change that line into `"ng": "nx"`

packages/@o3r/core/schematics/application/templates/package.json.template
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "<%= name %>",
+  "version": "0.0.0",
```
matthieu-crouzet, line 3:
```suggestion
  "version": "0.0.0-placeholder",
```

packages/@o3r/core/schematics/application/templates/package.json.template
```diff
@@ -0,0 +1,29 @@
...
+  "dependencies": {
+    "@angular/animations": "<%= rootDependencies['@angular/animations'] || rootDependencies['@angular/core'] || latest %>",
+    "@angular/common": "<%= rootDependencies['@angular/common'] || rootDependencies['@angular/core'] || latest %>",
+    "@angular/compiler": "<%= rootDependencies['@angular/compiler'] || rootDependencies['@angular/core'] || latest %>",
+    "@angular/core": "<%= rootDependencies['@angular/core'] || latest %>",
+    "@angular/forms": "<%= rootDependencies['@angular/forms'] || rootDependencies['@angular/core'] || latest %>",
+    "@angular/platform-browser": "<%= rootDependencies['@angular/platform-browser'] || rootDependencies['@angular/core'] || latest %>",
+    "@angular/platform-browser-dynamic": "<%= rootDependencies['@angular/platform-browser-dynamic'] || rootDependencies['@angular/core'] || latest %>",
+    "@angular/router": "<%= rootDependencies['@angular/router'] || rootDependencies['@angular/core'] || latest %>",
+    "rxjs": "<%= rootDependencies['rxjs'] || latest %>",
```
matthieu-crouzet, line 20: Do all our o3r app depends on rxjs ?

packages/@o3r/core/schematics/application/templates/package.json.template
```diff
@@ -0,0 +1,29 @@
...
+  "version": "0.0.0",
+  "scripts": {
+    "ng": "ng",
+    "start": "ng serve",
+    "build": "ng build",
+    "watch": "ng build --watch --configuration development"
+  },
+  "private": true,
+  "dependencies": {
+    "@angular/animations": "<%= rootDependencies['@angular/animations'] || rootDependencies['@angular/core'] || latest %>",
```
matthieu-crouzet, line 12: Is that mandatory ?

packages/@o3r/core/schematics/application/templates/package.json.template
vscaiceanu-1a: this is what Angular brings by default when you generate a new app https://github.com/angular/schematics-angular-builds/blob/main/workspace/files/package.json.template

packages/@o3r/core/schematics/application/templates/package.json.template
vscaiceanu-1a: This schematic won't work at all with nx. It's completely out of scope for this PR.

packages/@o3r/core/schematics/application/index.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    return chain([
+      (t, c) => externalSchematic('@schematics/angular', 'application', {
+        name: extendedOptions.name,
```
matthieu-crouzet, line 45: you should forward all options

packages/@o3r/core/schematics/application/index.ts
```diff
@@ -0,0 +1,58 @@
...
+      throw new Error('No workspace configuration file found');
+    }
+    if (isNxContext(tree)) {
+      throw new Error('NX is not supported.');
+    }
+
+    const defaultRoot = getPackagesBaseRootFolder(tree, context, config, 'application');
+
+    /** Path to the folder where to generate the new application */
+    const targetPath = path.posix.resolve('/', options.path || defaultRoot, cleanName);
```
fpaul-1A, line 40: on Windows, this will resolve as the actual absolute path (like `c:\...`) which won't work on the Tree
it would be nice to add some unit tests (even if it's not very stable on windows at the moment) to make sure it works as expected

packages/@o3r/core/schematics/application/templates/package.json.template
vscaiceanu-1a: https://github.com/AmadeusITGroup/otter/issues/720

packages/@o3r/core/schematics/application/index.ts
matthieu-crouzet:
```suggestion
    const targetPath = path.posix.join('/', options.path || defaultRoot, cleanName);
```

packages/@o3r/core/schematics/application/index.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    const rootDependencies = { ...rootPackageJson.dependencies, ...rootPackageJson.devDependencies };
+
+    const angularOptions = { ...extendedOptions, projectRoot: targetPath, style: 'scss' };
+
+    return chain([
+      (t, c) => externalSchematic('@schematics/angular', 'application', angularOptions)(t, c),
+      (t, c) => addProjectSpecificFiles(targetPath, rootDependencies)(t, c),
+      (t, c) => schematic('ng-add', extendedOptions)(t, c)
+    ])(tree, context);
```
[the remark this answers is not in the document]
vscaiceanu-1a: Actually, it's not. The order is no longer respected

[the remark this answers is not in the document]
matthieu-crouzet: chain keep the order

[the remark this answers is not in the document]
kpanot: With
```typescript
externalSchematic('@schematics/angular', 'application', {
        ...Object.entries(options).reduce(acc, [key, value] => (schematicSpecificOptions.include(key) ? acc : {...acc, [key]: value}), {})
```

packages/@o3r/core/schematics/application/index.ts
```diff
@@ -0,0 +1,68 @@
...
+    const angularAppSchema = context.engine.createCollection('@schematics/angular').createSchematic('application');
+
+    const getOptions = (schema: Schematic<any, any>) => {
+      const schemaOptions = schema.description.schemaJson?.properties || {};
+      return Object.keys(schemaOptions);
+    };
+
+    return chain([
+      (t, c) => externalSchematic('@schematics/angular', 'application', {
+        ...Object.entries(extendedOptions).reduce((acc, [key, value]) => (getOptions(angularAppSchema).includes(key) ? {...acc, [key]: value} : acc), {}),
```
kpanot, line 56: you can create a const to not recall `getOptions(angularAppSchema)` to every iteration

packages/@o3r/core/schematics/application/schema.json
```diff
@@ -0,0 +1,42 @@
...
+        "$source": "argv",
+        "index": 0
+      }
+    },
+    "description": {
+      "type": "string",
+      "description": "Description of the new application",
+      "default": ""
+    },
+    "prefix": {
```
kpanot, line 21: may not be needed

packages/@o3r/core/schematics/application/index.ts
```diff
@@ -0,0 +1,76 @@
...
+    const targetPath = path.posix.join('/', options.path || defaultRoot, cleanName);
+    const extendedOptions = { ...options, targetPath, name: cleanName, projectName: cleanName };
+
+    const rootPackageJson = tree.readJson('/package.json') as PackageJson;
+
+    const rootDependencies = { ...rootPackageJson.dependencies, ...rootPackageJson.devDependencies };
+
+    const angularAppSchema = context.engine.createCollection('@schematics/angular').createSchematic('application');
+
+    const getOptions = (schema: Schematic<any, any>) => {
```
cpaulve-1A, line 54: https://github.com/AmadeusITGroup/otter/blob/3d2b71b64dba8a27bf6beffe8889fcbe9dc6e197/packages/%40o3r/schematics/src/rule-factories/ng-add/index.ts#L58 

Is it possible to use the same method for those two cases ?

packages/@o3r/core/schematics/application/index.ts
```diff
@@ -0,0 +1,76 @@
...
+    const angularOptions = getOptions(angularAppSchema);
+
+    return chain([
+      externalSchematic('@schematics/angular', 'application', {
+        ...Object.entries(extendedOptions).reduce((acc, [key, value]) => (angularOptions.includes(key) ? {...acc, [key]: value} : acc), {}),
+        name: extendedOptions.name,
+        projectRoot: extendedOptions.targetPath,
+        style: 'scss'}),
+      addProjectSpecificFiles(targetPath, rootDependencies),
+      schematic('ng-add', {
```
kpanot, line 67: may explicit the type of the schematic options (same for `externalSchematic('@schematics/angular'....)`)

packages/@o3r/core/schematics/application/index.ts
```diff
@@ -0,0 +1,76 @@
...
+        rootDependencies
+      }),
+      move(targetPath),
+      renameTemplateFiles()
+    ]), MergeStrategy.Overwrite);
+  };
+
+  const ngCliUpdate: Rule = (tree, context) => {
+
+    if (options.style) {
```
matthieu-crouzet, line 30: I would display the warning only if `style` is different from `scss`

packages/@o3r/core/schematics/application/index.ts
```diff
@@ -0,0 +1,76 @@
...
+      context.logger.warn('The provided "style" option will be ignored. "SCSS" will be used.');
+    }
+
+    const config = getWorkspaceConfig(tree);
+    if (!config) {
+      throw new Error('No workspace configuration file found');
+    }
```
matthieu-crouzet, line 37: Why should we throw an error if no workspace config is found ?
we use it only to get the defaultRoot if no path is provided
```suggestion
    if (!options.path && !config) {
      throw new Error('The `path` option is not provided and no workspace configuration file found to define it');
    }
```

packages/@o3r/core/schematics/application/index.ts
```diff
@@ -0,0 +1,76 @@
...
+    return chain([
+      externalSchematic('@schematics/angular', 'application', {
+        ...Object.entries(extendedOptions).reduce((acc, [key, value]) => (angularOptions.includes(key) ? {...acc, [key]: value} : acc), {}),
+        name: extendedOptions.name,
+        projectRoot: extendedOptions.targetPath,
+        style: 'scss'}),
+      addProjectSpecificFiles(targetPath, rootDependencies),
+      schematic('ng-add', {
+        ...extendedOptions,
+        skipInstall: true
```
matthieu-crouzet, line 69: why do we skip install ?

packages/@o3r/core/schematics/application/index.ts
vscaiceanu-1a: Not really, the `ng-add` method does the filtering, here we simply return the array of keys.

I could factorize to get the schema options but I think it's overkill: 
```
const getSchemaOptions = (schema: Schematic<any, any>) => {
  return schema.description.schemaJson?.properties || {};
};
```

mrednic-1A (approved):

kpanot (approved):

PR #716, opened 2023-09-01
fix: ng-add, some files incorrectly generated in the context of monorepo

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/core/schematics/ng-add/project-setup/index.ts
+++ b/packages/@o3r/core/schematics/ng-add/project-setup/index.ts
@@ -37,11 +37,11 @@ export const prepareProject = (options: NgAddSchematicsSchema) => async (tree: T
   }
   const o3rCoreVersion = corePackageJsonContent.version;
   const {
-    addVsCodeRecommendations, applyEsLintFix, getProjectFromTree, install, mapImportV7toV8, ngAddPackages,
+    addVsCodeRecommendations, applyEsLintFix, getWorkspaceConfig, install, mapImportV7toV8, ngAddPackages,
     readPackageJson, removePackages, renamedPackagesV7toV8, updateImports, isMultipackagesContext, getO3rPeerDeps
   } = await import('@o3r/schematics');
   const installOtterLinter = await shouldOtterLinterBeInstalled(context);
-  const workspaceProject = tree.exists('angular.json') ? getProjectFromTree(tree, options.projectName) : undefined;
+  const workspaceProject = options.projectName && getWorkspaceConfig(tree)?.projects?.[options.projectName] || undefined;
   const projectType = workspaceProject?.projectType;
   const depsInfo = getO3rPeerDeps(corePackageJsonPath);
   const internalPackagesToInstallWithNgAdd = Array.from(new Set([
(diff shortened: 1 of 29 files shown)


vscaiceanu-1a (approved):

mrednic-1A (approved):

packages/@o3r/core/schematics/ng-add/project-setup/index.ts
```diff
@@ -41,7 +41,7 @@ export const prepareProject = (options: NgAddSchematicsSchema) => async (tree: T
     readPackageJson, removePackages, renamedPackagesV7toV8, updateImports, isMultipackagesContext, getO3rPeerDeps
   } = await import('@o3r/schematics');
   const installOtterLinter = await shouldOtterLinterBeInstalled(context);
-  const workspaceProject = tree.exists('angular.json') ? getProjectFromTree(tree, options.projectName) : undefined;
+  const workspaceProject = (tree.exists('angular.json') && options.projectName) ? getProjectFromTree(tree, options.projectName) : undefined;
```
kpanot, line 44: the function `getProjectFromTree` should be deprecated, the implementation that defaults on the first of the list should be delegated to the schematic.
I would suggest to do a `options.projectName && getWorkspaceConfig(tree)?.projects?.[options.projectName] || undefined`

packages/@o3r/schematics/src/utility/modules.ts
```diff
@@ -104,11 +104,15 @@ export function getMainFilePath(tree: Tree, context: SchematicContext) {
  * one of its modules.
  *
  * @param tree
+ * @param options @see RuleFactory.options
```
kpanot, line 107: it is not supposed to be `{@see RuleFactory.options}` ?

fpaul-1A: I've copy pasted what we already have everywhere else
if we want to update that everywhere this should be done in a separate PR

packages/@o3r/core/schematics/ng-add/project-setup/index.ts
mrednic-1A: IMO this should be a dedicated task, as **getProjectFromTree** is extensively used across the schematic scripts of updates and in generators.

packages/@o3r/schematics/src/rule-factories/ng-add/index.ts
```diff
@@ -29,13 +32,13 @@ export function ngAddPackages(packages: string[], options?: Omit<NgAddPackageOpt
 
   const getInstalledVersion = (packageName: string) => {
     try {
-      const packageJsonAfterInstall = JSON.parse(readFileSync('./package.json', {encoding: 'utf8'}));
+      const fileSystemPackageJson = JSON.parse(readFileSync(path.join(fsWorkingDirectory, 'package.json'), {encoding: 'utf8'}));
```
matthieu-crouzet, line 35:
```suggestion
      const fileSystemPackageJson = JSON.parse(readFileSync(path.posix.join(fsWorkingDirectory, 'package.json'), {encoding: 'utf8'}));
```

fpaul-1A: I am building an actual file system path, so I need to use `path.join`

mrednic-1A (approved):

PR #738, opened 2023-09-06
fix: tree update in component presenter schematic

kpanot:


kpanot:
Closed because already applied to previous version and cascaded

Issue #736, opened 2023-09-05
[Bug]: `npm create @o3r my-workspace` create a commit before o3r is added


matthieu-crouzet:
### Package name

create

### Package version

9.1

### Reproduction steps

`npm create @o3r my-workspace`

### Current result

Repository is initialized with one commit that contains only the angular part, not the @o3r/core part.
The @o3r/core part is not committed.

### Expected result

1 initial commit with everything inside
or 2 commits, one for the angular/nx part and one for the add of @o3r/core

### Additional comments

Bonus: support the option `--skip-git`, as angular and nx.

kpanot:
It looks more to feature than a bug.

matthieu-crouzet:
For me it's a bug that you have uncommitted file after the 'create'

PR #739, opened 2023-09-06
feat: add commit at Otter setup

kpanot:
## Proposed change

Add commit message at Otter Framework setup

## Related issues

- :rocket: Feature #736


packages/@o3r/core/schematics/ng-add/index.ts
```diff
@@ -50,6 +50,13 @@ export function ngAdd(options: NgAddSchematicsSchema): Rule {
         const { OTTER_MODULE_KEYWORD, OTTER_MODULE_SUPPORTED_SCOPES } = await import('@o3r/schematics');
         const displayModuleListRule = displayModuleList(OTTER_MODULE_KEYWORD, OTTER_MODULE_SUPPORTED_SCOPES);
         return () => displayModuleListRule(t, c);
+      },
+      (_, c) => {
+        if (!options.projectName && !options.skipGit && options.commit) {
```
fpaul-1A, line 55: `options.projectName` will be automatically filled by `$source` even if not sent (if there is only one project)
are we ok with not trying to setup git if there already is a project ?
it feels weird since we'll do the setup if there are 2 projects

matthieu-crouzet (approved):

packages/@o3r/core/schematics/ng-add/index.ts
kpanot: The purpose is to commit only when the `ng-add @o3r/core` is apply to the workspace.
In case of 2 projects, if no `projectName` specified, the `ng add` will be applied to the workspace so I would like to commit

packages/@o3r/core/schematics/ng-add/schema.ts
```diff
@@ -41,4 +41,10 @@ export interface NgAddSchematicsSchema extends SchematicOptionObject {
 
   /** Enable Apis manager */
   enableApisManager: boolean;
+
+  /** Initial git repository commit information. */
+  commit: boolean | { name: string; email: string };
```
cpaulve-1A, line 46: The comment is not so clear on the possible values. It is a bit strange to allow a boolean or an object so I would put more information on the commit

kpanot: it is actually the one from the official Angular `ng new` schematic.
(same for the fact to accept boolean and object, I just followed the exact same type)

mrednic-1A (approved):

Issue #723, opened 2023-09-04
[Bug]: `npm create @o3r` fails + no arguments forwarded to ng add o3r/core 


mrednic-1A:
### Package name

create

### Package version

9.2.0-alpha.44

### Reproduction steps

npm create @o3r my-workspace --no-enable-customization

### Current result

unknown option enableCustomization

### Expected result

workspace created and the option passed taken into consideration for the `ng add` of `o3r core`

### Additional comments
Use case is to run the `o3r create` passing the arguments on command line, not interactive mode.

The comand is failing because the argument is passed to `ng new` command ( to the angular generator) which is called by o3r create.
A possible solution is to filter the arguments and remove the options not taken by the `ng new`.

Also, the arguments passed in the command line should be forwarded to `ng add @o3r/core`.

kpanot:
Incorrect Reproduce steps, the correct one is:
`npm create https://github.com/o3r my-workspace -- --no-enable-customization`

PR #740, opened 2023-09-06
feat: propagate the argument to the ng add otter schematic

kpanot:
## Proposed change

Propagate the argument to the ng add otter schematic

## Related issues

- :rocket: Feature #723

### DIFF
--- a/packages/@o3r/create/package.json
+++ b/packages/@o3r/create/package.json
@@ -16,6 +16,7 @@
   },
   "dependencies": {
     "@angular/cli": "~16.2.0",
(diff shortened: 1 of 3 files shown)


fpaul-1A (approved):

matthieu-crouzet (approved):

matthieu-crouzet:
Maybe we could add some unit tests to this package to validate what is propagated to `ng new` and `ng add @o3r/core`

matthieu-crouzet:
Should not be done in this PR but it could be good that we document somewhere all parameters that could be given to `[packageManager] create @o3r`

PR #743, opened 2023-09-06
fix: remove x-prompt for Otter ng-add

kpanot:


matthieu-crouzet (approved):

mrednic-1A (dismissed):

packages/@o3r/core/schematics/index.it.spec.ts
```diff
@@ -23,7 +23,7 @@ describe('new otter application', () => {
     execAppOptions.cwd = appFolderPath;
   });
   test('should build empty app', () => {
-    packageManagerExec(`ng add --skip-confirmation @o3r/core@${o3rVersion} --preset=cms`, execAppOptions);
+    packageManagerExec(`ng add --skip-confirmation @o3r/core@${o3rVersion} --preset=cms --enable-analytics`, execAppOptions);
```
fpaul-1A, line 26: it's weird to add a deprecated flag to prevent the test from failing :)

kpanot: There is an issue in the component creation that does not support the `--enable-analytics=false`.
But the fix of the component should not be part of this PR.

fpaul-1A: we actually want to test that the component generator works with analytics
I think we should manually add @o3r/analytics instead of using the deprecated flag
```suggestion
    packageManagerExec(`ng add --skip-confirmation @o3r/core@${o3rVersion} --preset=cms`, 
    packageManagerExec(`ng add --skip-confirmation @o3r/analytics@${o3rVersion}`, execAppOptions);
```

fpaul-1A (approved):

mrednic-1A (approved):

PR #744, opened 2023-09-06
fix(playwright): use @playwright/test instead of playwright

### LINKED ISSUE #735: [Bug]: @o3r/testing using both playwright and @playwright/test
### Package name

testing

### Package version

9.0

### Reproduction steps

-

### Current result

@o3r/testing/core is importing interfaces from "playwright" dependency (see [element.ts](https://github.com/AmadeusITGroup/otter/blob/6bd765e2b8b378b9418a4d259ade3e0cda85e91f/packages/%40o3r/testing/src/core/playwright/element.ts#L1)).
@o3r/testing schematics are generating files with @playwright/test dependency (see [template of playwright-spec.ts](https://github.com/AmadeusITGroup/otter/blob/6bd765e2b8b378b9418a4d259ade3e0cda85e91f/packages/%40o3r/testing/schematics/playwright/scenario/templates/__name%40dasherize__.e2e-playwright-spec.ts#L1C9-L1C9)).

From playwright 1.34, having both playwright and @playwright/test installed at the same time is not recommended anymore (see [[Playwright Docs] Breaking changes section in release notes](https://playwright.dev/docs/release-notes#%EF%B8%8F-breaking-changes-1)) due to a conflict in binaries exposed by those two packages.

This leads to an error message at runtime explaining that one of the packages needs to be uninstalled. If "playwright" dependency is uninstalled, the e2e tests don't compile anymore due to @o3r/testing/core relying on "playwright" dep.

### Expected result

As a user of otter framework, I would like to be able to update playwright to version 1.34+.
Sticking with @playwright/test should be OK to resolve this issue.

### Additional comments

Thanks!

matthieu-crouzet:
## Proposed change
Use @playwright/test instead of playwright

## Related issues

- :bug: Fixes #735

### DIFF
--- a/packages/@o3r/testing/package.json
+++ b/packages/@o3r/testing/package.json
@@ -105,9 +105,9 @@
     "@o3r/core": "workspace:^",
     "@o3r/localization": "workspace:^",
     "@o3r/schematics": "workspace:^",
+    "@playwright/test": "~1.37.0",
     "@schematics/angular": "~16.2.0",
     "pixelmatch": "^5.2.1",
-    "playwright": "~1.37.0",
     "pngjs": "^7.0.0",
     "protractor": "^7.0.0",
     "rxjs": "^7.8.1",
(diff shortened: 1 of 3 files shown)


guillaumerochelle:
I confirm this fixes our issue and playwright dependency is not needed anymore in our project.

mrednic-1A (approved):

kpanot (approved):

PR #756, opened 2023-09-08
feat: create a dedicated error for cli script with environment information

matthieu-crouzet:
## Proposed change
Create a dedicated error for cli script with environment information

Example of result in terminal 
image

### DIFF
--- a/packages/@ama-sdk/schematics/package.json
+++ b/packages/@ama-sdk/schematics/package.json
@@ -31,9 +31,6 @@
     "@angular/compiler": {
       "optional": true
     },
-    "@o3r/schematics": {
-      "optional": true
-    },
     "@openapitools/openapi-generator-cli": {
       "optional": true
     }
--- a/packages/@o3r/schematics/package.json
+++ b/packages/@o3r/schematics/package.json
@@ -16,6 +16,9 @@
     "prepare:build:builders": "yarn cpy 'schematics/**/*.json' dist/schematics && yarn cpy 'collection.json' dist",
     "build:builders": "tsc -b tsconfig.builders.json --pretty && yarn generate-cjs-manifest"
   },
+  "bin": {
+    "environment": "./dist/src/cli/environment.js"
+  },
   "peerDependencies": {
     "@angular-devkit/architect": "~0.1602.0",
     "@angular-devkit/core": "~16.2.0",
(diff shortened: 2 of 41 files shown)


packages/@ama-sdk/schematics/package.json
```diff
@@ -43,6 +40,7 @@
     "@angular-devkit/core": "^16.0.5",
     "@angular-devkit/schematics": "^16.0.5",
     "@angular-devkit/schematics-cli": "^16.0.5",
+    "@o3r/schematics": "workspace:^",
```
kpanot, line 43: Should not depend on `@o3r/schematics` which brings Angular.
This actually the reason to split @o3r/schematics

packages/@o3r/schematics/package.json
```diff
@@ -16,6 +16,9 @@
     "prepare:build:builders": "yarn cpy 'schematics/**/*.json' dist/schematics && yarn cpy 'collection.json' dist",
     "build:builders": "tsc -b tsconfig.builders.json --pretty && yarn generate-cjs-manifest"
   },
+  "bin": {
+    "environment": "./dist/src/cli/environment.js"
```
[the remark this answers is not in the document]
matthieu-crouzet: `@o3r/dev-tools` is a devDep of all packages and so of `@o3r/schematics`
I could move the helper into @o3r/dev-tools` but it means that it will become an optional peerDep of all packages
that in fact is mandatory if we want to run any schematics or builders.

kpanot (dismissed):

packages/@o3r/schematics/src/utility/environment.ts
```diff
@@ -0,0 +1,45 @@
...
+  const nodeInfo = {
+    version: nodeVersion && `${nodeVersion.major}.${nodeVersion.minor}.${nodeVersion.patch}`
+  };
+  const packageManagerInfo = getPackageManagerInfo();
+  let otterCorePackageJsonPath: string | undefined;
+  try {
+    otterCorePackageJsonPath = require.resolve('@o3r/core/package.json');
+  } catch {}
+  const otterInfo = {
+    version: otterCorePackageJsonPath ? JSON.parse(require.resolve('@o3r/core/package.json')).version as string : undefined
```
fpaul-1A, line 25:
```suggestion
    version: otterCorePackageJsonPath ? JSON.parse(readFileSync(otterCorePackageJsonPath, {encoding: 'utf8'})).version as string : undefined
```
?

fpaul-1A (approved):

kpanot (approved):

PR #757, opened 2023-09-08
fix: scripts in package json at monorepo creation

### LINKED ISSUE #721: [Bug]: Package json scripts at monorepo root level
### Package name

core

### Package version

9.1.0

### Reproduction steps

run o3r create to create a monorepo skelethon files

`npm run '@o3r' create my-wk-name`

### Current result

This will create monorepo skeleton files and preperare o3r generators inside the monorepo.
The _package.json_ which is created contains the scripts generated by angular generator for an app which does not work with multiple pacakges.

### Expected result
The `scripts` to be empty (at the time being).

mrednic-1A:
## Related issues

- :bug: Fixes #721

### DIFF
--- a/packages/@o3r/core/schematics/rule-factories/monorepo/index.ts
+++ b/packages/@o3r/core/schematics/rule-factories/monorepo/index.ts
@@ -2,13 +2,12 @@ import { chain } from '@angular-devkit/schematics';
 import type { Rule, SchematicContext, Tree } from '@angular-devkit/schematics';
 import { SchematicsException } from '@angular-devkit/schematics';
 import type { PackageJson } from 'type-fest';
-import {DEFAULT_ROOT_FOLDERS, isNxContext, setupSchematicsDefaultParams, WorkspaceLayout, WorkspaceSchematics} from '@o3r/schematics';
+import { DEFAULT_ROOT_FOLDERS, isNxContext, setupSchematicsDefaultParams, WorkspaceLayout, WorkspaceSchematics } from '@o3r/schematics';
 
 /**
  * Update root package.json to include workspaces
  * if ng context Update angular.json to include the workspaces definitions into schematics default params
  * if nx context update nx.json workspacelayout
- *
  * @param directories workspaces directories to include in package.json
  */
 export function addWorkspacesToProject(directories: WorkspaceLayout = DEFAULT_ROOT_FOLDERS): Rule {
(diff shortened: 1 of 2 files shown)


packages/@o3r/create/src/index.ts
```diff
@@ -119,8 +120,26 @@ const addOtterCore = (relativeDirectory = '.') => {
   }
 };
 
+/**
+ * When the monorepo is created, the package json will keep only the ng script
+ * @param relativeDirectory package.json directory
+ */
+const cleanPackageJsonScripts = (relativeDirectory = '.') => {
```
kpanot, line 127: Should avoid as much as possible to do content change in this script.
I would suggest to do the change of the `package.json` (maybe by removing the specific scripts that does not make explicitly sense, not replacing the whole `script` field) in the `ng add @or3/core`

packages/@o3r/create/src/index.ts
```diff
@@ -119,8 +120,26 @@ const addOtterCore = (relativeDirectory = '.') => {
...
+  const packageJsonPath = resolve(process.cwd(), relativeDirectory, 'package.json');
+  try {
+    const packageJsonContent = JSON.parse(readFileSync(packageJsonPath, { encoding: 'utf-8' })) as PackageJson;
+    packageJsonContent.scripts = { ng: 'ng' };
+    writeFileSync(packageJsonPath, JSON.stringify(packageJsonContent, null, 2));
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.error(error);
+    process.exit(2);
```
matthieu-crouzet, line 136:
```suggestion
    process.exit(3);
```
2 is already use line 119

packages/@o3r/create/src/index.ts
mrednic-1A: I agree that it is not a good idea to complexify the `create`. After discussing the solution you proposed to remove the specific scripts it's what we need. I'll do the updates accordingly

packages/@o3r/create/src/index.ts
mrednic-1A: the modification on this file was fully discarded

matthieu-crouzet (approved):

packages/@o3r/core/schematics/rule-factories/monorepo/index.ts
```diff
@@ -49,3 +48,31 @@ export function addWorkspacesToProject(directories: WorkspaceLayout = DEFAULT_RO
...
+ */
+export function filterPackageJsonScripts(tree: Tree, _context: SchematicContext) {
+  const rootPackageJsonPath = '/package.json';
+  if (!tree.exists(rootPackageJsonPath)) {
+    throw new SchematicsException('Root package.json does not exist');
+  }
+
+  const rootPackageJsonObject = tree.readJson(rootPackageJsonPath) as PackageJson;
+  rootPackageJsonObject.scripts ||= {};
+  rootPackageJsonObject.scripts = Object.entries(rootPackageJsonObject.scripts).reduce<Record<string, string | undefined>>((acc, [scriptName, scriptValue]) => {
```
kpanot, line 65: would be simpler with a map:
```typescript
const defaultScriptToRemote = {
  build: 'ng build',
  test: 'ng test',
  start: 'ng serve',
  watch: 'ng build --watch --configuration development'
};

rootPackageJsonObject.scripts = Object.fromEntries(
  Object.entries(rootPackageJsonObject.scripts)
    .filter(([key, value]) => value !== defaultScriptToRemote[key])
);
```

fpaul-1A (approved):

kpanot (approved):

PR #763, opened 2023-09-08
fix: showcase build fail on initial setup

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/apps/showcase/project.json
+++ b/apps/showcase/project.json
@@ -96,9 +96,7 @@
           "showcase:generate-translations:development",
           "showcase:serve-app"
         ]
-      },
-      "dependsOn": [
-      ]
+      }
     },
     "serve-app": {
       "executor": "@angular-devkit/build-angular:dev-server",
@@ -131,7 +129,7 @@
         "exposedComponentSupport": true
       },
       "dependsOn": [
-        "^build-source"
+        "build-source"
       ]
     },
     "extract-rules-engine": {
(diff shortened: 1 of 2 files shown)


packages/@o3r/core/project.json
```diff
@@ -14,6 +14,7 @@
         "script": "postbuild"
       },
       "dependsOn": [
+        "^build",
```
kpanot, line 17: It is supposed to be the job of the `compile` task

apps/showcase/project.json
```diff
@@ -30,6 +30,7 @@
         "script": "postbuild:patch"
       },
       "dependsOn": [
+        "^build",
```
kpanot, line 33: Not needed here, it is part of the `build-source`.
I also think that the `build-source` default dependency from `nx.json` is already including the `^build`

packages/@o3r/core/project.json
fpaul-1A: indeed, this was unnecessary

nx.json
```diff
@@ -187,6 +187,9 @@
       ]
     },
     "build-source": {
```
kpanot, line 189: Maybe simpler to rename the `build-source` from the `showcase` to `compile` to be align with the other packages

kpanot (approved):

nx.json
fpaul-1A: I'd say it might be better to keep it as close as possible as the `angular.json` generated from new apps

mrednic-1A (approved):

PR #764, opened 2023-09-09
fix: add PR label calculation for fork PR

kpanot:


.github/workflows/pr-labels.yml
```diff
@@ -1,9 +1,12 @@
 name: Pull Request Label
 
-on: pull_request
+on:
+  workflow_run:
```
fpaul-1A, line 4: This workflow_run will execute on `main` branch regardless of the PR that triggered it, so you might need to change a few things on this script to target the right commit

PR #779, opened 2023-09-12
feat: support incremental ts build for non-ng-packagr tasks

kpanot:


kpanot:
Close because it is still limited to ng-packagr build that does not support incremental build

PR #783, opened 2023-09-13
perf: make component generators rely on tree instead of file system

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


packages/@o3r/core/schematics/page/index.ts
```diff
@@ -164,37 +164,28 @@ export function ngGeneratePage(options: NgGeneratePageSchematicsSchema): Rule {
...
+        o3rStylePath,
+        options
+      ),
+      getAddLocalizationRules(
+        componentPath,
+        options
+      ),
+      getAddFixtureRules(
+        componentPath,
+        {
```
kpanot, line 182: There is a reason to not pass the full option list as previously?

kpanot (approved):

PR #787, opened 2023-09-13
feat: migrate to nx dependency check

kpanot:


apps/chrome-devtools/project.json
```diff
@@ -165,7 +165,8 @@
       "options": {
         "eslintConfig": "apps/chrome-devtools/.eslintrc.js",
         "lintFilePatterns": [
-          "apps/chrome-devtools/src/**/*.ts"
+          "apps/chrome-devtools/src/**/*.ts",
+          "apps/chrome-devtools/package.json"
```
mrednic-1A, line 169: Dummy question: Why not adding the project.json, too ?

kpanot: Today we have no rule applied to the `project.json` (or any `.json` generally, only to the specific `package.json`)
cf: https://nx.dev/packages/eslint-plugin/documents/dependency-checks

fpaul-1A (approved):

fpaul-1A (approved):

apps/chrome-devtools/project.json
mrednic-1A: ok :) thanks for details

mrednic-1A (approved):

PR #789, opened 2023-09-13
fix(generator-sdk): invalid content-type for form-encoded POST requests

### LINKED ISSUE #788: [Bug]: invalid content-type when generating a POST request with openapi generator
### Package name

generator-sdk

### Package version

Latest alpha

### Reproduction steps

Generate an api library using the newly schematic `ng generate sdk` and use a spec yaml file that contains a POST request with `consumes: -application/x-www-form-urlencoded`

Launch the typescript services / models generation.

### Current result

Observe that the typescript api will contain the following:
```typescript
  const headers: { [key: string]: string | undefined } = { 
    'Content-Type': metadata?.headerContentType || 'application/json',
    ...(metadataHeaderAccept ? {'Accept': metadataHeaderAccept} : {})
  };
```

### Expected result

The part above should be:
```typescript
  const headers: { [key: string]: string | undefined } = { 
    'Content-Type': metadata?.headerContentType || 'application/x-www-form-urlencoded',
    ...(metadataHeaderAccept ? {'Accept': metadataHeaderAccept} : {})
  };
```

### Additional comments

This is due to a regression in the api.mustache file when switching to the openapi generator.
We are using HeaderJsonMimeType custom lambda but giving an invalid input.

quentinderoubaix:
## Proposed change

A proposed change to fix #788 consisting on removing the regression in the api.mustache file for openapi generator.

### DIFF
--- a/packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/resources/typescriptFetch/api/api.mustache
+++ b/packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/resources/typescriptFetch/api/api.mustache
@@ -74,7 +74,7 @@ export class {{classname}} implements Api {
     const getParams = this.client.extractQueryParams<{{#uppercaseFirst}}{{nickname}}{{/uppercaseFirst}}RequestData>(data, [{{#trimComma}}{{#queryParams}}'{{baseName}}', {{/queryParams}}{{/trimComma}}]{{^queryParams}} as never[]{{/queryParams}});
     const metadataHeaderAccept = metadata?.headerAccept || '{{#headerJsonMimeType}}{{#produces}}{{mediaType}}{{^-last}}, {{/-last}}{{/produces}}{{/headerJsonMimeType}}';
     const headers: { [key: string]: string | undefined } = { {{#trimComma}}
-      'Content-Type': metadata?.headerContentType || '{{#headerJsonMimeType}}{{#consumes}}{{mediaType}}, {{/consumes}}application/json{{/headerJsonMimeType}}',
+      'Content-Type': metadata?.headerContentType || '{{#headerJsonMimeType}}{{#consumes}}{{mediaType}}{{^-last}}, {{/-last}}{{/consumes}}{{/headerJsonMimeType}}',
       ...(metadataHeaderAccept ? {'Accept': metadataHeaderAccept} : {}),
       {{#headerParams}}'{{baseName}}': data['{{baseName}}'],{{/headerParams}}{{/trimComma}}
     };


packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/resources/typescriptFetch/api/api.mustache
```diff
@@ -74,7 +74,7 @@ export class {{classname}} implements Api {
     const getParams = this.client.extractQueryParams<{{#uppercaseFirst}}{{nickname}}{{/uppercaseFirst}}RequestData>(data, [{{#trimComma}}{{#queryParams}}'{{baseName}}', {{/queryParams}}{{/trimComma}}]{{^queryParams}} as never[]{{/queryParams}});
     const metadataHeaderAccept = metadata?.headerAccept || '{{#headerJsonMimeType}}{{#produces}}{{mediaType}}{{^-last}}, {{/-last}}{{/produces}}{{/headerJsonMimeType}}';
     const headers: { [key: string]: string | undefined } = { {{#trimComma}}
-      'Content-Type': metadata?.headerContentType || '{{#headerJsonMimeType}}{{#consumes}}{{mediaType}}, {{/consumes}}application/json{{/headerJsonMimeType}}',
+      'Content-Type': metadata?.headerContentType || '{{#headerJsonMimeType}}{{#consumes}}{{mediaType}}{{^-last}}, {{/-last}}{{/consumes}}{{/headerJsonMimeType}}',
```
kpanot, line 77: I guess that the previous purpose of `application/json` is to provide a default value in case `consumes` array is empty.
Your change suggest that the `consumes` is always populated with a default `application/json` if not specified in the spec, is it the case?
If not, an additiomal fallback mechanism to `application/json` may need to be added

quentinderoubaix: The ```headerJsonMimeType``` lambda returns only one mime, and it prioritizes the first json mime found.
In the regression, we give as inputs to this lambda an array composed of [...theMimesTakenFromTheSpecConsume, 'application/json']. So the lambda will always return a json mime, whatever the mimes defined in the spec are.

As you guessed, the above proposed change only removes the invalid concatenation and does not provide a fallback mechanism. For the fallback, we could update the HeaderJsonMimeType implementation to return the `application/json` mime if the input array is empty.

mrednic-1A:
Thanks @quentinderoubaix for your contribution.
I have included your fix in another PR #822 (as I could not push on your branch :D).
Therefore I am closing this one.

PR #801, opened 2023-09-15
fix: git ignore node_modules of monorepo projects

### LINKED ISSUE #795: [Bug]: node_modules should be ignored across workspaces
### Package name

core

### Package version

9.3.0

### Reproduction steps

When creating a new Otter monorepo, the `.gitignore` (that comes from the Angular generator) contains `/node_modules`.

### Current result

When adding an app/lib/sdk to the monorepo, in some cases, the `node_modules` folder is present at project level, for `bin`s for instance. These folders are not gitignored.

### Expected result

All `node_modules` folders across the repo should be gitignored.

### Additional comments

_No response_

mrednic-1A:
## Related issues

- :bug: Fixes #795 
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/core/schematics/rule-factories/generic-updates/index.ts
+++ b/packages/@o3r/core/schematics/rule-factories/generic-updates/index.ts
@@ -1,0 +1,30 @@
+import type { Rule, SchematicContext, Tree } from '@angular-devkit/schematics';
+
+/**
+ * Generic basic updates to be done on a repository
+ */
+export function genericUpdates(): Rule {
+
+  const updateGitIgnore = (tree: Tree, _context: SchematicContext) => {
+    // update gitignore
+    if (tree.exists('/.gitignore')) {
+      let gitignore = tree.readText('/.gitignore');
+      const foldersToExclude = ['node_modules', 'dist'];
+      const foldersToExcludeRegExp = new RegExp(`(\r\n|\n)/(${foldersToExclude.join('|')})(\r\n|\n)`, 'gm');
+      gitignore = gitignore.replaceAll(foldersToExcludeRegExp, '$1$2$3');
+      foldersToExclude.forEach(folderToExclude => {
+        if (!gitignore.includes(folderToExclude)) {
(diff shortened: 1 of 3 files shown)


packages/@o3r/core/schematics/rule-factories/generic-updates/index.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Generic basic updates to be done on a repository
+ */
+export function genericUpdates(): Rule {
+
+  const updateGitIgnore = (tree: Tree, _context: SchematicContext) => {
+    // update gitignore
+    if (tree.exists('/.gitignore')) {
+      let gitignore = tree.readText('/.gitignore');
+
+      gitignore = gitignore.replace(/(\r\n|\n)\/node_modules(\r\n|\n)/, '$1node_modules$2');
```
fpaul-1A, line 13: when working on the it-tests, I also had issues with the `/dist` folder only being ignored at root level

fpaul-1A, line 13: why not `/^\/node_modules$/` ?

mrednic-1A: I'll take that into consideration too. I'll do an update.

mrednic-1A: `tree.readText` is giving a single string as output, replacing the new lines.

Here is a small snapshot:
`# See http://help.github.com/ignore-files/ for more about ignoring files.\r\n\r\n# Compiled output\r\n/dist\r\n/tmp\r\n/out-tsc\r\n/bazel-out\r\n\r\n# Node\r\n/node_modules\r\nnpm-debug.log\r\nyarn-error.log\r\n\r\n# IDEs`

vscaiceanu-1a (dismissed):

fpaul-1A (approved):

kpanot (approved):

PR #766, opened 2023-09-11
feat: split core package to workspace module

kpanot:

Issue #599, opened 2023-07-31
Convert version-harmonize to Yarn Plugin


kpanot:
### Context

The version-harmonize tool, provided by `@o3r/dev-tools`, is aligning the versions of the different package's dependencies of the workspaces.
To be able to simplify and automatize the usage of the version harmonization (that can be done at the installation time), the version harmonize can be provided as yarn plugin.

### Proposal

Expose the version-harmonize script as [yarn plugin](https://yarnpkg.com/features/plugins).
The script may requires to have it own package

**Links**:
- [Writing our first plugin](https://yarnpkg.com/advanced/plugin-tutorial#writing-our-first-plugin)

vscaiceanu-1a:
Does this mean that our `npm` users won't be able to benefit from the version harmoniser?

kpanot:
The purpose is not to dedicated the feature to yarn only, but to expose the interface for a yarn plugin

kpanot:
Actually a good compromise could be to expose to version harmonize as ESLint rule

PR #807, opened 2023-09-19
Migrate version harmonize to linter

kpanot:
## Proposed change

Migrate version harmonize to linter

## Related issues

- :rocket: Feature #599
- :clock9: Waiting PR #787, #766 

## Missing

- [ ] Update ng-add workspace to include the harmonize linter (Waiting for #766 )


fpaul-1A (dismissed):

packages/@o3r/eslint-plugin/tsconfig.build.json
```diff
@@ -1,8 +1,10 @@
 {
   "extends": "../../../tsconfig.build",
   "compilerOptions": {
-    "declaration": false,
+    "declaration": true,
     "declarationMap": false,
+    "incremental": true,
+     "composite": true,
```
matthieu-crouzet, line 7: alignment issue 
```suggestion
    "composite": true,
```

matthieu-crouzet (approved):

matthieu-crouzet (approved):

packages/@o3r/eslint-plugin/src/rules/json/json-dependency-versions-harmonize/json-dependency-versions-harmonize.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+  dependencies: {
+    myDep: '^1.0.0',
+    myOtherDep: '~2.0.0'
+  }
+};
+
+const packageToLint = path.resolve(__dirname, 'local', 'packages', 'my-package', 'package.json');
+
+beforeAll(() => {
+  fs.mkdirSync(path.resolve(__dirname, 'local'));
```
fpaul-1A, line 34: it would probably be better to not rely on filesystem for the tests and mock `node:fs` module

packages/@o3r/eslint-plugin/src/rules/json/json-dependency-versions-harmonize/version-harmonize.ts
```diff
@@ -0,0 +1,98 @@
...
+    if (!parent || parent.hasDefaulted) {
+      return {
+        hasDefaulted: true,
+        packages: getPackages()
+      };
+    }
+  }
+};
+
+export const isRangeBetter = (currentRange?: string, range?: string) => {
```
fpaul-1A, line 61: missing doc

packages/@o3r/eslint-plugin/src/rules/json/json-dependency-versions-harmonize/version-harmonize.ts
```diff
@@ -0,0 +1,98 @@
...
+    if (semver.gt(minVersion, currentMinVersion)) {
+      return range;
+    } else if (semver.eq(minVersion, currentMinVersion) && semver.subset(range, currentRange)) {
+      return range;
+    }
+  }
+  return currentRange;
+};
+
+export const getBestRanges = (dependencyTypes: string[], packages: PackageProperty['packages']) => {
```
fpaul-1A, line 80: missing odc

fpaul-1A (dismissed):

packages/@o3r/eslint-plugin/src/rules/json/json-dependency-versions-harmonize/json-dependency-versions-harmonize.spec.ts
kpanot: Actually I just copy the UT from another plugin test we have.
Unfortunately, mocking `node:fs` may not be enough because of `globby` and mocking both of them will make the test more unitary but less meaningful :S

fpaul-1A: I'm not sure to understand your point, I think you do want the `fs` used by ESLint to be mocked as well so that it doesn't read from the disk

fpaul-1A (approved):

packages/@o3r/eslint-plugin/src/rules/json/json-dependency-versions-harmonize/json-dependency-versions-harmonize.spec.ts
kpanot: `TSESLint` is using `fs` internally to manage different files loading (and using `buffer`, not `string` content...).
As we cannot mock `fs` only for our code with Jest, it's not possible to correctly mock `fs` to handle everything needed by `TSESlint`.

docs/linter/eslint-plugin/rules/json-dependency-versions-harmonize.md
```diff
@@ -0,0 +1,99 @@
...
+    }
+  ]
+}
+```
+
+### dependencyTypes
+
+List of dependency types to update
+
+**Default**: ['optionalDependencies', 'dependencies', 'devDependencies', 'peerDependencies', 'generatorDependencies']*
```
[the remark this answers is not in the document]
kpanot: I just move the existing script but to reply, I would say "yes".
If it does not exist in the targeted package.json it will not do anything in any case.
(to compare, Renovate is based on the regex `/dependencies$/i`, so we are still less permissive)

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #809, opened 2023-09-20
test: add checks based on git diff in it tests

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/analytics/schematics/index.it.spec.ts
+++ b/packages/@o3r/analytics/schematics/index.it.spec.ts
@@ -1,6 +1,7 @@
 import {
   addImportToAppModule,
   getDefaultExecSyncOptions,
+  getGitDiff,
   packageManagerAdd,
   packageManagerExec,
   packageManagerInstall,
@@ -22,12 +23,16 @@ describe('new otter application with analytics', () => {
       appFolderPath = await prepareTestEnv(appName, 'angular-with-o3r-core');
       execAppOptions.cwd = appFolderPath;
     });
-    test('should add analytics to existing application', () => {
+    test('should add analytics to existing application', async () => {
       packageManagerExec(`ng add --skip-confirmation @o3r/analytics@${o3rVersion}`, execAppOptions);
 
       packageManagerExec('ng g @o3r/core:component test-component --use-otter-analytics=false', execAppOptions);
       packageManagerExec('ng g @o3r/analytics:add-analytics --path="src/components/test-component/test-component.component.ts"', execAppOptions);
-      addImportToAppModule(appFolderPath, 'TestComponentModule', 'src/components/test-component');
+      await addImportToAppModule(appFolderPath, 'TestComponentModule', 'src/components/test-component');
+
+      const diff = getGitDiff(appFolderPath);
+      expect(diff.modified).toContain('package.json');
+      expect(diff.added).toContain('src/components/test-component/test-component.analytics.ts');
 
       expect(() => packageManagerInstall(execAppOptions)).not.toThrow();
       expect(() => packageManagerRun('build', execAppOptions)).not.toThrow();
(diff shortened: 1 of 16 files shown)


packages/@o3r/analytics/schematics/ng-add/index.ts
```diff
@@ -1,11 +1,22 @@
...
  */
-export function ngAdd(): Rule {
+export function ngAdd(options: NgAddSchematicsSchema): Rule {
   /* ng add rules */
-  return () => {};
+  return async (tree) => {
+    const { ngAddPackages, getProjectFromTree } = await import('@o3r/schematics');
+    const packageJsonPath = path.resolve(__dirname, '..', '..', 'package.json');
+    const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, { encoding: 'utf-8' }));
+    const workspaceProject = tree.exists('angular.json') ? getProjectFromTree(tree, options.projectName) : undefined;
```
kpanot, line 17:
```suggestion
    const workspaceProject = getProjectFromTree(tree, options.projectName);
```

packages/@o3r/analytics/schematics/index.it.spec.ts
```diff
@@ -54,6 +55,12 @@ describe('new otter application with analytics', () => {
       );
       addImportToAppModule(appFolderPath, 'TestComponentContModule', 'projects/test-app/src/components/test-component');
 
+      const diff = getGitDiff(execAppOptions.cwd as string);
```
kpanot, line 58: This is a bit confusing because rely on the time the commit has been done (and the file it contains).
Maybe it would make more sense to have this diff in output of the `packageManagerExec` directly?

packages/@o3r/analytics/schematics/ng-add/index.ts
```diff
@@ -1,11 +1,22 @@
...
+    const workingDirectory = workspaceProject?.root;
+
+    return ngAddPackages(['@o3r/analytics'], {...options, workingDirectory, version: packageJson.version, skipNgAddSchematicRun: true});
```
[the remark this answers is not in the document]
fpaul-1A: `ng-add` will add the dependency at root level by itself, but I need to add it at project level also for monorepo
probably worth a comment a in the code

packages/@o3r/analytics/schematics/index.it.spec.ts
fpaul-1A: the only commit done is at the end of the prepareTestEnv, so the gitDiff contains only what have been changed inside the test itself
I want to test that at the end of the test, the project `dont-modify-me` which is the project[0] hasn't been modified, and I didn't want to test it after each and every command

packages/@o3r/analytics/schematics/ng-add/index.ts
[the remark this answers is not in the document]
kpanot: So if I understand correctly: running `npx ng add @o3r/analytics --projectName=my-app` will add the project at root level only, not in the sub-package?
Then what's about `npm -W my-app exec ng add @o3r/analytics` scenario?

packages/@o3r/analytics/schematics/index.it.spec.ts
kpanot: But we have a task to create a commit after the `ng add @o3r/core`, this may be the case of several rules actually.
I suggest to provide a return of `packageManagerExec` to not be dependent of the task implementation

packages/@o3r/analytics/schematics/ng-add/index.ts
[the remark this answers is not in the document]
fpaul-1A: both will add `@o3r/analytics` only in the root `package.json`, the ng-add is correctly executed for the project
I just want to add the dep in the `package.json` of the sub-package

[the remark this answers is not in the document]
fpaul-1A: as discussed, I extracted the part to only update package.json in tree in a new function `addDependenciesInPackageJson` to make it less misleading

packages/@o3r/analytics/schematics/index.it.spec.ts
fpaul-1A: I've added a git tag after the initialization of the test environment to handle the case where commits are added during the test

packages/@o3r/analytics/schematics/ng-add/index.ts
fpaul-1A: I've updated it according to https://github.com/AmadeusITGroup/otter/pull/811

sdo-1A (approved):

matthieu-crouzet (approved):

PR #813, opened 2023-09-21
fix: do not export ama-sdk-core as es module by default

mrednic-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@ama-sdk/core/package.json
+++ b/packages/@ama-sdk/core/package.json
@@ -29,9 +29,9 @@
     "build": "yarn nx build ama-sdk-core",
     "build:cjs": "swc src -d dist/cjs -C module.type=commonjs -q",
     "build:esm2015": "swc src -d dist/esm2015 -C module.type=es6 -q",
-    "build:esm2020": "tsc -b tsconfig.esm2020.json",
+    "build:esm2020": "tsc -b tsconfig.build.json",
     "postbuild": "patch-package-json-main",
-    "prepare:build:builders": "yarn cpy 'schematics/**/*.json' dist/schematics && yarn cpy 'collection.json' dist",
+    "prepare:build:builders": "yarn cpy 'schematics/**/*.json' dist/schematics && yarn cpy '{collection,package}.json' dist",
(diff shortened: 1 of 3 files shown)


packages/@ama-sdk/core/project.json
```diff
@@ -27,7 +27,7 @@
         "script": "build:esm2015"
       },
       "dependsOn": [
-        "compile"
```
kpanot, line 30: Why did you change the default `compile` name which comes with default config?

kpanot (approved):

packages/@ama-sdk/core/project.json
mrednic-1A: There is no technical reason behind the change. I changed it to be aligned on naming with the other configs.

packages/@ama-sdk/core/package.json
```diff
@@ -29,9 +29,9 @@
     "build": "yarn nx build ama-sdk-core",
     "build:cjs": "swc src -d dist/cjs -C module.type=commonjs -q",
     "build:esm2015": "swc src -d dist/esm2015 -C module.type=es6 -q",
-    "build:esm2020": "tsc -b tsconfig.esm2020.json",
+    "build:esm2020": "tsc -b tsconfig.build.json",
```
vscaiceanu-1a, line 32: `tsconfig.esm2020.json` still exist. Is is used somewhere else or can we remove it?

vscaiceanu-1a (approved):

fpaul-1A (approved):

packages/@ama-sdk/core/package.json
mrednic-1A: No, the file does not exist. The reason why we got nothing before is that the script `build:esm2020` was never called. (this build was done by the `compile` configuration which used the nx/js:tsc builder - which added the type: module in package.json btw. So I removed nx builder and align the naming with other builds and call this script defined in package.json)

vscaiceanu-1a: I understand. Thanks!

PR #830, opened 2023-09-27
feat: showcase layout

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


apps/showcase/src/styles.scss
```diff
@@ -1,28 +1,31 @@
-// Issue with pnp
-// @import "bootstrap/scss/bootstrap";
-// @import "bootstrap-icons/font/bootstrap-icons";
-@import "https://cdn.jsdelivr.net/npm/bootstrap@5.3.1/dist/css/bootstrap.min.css";
-@import "https://cdn.jsdelivr.net/npm/bootstrap-icons@1.10.2/font/bootstrap-icons.css";
+// FIXME sass-loader cannot resolve scss files with pnp
```
kpanot, line 1: seems not required anymore

apps/showcase/src/app/app.component.html
```diff
@@ -1,11 +1,38 @@
...
-      [linksGroups]="linksGroups"
-      [activeUrl]="activeUrl$ | async"
-    ></o3r-sidenav-pres>
+<header>
+  <div class="header-icon">
+    <img src="assets/otter-sticker.svg" width="48" height="48" alt="Otter logo" />
   </div>
-  <div class="page-content flex-fill">
-    <router-outlet></router-outlet>
+  Otter Showcase
```
kpanot, line 5: missing localization

apps/showcase/src/app/app.component.html
```diff
@@ -1,11 +1,38 @@
...
+    </div>
   </div>
 </div>
+<footer>
+  <div class="container">
+    <div class="row">
+      <div class="col-8">
+      </div>
+      <div class="col-4">
+        <h5>Links</h5>
```
kpanot, line 26: missing loc

apps/showcase/src/assets/otter-sticker.svg
kpanot, line 1: the left hear halo is a bit strange

apps/showcase/src/app/app.component.scss
```diff
@@ -1,5 +1,44 @@
...
+  top: 0;
+  height: 4rem;
+  width: 100%;
+  background: var(--bs-primary);
+  box-shadow: 0 0.5rem 1rem rgba(0,0,0,0.15), inset 0 -1px 0 rgba(255,255,255,0.15);
+  z-index: 1021;
+  display: flex;
+  align-items: center;
+  justify-content: flex-start;
+  padding-left: 1rem;
```
kpanot, line 12: should not be handle by bootstrap css class (same for margins)?

apps/showcase/src/app/app.component.html
cpaulve-1A: do we really use the translation reco from otter in the showcase?

apps/showcase/src/styles.scss
fpaul-1A: we can import `css` files but not `scss` files

apps/showcase/src/assets/otter-sticker.svg
fpaul-1A: I've just changed the color :D
anyway the icon is displayed as 48x48 so I don't think it's very noticeable

apps/showcase/src/app/app.component.html
kpanot: If the "showcase" is a showcase of the Otter capabilities, I would say yes :)

fpaul-1A: it needs to be done globally on all the texts of the showcase app, it should be done in a separate PR

apps/showcase/src/app/app.component.scss
fpaul-1A: removed

apps/showcase/src/app/app.component.html
fpaul-1A: localization will be handled separately

fpaul-1A:
> Could be good to have some screenshot :)

image

PR #832, opened 2023-09-28
feat: deploy showcase in github pages

fpaul-1A:
## Proposed change

https://fpaul-1a.github.io/otter

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/.github/workflows/deploy-showcase.yml
+++ b/.github/workflows/deploy-showcase.yml
@@ -1,0 +1,46 @@
+# Simple workflow for deploying static content to GitHub Pages
+name: Deploy static content to Pages
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["main"]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
+permissions:
+  contents: read
+  pages: write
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-showcase.yml
```diff
@@ -0,0 +1,46 @@
...
+
+jobs:
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    env:
+      NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
+    steps:
```
kpanot, line 32: Do you think it worth to deploy only if the app is touched?

kpanot:
Let's merge it after redesign:
- [ ] we are happy with the look and field

.github/workflows/deploy-showcase.yml
fpaul-1A: I would say yes, the app and its dependencies

fpaul-1A: this might not be as easy for a branch build

cpaulve-1A (approved):

vscaiceanu-1a (approved):

cpaulve-1A (approved):

PR #835, opened 2023-09-28
feat(dpf-18714): inNextMinutes operator

vscaiceanu-1a:
## Proposed change

Computing the input facts takes:
- 3.7748750001192093 ms for 1k rules
- 38.763249998912215 ms for 10k rules

### DIFF
--- a/apps/showcase/src/assets/rules/rulesets.json
+++ b/apps/showcase/src/assets/rules/rulesets.json
@@ -7,11 +7,10 @@
       "disabled": false,
       "rules": [
         {
-          "id": "5467e501-b9ff-414f-8026-56885d0d7a4c",
+          "id": "5467e501-b9ff-414f-8026-56885d0d7a4b",
           "name": "New-York availability",
           "disabled": false,
           "outputRuntimeFacts": [],
-          "inputFacts": ["outboundDate"],
           "inputRuntimeFacts": [],
           "rootElement": {
             "elementType": "RULE_BLOCK",
@@ -57,7 +56,6 @@
           "name": "Destination selected",
           "disabled": false,
           "outputRuntimeFacts": [],
-          "inputFacts": ["destination"],
           "inputRuntimeFacts": [],
           "rootElement": {
             "elementType": "RULE_BLOCK",
@@ -85,11 +83,10 @@
           }
         },
         {
-          "id": "5467e501-b9ff-414f-8026-56885d0d7a4c",
-          "name": "The otter is in vacations",
+          "id": "5467e501-b9ff-414f-8026-56885d0d7a4d",
+          "name":"The otter is on vacation",
           "disabled": false,
           "outputRuntimeFacts": [],
-          "inputFacts": ["outboundDate"],
           "inputRuntimeFacts": [],
           "rootElement": {
             "elementType": "RULE_BLOCK",
@@ -109,6 +106,41 @@
                     "value": "outboundDate"
                   },
                   "operator": "duringSummer"
+                }
+              ]
+            },
+            "successElements": [
+              {
+                "elementType": "ACTION",
+                "actionType": "UPDATE_ASSET",
+                "asset": "otter.svg",
+                "value": "otter-summer.svg"
+              }
+            ],
+            "failureElements": []
+          }
+        },
+        {
+          "id": "5467e501-b9ff-414f-8026-56885d0d7a4c",
+          "name": "The otter is late",
+          "disabled": false,
+          "outputRuntimeFacts": [],
+          "inputRuntimeFacts": [],
+          "rootElement": {
+            "elementType": "RULE_BLOCK",
+            "blockType": "IF_ELSE",
+            "condition": {
+              "all": [
+                {
+                  "lhs": {
+                    "type": "FACT",
+                    "value": "outboundDate"
+                  },
+                  "rhs": {
+                      "type": "LITERAL",
+                      "value": "2880"
+                  },
+                  "operator": "dateInNextMinutes"
                 }
               ]
             },
--- a/packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.ts
+++ b/packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.ts
@@ -277,6 +277,14 @@ export class RulesEngineExtractor {
               nbValues: 1
             }
           };
+          if (declaration.initializer && ts.isObjectLiteralExpression(declaration.initializer)) {
+            declaration.initializer.properties.forEach((property: ts.ObjectLiteralElementLike) => {
+              if (ts.isPropertyAssignment(property) && property.name.getText(source) === 'factImplicitDependencies') {
+                operator.factImplicitDependencies = JSON.parse(property.initializer.getText(source).replace(/'/g,'"'));
+              }
+            });
+          }
+
           if (operatorType === 'Operator') {
             operator.rightOperand = {
               types: allDefaultSupportedTypes,
--- a/packages/@o3r/rules-engine/src/engine/operator/operator.helpers.ts
+++ b/packages/@o3r/rules-engine/src/engine/operator/operator.helpers.ts
@@ -1,22 +1,34 @@
 import {DateInput, Operator, SupportedSimpleTypes} from './operator.interface';
+import type {Facts} from '../fact';
 
 /**
  * Execute Operator
- *
  * @param lhs Left hand side
  * @param rhs Right hand side
  * @param operator Operator to compare values
+ * @param operatorFacts Facts that operator can depend on
  */
-export function executeOperator<L = unknown, R = unknown>(lhs: L, rhs: R, operator: Operator<L, R>) {
+export function executeOperator<L = unknown, R = unknown>(lhs: L, rhs: R, operator: Operator<L, R>, operatorFacts?: Record<string, Facts | undefined>) {
   const validLhs = (!operator.validateLhs || operator.validateLhs(lhs));
   const validRhs = (!operator.validateRhs || operator.validateRhs(rhs));
+  let operatorFactValues: Record<string, Facts> | undefined;
+  if (operatorFacts && operator.factImplicitDependencies) {
+    operatorFactValues = operator.factImplicitDependencies.reduce((acc, dep) => {
+      if (operatorFacts[dep]) {
+        acc[dep] = operatorFacts[dep]!;
+      } else {
+        throw new Error(`The fact "${dep}" requested by ${operator.name} cannot be found.`);
+      }
+      return acc;
+    }, {} as Record<string, Facts>);
+  }
   if (!validLhs) {
-    throw new Error(`Invalid left operand : ${JSON.stringify(lhs)}`);
+    throw new Error(`Invalid left operand: ${JSON.stringify(lhs)}`);
   }
   if (!validRhs) {
-    throw new Error(`Invalid right operand : ${JSON.stringify(rhs)}`);
+    throw new Error(`Invalid right operand: ${JSON.stringify(rhs)}`);
   }
-  const obs = operator.evaluator(lhs, rhs);
+  const obs = operator.evaluator(lhs, rhs, operatorFactValues);
   return obs;
 }
 
--- a/packages/@o3r/rules-engine/src/engine/operator/operator.interface.ts
+++ b/packages/@o3r/rules-engine/src/engine/operator/operator.interface.ts
@@ -1,3 +1,5 @@
+import type { Facts } from '../fact/fact.interfaces';
+
 /**
  * Rule Engine operator
  */
@@ -11,7 +13,9 @@ export interface Operator<LeftExposed = unknown, RightExposed = unknown, LeftSup
   /** Right Hand Value validator function */
   validateRhs?: unknown extends RightSupported ? (operand: unknown) => boolean : (operand: unknown) => operand is RightSupported;
   /** Evaluate the values */
-  evaluator: (lhs: LeftSupported, rhs: RightSupported) => boolean;
+  evaluator: (lhs: LeftSupported, rhs: RightSupported, operatorFactValues?: Record<string, Facts>) => boolean;
+  /** List of facts names that the operator can depend on */
+  factImplicitDependencies?: string[];
 }
 
 /**
--- a/packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.spec.ts
+++ b/packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.spec.ts
@@ -3,7 +3,9 @@ import {
   dateAfter,
   dateBefore,
   dateEquals,
+  dateInNextMinutes,
   dateNotEquals,
+  dateNotInNextMinutes,
   inRangeDate
 } from './date-based.operators';
 
@@ -13,15 +15,15 @@ describe('Operators', () => {
   const now = new Date();
   const tomorrow = new Date(new Date(now).setHours(0, 0, 0, 0) + millisecondsInADay);
   const yesterday = new Date(new Date(now).setHours(0, 0, 0, 0) - millisecondsInADay);
+  const inTwoDays = new Date(new Date(now).setHours(0, 0, 0, 0) + 2 * millisecondsInADay);
 
   describe('inRangeDate', () => {
     it('should have a valid name', () => {
       expect(inRangeDate.name).toBe('inRangeDate');
     });
 
-    it('should invalid when dates is invalid', () => {
+    it('should invalid when dates are invalid', () => {
       expect(() => executeOperator('invalid date' as any, [] as any, inRangeDate)).toThrow();
-      expect(() => executeOperator(null as any, [] as any, inRangeDate)).toThrow();
       expect(() => executeOperator(null as any, [] as any, inRangeDate)).toThrow();
       expect(() => executeOperator(now, ['invalid date'] as any, inRangeDate)).toThrow();
       expect(() => executeOperator(now, ['invalid date', 'invalid date'], inRangeDate)).toThrow();
@@ -35,6 +37,112 @@ describe('Operators', () => {
       expect(executeOperator(now.getTime(), [tomorrow.getTime(), tomorrow.getTime()], inRangeDate)).toBeFalsy();
       expect(executeOperator('2000-01-01', ['1999-12-31', '2099-12-31'], inRangeDate)).toBeTruthy();
       expect(executeOperator('2000-01-01', ['2099-12-31', '3099-12-31'], inRangeDate)).toBeFalsy();
+    });
+  });
+
+  describe('dateInNextMinutes', () => {
+    it('should have a valid name', () => {
+      expect(dateInNextMinutes.name).toBe('dateInNextMinutes');
+    });
+
+    it('should invalid when dates are invalid', () => {
+      expect(() => executeOperator('invalid date' as any, [] as any, dateInNextMinutes)).toThrow();
+      expect(() => executeOperator(null as any, [] as any, dateInNextMinutes)).toThrow();
+      expect(() => executeOperator(yesterday, 'invalid date', dateInNextMinutes)).toThrow();
+      expect(() => executeOperator(yesterday, -1, dateInNextMinutes)).toThrow();
+
+    });
+
+    it('should correctly check date in next minutes', () => {
+      expect(executeOperator(tomorrow, (24 * 60 + 1), dateInNextMinutes, {o3rCurrentTime: now.getTime()})).toBeTruthy();
+      // for past events, the operator should return false
+      expect(executeOperator(yesterday, 0, dateInNextMinutes, {o3rCurrentTime: now.getTime()})).toBeFalsy();
+      // range is from now to +24h but event is in two days. The operator returns false
+      expect(executeOperator(inTwoDays, (24 * 60 + 1), dateInNextMinutes, {o3rCurrentTime: now.getTime()})).toBeFalsy();
+
+    });
+
+    it('should throw error when no operatorFactValues is provided', () => {
+      expect(() => executeOperator(now, 5, dateInNextMinutes)).toThrow('No operatorFactValues. Unable to retrieve the current time.');
+    });
+
+    it('should correctly check the edge case of current time', () => {
+      // Considering the exact current time as the leftDate, it should still be true if the range is 0 minutes
+      expect(executeOperator(now, 0, dateInNextMinutes, {o3rCurrentTime: now.getTime()})).toBeTruthy();
+
+      // Considering the exact current time but with 1 minute in the future, it should be true
+      expect(executeOperator(now, 1, dateInNextMinutes, {o3rCurrentTime: now.getTime()})).toBeTruthy();
+    });
+
+    it('should correctly check the edge case of exact target date time', () => {
+      const oneHourLater = new Date(now.getTime() + 60 * 60 * 1000); // 1 hour later from now
+      expect(executeOperator(oneHourLater, 60, dateInNextMinutes, {o3rCurrentTime: now.getTime()})).toBeTruthy();
+
+      const oneMinuteLater = new Date(now.getTime() + 60 * 1000); // 1 minute later from now
+      expect(executeOperator(oneMinuteLater, 1, dateInNextMinutes, {o3rCurrentTime: now.getTime()})).toBeTruthy();
+    });
+
+    it('should return false for dates just beyond the specified range', () => {
+      const oneMinutePastTarget = new Date(new Date(now).setMinutes(now.getMinutes() + 6)); // 6 minutes from now
+      expect(executeOperator(oneMinutePastTarget, 5, dateInNextMinutes, {o3rCurrentTime: now.getTime()})).toBeFalsy();
+    });
+
+    it('should validate rhs as numbers', () => {
+      expect(() => executeOperator(now, 'invalid' as any, dateInNextMinutes, {o3rCurrentTime: now.getTime()})).toThrow();
+    });
+  });
+
+  describe('dateNotInNextMinutes', () => {
+    it('should have a valid name', () => {
+      expect(dateNotInNextMinutes.name).toBe('dateNotInNextMinutes');
+    });
+
+    it('should invalid when dates are invalid', () => {
+      expect(() => executeOperator('invalid date' as any, 1, dateNotInNextMinutes)).toThrow();
+      expect(() => executeOperator(null as any, 0, dateNotInNextMinutes)).toThrow();
+      expect(() => executeOperator(now, 'invalid minutes' as any, dateNotInNextMinutes)).toThrow();
+      expect(() => executeOperator(now, 7, dateNotInNextMinutes)).toThrow();
+    });
+
+    it('should correctly check date not in next minutes', () => {
+      // event occuring tomorrow is not in 0 minutes from now
+      expect(executeOperator(tomorrow, 0, dateNotInNextMinutes, {o3rCurrentTime: now.getTime()})).toBeTruthy();
+      // event occuring in two days is not in 24h from now
+      expect(executeOperator(inTwoDays, 24 * 60 + 1, dateNotInNextMinutes, {o3rCurrentTime: now.getTime()})).toBeTruthy();
+      // event that occured yesterday is not in 10 minutes from now
+      expect(executeOperator(yesterday, 10, dateNotInNextMinutes, {o3rCurrentTime: now.getTime()})).toBeFalsy();
+    });
+
+    it('should throw error when no operatorFactValues is provided', () => {
+      expect(() => executeOperator(now, 5, dateNotInNextMinutes)).toThrow('No operatorFactValues. Unable to retrieve the current time.');
+    });
+
+    it('should correctly check for events exactly at the edge of the range', () => {
+      // Event occurring exactly 1 minute from now should return false, as it is within the range
+      const oneMinuteFromNow = new Date(now.getTime() + 60 * 1000);
+      expect(executeOperator(oneMinuteFromNow, 1, dateNotInNextMinutes, {o3rCurrentTime: now.getTime()})).toBeFalsy();
+
+      // Event occurring just after 1 minute from now should return true, as it is outside the range
+      const slightlyAfterOneMinuteFromNow = new Date(oneMinuteFromNow.getTime() + 1);
+      expect(executeOperator(slightlyAfterOneMinuteFromNow, 1, dateNotInNextMinutes, {o3rCurrentTime: now.getTime()})).toBeTruthy();
+    });
+
+    it('should return false for events within the specified range', () => {
+      const halfHourFromNow = new Date(now.getTime() + 30 * 60 * 1000);
+      expect(executeOperator(halfHourFromNow, 60, dateNotInNextMinutes, {o3rCurrentTime: now.getTime()})).toBeFalsy();
+    });
+
+    it('should validate rhs as numbers', () => {
+      expect(() => executeOperator(now, 'invalid' as any, dateNotInNextMinutes, {o3rCurrentTime: now.getTime()})).toThrow();
+    });
+    it('should return true for events happening way into the future', () => {
+      const nextYear = new Date(new Date().setFullYear(now.getFullYear() + 1));
+      expect(executeOperator(nextYear, 60, dateNotInNextMinutes, {o3rCurrentTime: now.getTime()})).toBeTruthy();
+    });
+
+    it('should return false for dates way in the past', () => {
+      const lastYear = new Date(new Date().setFullYear(now.getFullYear() - 1));
+      expect(executeOperator(lastYear, 60, dateNotInNextMinutes, {o3rCurrentTime: now.getTime()})).toBeFalsy();
     });
   });
 
--- a/packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
+++ b/packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
@@ -1,4 +1,4 @@
-import {isValidDateInput, isValidDateRange} from '../operator.helpers';
+import {isValidDateInput, isValidDateRange, numberValidator} from '../operator.helpers';
 import {DateInput, Operator} from '../operator.interface';
 
 /**
@@ -14,6 +14,58 @@ export const inRangeDate: Operator<Date, [DateInput, DateInput], DateInput> = {
   },
   validateLhs: isValidDateInput,
   validateRhs: isValidDateRange
+};
+
+
+/**
+ * Check if the value of the variable is in the next x minutes
+ *
+ * @title is in next minutes
+ *
+ * @returns false for dates before `now` and for dates after `now` + `nextMinutes`, true for dates between `now` and `now` + `nextMinutes`
+ */
+export const dateInNextMinutes: Operator<Date, string | number, DateInput> = {
+  name: 'dateInNextMinutes',
+  evaluator: (leftDateInput, minutes, operatorFactValues) => {
+    if (!operatorFactValues) {
+      throw new Error('No operatorFactValues. Unable to retrieve the current time.');
+    }
+    if (typeof operatorFactValues.o3rCurrentTime !== 'number') {
+      throw new Error('o3rCurrentTime value is not a number');
+    }
+    const currentTimeValue = operatorFactValues.o3rCurrentTime;
+    return inRangeDate.evaluator(leftDateInput, [currentTimeValue, currentTimeValue + +minutes * 60000]);
+  },
+  factImplicitDependencies: ['o3rCurrentTime'],
+  validateLhs: isValidDateInput,
+  validateRhs: numberValidator
+};
+
+/**
+ * Check if the value of the variable is not in the next x minutes
+ *
+ * @title is not in next minutes
+ *
+ * @returns false for dates before `now` and for dates between `now` and `now` + `nextMinutes`, true for dates after `now` + `nextMinutes`
+ */
+export const dateNotInNextMinutes: Operator<Date, string | number, DateInput> = {
+  name: 'dateNotInNextMinutes',
+  evaluator: (leftDateInput, minutes, operatorFactValues) => {
+    if (!operatorFactValues) {
+      throw new Error('No operatorFactValues. Unable to retrieve the current time.');
+    }
+    if (typeof operatorFactValues.o3rCurrentTime !== 'number') {
+      throw new Error('o3rCurrentTime value is not a number');
+    }
+    const currentTimeValue = operatorFactValues.o3rCurrentTime;
+    const now = new Date(currentTimeValue);
+    const leftDate = new Date(leftDateInput);
+    const targetDate = new Date(new Date(currentTimeValue).setMinutes(now.getMinutes() + +minutes));
+    return leftDate >= now && leftDate > targetDate;
+  },
+  factImplicitDependencies: ['o3rCurrentTime'],
+  validateLhs: isValidDateInput,
+  validateRhs: numberValidator
 };
 
 /**
@@ -81,5 +133,5 @@ export const dateNotEquals: Operator<Date, DateInput, DateInput> = {
 };
 
 export const dateBasedOperators = [
-  inRangeDate, dateAfter, dateBefore, dateEquals, dateNotEquals
+  inRangeDate, dateInNextMinutes, dateNotInNextMinutes, dateAfter, dateBefore, dateEquals, dateNotEquals
 ];
--- a/packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
+++ b/packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
@@ -4,7 +4,8 @@ import { combineLatest, Observable, of } from 'rxjs';
 import { distinctUntilChanged, map, pairwise, shareReplay, startWith, switchMap, tap } from 'rxjs/operators';
 import { handleRuleEvaluationDebug } from './debug/helpers';
 import type { RulesEngine } from './engine';
-import { EngineRuleset, EvaluationReason, Facts, RuleEvaluation, RuleEvaluationOutput } from './engine.interface';
+import type { Facts } from './fact';
+import { EngineRuleset, EvaluationReason, RuleEvaluation, RuleEvaluationOutput } from './engine.interface';
 import { executeOperator } from './operator/operator.helpers';
 import { Operator } from './operator/operator.interface';
 import {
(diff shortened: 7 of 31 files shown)


packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
```diff
@@ -16,6 +16,40 @@ export const inRangeDate: Operator<Date, [DateInput, DateInput], DateInput> = {
...
+    return leftDate <= targetDate;
+  },
+  validateLhs: isValidDateInput,
+  validateRhs: isValidDateWithMinutes
+};
+
+/**
+ * Check if the value of the variable is not in the next x minutes
+ *
+ * @title is not in next minutes
```
kpanot, line 39: This is really not clear.
Maybe cleaner to say that the time between the 2 dates are inferior to the specified minutes

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
```diff
@@ -16,6 +16,40 @@ export const inRangeDate: Operator<Date, [DateInput, DateInput], DateInput> = {
...
+ */
+export const dateNotInNextMinutes: Operator<Date, [DateInput, number], DateInput> = {
+  name: 'dateNotInNextMinutes',
+  evaluator: (leftDateInput, [rightDate, minutes]) => {
+    const targetDate = new Date(rightDate);
+    const leftDate = new Date(leftDateInput);
+    targetDate.setMinutes(targetDate.getMinutes() + +minutes);
```
kpanot, line 46: Not sure this is correct, you should add the minutes to the left date and ensure that the left date is still inferior to the right one.
(Maybe a UT testing the calcul would help)

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
```diff
@@ -16,6 +16,40 @@ export const inRangeDate: Operator<Date, [DateInput, DateInput], DateInput> = {
...
+ *
+ * @title is in next minutes
+ */
+export const dateInNextMinutes: Operator<Date, [DateInput, number], DateInput> = {
+  name: 'dateInNextMinutes',
+  evaluator: (leftDateInput, [rightDate, minutes]) => {
+    const targetDate = new Date(rightDate);
+    const leftDate = new Date(leftDateInput);
+    targetDate.setMinutes(targetDate.getMinutes() + +minutes);
+    return leftDate <= targetDate;
```
fpaul-1A, line 30: maybe it would be better to reuse `inRangeDate`
```typescript
evaluator: (leftDateInput, [rightDate, minutes]) => {
  return inRangeDate.evaluator(leftDateInput, [rightDate, new Date(rightDate).getTime() + minutes * 60000)]);
}
```

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -61,6 +62,137 @@ export class RulesetExecutor {
...
+              } else if (typeof currentObj[key] === 'object' && currentObj[key] !== null) {
+                exploreObject(currentObj[key]);
+              }
+            }
+          }
+        };
+        exploreObject(obj);
+        return Array.from(new Set(ruleInputFacts));
+      };
+      rule.inputFacts = findRuleInputFacts(rule.rootElement.condition);
```
fpaul-1A, line 93: `rule` can be read-only in some cases (when coming from the store with strictImmutability)
you may need to store it somewhere else

apps/showcase/src/assets/rules/rulesets.json
```diff
@@ -82,6 +82,42 @@
           "name": "The otter is in vacations",
           "disabled": false,
           "outputRuntimeFacts": [],
+          "inputFacts": ["outboundDate", "currentTime"],
```
fpaul-1A, line 85: this can now be removed ?

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
vscaiceanu-1a: to clarify: 
`inNextMinutes` returns:
- `false` for date before `now` and for dates after `now` + `nextMinutes`
- `true` for dates between `now and `now` + `nextMinutes`

`notInNextMinutes returns:
- `false` for date before `now` and for dates between `now and `now` + `nextMinutes`
- `true` for dates after `now` + `nextMinutes`

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
vscaiceanu-1a: to clarify: 
`inNextMinutes` returns:
- `false` for date before `now` and for dates after `now` + `nextMinutes`
- `true` for dates between `now and `now` + `nextMinutes`

`notInNextMinutes` returns:
- `false` for date before `now` and for dates between `now and `now` + `nextMinutes`
- `true` for dates after `now` + `nextMinutes`

packages/@o3r/rules-engine/src/engine/operator/facts/current-time/current-time.facts.ts
```diff
@@ -0,0 +1,9 @@
+import { FactDefinitions } from '../../../../engine/fact/index';
+
+export interface CurrentTimeFacts extends FactDefinitions {
+  /**
+   * The current time as a timestamp
+   **/
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  o3r_currentTime: number;
```
matthieu-crouzet, line 8: Why not `o3rCurrentTime` to follow the naming convention ?

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
matthieu-crouzet: Maybe that could be added in the description ?

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -61,6 +61,137 @@ export class RulesetExecutor {
...
+      inputFactsForRule[rule.id] = findRuleInputFacts(rule.rootElement);
+    });
+    const factsThatRerunEverything: string[] = [];
+    this.ruleset.rules.forEach((rule) => {
+      if (rule.outputRuntimeFacts.length > 0 || rule.inputRuntimeFacts.length > 0) {
+        factsThatRerunEverything.push(...inputFactsForRule[rule.id]);
+      } else {
+      }
+    });
+    const triggerFull: Observable<unknown[]> = factsThatRerunEverything.length === 0 ? of([]) :
```
matthieu-crouzet, line 104:
```suggestion
    const triggerFull$: Observable<unknown[]> = factsThatRerunEverything.length === 0 ? of([]) :
```

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -195,112 +326,4 @@ export class RulesetExecutor {
     throw new Error(`Unknown condition block met : ${JSON.stringify(nestedCondition)}`);
   }
 
-  /**
-   * Plug ruleset to fact streams and trigger a first evaluation
-   *
-   * @param ruleset
-   */
-  public prepareRuleset() {
```
matthieu-crouzet, line 203: we cannot remove public method like this

packages/@o3r/rules-engine/src/engine/operator/operator.helpers.ts
```diff
@@ -7,16 +8,28 @@ import {DateInput, Operator, SupportedSimpleTypes} from './operator.interface';
...
   const validRhs = (!operator.validateRhs || operator.validateRhs(rhs));
+  let operatorFactValues: Record<string, Facts> | undefined;
+  if (operatorFacts && operator.dependencies) {
+    operatorFactValues = operator.dependencies.reduce((acc, dep) => {
+      if (operatorFacts[dep]) {
+        acc[dep] = operatorFacts[dep]!;
+      }
+      return acc;
+    }, {} as Record<string, Facts>);
+    if (!(!operator.validateOperatorFactValues || operator.validateOperatorFactValues(operatorFactValues))) {
```
mrednic-1A, line 22: `if (operator.validateOperatorFactValues && !operator.validateOperatorFactValues(operatorFactValues)` ?

packages/@o3r/rules-engine/src/engine/operator/operator.interface.ts
```diff
@@ -10,8 +12,12 @@ export interface Operator<LeftExposed = unknown, RightExposed = unknown, LeftSup
   validateLhs?: unknown extends LeftSupported ? (operand: unknown) => boolean : (operand: unknown) => operand is LeftSupported;
   /** Right Hand Value validator function */
   validateRhs?: unknown extends RightSupported ? (operand: unknown) => boolean : (operand: unknown) => operand is RightSupported;
+  /** Operator Fact Values validator function */
+  validateOperatorFactValues?: unknown extends Record<string, Facts> ? (operators: unknown) => boolean : (operators: unknown) => operators is Record<string, Facts>;
   /** Evaluate the values */
-  evaluator: (lhs: LeftSupported, rhs: RightSupported) => boolean;
+  evaluator: (lhs: LeftSupported, rhs: RightSupported, operatorFactValues?: Record<string, Facts>) => boolean;
+  /** Dependencies */
```
mrednic-1A, line 19: `List of facts names used inside the operator` ?

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.spec.ts
```diff
@@ -38,6 +42,52 @@ describe('Operators', () => {
...
+      expect(() => executeOperator(now, 7, dateNotInNextMinutes)).toThrow();
+
+    });
+
+    it('should correctly check date not in next minutes', () => {
+      // event occuring tomorrow is not in 0 minutes from now
+      expect(executeOperator(tomorrow, 0, dateNotInNextMinutes, {o3r_currentTime: now})).toBeTruthy();
+      // event occuring in two days is not in 24h from now
+      expect(executeOperator(inTwoDays, (24 * 60 + 1), dateNotInNextMinutes, {o3r_currentTime: now})).toBeTruthy();
+      // event that occured yesterday is not in 0 minutes from now
```
mrednic-1A, line 86: '10 minutes from now' I guess

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.spec.ts
```diff
@@ -38,6 +42,52 @@ describe('Operators', () => {
...
+      expect(() => executeOperator(now, 'invalid minutes' as any, dateNotInNextMinutes)).toThrow();
+      expect(() => executeOperator(now, 7, dateNotInNextMinutes)).toThrow();
+
+    });
+
+    it('should correctly check date not in next minutes', () => {
+      // event occuring tomorrow is not in 0 minutes from now
+      expect(executeOperator(tomorrow, 0, dateNotInNextMinutes, {o3r_currentTime: now})).toBeTruthy();
+      // event occuring in two days is not in 24h from now
+      expect(executeOperator(inTwoDays, (24 * 60 + 1), dateNotInNextMinutes, {o3r_currentTime: now})).toBeTruthy();
```
mrednic-1A, line 85: I don't think you need parenthesis for 24 * 60 + 1

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
vscaiceanu-1a: it wasn't supposed to be public in the first place :) only used within this file

matthieu-crouzet: But it is now 
we can deprecate it in v9 and set it private in v10

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
```diff
@@ -16,6 +17,78 @@ export const inRangeDate: Operator<Date, [DateInput, DateInput], DateInput> = {
...
+
+  return Object.prototype.hasOwnProperty.call(value, 'o3rCurrentTime');
+}
+
+/**
+ * Check if the value of the variable is in the next x minutes
+ *
+ * @title is in next minutes
+ *
+ * returns:
```
matthieu-crouzet, line 43: why not using tsdoc with `@returns`

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
vscaiceanu-1a: Even if it is now, it was not supposed to be. 
The method is only called in the constructor of this class and this class is not exported
image

I moved it back down to have a correct diff on the PR and added a TODO to move it back to top as public in a separate PR.

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
```diff
@@ -16,6 +17,74 @@ export const inRangeDate: Operator<Date, [DateInput, DateInput], DateInput> = {
...
+  name: 'dateInNextMinutes',
+  evaluator: (leftDateInput, minutes, operatorFactValues) => {
+    if (!operatorFactValues) {
+      throw new Error('No operatorFactValues. Unable to retrieve the current time.');
+    }
+    const currentTimeValue = operatorFactValues.o3rCurrentTime as number;
+    const now = new Date(currentTimeValue);
+    const leftDate = new Date(leftDateInput);
+    const targetDate = new Date(new Date(currentTimeValue).setMinutes(now.getMinutes() + +minutes));
+    return leftDate >= now && leftDate <= targetDate;
```
fpaul-1A, line 55:
```suggestion
    const currentTimeValue = operatorFactValues.o3rCurrentTime as number;
    return inRangeDate.evaluator(leftDateInput, [currentTimeValue, currentTimeValue +minutes * 60000]);
```
?

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -195,38 +195,62 @@ export class RulesetExecutor {
...
+                exploreObject(currentObj[key]);
+              }
+            }
+          }
+        };
+        exploreObject(obj);
+        return Array.from(new Set(ruleInputFacts));
+      };
+
+      inputFactsForRule[rule.id] = findRuleInputFacts(rule.rootElement);
```
fpaul-1A, line 230: it seems weird to use an Array to add all the items and remove the duplicates at the end
It would make more sense to add the items directly in the Set and convert it to Array at the end:
```suggestion
      const findRuleInputFacts = (obj: AllBlock): string[] => {
        const ruleInputFacts = new Set<string>();
        // Recursive function to explore each object
        const exploreObject = (currentObj: any) => {
          if (currentObj && currentObj.type === 'FACT') {
            ruleInputFacts.add(currentObj.value);
          } else if (Array.isArray(currentObj)) {
            currentObj.forEach(exploreObject);
          } else {
            for (const key in currentObj) {
              if (key === 'operator') {
                const op = this.operators[currentObj[key]];
                if (op && op.dependencies) {
                  op.dependencies.forEach(rulesInputFacts.add, rulesInputFacts);
                }
              } else if (typeof currentObj[key] === 'object' && currentObj[key] !== null) {
                exploreObject(currentObj[key]);
              }
            }
          }
        };
        exploreObject(obj);
        return Array.from(ruleInputFacts);
      };

      inputFactsForRule[rule.id] = findRuleInputFacts(rule.rootElement);
```

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
```diff
@@ -16,6 +17,74 @@ export const inRangeDate: Operator<Date, [DateInput, DateInput], DateInput> = {
   validateRhs: isValidDateRange
 };
 
+/**
+ * Validate if the provided operatorFactValues contain 'o3rCurrentTime'
+ * @param value the value
+ */
+export function isValidCurrentTimeValue(value: unknown): value is Record<string, Facts> {
```
pginoux-1A, line 24: I think the name of the function is unclear, it doesn't check if the current time value is valid, it just checks if it contains the key

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -195,38 +195,62 @@ export class RulesetExecutor {
...
    */
   public prepareRuleset() {
-    const rulesWithContext: Rule[] = [];
-    const rulesWithoutContext: Rule[] = [];
+    const inputFactsForRule: Record<string, string[]> = {};
+    this.ruleset.rules.forEach((rule) => {
+      const findRuleInputFacts = (obj: AllBlock): string[] => {
+        const ruleInputFacts: string[] = [];
+        // Recursive function to explore each object
+        const exploreObject = (currentObj: any) => {
```
pginoux-1A, line 208: I'd be interested in having a check with a a big enough rulesets file to see how long it takes to be executed (you can just duplicate the IT tests json content).

vscaiceanu-1a: time with 1000 rules: 3.7748750001192093

total test time: 4.24 s

time with 10000: 38.763249998912215

total test time: 24.12s

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.spec.ts
```diff
@@ -1,9 +1,13 @@
+/* eslint-disable camelcase */
+/* eslint-disable @typescript-eslint/naming-convention */
```
matthieu-crouzet, line 2: not needed anymore ?

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -195,38 +195,62 @@ export class RulesetExecutor {
...
    */
   public prepareRuleset() {
-    const rulesWithContext: Rule[] = [];
-    const rulesWithoutContext: Rule[] = [];
+    const inputFactsForRule: Record<string, string[]> = {};
+    this.ruleset.rules.forEach((rule) => {
+      const findRuleInputFacts = (obj: AllBlock): string[] => {
+        const ruleInputFacts = new Set<string>();
+        // Recursive function to explore each object
+        const exploreObject = (currentObj: any) => {
```
matthieu-crouzet, line 208: [not mandatory] 
Maybe this method could be defined outside to avoid redefining it for every rules

matthieu-crouzet (approved):

packages/@o3r/rules-engine/src/engine/operator/facts/current-time/current-time-facts.service.ts
```diff
@@ -0,0 +1,25 @@
+import { Injectable } from '@angular/core';
+import { BehaviorSubject } from 'rxjs';
+import { FactsService } from '../../../../fact';
+import { RulesEngineService } from '../../../../services';
+import { CurrentTimeFacts } from './current-time.facts';
+
+@Injectable({
```
kpanot, line 7: Do we really want to be Angular injectable fact in the case of engine inner facts

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
```diff
@@ -16,6 +17,71 @@ export const inRangeDate: Operator<Date, [DateInput, DateInput], DateInput> = {
...
+  }
+
+  for (const key in value) {
+    if (typeof key !== 'string') {
+      return false;
+    }
+  }
+
+  return Object.prototype.hasOwnProperty.call(value, 'o3rCurrentTime');
+}
```
kpanot, line 36: This function is over complex and the return type is actually incorrect because not checked

```suggestion
function containsO3rCurrentTime(value: any) {
  return !!value?.o3rCurrentTime;
}
```

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
```diff
@@ -16,6 +17,71 @@ export const inRangeDate: Operator<Date, [DateInput, DateInput], DateInput> = {
...
+    const currentTimeValue = operatorFactValues.o3rCurrentTime as Date;
+    const now = new Date(currentTimeValue);
+    const leftDate = new Date(leftDateInput);
+    const targetDate = new Date(new Date(currentTimeValue).setMinutes(now.getMinutes() + +minutes));
+    return leftDate >= now && leftDate > targetDate;
+  },
+  dependencies: ['o3rCurrentTime'],
+  validateLhs: isValidDateInput,
+  validateRhs: numberValidator,
+  validateOperatorFactValues: containsO3rCurrentTime
```
kpanot, line 82: This name is not matching its purpose

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -195,38 +195,62 @@ export class RulesetExecutor {
...
    * Plug ruleset to fact streams and trigger a first evaluation
-   *
-   * @param ruleset
    */
   public prepareRuleset() {
-    const rulesWithContext: Rule[] = [];
-    const rulesWithoutContext: Rule[] = [];
+    const inputFactsForRule: Record<string, string[]> = {};
+    this.ruleset.rules.forEach((rule) => {
+      const findRuleInputFacts = (obj: AllBlock): string[] => {
```
kpanot, line 205: should defined outside of the loop

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
kpanot: This is actually mandatory for my point of view, for performance purpose

packages/@o3r/rules-engine/src/engine/operator/operator.helpers.ts
```diff
@@ -7,16 +8,28 @@ import {DateInput, Operator, SupportedSimpleTypes} from './operator.interface';
  * @param rhs Right hand side
  * @param operator Operator to compare values
  */
-export function executeOperator<L = unknown, R = unknown>(lhs: L, rhs: R, operator: Operator<L, R>) {
+export function executeOperator<L = unknown, R = unknown>(lhs: L, rhs: R, operator: Operator<L, R>, operatorFacts?: Record<string, Facts | undefined>) {
```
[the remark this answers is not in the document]
vscaiceanu-1a: because `factsValue: Record<string, Facts | undefined>` in `ruleset-executor.ts`

packages/@o3r/rules-engine/src/engine/operator/operator.helpers.ts
```diff
@@ -7,16 +8,28 @@ import {DateInput, Operator, SupportedSimpleTypes} from './operator.interface';
...
   const validLhs = (!operator.validateLhs || operator.validateLhs(lhs));
   const validRhs = (!operator.validateRhs || operator.validateRhs(rhs));
+  let operatorFactValues: Record<string, Facts> | undefined;
+  if (operatorFacts && operator.dependencies) {
+    operatorFactValues = operator.dependencies.reduce((acc, dep) => {
+      if (operatorFacts[dep]) {
```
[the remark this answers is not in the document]
vscaiceanu-1a: this is exactly the purpose of the `validateOperatorFactValues` on the next line

packages/@o3r/rules-engine/src/engine/operator/facts/current-time/current-time-facts.service.ts
vscaiceanu-1a: you have to register it in the app like any other fact service

kpanot:
Missing:
- [ ] JSON Schema update
- [ ] Documentation

packages/@o3r/rules-engine/src/engine/operator/operator.interface.ts
```diff
@@ -10,8 +12,12 @@ export interface Operator<LeftExposed = unknown, RightExposed = unknown, LeftSup
   validateLhs?: unknown extends LeftSupported ? (operand: unknown) => boolean : (operand: unknown) => operand is LeftSupported;
   /** Right Hand Value validator function */
   validateRhs?: unknown extends RightSupported ? (operand: unknown) => boolean : (operand: unknown) => operand is RightSupported;
+  /** Operator Fact Values validator function */
+  validateOperatorFactValues?: unknown extends Record<string, Facts> ? (operators: unknown) => boolean : (operators: unknown) => operators is Record<string, Facts>;
   /** Evaluate the values */
-  evaluator: (lhs: LeftSupported, rhs: RightSupported) => boolean;
+  evaluator: (lhs: LeftSupported, rhs: RightSupported, operatorFactValues?: Record<string, Facts>) => boolean;
+  /** List of facts names used inside the operator */
+  dependencies?: string[];
```
kpanot, line 20: Should have a more explicit name, such as `factImplicitDependencies`

kpanot, line 20: Should be part of the extracted metadata

packages/@o3r/rules-engine/src/engine/fact/current-time/current-time-facts.service.ts
```diff
@@ -0,0 +1,27 @@
+import { Injectable } from '@angular/core';
+import { BehaviorSubject } from 'rxjs';
+import { FactsService } from '../../../fact';
+import { RulesEngineService } from '../../../services';
+import { CurrentTimeFacts } from './current-time.facts';
+
+@Injectable({
```
kpanot, line 7: An Angular Injectable should not be in the `engine` folder.
I do even think that this fact implementation should not be in the Engine module at all

packages/@o3r/rules-engine/src/engine/fact/current-time/current-time.facts.ts
```diff
@@ -0,0 +1,12 @@
+import type { FactDefinitions } from '../index';
+
+
+/**
+ * Operator facts that provide the current time
+ */
+export interface CurrentTimeFacts extends FactDefinitions {
```
kpanot, line 7: If the fact is defined in this Engine and required by a default Operators, it needs to be registered by the Engine itself.
Up to the User to override this fact with its own impl to change the frequency

kpanot (approved):

packages/@o3r/rules-engine/src/engine/operator/operator.helpers.ts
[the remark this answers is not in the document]
kpanot: as discussed, this should be the job of the operatorExecutor, as the list of dependencies has been already provided by the operator.
The operators can check the value emitted by the fact but it does not need to check it existence itself

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
kpanot: As discuss, the full fonction may need to be removed

packages/@o3r/rules-engine/src/engine/fact/current-time/current-time.facts.ts
vscaiceanu-1a: got it

vscaiceanu-1a:
> Missing:
> 
> * [ ]  JSON Schema update
> * [ ]  Documentation
> * [ ]  (metadata extraction)

@kpanot which JSON Schema are you referring to? The ruleset one hasn't been modified in this PR.

packages/@o3r/rules-engine/src/engine/operator/operator.interface.ts
vscaiceanu-1a: Please explain why the `factImplicitDependencies` should be part of the extracted metadata. This field is not intended to be used in the CMS.

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
vscaiceanu-1a: renamed to `containsO3rCurrentTimeProperty`

packages/@o3r/rules-engine/src/engine/operator/operator.helpers.ts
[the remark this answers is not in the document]
vscaiceanu-1a: We cannot throw in the `else` because in `operatorFacts` we have all the input facts. The goal here is to execute `operator.evaluator` only if we have valid values for `lhs` `rhs`  and `operatorFactValues` (based on the operator's dependencies)

packages/@o3r/rules-engine/src/engine/operator/operator.interface.ts
vscaiceanu-1a: add `factImplicitDependencies` in the metadata so that applications can perform checks when using custom operators with dependencies

packages/@o3r/rules-engine/src/engine/operator/operator.helpers.ts
```diff
@@ -1,22 +1,34 @@
...
+export function executeOperator<L = unknown, R = unknown>(lhs: L, rhs: R, operator: Operator<L, R>, operatorFacts?: Record<string, Facts | undefined>) {
   const validLhs = (!operator.validateLhs || operator.validateLhs(lhs));
   const validRhs = (!operator.validateRhs || operator.validateRhs(rhs));
+  let operatorFactValues: Record<string, Facts> | undefined;
+  if (operatorFacts && operator.factImplicitDependencies) {
+    operatorFactValues = operator.factImplicitDependencies.reduce((acc, dep) => {
+      if (operatorFacts[dep]) {
+        acc[dep] = operatorFacts[dep]!;
+      } else {
+        throw new Error(`No value found for ${dep}.`);
```
kpanot, line 20:
```suggestion
        throw new Error(`The fact "${dep}" requested by ${operator.name} cannot be found.`);
```

packages/@o3r/rules-engine/src/engine/operator/operator.helpers.ts
```diff
@@ -1,22 +1,34 @@
 import {DateInput, Operator, SupportedSimpleTypes} from './operator.interface';
+import type {Facts} from '../engine.interface';
 
 /**
  * Execute Operator
- *
  * @param lhs Left hand side
  * @param rhs Right hand side
  * @param operator Operator to compare values
+ * @param operatorFacts Operator facts that operator can depend on
```
kpanot, line 9:
```suggestion
 * @param operatorFacts Facts that operator can depend on
```

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
```diff
@@ -16,6 +16,52 @@ export const inRangeDate: Operator<Date, [DateInput, DateInput], DateInput> = {
...
+ *
+ * @returns false for dates before `now` and for dates after `now` + `nextMinutes`, true for dates between `now` and `now` + `nextMinutes`
+ */
+export const dateInNextMinutes: Operator<Date, string | number, DateInput> = {
+  name: 'dateInNextMinutes',
+  evaluator: (leftDateInput, minutes, operatorFactValues) => {
+    if (!operatorFactValues) {
+      throw new Error('No operatorFactValues. Unable to retrieve the current time.');
+    }
+    const currentTimeValue = operatorFactValues.o3rCurrentTime as number;
```
kpanot, line 33: maybe you should check the value before hard casting

packages/@o3r/rules-engine/src/engine/operator/operator.interface.ts
```diff
@@ -11,7 +13,9 @@ export interface Operator<LeftExposed = unknown, RightExposed = unknown, LeftSup
   /** Right Hand Value validator function */
   validateRhs?: unknown extends RightSupported ? (operand: unknown) => boolean : (operand: unknown) => operand is RightSupported;
   /** Evaluate the values */
-  evaluator: (lhs: LeftSupported, rhs: RightSupported) => boolean;
+  evaluator: (lhs: LeftSupported, rhs: RightSupported, operatorFactValues?: Record<string, Facts>) => boolean;
+  /** List of facts names used inside the operator */
```
kpanot, line 17:
```suggestion
  /** List of facts names that the operator can depend on */
```

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
```diff
@@ -16,6 +16,52 @@ export const inRangeDate: Operator<Date, [DateInput, DateInput], DateInput> = {
...
+ */
+export const dateNotInNextMinutes: Operator<Date, string | number, DateInput> = {
+  name: 'dateNotInNextMinutes',
+  evaluator: (leftDateInput, minutes, operatorFactValues) => {
+    if (!operatorFactValues) {
+      throw new Error('No operatorFactValues. Unable to retrieve the current time.');
+    }
+    const currentTimeValue = operatorFactValues.o3rCurrentTime as number;
+    const now = new Date(currentTimeValue);
+    const leftDate = new Date(leftDateInput);
```
kpanot, line 56: why this `new Date`, `leftDateInput` is already a type `Date`

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -195,38 +195,58 @@ export class RulesetExecutor {
     throw new Error(`Unknown condition block met : ${JSON.stringify(nestedCondition)}`);
   }
 
+  // TODO: make private and move to top of the class in a separate PR to make the diff more meaningful in this one
```
kpanot, line 198: Or just do it in a separated commit if you are scared about the diff.
I am not sure this change should request: a "TODO" in the code + a Github issue + a new PR

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -195,38 +195,58 @@ export class RulesetExecutor {
...
   /**
    * Plug ruleset to fact streams and trigger a first evaluation
-   *
-   * @param ruleset
    */
   public prepareRuleset() {
-    const rulesWithContext: Rule[] = [];
-    const rulesWithoutContext: Rule[] = [];
+    const inputFactsForRule: Record<string, string[]> = {};
+    const exploreObject = (currentObj: any, ruleInputFacts: Set<string>) => {
```
kpanot, line 204: I am not sure to understand the purpose of this part :S.
Why do you need it as you already have the list of facts available for the ruleset?

packages/@o3r/rules-engine/src/services/rules-engine.service.ts
```diff
@@ -196,6 +200,10 @@ export class RulesEngineService implements OnDestroy {
     }));
   }
 
+  private registerOperatorFacts() {
+    this.upsertFacts([{ id: 'o3rCurrentTime', value$: currentTimeSubject$.asObservable() }]);
```
kpanot, line 204: The documentation `docs/rules-engine/facts.md` should contain the list of facts available per default

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
vscaiceanu-1a: The goal is to compute the list of input facts and no longer take into account the value of the `inputFacts` coming from the CMS.
We have even removed them from the showcase app ruleset https://github.com/AmadeusITGroup/otter/pull/835/files#diff-1c83c4284e0d72e53c44790c5bdd50daedb84c30b8e703c4e0552ecf2ab870ae

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
vscaiceanu-1a: `leftDateInput` is of type `type DateInput = Date | string | number;` as it is the case for the oder date-based operators

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
vscaiceanu-1a: I don't intend to create an issue for that. I still find it cleaner to do it in a separate pull request because
1. the change is not linked to this one
2. if I do it in a separate commit, a reviewer will have to be aware that a certain commit has to be selected in order to review the PR

packages/@o3r/rules-engine/src/engine/operator/operator.interface.ts
```diff
@@ -1,3 +1,5 @@
+import type { Facts } from '../engine.interface';
```
kpanot, line 1: This is actually quite close to be a circular dependency (on the file path, not the imported interface).
May be we should take care of it

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
vscaiceanu-1a: https://github.com/AmadeusITGroup/otter/issues/1044

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
vscaiceanu-1a: Concerning the types for the input, we support nested rules and the method will iterate on many types of Objects or Arrays.

Here's an example of a nested rule:
```typescript
        {
          'id': '6e8t54h6s4e-6erth46sre8th4-d46t8s13t5j2',
          'name': 'the second rule',
          'inputRuntimeFacts': ['UI_FACT_2'],
          'inputFacts': ['cart'],
          'outputRuntimeFacts': [],
          'rootElement': {
            'elementType': 'RULE_BLOCK',
            'blockType': 'IF_ELSE',
            'condition': {
              'any': [
                {
                  'lhs': {
                    'type': 'RUNTIME_FACT',
                    'value': 'UI_FACT_2'
                  },
                  'operator': 'equals',
                  'rhs': {
                    'type': 'LITERAL',
                    'value': false
                  }
                }
              ]
            },
            'successElements': [
              {
                'elementType': 'ACTION',
                'actionType': 'UPDATE_LOCALISATION',
                'key': 'my.loc.key2.success',
                'value': 'my.loc.value2.success'
              },
              {
                'elementType': 'ACTION',
                'actionType': 'UPDATE_LOCALISATION',
                'key': 'my.loc.key3.success',
                'value': 'my.loc.value3.success'
              },
              {
                'elementType': 'RULE_BLOCK',
                'blockType': 'IF_ELSE',
                'condition': {
                  'any': [
                    {
                      'lhs': {
                        'type': 'FACT',
                        'value': 'cart',
                        'path': '$.xmasHampers[0].hamperItems[1].id'
                      },
                      'operator': 'equals',
                      'rhs': {
                        'type': 'LITERAL',
                        'value': 'foieGras'
                      }
                    }
                  ]
                },
                'successElements': [
                  {
                    'elementType': 'ACTION',
                    'actionType': 'UPDATE_LOCALISATION',
                    'key': 'my.loc.key4.success',
                    'value': 'my.loc.value4.success'
                  }
                ],
                'failureElements': [
                  {
                    'elementType': 'ACTION',
                    'actionType': 'UPDATE_LOCALISATION',
                    'key': 'my.loc.key5.failure',
                    'value': 'my.loc.value5.failure'
                  }
                ]
              },
              {
                'elementType': 'ACTION',
                'actionType': 'UPDATE_LOCALISATION',
                'key': 'my.loc.key6.success',
                'value': 'my.loc.value6.success'
              }
            ],
            'failureElements': [
              {
                'elementType': 'ACTION',
                'actionType': 'SET_FACT',
                'fact': 'UI_FACT_2',
                'value': false
              }
            ]
          }
        }
```

The method will find the `cart` fact in a `successElements` array.

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -195,38 +196,65 @@ export class RulesetExecutor {
...
-    const rulesWithoutContext: Rule[] = [];
+    const inputFactsForRule: Record<string, string[]> = {};
+    /**
+     * Recursively explores a rule to identify and collect input facts.
+     * Input facts are identified based on the 'FACT' type and operator-specific implicit dependencies.
+     *
+     * @param currentObject The current object being explored.
+     * @param ruleInputFacts A set to store the identified input facts for the rule.
+     */
+    const collectRuleInputFacts = (currentObject: any, ruleInputFacts: Set<string>) => {
```
kpanot, line 212: `currentObject` should be properly typed.
(this function can also be a private field to not be redefined at each execution)

packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.ts
```diff
@@ -277,6 +277,14 @@ export class RulesEngineExtractor {
               nbValues: 1
             }
           };
+          if (declaration.initializer && ts.isObjectLiteralExpression(declaration.initializer)) {
+            declaration.initializer.properties.forEach((property: any) => {
```
kpanot, line 281: `: any` should be avoid, how do you guaranty that `property.name` exists otherwise?

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -61,6 +62,32 @@ export class RulesetExecutor {
...
   }
 
+  /**
+   * Recursively explores a rule to identify and collect input facts.
+   * Input facts are identified based on the 'FACT' type and operator-specific implicit dependencies.
+   *
+   * @param currentObject The current object being explored.
+   * @param ruleInputFacts A set to store the identified input facts for the rule.
+   */
+  private collectRuleInputFacts(currentObject: any, ruleInputFacts: Set<string>) {
```
kpanot, line 72: Avoid `: any`

kpanot (approved):

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.ts
```diff
@@ -16,6 +16,58 @@ export const inRangeDate: Operator<Date, [DateInput, DateInput], DateInput> = {
...
+      throw new Error('No operatorFactValues. Unable to retrieve the current time.');
+    }
+    if (typeof operatorFactValues.o3rCurrentTime !== 'number') {
+      throw new Error('o3rCurrentTime value is not a number');
+    }
+    const currentTimeValue = operatorFactValues.o3rCurrentTime;
+    const now = new Date(currentTimeValue);
+    const leftDate = new Date(leftDateInput);
+    const targetDate = new Date(new Date(currentTimeValue).setMinutes(now.getMinutes() + +minutes));
+    return leftDate >= now && leftDate > targetDate;
```
fpaul-1A, line 64:
```suggestion
    return !dateInNextMinutes.evaluator(leftDateInput, minutes, operatorFactValues);
```
?

fpaul-1A (approved):

kpanot (approved):

PR #836, opened 2023-09-28
feat: findFilenameInTree helper function

vscaiceanu-1a:
## Proposed change

Helper function useful for looking up for files inside a Tree.


packages/@o3r/schematics/src/utility/tree.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ *
+ * Helper function that looks for files in the Tree
+ *
+ * @param memory where to store the results
+ * @param directory where to perform the search
+ * @param fileMatchesCriteria a function defining the criteria to look for
+ * @param ignoreDirectories optional parameter to ignore folders
+ */
+export function findFilenameInTree(memory: Set<FileEntry>, directory: DirEntry, fileMatchesCriteria: (file: string) => boolean, ignoreDirectories: string[] = ['node_modules', '.git', '.yarn']) {
```
kpanot, line 12: Never export a recursive function with a memory. You should export a function without memory using the recursive function

vscaiceanu-1a: how about 
```typescript
import type { DirEntry, FileEntry } from '@angular-devkit/schematics';

/**
 *
 * Helper function that looks for files in the Tree
 *
 * @param directory where to perform the search
 * @param fileMatchesCriteria a function defining the criteria to look for
 * @param ignoreDirectories optional parameter to ignore folders
 */
export function findFilenameInTree(directory: DirEntry, fileMatchesCriteria: (file: string) => boolean, ignoreDirectories: string[] = ['node_modules', '.git', '.yarn']): Set<FileEntry> {
  const memory = new Set<FileEntry>();

  function recursiveSearch(directory: DirEntry): void {
    if (ignoreDirectories.some(i => directory.path.includes(i))) {
      return;
    }

    directory.subfiles
      .filter(fileMatchesCriteria)
      .forEach((file) => memory.add(directory.file(file)!));

    directory.subdirs
      .forEach((dir) => recursiveSearch(directory.dir(dir)));
  }

  recursiveSearch(directory);

  return memory;
}
`
``

Comment (kpanot) [reply packages/@o3r/schematics/src/utility/tree.ts:12]: In term of perf: you instancy an impure rec function at each call, you can improve it to simply:

```typescript
function findFilenameInTreeRec(memory: Set<FileEntry>, directory: DirEntry, fileMatchesCriteria: (file: string) => boolean, ignoreDirectories: string[] = ['node_modules', '.git', '.yarn']) {
  if (ignoreDirectories.some(i => directory.path.includes(i))) {return memory;}

  directory.subfiles
    .filter(fileMatchesCriteria)
    .forEach((file) => memory.add(directory.file(file)!));

  directory.subdirs
    .forEach((dir) => findFilenameInTree(memory, directory.dir(dir), fileMatchesCriteria, ignoreDirectories));

  return memory;
}

export function fndFilenameInTree(directory: DirEntry, fileMatchesCriteria: (file: string) => boolean, ignoreDirectories: string[] = ['node_modules', '.git', '.yarn']) {
  const memory = new Set<FileEntry>();
  return [...findFilenameInTreeRec(directory, fileMatchesCriteria)];
```

packages/@o3r/schematics/src/utility/loaders.ts
```diff
@@ -6,6 +7,34 @@ import * as path from 'node:path';
 import type { PackageJson } from 'type-fest';
 import type { WorkspaceProject, WorkspaceSchema } from '../interfaces/index';
 
+function findFilenameInTreeRec(memory: Set<FileEntry>, directory: DirEntry, fileMatchesCriteria: (file: string) => boolean, ignoreDirectories: string[]) {
+  if (ignoreDirectories.some(i => directory.path.includes(i))) {
```
matthieu-crouzet, line 11: `includes` or `startsWith` ?

vscaiceanu-1a: I prefer includes so that we apply it to all levels (ex node_modules inside app folders)

packages/@o3r/schematics/src/utility/loaders.ts
```diff
@@ -6,6 +7,34 @@ import * as path from 'node:path';
...
+    return memory;
+  }
+
+  directory.subfiles
+    .filter(fileMatchesCriteria)
+    .forEach((file) => memory.add(directory.file(file)!));
+
+  directory.subdirs
+    .forEach((dir) => findFilenameInTreeRec(memory, directory.dir(dir), fileMatchesCriteria, ignoreDirectories));
```
fpaul-1A, line 20: hopefully there is no symlinks in the tree, so no potential infinite loops ?

vscaiceanu-1a: I think you can control this through the `ignoreDirectories` param. Do you think it's safer to add a depth limit?

kpanot: Good point, how do symlinks are translated inside the tree object?

vscaiceanu-1a: I did a test locally with a symlinked package in my node_modules and the tree is flattened so there's no risk for infinite loop

fpaul-1A: did you test with 2 symlinked folder referencing each other ?
```
node_modules
|  a
|  |  node_modules
|  |  |  b (symlink)
|  b
|  |  node_modules
|  |  |  a (symlink)
```

vscaiceanu-1a: No, I have not tried.
Should we ever access the `node_modules` folder? I haven't identified a use case for it. 
I propose that we always ignore the `node_modules` folders and we document this in the JSDoc. 
Not excluding `node_modules` considerably slows down this method.

fpaul-1A (dismissed):

packages/@o3r/schematics/src/utility/loaders.ts
vscaiceanu-1a: > did you test with 2 symlinked folder referencing each other ?
> 
> ```
> lib
> |  a
> |  |  node_modules
> |  |  |  b (symlink)
> |  b
> |  |  node_modules
> |  |  |  a (symlink)
> ```

Tests done with 2 folders linking one each other and we didn't run into no infinite loops.

packages/@o3r/schematics/src/utility/loaders.ts
matthieu-crouzet: findFilenameInTreeRec(..., ['directory-to-ignore'])
actual code will ignore 'path/not-directory-to-ignore'

```suggestion
  if (ignoreDirectories.some(i => directory.path.split(path.posix.sep).includes(i))) {
```

mrednic-1A (dismissed):

packages/@o3r/schematics/src/utility/loaders.ts
vscaiceanu-1a: you are right

fpaul-1A (approved):

kpanot (approved):

PR #847, opened 2023-10-02
fix: run linter correctly during schematics

matthieu-crouzet:
## Proposed change
Currently the linter is not running while using schematics
To be able to run correctly files need to be "printed" and not be only in tree memory, so it has to be a schematic run as a task to be done after the tree is printed


packages/@o3r/schematics/src/tasks/eslint/index.ts
```diff
@@ -1,8 +1,11 @@
+import * as path from 'node:path';
 import { TaskConfiguration, TaskConfigurationGenerator } from '@angular-devkit/schematics';
 import { NodePackageName, NodePackageTaskOptions } from '@angular-devkit/schematics/tasks/package-manager/options';
-import * as path from 'node:path';
 import { getPackageManager } from '../../utility/package-manager-runner';
 
+/**
+ * @deprecated will be removed in v10.
```
kpanot, line 7: Maybe you should give advice to what to use instead

packages/@o3r/schematics/schematics/eslint/index.ts
```diff
@@ -0,0 +1,13 @@
+import type { Rule } from '@angular-devkit/schematics';
+import { eslintFix } from '@o3r/schematics';
+import type { EslintSchematicsSchema } from './schema';
+
+/**
+ * Eslint Schematics
+ */
+export function eslint(options: EslintSchematicsSchema): Rule {
```
kpanot, line 8: This is really similar to [@angular-eslint/builder](https://github.com/angular-eslint/angular-eslint).
Is there is reason to have our custom schematic to run eslint?

packages/@o3r/schematics/src/rule-factories/eslint-fix/index.ts
```diff
@@ -9,29 +9,74 @@ import { getPackageManagerExecutor } from '../../utility/package-manager-runner'
...
+   */
+  configFile?: string;
+}
+
+/**
+ * Apply eslint on `files`
+ * @param files
+ * @param linterOptions
+ */
+export function eslintFix(files: string[], linterOptions: LinterOptions): Rule {
```
kpanot, line 37: This is quite confusing for 2 reasons:
- It is exposed as a rule but it is child-processing a script on the file system.
- It is using process command instead of going via the Task Api that can not make the run schedulable by the rule using it.

kpanot:
The problem that this PR try to solve is not clear :S.
The Schematics Task are actually run after the tree is printed to the file system. So the current exposition of ESLint (as Schematic task) as it is today should not be the original issue this PR wants to fix.

I am also not sure we want to expose a new schematic for eslint fixing instead of using the Angular's one :S.

PR #856, opened 2023-10-04
fix: enhance doc dynamic content module

pginoux-1A:
## Proposed change

Enhance read me for dynamic content module doc

### DIFF
--- a/packages/@o3r/dynamic-content/README.md
+++ b/packages/@o3r/dynamic-content/README.md
@@ -38,19 +38,22 @@ A pipe to be used in your component templates:
 
 ```html
 <img src="{{'assets-otter/imgs/logo.png' | dynamicContent}}" /> or
-<img [src]="'assets-otter/imgs/logo.png' | dynamicContent" />
+<img ngSrc="{{'assets-otter/imgs/logo.png' | dynamicContent}}"/>
 ```
 
 and a service to be used in your component classes, for example:
 
 ```typescript
-
+import { firstValueFrom } from 'rxjs';
 @Component({
   /* ... */
 })
 export class MyComponent {
   constructor(private service: DynamicContentService) {
-    const imgSrc = this.service.getMediaPath('assets/assets-otter/imgs/logo.png');
+  }
+  async getLogoPath() {
+    const result = await firstValueFrom(this.service.getMediaPathStream('assets/assets-otter/imgs/logo.png'));
+    //...
   }
 }
 ```


cpaulve-1A (dismissed):

mrednic-1A (approved):

packages/@o3r/dynamic-content/README.md
```diff
@@ -38,19 +38,22 @@ A pipe to be used in your component templates:
...
+import { firstValueFrom } from 'rxjs';
 @Component({
   /* ... */
 })
 export class MyComponent {
   constructor(private service: DynamicContentService) {
-    const imgSrc = this.service.getMediaPath('assets/assets-otter/imgs/logo.png');
+  }
+  async getLogoPath() {
+    const result = await firstValueFrom(this.service.getMediaPathStream('assets/assets-otter/imgs/logo.png'));
```
kpanot, line 55: I think we got the same discussion in another doc but we should not give an example something that we do not recommend because not compatible with the rule engine.
Maybe an example based on `| async` would be more clean.

pginoux-1A: That's more or less the difference between the media and content path. Everything that is out of the assets folder can't be override (else you could for example override the rulesets.json file itself, leading in very unpredictable behavior. Indeed for a logo like in the example, it doesn't really make sense to use the media path stream. Would you agree just giving an example with the rulesets.json in the app component as an example ?

kpanot: The argument would be the same as previous: we should not provide example that we already know that it will generate issue if it is used as it is :S

pginoux-1A: I am not sure I get it, rulesets.json can't be overridden, isn't it the typical example where getMediaPathStream can be used instead of getContentPathStream in the app component in order to retrieve the rulesets ?

kpanot: I may not have been enough explicite :S.
I am not talking about `rulesets.json` but about `assets/assets-otter/imgs/logo.png`.
A Rules Engine's rule, via the `UPDATE_ASSET` action, can override the path `assets/assets-otter/imgs/logo.png` at any moment. (This is actually the case for all paths returned by `getMediaPathStream`).

The point here is that, if you change the observable to promise, the `UPDATE_ASSET` will not have the expected effect anymore.
It is not a problem for me that an application does the promise conversion (because they are not using the Rules Engine for example) but I don't think we should provide this example that will be wrongly used by people and will be take as guideline that are actually not what we want to promote.

pginoux-1A: I agreed for the logo, cf my second message: "Indeed for a logo like in the example, it doesn't really make sense to use the media path stream. Would you agree just giving an example with the rulesets.json in the app component as an example ?" - That's why I proposed to update to example to put the ruleset.json instead.

pginoux-1A: The thing is, getMediaPathStream should not be used in "general cases". It explicit in the documentation that overrides are not considered when using this function. If we want to give an example it will be a specific use case for sure.

kpanot: I don't think so, and according to the documentation on the function: "any override is applied to the content, returns the override path instead".
I am not sure to understand what make you say that "getMediaPathStream should not be used in general cases".

PR #862, opened 2023-10-05
fix: invalid peer dependencies

cpaulve-1A:
## Proposed change

- Remove jest-preset-stylint from the peer dependencies of @o3r/stylelint-plugin as it is only used for testing of the module (kept as devDependencies)
- Set pixelmatch and pngjs as optional in @o3r/testing as there are only used in the @o3r/testing/visual-test subpackage

### DIFF
--- a/packages/@o3r/testing/package.json
+++ b/packages/@o3r/testing/package.json
@@ -139,6 +139,12 @@
     "@schematics/angular": {
       "optional": true
     },
+    "pixelmatch": {
+      "optional": true
+    },
+    "pngjs": {
+      "optional": true
+    },
     "protractor": {
       "optional": true
     },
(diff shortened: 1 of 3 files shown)


matthieu-crouzet (approved):

fpaul-1A (approved):

mrednic-1A (approved):

packages/@o3r/testing/package.json
```diff
@@ -139,6 +139,12 @@
     "@schematics/angular": {
       "optional": true
     },
+    "pixelmatch": {
```
kpanot, line 142: These deps are used by modules exposed in sub-entry only?
If the module using these deps are exposed at barrel level, you won't be able to turn them optional :S

kpanot (approved):

PR #865, opened 2023-10-05
fix: run linter task during schematics

matthieu-crouzet:
## Proposed change
run linter task during schematics

### DIFF
--- a/packages/@o3r/schematics/src/tasks/eslint/index.ts
+++ b/packages/@o3r/schematics/src/tasks/eslint/index.ts
@@ -1,13 +1,44 @@
 import { TaskConfiguration, TaskConfigurationGenerator } from '@angular-devkit/schematics';
 import { NodePackageName, NodePackageTaskOptions } from '@angular-devkit/schematics/tasks/package-manager/options';
-import * as path from 'node:path';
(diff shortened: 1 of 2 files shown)


mrednic-1A (approved):

packages/@o3r/schematics/src/tasks/eslint/index.ts
```diff
@@ -1,22 +1,59 @@
...
+   *
+   * @default true
+   */
+  continueOnError?: boolean;
+  /**
+   * If enabled, only errors are reported (--quiet option of ESLint CLI)
+   *
+   * @default true
+   */
+  hideWarnings?: boolean;
```
kpanot, line 27: maybe you can call it `showWarning` and `exitOnError` to default to `false` :D

matthieu-crouzet: As discussed 
Deprecated `hideWarnings` in favor to `showWarning` is not that important.
Change `continueOnError` to `exitOnError` is not a good idea because we loose the fact that we will exit successfully even if there are errors.

kpanot (approved):

mrednic-1A (approved):

PR #870, opened 2023-10-05
fix: remove mandatory @o3r/schematics dependency on @ama-sdk/core

cpaulve-1A:
## Proposed change

@o3r/schematics is only required to install @ama-sdk/core via the ng-add command and should not be a mandatory dependency but an optional peer dependency

### DIFF
--- a/packages/@ama-sdk/core/package.json
+++ b/packages/@ama-sdk/core/package.json
@@ -47,7 +47,6 @@
     "prepare:publish": "prepare-publish ./dist"
   },
   "dependencies": {
-    "@o3r/schematics": "workspace:^",
     "@swc/helpers": "^0.5.0",
     "tslib": "^2.5.3",
     "uuid": "^9.0.0"
@@ -56,6 +55,7 @@
     "@angular-devkit/schematics": "~16.2.0",
     "@angular/cli": "~16.2.0",
     "@angular/common": "~16.2.0",
+    "@o3r/schematics": "workspace:^",
     "@schematics/angular": "~16.2.0",
     "isomorphic-fetch": "^3.0.0",
     "rxjs": "^7.8.1",
@@ -69,6 +69,9 @@
       "optional": true
     },
     "@angular/common": {
+      "optional": true
+    },
+    "@o3r/schematics": {
       "optional": true
     },
     "@schematics/angular": {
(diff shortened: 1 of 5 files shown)


matthieu-crouzet (dismissed):

fpaul-1A (approved):

packages/@ama-sdk/core/package.json
```diff
@@ -47,7 +47,6 @@
     "prepare:publish": "prepare-publish ./dist"
   },
   "dependencies": {
-    "@o3r/schematics": "workspace:^",
```
kpanot, line 50: is it added by the `shell` generator?
Otherwise the `create` generation will fail.

packages/@ama-sdk/core/schematics/ng-add/index.ts
```diff
@@ -7,36 +7,52 @@ import * as ts from 'typescript';
  */
 export function ngAdd(): Rule {
 
-  const removeImports = async () => {
-    const {removePackages} = await import('@o3r/schematics');
-    return removePackages(['@dapi/sdk-core']);
+  const removeImports = async (_tree: Tree, context: SchematicContext) => {
```
kpanot, line 10:
```suggestion
  const removeImports = async (_, context) => {
```

packages/@ama-sdk/core/package.json
cpaulve-1A: let me check

matthieu-crouzet (approved):

packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/package.json.template
```diff
@@ -78,6 +78,7 @@
     "@commitlint/config-conventional": "^17.0.0",
     "@ama-sdk/schematics": "~<%= sdkCoreVersion %>",
     "@ama-sdk/core": "~<%= sdkCoreVersion %>",
+    "@o3r/schematics": "^<%= sdkCoreVersion %>",
```
mrednic-1A, line 81: __Open question:__ What if we put it directly in the `@ama-sdk/create`? We already have `@schematics/angular` as dep and other `@angular-devkit` deps. 
Because here it is not really a dev dependency of the sdk.

kpanot: What problem it would fix?
Actually the `@ama-sdk/create` is just the shell script to child process the installation schematics, its dependencies can not benefice to the installation process :S

kpanot (approved):

packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/package.json.template
mrednic-1A: I was thinking that we put a dep in the generated sdk, which is not needed there. And in `@ama-sdk/create` we already have dependencies which are not really needed for the shell script itself , but needed for the generators which are launched.

mrednic-1A (approved):

PR #846, opened 2023-10-02
feat(styling): add theming details input feature

kpanot:

PR #886, opened 2023-10-11
feat(styling): deprecated Otter Theme in default export

kpanot:
## Proposed change

Remove Otter Theming and expose it as sub entries to remove the dependency to Angular Material

## Related issues

- :rocket: PR #846


kpanot:
Duplicate of #897

PR #913, opened 2023-10-16
feat(extractor): add mapping between localization key and component

matthieu-crouzet:
## Proposed change
Add mapping between localization key and component to allow contextualization in CMS


matthieu-crouzet:
Solution implemented in https://github.com/AmadeusITGroup/otter/pull/915

PR #915, opened 2023-10-16
feat(extractor): add mapping between localization key and component (localizationKeys)

matthieu-crouzet:
## Proposed change
Add mapping between localization key and component to allow contextualization in CMS

- [x] Validate format with the CMS team  (other proposal https://github.com/AmadeusITGroup/otter/pull/913)

Example of component class metadata 
image

### DIFF
--- a/packages/@o3r/components/schemas/component.metadata.schema.json
+++ b/packages/@o3r/components/schemas/component.metadata.schema.json
@@ -125,6 +125,13 @@
       "linkableToRuleset": {
         "type": "boolean",
         "description": "Determine if the component is activating a ruleset"
+      },
+      "localizationKeys": {
+        "type": "array",
+        "description": "List of localization keys declared by the component",
+        "items": {
+          "type": "string"
+        }
       }
     }
   }
--- a/packages/@o3r/extractors/src/utils/localization.ts
+++ b/packages/@o3r/extractors/src/utils/localization.ts
@@ -1,0 +1,37 @@
+import * as ts from 'typescript';
+
+const localizationDecoratorName = 'Localization';
+
+/**
+ * Retrieve the localization json files from TS Code
+ *
+ * @param node TSNode of the angular component class
+ */
+export function getLocalizationFileFromAngularElement(node: ts.ClassDeclaration): string[] | undefined {
+  const localizationPaths: string[] = [];
+  node.forEachChild((item) => {
+    if (!ts.isPropertyDeclaration(item)) {
+      return;
+    }
+
+    item.forEachChild((decorator) => {
+      if (
+        !ts.isDecorator(decorator)
+        || !ts.isCallExpression(decorator.expression)
+        || !ts.isIdentifier(decorator.expression.expression)
+        || decorator.expression.expression.escapedText !== localizationDecoratorName
+      ) {
+        return;
+      }
+
+      const firstArg = decorator.expression.arguments[0];
+      if (!firstArg || !ts.isStringLiteral(firstArg)) {
+        return;
+      }
+
+      localizationPaths.push(firstArg.text);
+    });
+  });
+
+  return localizationPaths.length ? localizationPaths : undefined;
+}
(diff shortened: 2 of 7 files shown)


apps/showcase/project.json
```diff
@@ -221,7 +221,8 @@
     "extract-translations": {
       "executor": "@o3r/localization:extractor",
       "inputs": [
-        "source"
+        "source",
+        "{projectRoot}/src/**/*.localization.json"
```
kpanot, line 225: should be already included in `source`

packages/@o3r/extractors/src/utils/localization.ts
```diff
@@ -0,0 +1,32 @@
...
+      return;
+    }
+
+    item.forEachChild((decorator) => {
+      if (!ts.isDecorator(decorator)) {
+        return;
+      }
+
+      const text = decorator.getText(source);
+      const result = /^@Localization *\( *['"](.*)['"] *\)/.exec(text);
```
kpanot, line 22: Could go deeper in the node instead (it would support EOL)

deathpat:
The new schema proposed is ok for us !

packages/@o3r/extractors/src/utils/localization.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (!ts.isPropertyDeclaration(item)) {
+      return;
+    }
+
+    item.forEachChild((decorator) => {
+      if (
+        !ts.isDecorator(decorator)
+        || !ts.isCallExpression(decorator.expression)
+        || !ts.isIdentifier(decorator.expression.expression)
+        || decorator.expression.expression.escapedText !== 'Localization'
```
kpanot, line 21: Extract `Localization` to constant.
(The best scenario would be to use the `Localization.name` imported from `@o3r.localization` but not so easy in  `@o3r/component` except with a dynamic import and `@o3r.localization` as optional peerDep)

packages/@o3r/extractors/src/utils/localization.ts
```diff
@@ -0,0 +1,36 @@
...
+      ) {
+        return;
+      }
+
+      const firstArg = decorator.expression.arguments[0];
+      if (!firstArg || !ts.isStringLiteral(firstArg)) {
+        return;
+      }
+
+      localizationPaths.push(firstArg.text.replaceAll(/['"]/g, ''));
```
kpanot, line 31: Is the replace needed? `text` is not supposed to be the content directly?

packages/@o3r/extractors/src/utils/localization.ts
matthieu-crouzet: This file is in @o3r/extractors so it's not possible as @o3r/localization depends on @o3r/extractors

kpanot (dismissed):

packages/@o3r/extractors/src/utils/localization.ts
```diff
@@ -0,0 +1,38 @@
+import * as ts from 'typescript';
+
+const localizationDecoratorName = 'Localization';
+
+/**
+ * Retrieve the localization json files from TS Code
+ *
+ * @param node TSNode of the angular component class
+ * @param source Ts file source
```
vscaiceanu-1a, line 9: No source used here

packages/@o3r/components/schemas/component.metadata.schema.json
```diff
@@ -125,6 +125,12 @@
       "linkableToRuleset": {
         "type": "boolean",
         "description": "Determine if the component is activating a ruleset"
+      },
+      "localizationKeys": {
```
vscaiceanu-1a, line 129: a description field would be welcome here

packages/@o3r/localization/builders/helpers/localization.generator.ts
```diff
@@ -300,9 +269,9 @@ export class LocalizationExtractor {
     const localizationFiles = tsFiles
       .map((file) => ({file, source: program.getSourceFile(file)}))
       .map(({ file, source }) => ({ file, classes: source && this.getAngularClassNode(source), source}))
-      .filter(({ classes, source }) => !!classes && !!source)
-      .map(({file, classes, source}) => classes!
-        .map((classItem) => this.getLocalizationFileFromAngularElement(classItem, source!))
+      .filter(({ classes }) => !!classes)
+      .map(({file, classes}) => classes!
```
vscaiceanu-1a, line 273:
```suggestion
      .map(({ file, classes }) => classes!
```
spacing

kpanot (approved):

fpaul-1A (approved):

PR #921, opened 2023-10-17
fix: rules engine triggers with duplicates

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/rules-engine/src/engine/debug/engine.debug.ts
+++ b/packages/@o3r/rules-engine/src/engine/debug/engine.debug.ts
@@ -185,6 +185,7 @@ export class EngineDebugger {
     }
 
     return {
+      executionCounter,
       rulesetOutputExecution,
       allExecutionsValid,
       rulesetTriggers
--- a/packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
+++ b/packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
@@ -226,7 +226,8 @@ export class RulesetExecutor {
       }
 
       return combineLatest(this.ruleset.rules.map((rule) => {
-        const values$ = rule.inputFacts.map((fact) => this.rulesEngine.retrieveOrCreateFactStream(fact));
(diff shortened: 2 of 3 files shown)


matthieu-crouzet (approved):

matthieu-crouzet (approved):

packages/@o3r/rules-engine/src/services/rules-engine.service.ts
```diff
@@ -86,7 +86,8 @@ export class RulesEngineService implements OnDestroy {
 
     this.subscription.add(
       this.store.pipe(
-        select(selectAllRulesets)
+        select(selectAllRulesets),
+        distinctUntilChanged((a, b) => JSON.stringify(a) === JSON.stringify(b))
```
kpanot, line 90: Why do you need it?
The purpose of having memoize selector is to not have to do this kind of trick :S

fpaul-1A: the store emits twice with the same value, not sure why

kpanot: Would be interesting to find the reason of this double emission actually :S

fpaul-1A: After further investigation, a new event is emitted every time the `updateReducers` is called on ngrx side (for each `StoreModule.forFeature` that is imported for example)
The value emitted seems to be a deepCopy of the same object every time

fpaul-1A: The multiple emit comes from the default `mergeReducers` function from store-sync creating deep copy of the store
I've updated it to only create deep copy of the store entries that need to be merged
so I removed this `distinctUntilChanged` (assuming people use our fork of store-sync or override the default mergeReducers function accordingly)

kpanot (approved):

matthieu-crouzet (approved):

PR #922, opened 2023-10-17
feat(vscode-ext): use eslint config for @o3rWidget and @o3rWidgetParam intellisense

matthieu-crouzet:
## Proposed change
use eslint config for @o3rWidget and @o3rWidgetParam intellisense

https://github.com/AmadeusITGroup/otter/assets/52541061/0b7f5d0a-52c9-4a9b-a3f4-a71e80e68e98

### DIFF
--- a/apps/vscode-extension/package.json
+++ b/apps/vscode-extension/package.json
@@ -307,6 +307,9 @@
     "watch": "yarn nx run vscode-extension:compile:watch",
     "prepare:publish": "prepare-publish ./dist"
   },
+  "dependencies": {
+    "eslint": "^8.42.0"
+  },
   "devDependencies": {
     "@angular-devkit/build-angular": "~16.2.0",
     "@angular-devkit/core": "~16.2.0",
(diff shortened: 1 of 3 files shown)


packages/@o3r/schematics/src/utility/environment.ts
```diff
@@ -20,7 +20,7 @@ export const getEnvironmentInfo = () => {
   const packageManagerInfo = getPackageManagerInfo();
   let otterCorePackageJsonPath: string | undefined;
   try {
-    otterCorePackageJsonPath = require.resolve('@o3r/core/package.json');
```
kpanot, line 23: Hmmmm, why?

apps/vscode-extension/package.json
```diff
@@ -325,14 +328,14 @@
     "@o3r/dev-tools": "workspace:^",
     "@o3r/eslint-plugin": "workspace:^",
     "@o3r/schematics": "workspace:^",
+    "@types/eslint": "^8.42.0",
```
kpanot, line 331: is it needed?
It is not already provided in `eslint`?

packages/@o3r/schematics/src/utility/environment.ts
matthieu-crouzet: I forget to put it back :)
it was not working while testing with node-modules instead of pnp because nx see that as a circular dependencies between @o3r/schematics and @o3r/core

apps/vscode-extension/package.json
matthieu-crouzet: It doesn't seem to be provided

kpanot (approved):

fpaul-1A (approved):

PR #927, opened 2023-10-18
feat(vscode-ext): snippets for configuration tags

matthieu-crouzet:
## Proposed change
snippets for configuration tags

### DIFF
--- a/apps/vscode-extension/src/intellisense/configuration.ts
+++ b/apps/vscode-extension/src/intellisense/configuration.ts
@@ -43,16 +43,36 @@ const getConfigurationTagsFromEslintConfig = (eslintConfig: any, comment: string
   const widgetParamsToPropose = Object.keys(widgetConfig).filter((paramName) => !paramsPresent.has(paramName));
 
   return {
+    tags: {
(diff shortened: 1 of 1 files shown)


apps/vscode-extension/src/intellisense/configuration.ts
```diff
@@ -43,16 +43,36 @@ const getConfigurationTagsFromEslintConfig = (eslintConfig: any, comment: string
   const widgetParamsToPropose = Object.keys(widgetConfig).filter((paramName) => !paramsPresent.has(paramName));
 
   return {
+    tags: {
```
fpaul-1A, line 46: there is no risk of conflict if the snippet already exist elsewhere with the same name ?

fpaul-1A (approved):

apps/vscode-extension/src/intellisense/configuration.ts
matthieu-crouzet: It will override the previous one 
maybe in another PR we can think about adding these tags only when we are in config interface or property comment

PR #928, opened 2023-10-18
fix(loc-extractor): do not resolve libraries files (isApplication)

matthieu-crouzet:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/localization/builders/helpers/localization.generator.ts
+++ b/packages/@o3r/localization/builders/helpers/localization.generator.ts
@@ -6,6 +6,7 @@ import * as fs from 'node:fs';
 import * as glob from 'globby';
 import * as path from 'node:path';
 import * as ts from 'typescript';
+import type { LocalizationExtractorBuilderSchema } from '../localization-extractor/schema';
 
 /** List of Angular decorator to look for */
 const ANGULAR_ANNOTATION = ['Component', 'Injectable', 'Pipe'];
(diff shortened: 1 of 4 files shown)


packages/@o3r/localization/builders/helpers/localization.generator.ts
```diff
@@ -6,6 +6,7 @@ import * as fs from 'node:fs';
 import * as glob from 'globby';
 import * as path from 'node:path';
 import * as ts from 'typescript';
+import { LocalizationExtractorBuilderSchema } from '../localization-extractor/schema';
```
fpaul-1A, line 9:
```suggestion
import type { LocalizationExtractorBuilderSchema } from '../localization-extractor/schema';
```

packages/@o3r/localization/builders/helpers/localization.generator.ts
```diff
@@ -121,6 +135,7 @@ export class LocalizationExtractor {
         }
         return res;
       })
+      .filter(({ref}) => this.options.libraries.every((lib) => !ref.startsWith(lib)))
```
kpanot, line 138: would be cleaner to have an optional `option` instead of having defaults values that are not really meaningful.
It would lead to:

```suggestion
      .filter(({ref}) => (this.options.?.libraries || []).every((lib) => !ref.startsWith(lib)))
```

packages/@o3r/localization/builders/helpers/localization.generator.ts
kpanot, line 9:
```suggestion
import type { LocalizationExtractorBuilderSchema } from '../localization-extractor/schema';
```

fpaul-1A (dismissed):

matthieu-crouzet:
Will be fixed by https://github.com/AmadeusITGroup/otter/pull/936

PR #936, opened 2023-10-19
fix(loc-extractor): do not resolve libraries files (empty libraries)

matthieu-crouzet:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/localization/builders/helpers/localization.generator.ts
+++ b/packages/@o3r/localization/builders/helpers/localization.generator.ts
@@ -6,6 +6,7 @@ import * as fs from 'node:fs';
 import * as glob from 'globby';
 import * as path from 'node:path';
 import * as ts from 'typescript';
+import type { LocalizationExtractorBuilderSchema } from '../localization-extractor/schema';
 
 /** List of Angular decorator to look for */
 const ANGULAR_ANNOTATION = ['Component', 'Injectable', 'Pipe'];
@@ -59,7 +60,7 @@ export class LocalizationExtractor {
 
   private logger: logging.LoggerApi;
 
-  constructor(tsconfigPath: string, logger: logging.LoggerApi) {
+  constructor(tsconfigPath: string, logger: logging.LoggerApi, private options?: Partial<LocalizationExtractorBuilderSchema>) {
     this.tsconfigPath = tsconfigPath;
     this.logger = logger;
   }
@@ -123,14 +124,18 @@ export class LocalizationExtractor {
       })
       .map(({ref}) => {
         if (!ref.startsWith('.')) {
-          try {
-            return require.resolve(ref);
-          } catch {
-            // Should be a local file
+          if (this.options?.libraries?.length && this.options.libraries.every((lib) => !ref.startsWith(lib))) {
+            try {
+              return require.resolve(ref);
+            } catch {
+              // Should be a local file
+            }
           }
+          return undefined;
         }
         return folder ? path.resolve(folder, ref) : ref;
-      });
+      })
+      .filter((ref): ref is string => !!ref);
 
     return referencedFiles.length ? referencedFiles : undefined;
   }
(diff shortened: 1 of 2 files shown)


packages/@o3r/localization/builders/helpers/localization.generator.ts
```diff
@@ -362,7 +375,15 @@ export class LocalizationExtractor {
           .forEach((locKey) => addMetadata(this.generateMetadataItemFromLocalization(localizationMap[locFile].data[locKey], locKey), locFile))
       );
 
-    const localizationMetadata = Object.keys(metadata).map((k) => metadata[k]);
+    const localizationMetadata = Object.values(metadata);
+
+    localizationMetadata.forEach((data) => {
+      if (data.ref && !data.ref.includes(referenceSeparator) && !metadata[data.ref]) {
+        this.logger.error(`The key ${data.ref} is unknown but referenced by ${data.key}.`);
+        throw new O3rCliError('Unknown referenced key');
```
kpanot, line 383: Do you want to throw at the first key of we way to report them all before throwing?

packages/@o3r/localization/builders/helpers/localization.generator.ts
```diff
@@ -123,14 +126,18 @@ export class LocalizationExtractor {
       })
       .map(({ref}) => {
         if (!ref.startsWith('.')) {
+          if ((this.options?.libraries || []).some((lib) => ref.startsWith(lib))) {
```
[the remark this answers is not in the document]
matthieu-crouzet: For me we should not resolve if it's in the libraries list because later in the script we add all metadata from the librairies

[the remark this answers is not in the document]
matthieu-crouzet: https://github.com/AmadeusITGroup/otter/blob/0b96663ac08d9c925444c75cd952c882457179c2/packages/%40o3r/localization/builders/helpers/localization.generator.ts#L359-L363

packages/@o3r/localization/builders/helpers/localization.generator.ts
```diff
@@ -123,14 +124,18 @@ export class LocalizationExtractor {
...
```
fpaul-1A, line 127: it seems `this.options.libraries` cannot be undefined
```suggestion
          if (this.options?.libraries.some((lib) => ref.startsWith(lib))) {
```
?

packages/@o3r/localization/builders/helpers/localization.generator.ts
```diff
@@ -176,7 +181,10 @@ export class LocalizationExtractor {
     }
 
     if (loc.$ref) {
-      res.ref = loc.$ref.split('#/')[1];
+      const [refPath, refKey] = loc.$ref.split('#/');
```
fpaul-1A, line 184:
```suggestion
      const [refPath, refKey] = loc.$ref.split('#/', 2);
```

packages/@o3r/localization/builders/helpers/localization.generator.ts
matthieu-crouzet: options: Partial<LocalizationExtractorBuilderSchema>
so it can be undefined

kpanot (approved):

fpaul-1A (approved):

kpanot (approved):

packages/@o3r/localization/builders/helpers/localization.generator.ts
```diff
@@ -123,14 +124,18 @@ export class LocalizationExtractor {
       })
       .map(({ref}) => {
         if (!ref.startsWith('.')) {
-          try {
-            return require.resolve(ref);
-          } catch {
-            // Should be a local file
+          if (this.options?.libraries?.length && this.options.libraries.every((lib) => !ref.startsWith(lib))) {
```
fpaul-1A, line 127: shouldn't it be:
```suggestion
          if (!this.options?.libraries?.length || this.options.libraries.every((lib) => !ref.startsWith(lib))) {
```
or am I misunderstanding ?
it seems we use require only if the ref is not in the list of options.libraries, so I would expect to do it also if the list is empty

matthieu-crouzet: When there are no libraries we don't want to `require.resolve`, and we don't want too if we know the library because we will already get the information from its metadata

fpaul-1A (approved):

PR #939, opened 2023-10-20
fix: ignore files when generating with yarn

mrednic-1A:
## Proposed change
- updates on _gitignore_ when package manager is yarn
- sdk generator in monorepo to not touch yarn config

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/core/schematics/rule-factories/generic-updates/index.ts
+++ b/packages/@o3r/core/schematics/rule-factories/generic-updates/index.ts
@@ -1,9 +1,12 @@
 import type { Rule, SchematicContext, Tree } from '@angular-devkit/schematics';
+import { getPackageManager, WorkspaceSchema } from '@o3r/schematics';
 
 /**
  * Generic basic updates to be done on a repository
+ *
+ * @param workspaceConfig
  */
-export function genericUpdates(): Rule {
+export function genericUpdates(workspaceConfig?: WorkspaceSchema | null): Rule {
 
   const updateGitIgnore = (tree: Tree, _context: SchematicContext) => {
     // update gitignore
@@ -20,6 +23,29 @@ ${folderToExclude}
         `;
         }
       });
+      const packageManager = getPackageManager({workspaceConfig});
+      if (packageManager === 'yarn') {
+        gitignore += `
+
+### yarn ###
+# https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored
+
+.yarn/*
+!.yarn/releases
+!.yarn/patches
+!.yarn/plugins
+!.yarn/sdks
+!.yarn/versions
+
+# if you are NOT using Zero-installs, then:
+# comment the following lines
+# !.yarn/cache
+
+# and uncomment the following lines
+.pnp.*
+
+`;
+      }
       tree.overwrite('/.gitignore', gitignore);
     }
     return tree;
--- a/packages/@o3r/core/schematics/sdk/index.ts
+++ b/packages/@o3r/core/schematics/sdk/index.ts
@@ -39,15 +39,29 @@ export function generateSdk(options: NgGenerateSdkSchema): Rule {
       renameTemplateFiles()
     ]), MergeStrategy.Overwrite);
 
+    const packageManager = getPackageManager({ workspaceConfig });
+    const yarnrcPath = '.yarnrc.yml';
+    const yarnrcBeforeSdkGeneration = tree.exists(yarnrcPath) ? tree.readText(yarnrcPath) : '';
+
     return chain([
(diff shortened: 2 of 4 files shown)


matthieu-crouzet (approved):

packages/@o3r/core/schematics/rule-factories/generic-updates/index.ts
```diff
@@ -20,6 +23,29 @@ ${folderToExclude}
...
+!.yarn/plugins
+!.yarn/sdks
+!.yarn/versions
+
+# if you are NOT using Zero-installs, then:
+# comment the following lines
+# !.yarn/cache
+
+# and uncomment the following lines
+# .pnp.*
```
fpaul-1A, line 45: seems like one the 2 should be uncommented, no ?

mrednic-1A: Indeed :). Updated

fpaul-1A (approved):

packages/@o3r/core/schematics/rule-factories/generic-updates/index.ts
```diff
@@ -1,9 +1,12 @@
 import type { Rule, SchematicContext, Tree } from '@angular-devkit/schematics';
+import { getPackageManager, WorkspaceSchema } from '@o3r/schematics';
 
 /**
  * Generic basic updates to be done on a repository
+ *
+ * @param workspaceConfig
  */
-export function genericUpdates(): Rule {
+export function genericUpdates(workspaceConfig?: WorkspaceSchema | null): Rule {
```
kpanot, line 9: This change should not be in `@o3r/workspace` package?
This package manager related gitignore seem to be a workspace rule only :S

packages/@o3r/core/schematics/rule-factories/generic-updates/index.ts
```diff
@@ -20,6 +23,29 @@ ${folderToExclude}
         `;
         }
       });
+      const packageManager = getPackageManager({workspaceConfig});
+      if (packageManager === 'yarn') {
+        gitignore += `
+
+### yarn ###
```
kpanot, line 30: Question (that is not directly relative this PR): Do we execute a `yarn set version @${version}` before generating a yarn2+ gitigore?`

mrednic-1A: Indeed, we do not execute.
I have created an issue about (to fix the yarn version inside the repository for standalone sdks).
https://github.com/AmadeusITGroup/otter/issues/940

packages/@o3r/core/schematics/rule-factories/generic-updates/index.ts
mrednic-1A: Indeed :)
but in this version we don;t have the 'o3r/workspace'. It was included in '>= 9.4'.
I'll take care of it in the cascading.

packages/@o3r/core/schematics/sdk/index.ts
```diff
@@ -39,15 +39,29 @@ export function generateSdk(options: NgGenerateSdkSchema): Rule {
...
         name: scope,
         directory: targetPath,
-        packageManager: getPackageManager({workspaceConfig}),
+        packageManager,
         skipInstall: !!options.specPath || options.skipInstall
       }),
+      packageManager === 'yarn' ? (t) => {
+        if (yarnrcBeforeSdkGeneration) {
+          // discard changes done by sdk shell generator standalone on yarnrc
+          t.overwrite(yarnrcPath, yarnrcBeforeSdkGeneration);
```
cpaulve-1A, line 58: Why filter out the yarnrc from the shell instead of not adding it in the first place (via a param for example)?
Wouldn't you also get the same problem with the npmrc file?

mrednic-1A: The 'yarnrc' is added by the shell generator. I did not want to do it with a parameter, just to detect if we are running the sdk generator in standlone repo or in a monorepo. We already have the case to remove some generated files (generated by shell generator for standlone repo) and we can do that because we know we are in a generator which runs in a monorepo. Like that we don;t polute the shell generator with another param.

`Wouldn't you also get the same problem with the npmrc file?`
We'll not have the same problem because in the shell generator we check if we use yarn as package manager.

cpaulve-1A (approved):

kpanot (approved):

packages/@o3r/core/schematics/rule-factories/generic-updates/index.ts
mrednic-1A: ref: https://github.com/AmadeusITGroup/otter/pull/951

PR #941, opened 2023-10-20
chore: remove deprecated from v10

matthieu-crouzet:
## Proposed change
Remove all deprecated from v10


packages/@ama-sdk/core/src/plugins/index.ts
kpanot, line 1: These plugins should be removed from the package readme as well

packages/@o3r/core/src/core/interfaces/configuration.ts
```diff
@@ -13,32 +13,13 @@ export interface NestedConfiguration {
...
-/**
- * Configuration mode
- */
-export type ConfigurationMode = 'legacy' | 'strict';
-
-/**
- * Interface of configuration that is supported by the cms
- * NOTE: specify "legacy" as parameter type to disable strict mode and allow values not supported by the CMS. This option is strongly not recommended and deprecated.
- */
-export type Configuration<T extends ConfigurationMode = 'strict'> = T extends 'strict' ? StrictConfiguration : LegacyConfiguration;
```
kpanot, line 40: Hmmm I wonder if it would not be safer to keep it with `export type Configuration<T extends 'strict' = 'strict'> = T extends 'strict' ? StrictConfiguration : never;` because this parameter has not be officially deprecated :S.
Or provide an `ngUpdate` to find/replace `Configuration<'strict'>` to `Configuration`

packages/@o3r/core/schematics/ng-add/schema.ts
vscaiceanu-1a, line 1: should be removed from `schema.json` as well

packages/@o3r/core/migration.json
```diff
@@ -5,6 +5,11 @@
       "version": "8.2.0-rc.3",
       "description": "Updates of the Otter Library to v8.2.*",
       "factory": "./schematics/ng-update/index#updateV8_2"
+    },
+    "migration-v10_0": {
+      "version": "10.0.0",
```
vscaiceanu-1a, line 10: "version": "10.0.0-next.0",

packages/@o3r/core/src/core/interfaces/configuration.ts
matthieu-crouzet: ngUpdate provided

packages/@o3r/core/migration.json
matthieu-crouzet: We already have `10.0.0-alpha.0` `10.0.0-rc.0` 
we should decide which one we want to put in place 
issue with `10.0.0-next.0` is that migrating to this version will not include this script has it will be deploy at least on `10.0.0-next.22`

packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/readme.md
```diff
@@ -20,7 +20,7 @@ The following folders contain the generated code:
 The Code can be regenerated by running the following command:
 
 ```shell
-yarn schematics @ama-sdk/sdk:typescript-core --spec-path [path to your swagger file]
+yarn schematics @ama-sdk/schematics:typescript-core --spec-path [path to your swagger file]
```
kpanot, line 23: Can be fixed by `yarn create @ama-cli typescript --spec-path [path to your swagger file]`

docs/swagger-builder/readme.md
```diff
@@ -99,7 +99,7 @@ To perform validity check of a set of Swagger Specifications, additional tools a
 A code generator is provided to generate an extension of the **Amadeus Digital Experience API**.
 
 ```bash
-yarn yo @ama-sdk/sdk:api-extension
+yarn dlx -p @angular-devkit/schematics-cli schematics @ama-sdk/schematics:api-extension
```
cpaulve-1A, line 102: Are we sure this is the only part impacted by the sdk-generator/schematics change?

matthieu-crouzet: No if you know well the topic I would appreciate if you can review this doc :)

cpaulve-1A: Let's do it in a another PR

packages/@o3r/core/schematics/ng-update/v10.0/configuration.ts
```diff
@@ -0,0 +1,24 @@
+import type { Rule, Tree } from '@angular-devkit/schematics';
+import {getFilesWithExtensionFromTree} from '@o3r/schematics';
+
+const matchingRegex = /(Configuration)<\s+['"]strict['"]\s+>/g;
```
kpanot, line 4: I wonder it you can not remove `legacy` as well as there are both deprecated:

```suggestion
const matchingRegex = /(Configuration)<\s*['"](?:strict|legacy)['"]\s*>/g;
```

(not sure to understand why `\s+` instead of `\s*` as well)

cpaulve-1A (approved):

cpaulve-1A (approved):

PR #963, opened 2023-10-26
fix(showcase): build fails when localization-metadata are missing

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/apps/showcase/project.json
+++ b/apps/showcase/project.json
@@ -32,8 +32,8 @@
       "dependsOn": [
         "extract-translations",
         "extract-components",
-        "compile",
-        "generate-translations"
+        "generate-translations",
+        "compile"
       ]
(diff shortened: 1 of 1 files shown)


matthieu-crouzet (approved):

apps/showcase/project.json
```diff
@@ -32,8 +32,8 @@
       "dependsOn": [
         "extract-translations",
         "extract-components",
-        "compile",
-        "generate-translations"
+        "generate-translations",
```
kpanot, line 35: This change should not have any effect, so no problem :D

kpanot (approved):

matthieu-crouzet (approved):

PR #976, opened 2023-10-27
chore: move monorepo jest helpers methods to o3r-workspace

matthieu-crouzet:
## Proposed change
move monorepo jest helpers methods to o3r-workspace


apps/showcase/jest.config.js
```diff
@@ -1,4 +1,4 @@
-const { getJestModuleNameMapper } = require('@o3r/dev-tools');
+const { getJestModuleNameMapper } = require('@o3r/workspace');
```
kpanot, line 1: Are we sure we want to move it to `workspace`?
If we do so, it meets that the all the generated app/libs should have a dep to `workspace`.
Which is not what we want actually.

I understand that we need to move it out of the `dev-tools` package and that the `core` is not an option (becasue build in ESM).
What is about `testing`?

packages/@o3r/workspace/schematics/sdk/index.ts
```diff
@@ -44,7 +43,7 @@ export function generateSdk(options: NgGenerateSdkSchema): Rule {
     const yarnrcBeforeSdkGeneration = tree.exists(yarnrcPath) ? tree.readText(yarnrcPath) : '';
 
     return chain([
-      externalSchematic<NgGenerateTypescriptSDKShellSchematicsSchema>('@ama-sdk/schematics', 'typescript-shell', {
```
kpanot, line 47: Why is it removed?
If the remove of the type is justified, is it not possible to remove the dep to `@ama-sdk/schematics` as well?

PR #977, opened 2023-10-27
fix: fixes in generated tsconfigs

mrednic-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


packages/@o3r/core/schematics/library/rules/rules.ng.ts
```diff
@@ -106,7 +86,7 @@ export function ngGenerateModule(options: NgGenerateModuleSchema & { targetPath:
           Object.entries(configFile.compilerOptions.paths).filter(([pathName, _]) => pathName !== options.name));
         configFile.compilerOptions.paths[options.packageJsonName] = [
           path.posix.join(relativeTargetPath, 'dist'),
-          path.posix.join(relativeTargetPath, 'src', 'public_api')
+          path.posix.join(relativeTargetPath, 'src', 'public-api')
```
kpanot, line 89: if we want to align with our packages, we should generate `public_api.json`.
Is this name has been also updated in the ng-packagr.json (in case of lib only)?

packages/@o3r/create/src/index.ts
```diff
@@ -115,7 +115,7 @@ const isNgNewOptions = (arg: string) => {
 };
 
 const createNgProject = () => {
-  const { error } = spawnSync(process.execPath, [binPath, 'new', ...args.filter(isNgNewOptions)], {
+  const { error } = spawnSync(process.execPath, [binPath, 'new', ...args.filter(isNgNewOptions), '--skip-install'], {
```
kpanot, line 118: why `--skip-install`? Your issue was not regarding `--no-commit`?

packages/@o3r/core/schematics/library/rules/rules.ng.ts
mrednic-1A: In fact it is the default value generated by `@schematics/angular` generator for library. This is forwarded for the file name and for the entry in `ng-packagr.json`.
https://github.com/angular/schematics-angular-builds/blob/bb494b9b765e37b48720e364db8d3f4aedd04f23/library/schema.json#L23

packages/@o3r/create/src/index.ts
mrednic-1A: The problem I had was the fact that the update of gitignore done by 'ng add o3r/core' was after the `install` task done by '@angular/schematics'. Looking closely, the install task is not needed in the context of o3r/create. 
So we speed a bit the generation and I don;t have the problem with installation files not being git ignored.

vscaiceanu-1a (approved):

kpanot (approved):

PR #989, opened 2023-10-31
feat(extractor): support o3rRequired tag on config property for the required field in metadata

matthieu-crouzet:
## Proposed change
Support o3rRequired tag on config property for the required field in metadata


packages/@o3r/extractors/src/utils/config-doc.ts
```diff
@@ -119,6 +122,14 @@ export function getTagsFromDocComment(docComment: DocComment): string[] | undefi
   return (tags && tags.indexOf(']') > 0) && tags.split(']')[0].match(/\w+/g) || undefined;
 }
 
+/**
+ * Is the tag `@o3rRequired` present in `docText`
+ * @param docText
+ */
+export function isO3rRequiredTagPresent(docText: string): boolean {
+  return !!docText.match(/@o3rRequired/);
```
fpaul-1A, line 130:
```suggestion
  return /@o3rRequired/.test(docText);
```

kpanot (approved):

PR #993, opened 2023-11-02
chore: move monorepo jest helper to the workspace package

matthieu-crouzet:
## Proposed change
move monorepo jest helper to the workspace package


package.json
```diff
@@ -11,7 +11,7 @@
     "get:current-dir": "node -e 'process.stdout.write(process.cwd());'",
     "build": "yarn nx run-many --target=build --parallel $(yarn get:cpus-number)",
     "build:affected": "yarn nx affected --target=build --parallel $(yarn get:cpus-number) --output-style stream",
-    "build:tools": "yarn nx run-many --target=build --projects=eslint-plugin,dev-tools --parallel $(yarn get:cpus-number)",
+    "build:tools": "yarn nx run-many --target=build --projects=eslint-plugin,dev-tools,workspace --parallel $(yarn get:cpus-number)",
```
kpanot, line 14: How did it impact the postinstall (in term of time) for a build without cache?
If I remember the workspace is quite long to build because of its deps 😭 
If it is bad, we should think about having these scripts in build-helpers 😔 .

Can dev-tools be removed of the build packages?

packages/@ama-sdk/showcase-sdk/package.json
```diff
@@ -77,6 +74,7 @@
     "@nx/jest": "~16.10.0",
     "@o3r/eslint-config-otter": "workspace:^",
     "@o3r/eslint-plugin": "workspace:^",
+    "@o3r/workspace": "workspace:^",
```
kpanot, line 77: You should not have this deep in an app.
Why do you need it?
If it is for the set-version, do you need a set-version here instead of the workspace level only?

package.json
matthieu-crouzet: When removing dev-tools, running `build:tools --skip-nx-cache` takes 1m0s on my machine
On release 10-next, running the same command take 44s on my machine 
I think it's acceptable

packages/@ama-sdk/showcase-sdk/package.json
matthieu-crouzet: When we generate an sdk we generate it with the script `set:version` it may not be needed if we are in a library but only in standalone mode. We can open an issue to fixed that point in another PR

package.json
kpanot: I would say that both or not actually but ok 😁

packages/@ama-sdk/showcase-sdk/package.json
kpanot: Agree for the task to update the generators. But this fix can be applied to our SDK in this PR

matthieu-crouzet: I did it for our SDK

matthieu-crouzet: https://github.com/AmadeusITGroup/otter/issues/998

vscaiceanu-1a (approved):

packages/@o3r/dev-tools/src/utils/check-peer-deps.ts
```diff
@@ -4,7 +4,7 @@ import { satisfies } from 'semver';
 
 /**
  * Interface containing a npm package name, needed version and optionally found version
- * @deprecated will be removed in Otter v12.
+ * @deprecated You can use the one expose in `@o3r/schematics`, will be removed in Otter v12.
```
fpaul-1A, line 7:
```suggestion
 * @deprecated You can use the one exposed in `@o3r/schematics`, will be removed in Otter v12.
```

packages/@o3r/dev-tools/src/utils/check-peer-deps.ts
```diff
@@ -18,7 +18,7 @@ export interface PackageVersion {
 /**
  * Check if the first level of peer deps of a given package are installed.
  * List all not installed packages or packages with version mismatch
- * @deprecated will be removed in Otter v12.
+ * @deprecated You can use the one expose in `@o3r/schematics`, will be removed in Otter v12.
```
fpaul-1A, line 21:
```suggestion
 * @deprecated You can use the one exposed in `@o3r/schematics`, will be removed in Otter v12.
```

packages/@o3r/dev-tools/src/utils/is-package-installed.ts
```diff
@@ -1,8 +1,7 @@
 /**
  * Check if an npm package is installed
- *
  * @param packageName The package to check
- * @deprecated will be removed in Otter v12.
+ * @deprecated You can use the one expose in `@o3r/schematics`, will be removed in Otter v12.
```
fpaul-1A, line 4:
```suggestion
 * @deprecated You can use the one exposed in `@o3r/schematics`, will be removed in Otter v12.
```

packages/@o3r/dev-tools/src/utils/string.ts
```diff
@@ -1,7 +1,6 @@
 /**
  * Specific padding for number
- * @deprecated will be removed in Otter v12.
- *
+ * @deprecated You can use the one expose in `@ama-sdk/core`, will be removed in Otter v12.
```
fpaul-1A, line 3:
```suggestion
 * @deprecated You can use the one exposed in `@ama-sdk/core`, will be removed in Otter v12.
```

fpaul-1A (approved):

vscaiceanu-1a (approved):

PR #1010, opened 2023-11-06
fix(rules-engine): avoid evaluating conditions when not needed

fpaul-1A:
## Proposed change

We can stop evaluating conditions as soon as one is false in an 'ALL' condition (as soon as one true for 'ANY' condition)
This avoid unnecessary errors when combining `isDefined` with another operator that expects the input fact to be defined for example

E.g.
```json
{
  "elementType": "RULE_BLOCK",
  "blockType": "IF_ELSE",
  "condition": {
  "all": [
    {
	  "lhs": {
	    "type": "FACT",
	    "value": "outboundDate"
	  },
	  "operator": "isDefined"
    },
    {
	  "lhs": {
	    "type": "FACT",
	    "value": "outboundDate"
	  },
	  "operator": "duringSummer"
    }
  ]
}
```
this is currently throwing an error that should not be thrown when `outboundDate` is undefined
`Error: Invalid left operand : undefined`

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
+++ b/packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
@@ -189,8 +189,8 @@ export class RulesetExecutor {
       return !this.evaluateCondition(nestedCondition.not, factsValue, runtimeFactValues);
     }
     if (nestedCondition.all || nestedCondition.any) {
-      const children = (nestedCondition.all || nestedCondition.any).map((condition) => this.evaluateCondition(condition, factsValue, runtimeFactValues));
-      return isAllConditions(nestedCondition) ? children.every((value) => value) : children.some((value) => value);
+      const evaluate = (condition: NestedCondition) => this.evaluateCondition(condition, factsValue, runtimeFactValues);
+      return isAllConditions(nestedCondition) ? nestedCondition.all.every(evaluate) : nestedCondition.any.some(evaluate);
     }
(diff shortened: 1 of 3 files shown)


kpanot (approved):

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -189,8 +189,8 @@ export class RulesetExecutor {
       return !this.evaluateCondition(nestedCondition.not, factsValue, runtimeFactValues);
     }
     if (nestedCondition.all || nestedCondition.any) {
-      const children = (nestedCondition.all || nestedCondition.any).map((condition) => this.evaluateCondition(condition, factsValue, runtimeFactValues));
-      return isAllConditions(nestedCondition) ? children.every((value) => value) : children.some((value) => value);
+      const evaluate = (condition: NestedCondition) => this.evaluateCondition(condition, factsValue, runtimeFactValues);
+      return isAllConditions(nestedCondition) ? nestedCondition.all.every(evaluate) : nestedCondition.any.some(evaluate);
```
kpanot, line 193: Maybe you can add a IT to ensure the non unnecessary evaluation of the rules

kpanot (approved):

matthieu-crouzet (approved):

kpanot (approved):

PR #1038, opened 2023-11-10
Angular 17

fpaul-1A:
## Proposed change

Migration to Angular 17

- ~~Pending ngrx release 17.0~~
- ng-bootstrap update requires bootstrap 5.3 which is currently not supported by design-factory
- ~~some issues during the resolving of sass files for the showcase app~~

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


packages/@o3r/workspace/schematics/sdk/rules/rules.nx.ts
```diff
@@ -1,6 +1,6 @@
-import type { Rule } from '@angular-devkit/schematics';
-import type { NgGenerateSdkSchema } from '../schema';
-import * as path from 'node:path';
+import type { Rule } from "@angular-devkit/schematics";
```
matthieu-crouzet, line 1: small issue `'` has been changed to `"`

nx.json
```diff
@@ -299,28 +195,89 @@
...
+    "extract-rules-engine": {
+      "cache": true
+    },
+    "build-swagger": {
+      "cache": true
+    },
+    "i18n": {
+      "cache": true
+    },
+    "@nx/jest:jest": {
```
kpanot, line 269: This is the name of a `executor`, not a `task. Do rules can be applied according to executor now? Should them be defined in `targetDefaults` as well?

nx.json
```diff
@@ -299,28 +195,89 @@
...
+      "cache": true,
+      "inputs": ["default", "^default"],
+      "options": {
+        "passWithNoTests": true
```
kpanot, line 273: I think `"passWithNoTests": true` should be the exception, not the default value

nx.json
```diff
@@ -299,28 +195,89 @@
...
+    "build-swagger": {
+      "cache": true
+    },
+    "i18n": {
+      "cache": true
+    },
+    "@nx/jest:jest": {
+      "cache": true,
+      "inputs": ["default", "^default"],
+      "options": {
```
kpanot, line 272: `options` and `configurations ` are now available globally via the `nx.json`?

nx.json
fpaul-1A: yes executor can be used in targetDefaults according to the docs but it seems that Nx won't try to read the config for the task name if a config for the executor is found so I think I should remove that part

https://nx.dev/reference/nx-json#target-defaults
> Note that the inputs and outputs are respecified on the @nx/js:tsc default configuration. This is required, as when reading target defaults Nx will only ever look at one key. If there is a default configuration based on the executor used, it will be read first. If not, Nx will fall back to looking at the configuration based on target name. For instance, running nx build project will read the options from targetDefaults[@nx/js:tsc] if the target configuration for build uses the @nx/js:tsc executor. It would not read any of the configuration from the build target default configuration unless the executor does not match.
```

File: nx.json (hunk 3 above)
Comment (fpaul-1A) [reply nx.json:273]: agreed

File: nx.json (hunk 4 above)
Comment (fpaul-1A) [reply nx.json:272]: I don't know if this is anything new but according to the docs we can do that
https://nx.dev/recipes/running-tasks/reduce-repetitive-configuration#reduce-configuration-with-targetdefaults

File: packages/@o3r/workspace/schematics/sdk/rules/rules.nx.ts (hunk 1 above)
Comment (fpaul-1A) [reply packages/@o3r/workspace/schematics/sdk/rules/rules.nx.ts]: nice catch

File: nx.json (hunk 4 above)
Comment (kpanot) [reply nx.json]: Nice, this can help to reduce a lot of duplicated config

File: packages/@ama-sdk/core/src/plugins/mock-intercept/mock-intercept.fetch.ts [hunk 5]
```diff
@@ -23,7 +23,8 @@ export class MockInterceptFetch implements FetchPlugin {
         await this.options.adapter.initialize();
 
         let responsePromise = fetchCall;
-        if (!context.options.headers || !(context.options.headers instanceof Headers) || !context.options.headers.has(CUSTOM_MOCK_OPERATION_ID_HEADER)) {
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+        if (!context.options.headers || !(context.options.headers instanceof Headers) || !(context.options.headers as Headers).has(CUSTOM_MOCK_OPERATION_ID_HEADER)) {
```
fpaul-1A, line 27: workaround

packages/@ama-sdk/core/src/plugins/mock-intercept/mock-intercept.fetch.ts
```diff
@@ -33,7 +34,8 @@ export class MockInterceptFetch implements FetchPlugin {
           responsePromise = new Promise<Response>((resolve) => setTimeout(resolve, delay)).then(() => resp);
         }
 
-        const operationId = context.options.headers.get(CUSTOM_MOCK_OPERATION_ID_HEADER)!;
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+        const operationId = (context.options.headers as Headers).get(CUSTOM_MOCK_OPERATION_ID_HEADER)!;
         try {
```
fpaul-1A, line 39: workaround

nx.json
```diff
@@ -135,9 +82,9 @@
   },
   "targetDefaults": {
     "lint": {
-      "inputs": [
-        "lint"
-      ]
+      "inputs": ["lint"],
+      "cache": true,
+      "dependsOn": ["^build"]
```
fpaul-1A, line 87: workaround

apps/github-cascading-app/package.json
```diff
@@ -18,30 +18,30 @@
   },
   "devDependencies": {
     "@actions/github": "^5.1.1",
-    "@angular-devkit/build-angular": "~16.2.0",
-    "@angular-devkit/core": "~16.2.0",
+    "@angular-devkit/build-angular": "~17.0.0",
+    "@angular-devkit/core": "~17.0.0",
     "@angular-eslint/builder": "~16.2.0",
```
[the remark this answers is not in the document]
fpaul-1A: I'm not sure, I've updated it to match the others

packages/@ama-sdk/schematics/jest.config.js
```diff
@@ -3,5 +3,6 @@ const getJestConfig = require('../../../jest.config.ut').getJestConfig;
 /** @type {import('ts-jest/dist/types').JestConfigWithTsJest} */
 module.exports = {
   ...getJestConfig(__dirname, false),
-  displayName: require('./package.json').name
+  displayName: require('./package.json').name,
+  testTimeout: 30 * 1000
```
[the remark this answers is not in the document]
fpaul-1A: I got some random timeout issues, I suppose it's due to the fact that we actually write files on the disk for these tests

vscaiceanu-1a:
could you ensure that `@o3r create` works with Angular 17?

nx.json
```diff
@@ -166,85 +121,58 @@
...
+      },
+      "configurations": {
+        "ci": {
+          "ci": true,
+          "codeCoverage": true
+        }
+      },
+      "cache": true
     },
     "prepare-publish": {
```
kpanot, line 136: missing `cache: true`, is it on purpose?

fpaul-1A:
> could you ensure that `@o3r create` works with Angular 17?

I've updated the code to set standalone as default for new application
If we attempt to force `--standalone=true` we end-up with the following error:
> Could not find bootstrap module
(until we support it)

package.json
```diff
@@ -202,25 +205,28 @@
...
     "jsonc-eslint-parser": "~2.4.0",
+    "jsonpath-plus": "^7.0.0",
     "lint-staged": "^13.0.0",
     "minimist": "^1.2.6",
-    "ng-packagr": "~16.2.0",
+    "ng-packagr": "~17.0.0",
+    "ngx-highlightjs": "^10.0.0",
     "npm-run-all": "^4.1.5",
-    "nx": "~16.10.0",
-    "nx-cloud": "~16.4.0",
```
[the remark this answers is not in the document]
fpaul-1A: `nx-cloud` is now part of `nx` package directly
https://blog.nrwl.io/nx-17-0-has-landed-1dad19176ff3#71d9

matthieu-crouzet (approved):

Issue #882, opened 2023-10-10
[Feature]: sdks - apps generation alignment


mrednic-1A:
### Context

We are in a workspace generated with o3r generator.
When generating an app, the testing and linting parts are not added by default.
We display logs with the commands to run after the application is generated.
For an sdk, we generate directly with 'jest' framework included and with '@o3r/eslint-*' packages included as well.

The idea of this feature is to align what is generated for the app with what is generated for an sdk.

### Proposal

To align the generations and enforce guidelines, 'jest' framework (and configs) should be added by default at the workspace generation. 
Like that when generating an app, the user will be asked if he wants to add 'o3r/testing' module on top of jest (today behavior if jest is already installed). For the sdk generation nothing changes.
For the linting part, for the apps we stay as today (nothing installed by default) and for the sdks we remove the defaults added with the generation (we'll log the same message that the 'o3r/eslint-*' packages can be added later.

PR #1045, opened 2023-11-13
fix: testing config at generation time

mrednic-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/core/schematics/sdk/rules/update-ts-paths.rule.ts
+++ b/packages/@o3r/core/schematics/sdk/rules/update-ts-paths.rule.ts
@@ -34,6 +34,9 @@ export function updateTsConfig(targetPath: string, projectName: string, scope: s
       `${relativeTargetPath}/dist`,
       `${relativeTargetPath}/src/index`
     ];
+    configWithPath.content.compilerOptions.paths[`${scope ? `@${scope}/` : ''}${projectName}/fixtures`] = [
+      `${relativeTargetPath}/src/fixtures/jest`
+    ];
     configWithPath.content.compilerOptions.paths[`${scope ? `@${scope}/` : ''}${projectName}/*`] = [
       `${relativeTargetPath}/dist/*`,
       `${relativeTargetPath}/src/*`
--- a/packages/@o3r/testing/schematics/ng-add/fixture/index.ts
+++ b/packages/@o3r/testing/schematics/ng-add/fixture/index.ts
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/naming-convention */
 import { chain, Rule, SchematicContext, Tree } from '@angular-devkit/schematics';
-import { getWorkspaceConfig } from '@o3r/schematics';
+import { getTestFramework, getWorkspaceConfig } from '@o3r/schematics';
 
 import * as ts from 'typescript';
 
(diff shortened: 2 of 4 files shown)


matthieu-crouzet (approved):

packages/@o3r/testing/schematics/ng-add/index.ts
```diff
@@ -114,7 +114,8 @@ export function ngAdd(options: NgAddSchematicsSchema): Rule {
           const jestConfigFilesForProject = () => mergeWith(apply(url('./templates/project'), [
             template({
               ...options,
-              rootRelativePath: path.posix.relative(workingDirectory, tree.root.path.replace(/^\//, './'))
+              rootRelativePath: path.posix.relative(workingDirectory, tree.root.path.replace(/^\//, './')),
+              isAngularSetup: tree.exists('/angular.json')
```
fpaul-1A, line 118: I don't think this condition is accurate
This condition will almost always be true, but we only want to set the angular setup on jest.config if it's needed (not the case for the sdk for example)
Maybe we could check if @angular is present in the dependencies of the project ?

packages/@o3r/core/schematics/sdk/rules/update-ts-paths.rule.ts
```diff
@@ -34,6 +34,9 @@ export function updateTsConfig(targetPath: string, projectName: string, scope: s
       `${relativeTargetPath}/dist`,
       `${relativeTargetPath}/src/index`
     ];
+    configWithPath.content.compilerOptions.paths[`${scope ? `@${scope}/` : ''}${projectName}/fixtures`] = [
+      `${relativeTargetPath}/src/fixtures/jest`
```
fpaul-1A, line 38: is it ok to add this without knowing if `jest` will be used to run the tests ?

packages/@o3r/testing/schematics/ng-add/index.ts
mrednic-1A: sdk generation is a special case. Today it is generated with its own testing setup, nothing to do with `o3r/testing` package. 
For a monorepo case we have a bit more complex story which is described in this issue https://github.com/AmadeusITGroup/otter/issues/882 to align the generation of the possible projects (app/lib/sdk)

packages/@o3r/testing/schematics/ng-add/fixture/index.ts
```diff
@@ -49,23 +46,60 @@ export function updateFixtureConfig(options: { projectName?: string | null | und
         return libs;
       }, []);
 
-      tsconfigCompilerOptions.paths = Object.entries(tsconfigCompilerOptions.paths || {}).reduce<Record<string, unknown>>((acc, [key, value]) => {
-        if (oldPaths.indexOf(key) === -1) {
-          acc[key] = value;
+      const testFramework = options.testingFramework || getTestFramework(getWorkspaceConfig(tree), context);
+      if (testFramework === 'jest') {
+        tsconfigCompilerOptions.types ||= [];
```
fpaul-1A, line 51: this may break the build in some cases
> By default all visible ”@types” packages are included in your compilation. Packages in node_modules/@types of any enclosing folder are considered visible. For example, that means packages within ./node_modules/@types/, ../node_modules/@types/, ../../node_modules/@types/, and so on.
If types is specified, only packages listed will be included in the global scope.

packages/@o3r/core/schematics/sdk/rules/update-ts-paths.rule.ts
mrednic-1A: For now sdk comes generated directly with 'jest', so the change is fixing the missing path. 
I link again the issue of the alignement of project generation https://github.com/AmadeusITGroup/otter/issues/882

fpaul-1A (approved):

PR #1046, opened 2023-11-13
fix: enhance error message for the Otter Chrome Extension

vscaiceanu-1a:
## Proposed change

Some users think the Chrome Extension isn't working but in fact they didn't give permissions to access their Otter based app. The goal of this PR is to enhance the error message when the Extension cannot reach the app.

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/apps/chrome-devtools/src/components/app-connection/app-connection.component.html
+++ b/apps/chrome-devtools/src/components/app-connection/app-connection.component.html
@@ -4,7 +4,7 @@
   </div>
 
   <div *ngSwitchCase="'timeout'" class="alert alert-danger m-2" role="alert">
-    The application does not seem to be compatible with the Otter Chrome DevTools.
+    The application does not seem to be compatible with the Otter Chrome DevTools or the Otter Chrome Extesion does not have permission to access this page.
   </div>
 
   <ng-container *ngSwitchCase="'connected'">


apps/chrome-devtools/src/components/app-connection/app-connection.component.html
```diff
@@ -4,7 +4,7 @@
   </div>
 
   <div *ngSwitchCase="'timeout'" class="alert alert-danger m-2" role="alert">
-    The application does not seem to be compatible with the Otter Chrome DevTools.
+    The application does not seem to be compatible with the Otter Chrome DevTools or the Otter Chrome Extesion does not have permission to access this page.
```
kpanot, line 7: Instead of adding a message, maybe you can just emit and display the error when then `onConnect` 's `chrome.scripting.executeScript` throw

vscaiceanu-1a: Are you saying that there's a way to detect whether  the user did not give permission to access the page or the application is not Otter compatible?

kpanot: When the permission is not given, the call to the Chrome Api is failing (according to the doc).
According to the case descriotion, the missing permission seems to be the page writring permission allowing `chrome.scripting.executeScript` so my question is: why nnot handling this function call failure instead of creating a message containing another possible failing case (that we can actually detect)

cpaulve-1A, line 7:
```suggestion
    The application does not seem to be compatible with the Otter Chrome DevTools or the Otter Chrome Extension does not have permission to access this page.
```

sdo-1A (approved):

PR #1047, opened 2023-11-13
fix: prepareRuleset is private method

### LINKED ISSUE #1044: [Bug]: prepareRuleset method should be private
### Package name

rules-engine

### Package version

9.4.0-alpha

### Reproduction steps

`prepareRuleset` method under `packages/@o3r/rules-engine/src/engine/ruleset-executor.ts` should pe private and not public

### Current result

`prepareRuleset` is public but used only within `ruleset-executor.ts`

### Expected result

`ruleset-executor.ts` to be private

### Additional comments

_No response_

vscaiceanu-1a:
## Related issues

- :bug: Fixes #1044

### DIFF
--- a/packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
+++ b/packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
@@ -62,6 +62,113 @@ export class RulesetExecutor {
   }
 
   /**
+   * Plug ruleset to fact streams and trigger a first evaluation
+   *
+   */
+  private prepareRuleset() {
+    const rulesWithContext: Rule[] = [];
+    const rulesWithoutContext: Rule[] = [];
+    const factsThatRerunEverything: string[] = [];
+    this.ruleset.rules.forEach((rule) => {
+      if (rule.outputRuntimeFacts.length > 0 || rule.inputRuntimeFacts.length > 0) {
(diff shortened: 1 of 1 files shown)


matthieu-crouzet:
Maybe you should target v10 instead of main for this breaking change

vscaiceanu-1a:
> Maybe you should target v10 instead of main for this breaking change

I think you missed a comment from here https://github.com/AmadeusITGroup/otter/pull/835#discussion_r1371864488
We are 100% sure the method is only used within this class. This means that this is **not** a breaking change and it can safely be removed straight away.

kpanot (approved):

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -61,6 +61,113 @@ export class RulesetExecutor {
     this.engineRuleset = this.prepareRuleset();
   }
 
+  /**
+   * Plug ruleset to fact streams and trigger a first evaluation
+   *
+   */
+  private prepareRuleset() {
```
kpanot, line 68: Would be cleaner to keep the public (as deprecated) field using the private one for pure backward compatible reason (event if calling this function from outside does not really make sense).

mrednic-1A (approved):

PR #1083, opened 2023-11-22
fix: additional properties not working for new application or component

fpaul-1A:
## Proposed change

Fix following issue:
> npm create @o3r test
> cd test
> npm exec ng g application test-app -- --standalone=false
> Error: Unknown argument: standalone

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/workspace/schematics/application/schema.json
+++ b/packages/@o3r/workspace/schematics/application/schema.json
@@ -31,6 +31,16 @@
       "description": "Do not install dependency packages.",
       "type": "boolean",
       "default": false
+    },
+    "strict": {
+      "description": "Creates an application with stricter bundle budgets settings.",
+      "type": "boolean",
+      "default": true
+    },
+    "standalone": {
+      "description": "Creates an application based upon the standalone API, without NgModules.",
+      "type": "boolean",
+      "default": false
     }
   },
   "required": ["name"],


matthieu-crouzet (dismissed):

vscaiceanu-1a (dismissed):

vscaiceanu-1a (approved):

fpaul-1A:
> Is it a change in Ng17 that makes the `additionalProperty: true` ignored by the CLI option validation process? Or did we have an issue before? What's about our other schematics that are supposed to propagate the options to the child Angular schematics?

We already have this issue in the current releases
We already define all properties for the case of angular components

kpanot:
> > Is it a change in Ng17 that makes the `additionalProperty: true` ignored by the CLI option validation process? Or did we have an issue before? What's about our other schematics that are supposed to propagate the options to the child Angular schematics?
> 
> 
> 
> We already have this issue in the current releases
> 
> We already define all properties for the case of angular components

I see 😟 
Can you then link the issue associated (and/or PR relative to the component and library fix)?

packages/@o3r/workspace/schematics/application/schema.json
```diff
@@ -31,6 +31,16 @@
       "description": "Do not install dependency packages.",
       "type": "boolean",
       "default": false
+    },
+    "strict": {
+      "description": "Creates an application with stricter bundle budgets settings.",
+      "type": "boolean",
```
kpanot, line 37: I suppose the `$ref` to Angular Schematics schema.json was not working?

kpanot (approved):

packages/@o3r/workspace/schematics/application/schema.json
fpaul-1A: indeed

PR #1085, opened 2023-11-23
feat(linter): new template rule to detect innerhtml

matthieu-crouzet:
## Proposed change
New template rule to detect innerhtml

## Related issues
- :rocket: Feature #1086

### DIFF
--- a/docs/linter/eslint-plugin/rules/no-inner-html.md
+++ b/docs/linter/eslint-plugin/rules/no-inner-html.md
@@ -1,0 +1,22 @@
+# no-inner-html
+
+This rule ensures that your template does not use innerHTML.
+For performance reason, it makes sense for certain application to encourage to use 'innerText' over 'innerHTML'.
+
+## Prerequisite
+
+The linter is dedicated to template files and requires the setup of the [@angular-eslint/template-parser](https://www.npmjs.com/package/@angular-eslint/template-parser).
+
+Example of a ESLint configuration:
+
+```json
+{
+  "parser": "@angular-eslint/template-parser",
+  "plugins": [
+    "@o3r"
+  ],
+  "rules": {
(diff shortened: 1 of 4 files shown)


packages/@o3r/eslint-plugin/src/index.ts
```diff
@@ -1,13 +1,15 @@
...
 import o3rWidgetTags from './rules/typescript/o3r-widget-tags/o3r-widget-tags';
+import noInnerHTML from './rules/template/no-inner-html/no-inner-html';
 import templateAsyncNumberLimitation from './rules/template/template-async-number-limitation/template-async-number-limitation';
 import jsonDependencyVersionsHarmonize from './rules/json/json-dependency-versions-harmonize/json-dependency-versions-harmonize';
 import matchingConfigurationName from './rules/typescript/matching-configuration-name/matching-configuration-name';
 
 module.exports = {
   rules: {
     'no-folder-import-for-module': noFolderImportForModule,
+    'no-inner-html': noInnerHTML,
```
kpanot, line 12: Missing .md file for the related doc

sdo-1A (approved):

docs/linter/eslint-plugin/rules/no-inner-html.md
```diff
@@ -0,0 +1,20 @@
+# no-inner-html
+
+This rule ensures that your template does not use innerHTML.
```
kpanot, line 3: Could be good to indicate the reason/motivation of it.
Having an innerHtml in the template is it a bad practice? Is it something we do want to encourage to all the application (or only for RefX or Application having CMS capabilities)?

If this is only for certain application (or in certain condition) maybe it should be off per default (and be activated on demand)

fpaul-1A (approved):

kpanot (approved):

PR #1089, opened 2023-11-24
feat(config): linter to detect invalid config property type

matthieu-crouzet:
## Proposed change
Linter to detect invalid config property type.
Every property in [StrictConfiguration](https://github.com/AmadeusITGroup/otter/blob/main/packages/%40o3r/core/src/core/interfaces/configuration.ts#L27) can be for example `string` or `number` but not the both at the same time. 
It cannot be check with TypeScript, so I propose a linter for that.

### DIFF
--- a/docs/linter/eslint-plugin/rules/no-multiple-type-configuration-property.md
+++ b/docs/linter/eslint-plugin/rules/no-multiple-type-configuration-property.md
@@ -1,0 +1,32 @@
+# @o3r/no-multiple-type-configuration-property
+
+Ensures that the configuration property does not accept multiple types.
+
+## How to use
+
+```json
+{
+  "@o3r/no-multiple-type-configuration-property": [
+    "error",
+    {
+      "supportedInterfaceNames": ["NestedConfiguration", "Configuration", "CustomConfigurationInterface"]
+    }
+  ]
+}
+```
+
+## Valid code example
+
+```typescript
+export interface MyFirstConfig extends Configuration {
+  myProp: string;
+}
+```
+
+## Invalid code example
+
+```typescript
+export interface MyConfig extends Configuration {
+  myProp: string | number;
+}
+```
--- a/packages/@o3r/eslint-plugin/src/rules/typescript/no-multiple-type-configuration-property/no-multiple-type-configuration-property.ts
+++ b/packages/@o3r/eslint-plugin/src/rules/typescript/no-multiple-type-configuration-property/no-multiple-type-configuration-property.ts
@@ -1,0 +1,79 @@
+import { TSESTree } from '@typescript-eslint/experimental-utils';
+import { createRule, defaultSupportedInterfaceNames, isExtendingConfiguration } from '../../utils';
+
+const separatorRegExp = /\s*[|&]\s*/;
+
+export interface NoMultipleTypeConfigurationPropertyOption {
+  supportedInterfaceNames?: string[];
+}
+
+export default createRule<NoMultipleTypeConfigurationPropertyOption[], 'error' | 'suggestion'>({
+  name: 'no-multiple-type-configuration-property',
+  meta: {
+    hasSuggestions: true,
+    type: 'problem',
+    docs: {
+      description: 'Ensures that the configuration property does not accept multiple types.',
+      recommended: 'error'
+    },
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          supportedInterfaceNames: {
+            type: 'array',
+            items: {
+              type: 'string'
+            },
+            default: defaultSupportedInterfaceNames
+          }
+        }
+      }
+    ],
+    messages: {
+      error: 'Configuration cannot be the union of multiple type',
+      suggestion: 'Replace {{currentValue}} by {{recommendedValue}}'
+    }
+  },
+  defaultOptions: [],
+  create: (context) => {
+    const supportedInterfaceNames = context.options.reduce((acc: string[], option) => acc.concat(option.supportedInterfaceNames || []), []);
+    const sourceCode = context.getSourceCode();
+
+    const rule = (node: TSESTree.TSUnionType | TSESTree.TSIntersectionType) => {
+      const interfaceDeclNode = node.parent?.parent?.parent?.parent;
+      if (!isExtendingConfiguration(interfaceDeclNode, supportedInterfaceNames)) {
+        return; // Not in a configuration interface
+      }
+
+      if (
+        node.types.every((type) => type.type === TSESTree.AST_NODE_TYPES.TSLiteralType && type.literal.type === TSESTree.AST_NODE_TYPES.Literal)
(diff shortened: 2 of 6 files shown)


packages/@o3r/eslint-plugin/src/rules/typescript/no-multiple-type-configuration-property/no-multiple-type-configuration-property.ts
```diff
@@ -0,0 +1,67 @@
...
+      error: 'Configuration could not be the union of multiple type',
+      suggestion: 'Replace {{currentValue}} by {{recommendedValue}}'
+    }
+  },
+  defaultOptions: [],
+  create: (context) => {
+    return {
+      // eslint-disable-next-line @typescript-eslint/naming-convention
+      TSUnionType: (node) => {
+        const supportedInterfaceNames = context.options.reduce((acc: string[], option) => acc.concat(option.supportedInterfaceNames || []), []);
```
kpanot, line 41: could be defined outside of this function

packages/@o3r/eslint-plugin/src/rules/typescript/no-multiple-type-configuration-property/no-multiple-type-configuration-property.ts
```diff
@@ -0,0 +1,67 @@
...
+    type: 'problem',
+    docs: {
+      description: 'Ensures that the configuration property does not accept multiple type.',
+      recommended: 'error'
+    },
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          supportedInterfaceNames: {
```
kpanot, line 22: missing description

packages/@o3r/eslint-plugin/src/rules/typescript/no-multiple-type-configuration-property/no-multiple-type-configuration-property.ts
```diff
@@ -0,0 +1,67 @@
...
+    messages: {
+      error: 'Configuration could not be the union of multiple type',
+      suggestion: 'Replace {{currentValue}} by {{recommendedValue}}'
+    }
+  },
+  defaultOptions: [],
+  create: (context) => {
+    return {
+      // eslint-disable-next-line @typescript-eslint/naming-convention
+      TSUnionType: (node) => {
```
kpanot, line 40: Should not check intersection type as well?

packages/@o3r/eslint-plugin/src/rules/typescript/no-multiple-type-configuration-property/no-multiple-type-configuration-property.ts
```diff
@@ -0,0 +1,79 @@
...
+
+      if (
+        node.types.every((type) => type.type === TSESTree.AST_NODE_TYPES.TSLiteralType && type.literal.type === TSESTree.AST_NODE_TYPES.Literal)
+        && [...(new Set((node.types as TSESTree.TSLiteralType[]).map((literalType) => typeof (literalType.literal as TSESTree.Literal).value)))].length === 1
+      ) {
+        return; // Only the same literal type
+      }
+
+      const text = context.getSourceCode().getText(node);
+      const separatorRegExp = /\s*(\||&)\s*/;
```
kpanot, line 55: Could be defined outside of the function (lets reduce the cost of the linter evaluation as much as possible for our poor IDEs)

packages/@o3r/eslint-plugin/src/rules/utils.ts
```diff
@@ -12,3 +12,19 @@ export const createRule = ESLintUtils.RuleCreator((name) => {
...
   return `https://github.com/AmadeusITGroup/otter/tree/release/${version}/docs/linter/eslint-plugin/rules/${name}.md`;
 });
+
+/**
+ * Returns true if the node extends one of the `supportedInterfaceNames`
+ * @param interfaceDeclNode
+ * @param supportedInterfaceNames
+ */
+export const isExtendingConfiguration = (interfaceDeclNode: TSESTree.Node | undefined, supportedInterfaceNames: string[] = []) => {
+  const supportedInterfaceNamesSet = new Set([...supportedInterfaceNames, 'Configuration', 'NestedConfiguration']);
```
kpanot, line 22: Could be aligned with the default value via a const.

packages/@o3r/eslint-plugin/src/rules/typescript/no-multiple-type-configuration-property/no-multiple-type-configuration-property.ts
```diff
@@ -0,0 +1,79 @@
...
+    type: 'problem',
+    docs: {
+      description: 'Ensures that the configuration property does not accept multiple type.',
+      recommended: 'error'
+    },
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          supportedInterfaceNames: {
```
kpanot, line 22: Actually the name is not really correct, as we always check the `NestedConfiguration` and `Configuration`.
Something like `additiionalSupportedInterfaceName` would be more correct :S

packages/@o3r/eslint-plugin/src/rules/typescript/no-multiple-type-configuration-property/no-multiple-type-configuration-property.ts
```diff
@@ -0,0 +1,81 @@
...
+      ) {
+        return; // Only the same literal type
+      }
+
+      const text = sourceCode.getText(node);
+      context.report({
+        messageId: 'error',
+        node,
+        loc: node.loc,
+        fix: (fixer) => fixer.replaceTextRange(node.range, text.split(separatorRegExp)[0]),
```
fpaul-1A, line 62: are we sure that taking the first type of the list is correct ?
I don't think there should be an automatic fix for that if we're not sure it's the right type to use

packages/@o3r/eslint-plugin/src/rules/typescript/no-multiple-type-configuration-property/no-multiple-type-configuration-property.ts
```diff
@@ -0,0 +1,81 @@
...
+            items: {
+              type: 'string'
+            },
+            default: defaultSupportedInterfaceNames
+          }
+        }
+      }
+    ],
+    messages: {
+      error: 'Configuration could not be the union of multiple type',
```
fpaul-1A, line 35:
```suggestion
      error: 'Configuration cannot be the union of multiple type',
```

fpaul-1A (dismissed):

docs/linter/eslint-plugin/rules/no-multiple-type-configuration-property.md
```diff
@@ -0,0 +1,32 @@
+# @o3r/no-multiple-type-configuration-property
+
+Ensures that the configuration property does not accept multiple type.
```
cpaulve-1A, line 3:
```suggestion
Ensures that the configuration property does not accept multiple types.
```

kpanot (approved):

fpaul-1A (approved):

PR #1098, opened 2023-11-29
fix(ama-sdk-schematics): add target folder to nx inputs of prepare-bu…

fpaul-1A:
…ild-builders

## Proposed change

Any gitignored file will not be considered for the inputs of Nx tasks
An easy fix is to add an exception to the `.nxignore` file

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/.nxignore
+++ b/.nxignore
@@ -3,3 +3,4 @@ packages/@ama-sdk/generator-sdk/src/**/templates/**/*.{j,t}s
 packages/@*/*/{builders,schematics}/**/templates/**/*.ts
 packages/@*/*/{builders,schematics}/**/mocks/**/*.ts
 packages/@*/*/build/**
+!**/schematics/**/target


kpanot (approved):

.nxignore
```diff
@@ -3,3 +3,4 @@ packages/@ama-sdk/generator-sdk/src/**/templates/**/*.{j,t}s
 packages/@*/*/{builders,schematics}/**/templates/**/*.ts
 packages/@*/*/{builders,schematics}/**/mocks/**/*.ts
 packages/@*/*/build/**
+!**/schematics/**/target
```
kpanot, line 6: Could not be done in a file at `packages/@ama-sdk/schematics/.nxignore` ?

fpaul-1A: I didn't try but according to the docs https://nx.dev/reference/nxignore, this file is only expected at root level

matthieu-crouzet (approved):

PR #1101, opened 2023-11-29
feat(config): support global configuration categories

matthieu-crouzet:
## Context
Today categories are defined on each interface, listing for each one a mapping between the key and the displayed value.
It means that even if those categories are used application wise, they needs to be copy/pasted on all the config interfaces that uses it.

## Proposed change
Define global categories in the extractor options.

### DIFF
--- a/apps/showcase/src/components/showcase/configuration/configuration-pres.config.ts
+++ b/apps/showcase/src/components/showcase/configuration/configuration-pres.config.ts
@@ -16,10 +16,14 @@ export interface DestinationConfiguration extends NestedConfiguration {
 
 /**
  * Component configuration example
+ * <o3rCategories>
+ *  <localCategory>Local category 8</localCategory>
+ * </o3rCategories>
  */
 export interface ConfigurationPresConfig extends Configuration {
   /**
    * Default date selected compare to today
+   * @o3rCategory localCategory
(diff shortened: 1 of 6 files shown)


apps/showcase/src/components/showcase/configuration/configuration-pres.config.ts
```diff
@@ -16,10 +16,14 @@ export interface DestinationConfiguration extends NestedConfiguration {
 
 /**
  * Component configuration example
+ * <o3rCategories>
```
kpanot, line 19: we may need to create a task to migrate this notation to the `@o3r...` based

packages/@o3r/components/builders/component-extractor/schema.ts
```diff
@@ -34,4 +35,10 @@ export interface ComponentExtractorBuilderSchema extends JsonObject {
 
   /** Include placeholder metadata file */
   placeholdersMetadataFilePath: string | null;
+
+  /**
+   * List of global categories with description
+   * Adding `& JsonObject` as workaround
```
kpanot, line 41: should not be in a `/**` comment but a simple `/*` comment

sdo-1A (approved):

PR #1103, opened 2023-11-30
fix: set enableScripts to false by default

fpaul-1A:
## Proposed change

By default, any third party dependency can execute arbitrary code during the install process (with `postinstall` script)

We only want to allow that if needed for packages that we can trust
It can be re-activated for a specific package using `dependenciesMeta`

(It does not impact the `postinstall` scripts of the workspace which are considered safe)

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,5 +1,7 @@
 compressionLevel: mixed
 
+enableScripts: false
+
 enableGlobalCache: false
 
 nodeLinker: pnp


kpanot:
Can you indicate the reason in the comment of the PR (or link the related task)?

kpanot (approved):

matthieu-crouzet (approved):

PR #1118, opened 2023-12-07
feat(eslint-plugin): yarnrc version harmonize

### LINKED ISSUE #933: [Feature]: Automatically align `.yarnrc.yml` dependencies ranges with workspaces dependencies ranges
### Context

Currently we do have discrepancies between the ranges of the dependencies declared for the sub dependencies (in `yarnrc.yml` file) and the dependencies defined in the different `package.json` files.
This discrepancies can lead to an misalignment of dependencies of the different packages.

### Proposal

- [x] Provide a tool to aligned this dependencies (linter rules can be a solution)

kpanot:
## Proposed change

Provides a linter rule to align yarn package extensions versions with monorepo's one

## Related issues

- :rocket: Feature resolves #933

### DIFF
--- a/packages/@o3r/eslint-plugin/src/rules/yaml/utils.ts
+++ b/packages/@o3r/eslint-plugin/src/rules/yaml/utils.ts
@@ -1,0 +1,49 @@
+import { ParserServices, TSESLint } from '@typescript-eslint/experimental-utils';
+
+/** Basic interface for the Parser Services object provided by yaml-eslint-parser */
+interface YamlParserServices extends ParserServices {
+  isYAML: boolean;
+}
+
+/**
+ * Determine if yaml-eslint-parser is used
+ * @param parserServices Parser services object
+ */
+export function isYamlParserServices(parserServices: any): parserServices is YamlParserServices {
+  return !!parserServices && parserServices.isYAML;
+}
+
+/**
+ * Retrieve the yaml parser services object or throw if the invalid parser is used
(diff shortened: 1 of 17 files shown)


packages/@o3r/eslint-plugin/src/rules/yaml/utils.ts
```diff
@@ -0,0 +1,50 @@
+import { ParserServices, TSESLint } from '@typescript-eslint/experimental-utils';
+// import type { AST } from 'yaml-eslint-parser';
```
fpaul-1A, line 2: I see dead code

kpanot: oops

fpaul-1A (approved):

fpaul-1A (approved):

packages/@o3r/workspace/package.json
```diff
@@ -16,8 +16,8 @@
     "build:builders": "tsc -b tsconfig.builders.json --pretty && yarn generate-cjs-manifest",
     "postbuild": "patch-package-json-main"
   },
-  "default": "./src/public_api.js",
-  "types": "./src/public_api.d.ts",
+  "main": "./dist/src/public_api.js",
+  "types": "./dist/src/public_api.d.ts",
```
mrednic-1A, line 20: Is './dist' removed by `prepare:publish` script ?

Issue #1120, opened 2023-12-07
[Feature]: "Matches" as a basic operator for strings in rules


AngieMoomin:
### Context

Currently it is not possible to have in the conditions something like the following: 

> **stringFact** matches **pattern**

Where "stringFact" is whatever string and "pattern" is a regular expression like "/^8[0-9]{3}/"

### Proposal

Add in the basic operators the following: 

```
export const matches: Operator<string, string> = {

  name: 'matches',
  evaluator: (value, pattern) => {
    const regExp = new RegExp(pattern);
    return regExp.test(value));
  },
  validateLhs: isString,
  validateRhs: isString
};
```

PR #1121, opened 2023-12-07
feat: matches basic operator

sdo-1A:
## Proposed change

Currently it is not possible to have in the conditions something like the following: 

> **stringFact** matches **pattern**

Where "stringFact" is whatever string and "pattern" is a regular expression like "/^8[0-9]{3}/"

## Related issues

- :rocket: Feature #1120

### DIFF
--- a/docs/rules-engine/operators.md
+++ b/docs/rules-engine/operators.md
@@ -59,6 +59,7 @@ Example of usage :
 | Number[]    | allRangeNumber            | all between            | Check if every values of the variable are included in a specified range |
 | Number[]    | oneRangeNumber            | one between            | Check if one of the values of the variable is included in a specified range |
 | String      | inString                  | within                 | Check if the text variable is part of the specified value |
+| String      | matchesPattern            | matches                | Check if the text variable matches a specific pattern |
 | String      | notInString               | not within             | Check if the text variable is not part in the specified value |
 | String      | stringContains            | contains               | Check if the specified text value is included in the text variable |
 | String      | notStringContains         | does not contain       | Check if the specified text value is not included in the text variable |
@@ -71,6 +72,14 @@ Example of usage :
 | All[]       | lengthGreaterThanOrEquals | number of ≥            | Check if the number of values of the variable is greater or equal to a specific value |
 | All[]       | lengthGreaterThan         | number of >            | Check if the number of values of the variable is greater than a specific value |
 
+> **Note**: For the operators comparing a text variable to a pattern (such as `matchesPattern`, `oneMatches`, and `allMatch`), 
+> we support the ES RegExp `/^myRegExp.*$/i` (containing the pattern and optional flags) or just the RegExp content `^myregexp.*$`.
+> 
+> The special characters used in the pattern should contain a double backslash (`\\`).
+> For example, to check if a string contains a `\t`, the pattern would need to include `\\t`.
+> 
+> > Also, to avoid the wrong detection of an ES RegExp instead of RegExp content, a content beginning with a slash `/` character
+> > (such as a path `/path/to/file`) should be preceded by a double backslash `\\` (for example `\\/path/to/file`)
 
 You can create your own operator in your application and add it to the engine.
 Note that the @title provides a string for the operator that can be displayed in an edition UI for better clarity (ex: < for lessThan).
@@ -180,4 +189,3 @@ ngOnInit() {
 ```
 
 Here, `CurrentTimeFactsService` also provides a `tick()` method that, when called, it recomputes the current time. It is up to the application to decide how ofter the current time should be recomputed (at a given time interval, on page navigation, etc).
-
--- a/packages/@o3r/rules-engine/src/engine/operator/operator.helpers.ts
+++ b/packages/@o3r/rules-engine/src/engine/operator/operator.helpers.ts
@@ -107,3 +107,19 @@ export function isSupportedSimpleTypes(value: unknown): value is SupportedSimple
 export function isString(value: unknown): value is string {
   return typeof value === 'string';
 }
+
+/**
+ * Parse input to return RegExp
+ *
+ * @param value value to test whether pattern exists (can be string or array of strings)
+ * @param inputString regexp pattern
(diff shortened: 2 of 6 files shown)


packages/@o3r/rules-engine/src/engine/operator/operators/basic.operators.ts
```diff
@@ -91,7 +91,22 @@ export const isUndefined: UnaryOperator<any> = {
...
 
+/**
+ * Check if the text variable matches the specified RegExp pattern
+ *
+ * @title matches the pattern
+ */
+export const matchesPattern: Operator<string, string> = {
+  name: 'matchesPattern',
+  evaluator: (value, pattern) => {
+    const regExp = new RegExp(pattern);
```
kpanot, line 102: Not as simple: `new RegExp('/a/i')` is not equivalent to `/a/i` but `/\/a\/i/`.
You will need to handle the 2 `/` of the regexp and the options given in the end (that are the second parameter of the `new RegExpp()`)

packages/@o3r/rules-engine/src/engine/operator/operators/array-based.operators.ts
```diff
@@ -119,7 +119,11 @@ export const allLower: Operator<number[], number | string> = {
 export const allMatch: Operator<string[], string> = {
   name: 'allMatch',
   evaluator: (array, inputString) => {
-    const regExp = new RegExp(inputString);
+    let regExp = new RegExp(inputString);
```
kpanot, line 122: can be instantiated without be used.
```typescript
    if (inputString.startWith('/')) {
      const finalSlash = inputString.lastIndexOf('/');
      const regexpContent = inputString.slice(1, finalSlash - 1);
      const regexpOption = inputString.slice(finalSlash + 1);
      return new RegExp(regexpContent, regexpOption);
    }
    return new RegExp(inputString);
```

docs/rules-engine/operators.md
```diff
@@ -71,6 +72,8 @@ Example of usage :
 | All[]       | lengthGreaterThanOrEquals | number of ≥            | Check if the number of values of the variable is greater or equal to a specific value |
 | All[]       | lengthGreaterThan         | number of >            | Check if the number of values of the variable is greater than a specific value |
 
+> **Note**: For the operators comparing a text variable to a pattern, the special characters used in the patterns should contain a double backslash (`\\`).
+> For example, to check if a string contains a `/`, the pattern would need to include `\\/`.
```
kpanot, line 76: You should also specify that you support `/^myRegExp.*$/i`(ES RegExp) and `^myregexp.*$` (regexp content).
(also, with my suggestion, the `\\/` is not mandatory, it is already not mandatory in the *regexp content* format)

packages/@o3r/rules-engine/src/engine/operator/operators/array-based.operators.ts
```diff
@@ -119,7 +119,11 @@ export const allLower: Operator<number[], number | string> = {
...
+    if (inputString[0] === '/') {
```
cpaulve-1A, line 123: What if you want to match an url, wouldn't it detect it as a regex? Ex: /this/is/a/url.com

Would that be a valid scenario?

sdo-1A: No since slashes are special characters in a regex so it would need an escape character before
Ex: \/this\/is\/a\/url.com

packages/@o3r/rules-engine/src/engine/operator/operators/array-based.operators.ts
```diff
@@ -119,8 +119,14 @@ export const allLower: Operator<number[], number | string> = {
...
-    const regExp = new RegExp(inputString);
-    return array.every((elementValue) => regExp.test(elementValue));
+    if (inputString.startsWith('/')) {
+      const finalSlash = inputString.lastIndexOf('/');
+      const regexpPattern = inputString.slice(1, finalSlash);
+      const regexpFlags = inputString.slice(finalSlash + 1);
+      const regexp = new RegExp(regexpPattern, regexpFlags);
+      return array.every((elementValue) => regexp.test(elementValue));
+    }
+    return array.every((elementValue) => RegExp(inputString).test(elementValue));
```
kpanot, line 129: Reduce the cost if you instantiate the regExp only once:
```suggestion
    const regExp = new RegExp(inputString);
    return array.every((elementValue) => regExp.test(elementValue));
```

kpanot (approved):

packages/@o3r/rules-engine/src/engine/operator/operators/basic.operators.ts
```diff
@@ -91,7 +91,27 @@ export const isUndefined: UnaryOperator<any> = {
...
+ */
+export const matchesPattern: Operator<string, string> = {
+  name: 'matchesPattern',
+  evaluator: (value, inputString) => {
+    if (inputString.startsWith('/')) {
+      const finalSlash = inputString.lastIndexOf('/');
+      const regexpPattern = inputString.slice(1, finalSlash);
+      const regexpFlags = inputString.slice(finalSlash + 1);
+      return RegExp(regexpPattern, regexpFlags).test(value);
+    }
```
fpaul-1A, line 107: this logic seems to be duplicated 3 times, might be good to extract the code in an helper
```typescript
const inputRegExp = parseRegExp(inputString);
```
or something

packages/@o3r/rules-engine/src/engine/operator/operator.helpers.ts
```diff
@@ -107,3 +107,19 @@ export function isSupportedSimpleTypes(value: unknown): value is SupportedSimple
 export function isString(value: unknown): value is string {
   return typeof value === 'string';
 }
+
+/**
+ * Validate that a value is a string
```
matthieu-crouzet, line 112: Copy pasta 
image

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #1122, opened 2023-12-08
test: add unit test for builders

fpaul-1A:
## Proposed change

Add unit tests for most builders
Use `memfs` and `unionfs` to avoid writing anything on disk

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/components/testing/setup-jest.ts
+++ b/packages/@o3r/components/testing/setup-jest.ts
@@ -1,2 +1,2 @@
+import '@o3r/test-helpers/setup-jest';
 import 'jest-preset-angular/setup-jest';
-
--- a/packages/@o3r/core/testing/setup-jest.ts
+++ b/packages/@o3r/core/testing/setup-jest.ts
@@ -1,16 +1,2 @@
+import '@o3r/test-helpers/setup-jest';
 import 'jest-preset-angular/setup-jest';
-
-jest.mock('node:fs', () => {
-  const originalModule = jest.requireActual('node:fs');
-  let mockedFiles: Record<string, string> = {};
-  return {
-    // eslint-disable-next-line @typescript-eslint/naming-convention
-    __esModule: true,
-    ...originalModule,
-    readFileSync: (filePath: string, options: any) => mockedFiles[filePath] ?? originalModule.readFileSync.call(originalModule, filePath, options),
-    // eslint-disable-next-line @typescript-eslint/naming-convention
-    __mockFiles: (mockFiles: Record<string, string>) => {
-      mockedFiles = mockFiles;
-    }
-  };
-});
--- a/packages/@o3r/localization/builders/i18n/index.spec.ts
+++ b/packages/@o3r/localization/builders/i18n/index.spec.ts
@@ -1,0 +1,51 @@
+import { Architect } from '@angular-devkit/architect';
+import { TestingArchitectHost } from '@angular-devkit/architect/testing';
+import { schema } from '@angular-devkit/core';
+import { cleanVirtualFileSystem, useVirtualFileSystem } from '@o3r/test-helpers';
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import { I18nBuilderSchema } from './schema';
+
+describe('Localization i18n Builder', () => {
+  const workspaceRoot = path.join('..', '..', '..', '..', '..');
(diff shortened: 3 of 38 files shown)


kpanot (approved):

packages/@o3r/core/testing/setup-jest.ts
```diff
@@ -1,16 +1,8 @@
...
-    __esModule: true,
-    ...originalModule,
-    readFileSync: (filePath: string, options: any) => mockedFiles[filePath] ?? originalModule.readFileSync.call(originalModule, filePath, options),
-    // eslint-disable-next-line @typescript-eslint/naming-convention
-    __mockFiles: (mockFiles: Record<string, string>) => {
-      mockedFiles = mockFiles;
-    }
-  };
-});
+global.setImmediate ||= ((fn: any, ...args: any) => {
```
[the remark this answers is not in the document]
fpaul-1A: I am not sure how to set it at workspace level but yes I can move the code in `test-helpers` and import it in the `setup-jest.ts`s

packages/@o3r/components/testing/setup-jest.ts
```diff
@@ -1,2 +1,8 @@
 import 'jest-preset-angular/setup-jest';
 
+global.setImmediate ||= ((fn: any, ...args: any) => {
```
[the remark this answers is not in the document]
fpaul-1A: I would not say it's needed, but there is no reason to override it if we have a native implementation
this function is supposed to available natively on Node but it is not when using Jest (presumably because of jsdom environment)
I don't want to override this method unless is does not exist

packages/@o3r/components/testing/setup-jest.ts
```diff
@@ -1,2 +1,2 @@
+import '@o3r/test-helpers/src/setup-jest';
```
kpanot, line 1: please avoid deep import, you can define a mapping in the `@o3r/test-helpers` package.json's `exports` properties

packages/@o3r/test-helpers/src/setup-jest/index.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * If setImmediate not available, fallback to setTimeout (
+ */
+global.setImmediate ||= ((fn: any, ...args: any) => {
```
kpanot, line 4: would not be cleaner to be `globalThis`?

packages/@o3r/localization/builders/i18n/index.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+      defaultLanguageFile: 'en-GB.json'
+    };
+    const run = await architect.scheduleBuilder('.:i18n', options);
+    const output = await run.result;
+    expect(output.error).toBeUndefined();
+    await run.stop();
+
+    const i18nOutput = JSON.parse(virtualFileSystem.readFileSync(path.join(i18nFolder, 'en-GB.json'), {encoding: 'utf8'}));
+    expect(typeof i18nOutput).toBe('object');
+    expect(Object.keys(i18nOutput)[0]).toMatch(/o3r-.*/);
```
cpaulve-1A, line 49: should we test cases with dictionnaries?

fpaul-1A: this PR is mostly for the setup, we can definitely add more tests after

kpanot (approved):

packages/@o3r/localization/builders/localization/index.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+      outputFile: path.resolve(__dirname, `${workspaceRoot}/apps/showcase/localisation.metadata.json`)
+    });
+    await virtualFileSystem.promises.mkdir(path.resolve(__dirname, `${workspaceRoot}/apps/showcase`), {recursive: true});
+    await virtualFileSystem.promises.writeFile(path.resolve(__dirname, `${workspaceRoot}/apps/showcase/localisation.metadata.json`), '[]');
+  });
+  afterEach(() => {
+    cleanVirtualFileSystem();
+  });
+
+  it('should genere the localizations', async () => {
```
mrednic-1A, line 37: generate

mrednic-1A (approved):

PR #1133, opened 2023-12-13
fix(component): clean up tsconfig file list

kpanot:
## Proposed change

Clean up component tsconfig file list

### DIFF
--- a/packages/@o3r/components/tsconfig.builders.json
+++ b/packages/@o3r/components/tsconfig.builders.json
@@ -16,8 +16,7 @@
   ],
   "exclude": [
     "**/*.spec.ts",
-    "**/templates/**/*.ts",
-    "**/mocks/**/*.ts",
-    "schematics/**/templates/**"
+    "**/templates/**",
+    "**/mocks/**/*.ts"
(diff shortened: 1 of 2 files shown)


packages/@o3r/components/tsconfig.builders.json
```diff
@@ -16,8 +16,7 @@
   ],
   "exclude": [
     "**/*.spec.ts",
-    "**/templates/**/*.ts",
-    "**/mocks/**/*.ts",
-    "schematics/**/templates/**"
+    "**/templates/**",
+    "**/mocks/**/*.ts"
```
matthieu-crouzet, line 20: Why do we keep the `*.ts` for `mocks` but not for `templates`

matthieu-crouzet (approved):

packages/@o3r/components/tsconfig.builders.json
kpanot: Not needed to keep it but I just fixed the `/template/` pattern without touching others

fpaul-1A (approved):

PR #1135, opened 2023-12-13
fix(rules-engine): deprecate executor's prepareRuleset function

### LINKED ISSUE #1044: [Bug]: prepareRuleset method should be private

kpanot:
## Proposed change

Deprecate executor's prepareRuleset function.
It can be removed in v11 as this function is not made to be used outside of the class (and would not have any effect if done)

## Related issues

- :rocket: Feature resolves #1044

### DIFF
--- a/packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
+++ b/packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
@@ -51,7 +51,6 @@ export class RulesetExecutor {
 
   /**
    * Create a new ruleset executor
-   *
    * @param ruleset Ruleset to evaluate
    * @param rulesEngine Instance of the rules engine
    */
@@ -59,13 +58,12 @@ export class RulesetExecutor {
     this.ruleset = ruleset;
     this.rulesEngine = rulesEngine;
     this.operators = rulesEngine.operators;
-    this.engineRuleset = this.prepareRuleset();
+    this.engineRuleset = this.plugRuleset();
   }
 
   /**
    * Recursively explores a rule to identify and collect input facts.
    * Input facts are identified based on the 'FACT' type and operator-specific implicit dependencies.
-   *
    * @param currentObject The current object being explored.
    * @param ruleInputFacts A set to store the identified input facts for the rule.
    */
@@ -90,7 +88,6 @@ export class RulesetExecutor {
 
   /**
    * Report performance mark for a rule run
-   *
    * @param rule Rule to measure
    * @param status status of the rule evaluation
    */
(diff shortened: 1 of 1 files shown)


packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -330,4 +320,12 @@ export class RulesetExecutor {
       rulesResultsSubject$: result$
     } as EngineRuleset;
   }
+
+  /**
+   * Plug ruleset to fact streams and trigger a first evaluation
+   * @deprecated This function is not made to be accessible from Outside of the class, will be removed in v11
```
matthieu-crouzet, line 326: I think we can remove it in v10

matthieu-crouzet (dismissed):

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
mrednic-1A: v10 or v12 :) not v11 :D

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -186,7 +178,6 @@ export class RulesetExecutor {
 
   /**
    * Returns true if the action sets a temporary fact
-   *
```
cpaulve-1A, line 189: Isn't it an eslint rule?

kpanot: It is

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
kpanot: v10 really is too short and v12 is a bit too far (for a function that does not make sense to be used).
If the reason to not be v11 is that "we do not make change in v11", that does not make sens to flag it this realese to remove it in v10 neither...
I don't want to argue and change it to `v10` but we may think to be less rigid on that "impair" release.

fpaul-1A (approved):

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
mrednic-1A: ok for me, I see your point

mrednic-1A (approved):

PR #1137, opened 2023-12-13
feat(contextualization): expose api on window

matthieu-crouzet:
## Proposed change
Expose an api on window for the contextualization functionalities

### DIFF
--- a/apps/showcase/src/app/app.component.ts
+++ b/apps/showcase/src/app/app.component.ts
@@ -43,14 +43,17 @@ export class AppComponent implements OnDestroy {
 
   private subscriptions = new Subscription();
 
-  constructor(router: Router, private offcanvasService: NgbOffcanvas) {
+  constructor(
+    router: Router,
+    private offcanvasService: NgbOffcanvas
+  ) {
     const onNavigationEnd$ = router.events.pipe(
       filter((event): event is NavigationEnd => event instanceof NavigationEnd),
       share()
     );
     this.activeUrl$ = onNavigationEnd$.pipe(
       map((event) => event.urlAfterRedirects),
-      shareReplay({bufferSize: 1, refCount: true})
+      shareReplay({ bufferSize: 1, refCount: true })
     );
     this.subscriptions.add(onNavigationEnd$.subscribe((event) => {
       if (this.offcanvasRef) {
--- a/packages/@o3r/configuration/src/devkit/configuration-devtools.console.service.ts
+++ b/packages/@o3r/configuration/src/devkit/configuration-devtools.console.service.ts
@@ -1,15 +1,15 @@
 /* eslint-disable no-console */
 import { Inject, Injectable, Optional } from '@angular/core';
-import type { Configuration, CustomConfig, DevtoolsServiceInterface, WindowWithDevtools } from '@o3r/core';
+import type { Configuration, ContextualizationDataset, CustomConfig, DevtoolsServiceInterface, WindowWithDevtools } from '@o3r/core';
 import { firstValueFrom } from 'rxjs';
-import { ConfigurationDevtoolsServiceOptions } from './configuration-devtools.interface';
+import { ConfigurationContextualizationDevtools, ConfigurationDevtoolsServiceOptions } from './configuration-devtools.interface';
 import { OtterConfigurationDevtools } from './configuration-devtools.service';
 import { OTTER_CONFIGURATION_DEVTOOLS_DEFAULT_OPTIONS, OTTER_CONFIGURATION_DEVTOOLS_OPTIONS } from './configuration-devtools.token';
 
 @Injectable({
   providedIn: 'root'
 })
-export class ConfigurationDevtoolsConsoleService implements DevtoolsServiceInterface {
+export class ConfigurationDevtoolsConsoleService implements DevtoolsServiceInterface, ConfigurationContextualizationDevtools {
 
   /** Name of the Window property to access to the devtools */
   public static readonly windowModuleName = 'configuration';
@@ -20,7 +20,13 @@ export class ConfigurationDevtoolsConsoleService implements DevtoolsServiceInter
   ) {
     this.options = { ...OTTER_CONFIGURATION_DEVTOOLS_DEFAULT_OPTIONS, ...options };
 
-    if (this.options.isActivatedOnBootstrap) {
+    if (
+      this.options.isActivatedOnBootstrap
+      || (
+        this.options.isActivatedOnBootstrapWhenCMSContext
+        && (document.body.dataset as ContextualizationDataset).cmscontext === 'true'
+      )
+    ) {
       this.activate();
     }
   }
@@ -112,7 +118,7 @@ export class ConfigurationDevtoolsConsoleService implements DevtoolsServiceInter
     const content = await this.configurationDevtools.getConfiguration();
 
     console.log('BOOKMARK');
-    console.log(`javascript:window._OTTER_DEVTOOLS_.loadConfiguration('${JSON.stringify(content).replace(/[']/g, '\\\'')}')`);
+    console.log(`javascript:window._OTTER_DEVTOOLS_.updateConfigurations('${JSON.stringify(content).replace(/[']/g, '\\\'')}')`);
   }
 
   /**
@@ -125,10 +131,18 @@ export class ConfigurationDevtoolsConsoleService implements DevtoolsServiceInter
 
   /**
    * Load a json configuration
-   *
    * @param configurations configurations to load
+   * @deprecated please use `updateConfigurations` instead, will be removed in Otter v12.
    */
   public loadConfiguration(configurations: string | CustomConfig<Configuration>[]): void {
+    this.configurationDevtools.loadConfiguration(configurations);
+  }
+
+  /**
+   * Replace N configurations in one shot
+   * @param configurations array of configurations to update
+   */
+  public updateConfigurations(configurations: string | CustomConfig<Configuration>[]): void {
     this.configurationDevtools.loadConfiguration(configurations);
   }
 }
--- a/packages/@o3r/core/src/core/devkit/devtools.interface.ts
+++ b/packages/@o3r/core/src/core/devkit/devtools.interface.ts
@@ -8,11 +8,29 @@ export interface WindowWithDevtools extends Window {
 /** Common option used by the different DevKit services */
 export interface DevtoolsCommonOptions {
   /**
-   * Activated the application bootstrap
+   * Activated on the application bootstrap
    *
    * @default false
    */
   isActivatedOnBootstrap: boolean;
+}
+
+/** Common option used by the different Contextualization DevKit services */
+export interface ContextualizationDevtoolsCommonOptions {
+  /**
+   * Activated on the application bootstrap when integrated in CMS context
+   *
+   * @default true
+   */
+  isActivatedOnBootstrapWhenCMSContext: boolean;
+}
+
+/**
+ * Dataset injected on the page when in CMS context
+ */
+export interface ContextualizationDataset {
+  /** `"true"` when in CMS context */
+  cmscontext?: string;
 }
 
 /** Interface describing an Otter Devtools service */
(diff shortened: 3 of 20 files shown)


apps/showcase/src/app/app.component.ts
```diff
@@ -43,7 +45,14 @@ export class AppComponent implements OnDestroy {
...
   private subscriptions = new Subscription();
 
-  constructor(router: Router, private offcanvasService: NgbOffcanvas) {
+  constructor(
+    router: Router,
+    localizationDevtoolsContextualizationService: LocalizationDevtoolsContextualizationService,
+    configurationDevtoolsContextualizationService: ConfigurationDevtoolsContextualizationService,
+    private offcanvasService: NgbOffcanvas
+  ) {
+    localizationDevtoolsContextualizationService.activate();
```
mrednic-1A, line 54: you could add some similar doc as for this one
https://github.com/AmadeusITGroup/otter/blob/main/docs/localization/LOCALIZATION.md#enable-chrome-extension-debugging

matthieu-crouzet: First I would like we discuss about how we expose this api 
I did it this way to test it 
but I'm not sure it will be the final way to do it

packages/@o3r/configuration/src/devkit/configuration-devtools.contextualization.module.ts
```diff
@@ -0,0 +1,34 @@
...
+  ]
+})
+export class ConfigurationContextualizationDevtoolsModule {
+
+  /**
+   * Initialize Otter Devtools
+   *
+   * @param options
+   */
+  public static instrument(options: Partial<ConfigurationContextualizationServiceOptions>): ModuleWithProviders<ConfigurationContextualizationDevtoolsModule> {
```
cpaulve-1A, line 25: why not put the option optionals?

kpanot, line 25: could turn `options` optional to avoid the `.instrument({})` in the app

packages/@o3r/localization/src/devkit/localization-devtools.contextualization.service.ts
```diff
@@ -0,0 +1,99 @@
+import { ApplicationRef, Inject, Injectable, Optional } from '@angular/core';
+import type { DevtoolsServiceInterface, WindowWithContextualization } from '@o3r/core';
+import { LocalizationService } from '../tools';
+import { LocalizationDevtoolsServiceOptions } from './localization-devkit.interface';
+import { OTTER_LOCALIZATION_CONTEXTUALIZATION_DEFAULT_OPTIONS, OTTER_LOCALIZATION_CONTEXTUALIZATION_OPTIONS } from './localization-devtools.token';
+import { lastValueFrom, Subscription, take } from 'rxjs';
+
+@Injectable()
+export class LocalizationDevtoolsContextualizationService implements DevtoolsServiceInterface {
```
kpanot, line 9: Why creating a new class instead of adding the function in the `LocalizationDevtoolsConsoleService` (just need to create an interface to ensure having the function required by the editor)?

packages/@o3r/configuration/src/devkit/configuration-devtools.contextualization.module.ts
matthieu-crouzet: If you don't need to pass options, then you don't need to call instrument, importing the module will be enough

packages/@o3r/localization/src/devkit/localization-devtools.contextualization.service.ts
matthieu-crouzet: I don't want both class to be instanciate and activated with the same condition

packages/@o3r/configuration/src/devkit/configuration-devtools.contextualization.module.ts
kpanot: but it is what you do in the showcase app

packages/@o3r/localization/src/devkit/localization-devtools.contextualization.service.ts
kpanot: But they give access to the same level of custo (via the same entry) and it will just force the app dev to import 2 devkit to activate instead of 1

apps/showcase/src/app/app.component.ts
```diff
@@ -43,14 +45,19 @@ export class AppComponent implements OnDestroy {
 
   private subscriptions = new Subscription();
 
-  constructor(router: Router, private offcanvasService: NgbOffcanvas) {
+  constructor(
+    router: Router,
+    public readonly configConsoleDevtools: ConfigurationDevtoolsConsoleService,
```
cpaulve-1A, line 50: do you still need to import them?

apps/showcase/src/app/app.component.ts
```diff
@@ -43,14 +45,19 @@ export class AppComponent implements OnDestroy {
...
+    public readonly localizationConsoleDevtools: LocalizationDevtoolsConsoleService,
```
fpaul-1A, line 51: are we injecting this to trigger the constructor of the service as a side effect ?
I don't really like that, I'd rather have an explicit function like `.activate()` or something

packages/@o3r/configuration/src/devkit/configuration-devtools.console.service.ts
```diff
@@ -20,7 +20,13 @@ export class ConfigurationDevtoolsConsoleService implements DevtoolsServiceInter
   ) {
     this.options = { ...OTTER_CONFIGURATION_DEVTOOLS_DEFAULT_OPTIONS, ...options };
 
-    if (this.options.isActivatedOnBootstrap) {
+    if (
+      this.options.isActivatedOnBootstrap
+      || (
+        this.options.isActivatedOnBootstrapWhenCMSContext
+        && (document.body.dataset as ContextualizationDataset).cmscontext === 'true'
```
fpaul-1A, line 27: are we sure `document.body.dataset` cannot be undefined here ?

packages/@o3r/core/src/core/devkit/devtools.interface.ts
```diff
@@ -15,6 +15,23 @@ export interface DevtoolsCommonOptions {
   isActivatedOnBootstrap: boolean;
 }
 
+/** Common option used by the different Contextualization DevKit services */
+export interface ContextualizationDevtoolsCommonOptions {
+  /**
+   * Activated the application bootstrap if integrated in CMS context
```
fpaul-1A, line 21: I understand the name of the attribute more than the comment :D

packages/@o3r/localization/src/devkit/localization-devtools.console.service.ts
```diff
@@ -1,21 +1,32 @@
...
+    private readonly localizationService: LocalizationService,
+    private readonly appRef: ApplicationRef,
     @Optional() @Inject(OTTER_LOCALIZATION_DEVTOOLS_OPTIONS) private options: LocalizationDevtoolsServiceOptions = OTTER_LOCALIZATION_DEVTOOLS_DEFAULT_OPTIONS
   ) {
-    if (this.options.isActivatedOnBootstrap) {
+    if (
+      this.options.isActivatedOnBootstrap
+      || (
+        this.options.isActivatedOnBootstrapWhenCMSContext
+        && (document.body.dataset as ContextualizationDataset).cmscontext === 'true'
```
fpaul-1A, line 27: same here, is this possibly undefined ?

packages/@o3r/configuration/src/devkit/configuration-devtools.console.service.ts
matthieu-crouzet: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset
dataset is always present on every `HTMLElement`

packages/@o3r/localization/src/devkit/localization-devtools.console.service.ts
matthieu-crouzet: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset
dataset is always present on every HTMLElement

packages/@o3r/localization/src/devkit/localization-devkit.interface.ts
```diff
@@ -20,3 +28,46 @@ export type AvailableLocalizationMessageContents =
...
+  /**
+   * Show localization keys
+   * @param value value enforced by the DevTools extension
+   */
+  showLocalizationKeys: (value?: boolean) => void;
+
+  /**
+   * Returns the current language
+   */
+  getCurrentLanguage: () => void;
```
fpaul-1A, line 46: shouldn't this return something ?

packages/@o3r/localization/src/devkit/localization-devkit.interface.ts
```diff
@@ -20,3 +28,46 @@ export type AvailableLocalizationMessageContents =
...
+
+/**
+ * Contextualization devtools exposed for localization in CMS integration
+ */
+export interface LocalizationContextualizationDevtools {
+  /**
+   * Show localization keys
+   * @param value value enforced by the DevTools extension
+   */
+  showLocalizationKeys: (value?: boolean) => void;
```
fpaul-1A, line 41: Maybe we should make all these functions asynchronous, just in case we'll need that at some point to avoid breaking changes

packages/@o3r/configuration/src/devkit/configuration-devtools.console.service.ts
kpanot, line 27: Why is a string `'true'`? The boolean are not supported? :S

packages/@o3r/core/src/core/devkit/devtools.interface.ts
```diff
@@ -8,13 +8,30 @@ export interface WindowWithDevtools extends Window {
...
+   * @default true
+   */
+  isActivatedOnBootstrapWhenCMSContext: boolean;
+}
+
+/**
+ * Dataset injected on the page when in CMS context
+ */
+export interface ContextualizationDataset {
+  cmscontext?: string;
```
kpanot, line 32: missing comment

packages/@o3r/configuration/src/devkit/configuration-devtools.console.service.ts
matthieu-crouzet: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset
it seems that `dataset` is a map of string

apps/showcase/src/app/app.component.ts
matthieu-crouzet: Done in the `main.ts`
Should we do it automatically when run `ng-add` of CMS packages ?

packages/@o3r/schematics/src/rule-factories/dev-tools/devtools-registration.ts
```diff
@@ -52,6 +53,64 @@ const registerModule = (options: DevtoolRegisterOptions): Rule => (tree, context
...
+  if (!mainFilePath || !tree.exists(mainFilePath)) {
+    return tree;
+  }
+
+  const content = tree.readText(mainFilePath);
+  if (content.includes(`inject(${options.serviceName})`)) {
+    return tree;
+  }
+
+  const match = /platformBrowserDynamic\(\).bootstrapModule\(\w+\)/.exec(content);
```
fpaul-1A, line 76: we should probably support standalone apps as well, since it's now the default with Angular 17+

cpaulve-1A (approved):

kpanot (approved):

packages/@o3r/localization/src/devkit/localization-devkit.interface.ts
```diff
@@ -20,3 +28,46 @@ export type AvailableLocalizationMessageContents =
...
+
+/**
+ * Contextualization devtools exposed for localization in CMS integration
+ */
+export interface LocalizationContextualizationDevtools {
+  /**
+   * Show localization keys
+   * @param value value enforced by the DevTools extension
+   */
+  showLocalizationKeys: (value?: boolean) => Promise<void>;
```
kpanot, line 41: Could be `void | Promise<void>` (same for others) to avoid the `return Promise.resolve(....);` on function with return value

kpanot (approved):

vscaiceanu-1a (approved):

PR #1138, opened 2023-12-13
fix(showcase): lighthouse accessibility report

fpaul-1A:
## Proposed change

Fixes deployed here https://fpaul-1a.github.io/otter

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/apps/showcase/src/app/app.component.html
+++ b/apps/showcase/src/app/app.component.html
@@ -1,5 +1,7 @@
 <header>
-  <button class="btn btn-lg d-xl-none mobile-menu-btn" (click)="open(content)"><i class="bi bi-list"></i></button>
+  <button class="btn btn-lg d-xl-none mobile-menu-btn" (click)="open(content)" title="Open side navigation">
+    <i class="bi bi-list"></i>
+  </button>
   <div class="header-title p-xl-3">
     <div class="header-icon">
       <img src="assets/otter-sticker.svg" width="48" height="48" alt="Otter logo" />
@@ -27,7 +29,7 @@
       <div class="col-12 col-md-8">
       </div>
       <div class="col-12 col-md-4">
-        <h5>Links</h5>
+        <h2 class="h5">Links</h2>
         <ul class="list-unstyled">
           <li>
             <a href="https://github.com/AmadeusITGroup/otter">Github</a>
--- a/apps/showcase/src/app/configuration/configuration.template.html
+++ b/apps/showcase/src/app/configuration/configuration.template.html
@@ -30,8 +30,10 @@
       <div class="row">
         <o3r-copy-text-pres language="html" [text]="codeConfig()" class="col-12 col-md-8 col-lg-9"></o3r-copy-text-pres>
         <div class="d-flex gap-2 align-self-start mt-md-2 pb-3 col-12 col-md-4 col-lg-3">
-          <button type="button" class="btn btn-primary" [class.disabled]="config()" (click)="toggleConfig()">Override</button>
-          <button type="button" class="btn btn-danger" [class.disabled]="!config()" (click)="toggleConfig()">Clear</button>
+          <button type="button" class="btn btn-primary" disabled="{{config()}}"
+                  [class.disabled]="config()" (click)="toggleConfig()">Override</button>
+          <button type="button" class="btn btn-danger" disabled="{{!config()}}"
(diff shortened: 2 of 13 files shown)


cpaulve-1A (approved):

apps/showcase/src/app/app.component.html
```diff
@@ -27,7 +29,7 @@
       <div class="col-12 col-md-8">
       </div>
       <div class="col-12 col-md-4">
-        <h5>Links</h5>
+        <h2 class="h5">Links</h2>
```
mrednic-1A, line 32: Do we know why this is better for the accessibility score ? Do we have previously an h1 and it expects an h2 ?

mrednic-1A (approved):

apps/showcase/src/app/app.component.html
fpaul-1A: the headings needs to be in the right order for better navigation and integration with screen-readers
here is the link https://dequeuniversity.com/rules/axe/4.8/heading-order proposed by lighthouse to explain it

apps/showcase/src/app/configuration/configuration.template.html
```diff
@@ -30,8 +30,10 @@ <h2 id="configuration-examples">Examples</h2>
       <div class="row">
         <o3r-copy-text-pres language="html" [text]="codeConfig()" class="col-12 col-md-8 col-lg-9"></o3r-copy-text-pres>
         <div class="d-flex gap-2 align-self-start mt-md-2 pb-3 col-12 col-md-4 col-lg-3">
-          <button type="button" class="btn btn-primary" [class.disabled]="config()" (click)="toggleConfig()">Override</button>
-          <button type="button" class="btn btn-danger" [class.disabled]="!config()" (click)="toggleConfig()">Clear</button>
+          <button type="button" class="btn btn-primary" disabled="{{config()}}"
```
matthieu-crouzet, line 33: `[attr.disabled]="config()"`
not mandatory

apps/showcase/src/app/rules-engine/rules-engine.template.html
```diff
@@ -30,55 +30,54 @@ <h2 id="rules-engine-example">Example</h2>
         Do not hesitate to run the application locally, if not installed yet, follow the <a routerLink="/run-app-locally">instructions</a>.
       </p>
       <a href="https://github.com/AmadeusITGroup/otter/blob/main/apps/showcase/src/components/showcase/rules-engine" target="_blank" rel="noopener">Source code</a>
-      <ul ngbNav class="nav nav-tabs mt-3">
```
matthieu-crouzet, line 33: Why do we delete all this `<ul>`

apps/showcase/src/components/showcase/basic/basic-pres.template.html
```diff
@@ -11,7 +11,7 @@ <h2 class="bg-body-tertiary text-secondary border border-light-subtle border-3 r
             <div aria-hidden="true" class="bg-body-tertiary position-absolute start-0 top-50 border border-light-subtle border-3 border-end-0 border-bottom-0"></div>
           </h2>
         </div>
-        <h5 class="row card-text mt-auto ms-1">Where do you want to go?</h5>
+        <h3 class="row card-text mt-auto ms-1 h5">Where do you want to go?</h3>
```
matthieu-crouzet, line 14: Do we really want to keep headings in the showcase components ?
Because they are not heading related to the app but only for a showcase purpose

apps/showcase/src/app/rules-engine/rules-engine.template.html
fpaul-1A: this is already handled by the directive ngbNav using role tablist and role tab

apps/showcase/src/components/showcase/basic/basic-pres.template.html
fpaul-1A: either way the headings needs to accurately define the structure of the page, which I think is the case here
I can change it if you disagree

matthieu-crouzet: Yes, I agree it is the good value according to the structure of the page.
Maybe we should use `<p>` or `<div>` instead `<h[x]`> in these showcase components. 
As you want :)

matthieu-crouzet (approved):

Issue #1127, opened 2023-12-12
[Bug]: Ng 17 deprecates `ComponentFactoryResolver` used by c11n


kpanot:
### Package name

core

### Package version

10.0.0

### Reproduction steps

Interface `ComponentFactoryResolver` used in `packages/@o3r/components/src/tools/component-replacement/c11n.directive.ts`.

### Current result

`ComponentFactoryResolver` is deprecated.

### Expected result

Avoid the usage of `ComponentFactoryResolver`.

PR #1143, opened 2023-12-14
Fix/c11n up to date

mrednic-1A:
## Proposed change

Ng 17 deprecates ComponentFactoryResolver used by c11n 

## Related issues

- :bug: Fixes [1127](https://github.com/AmadeusITGroup/otter/issues/1127)

### DIFF
--- a/apps/showcase/src/app/component-replacement/component-replacement.template.html
+++ b/apps/showcase/src/app/component-replacement/component-replacement.template.html
@@ -1,0 +1,43 @@
+<h1>Component Replacement</h1>
+<div class="row">
+  <div class="right-nav order-1 order-lg-2 col-12 col-lg-2 sticky-lg-top pt-5 pt-lg-0">
+    <o3r-in-page-nav-pres
+      id="component-replacement-nav"
+      [links]="links$ | async"
+    >
+    </o3r-in-page-nav-pres>
+  </div>
+  <div class="order-2 order-lg-1 col-12 col-lg-10">
+    <h2 id="component-replacement-description">Description</h2>
+    <div>
+      <p>This module provides a way to replace the presentation layer of a component inside an application.</p>
+    </div>
+
+    <h2 id="component-replacement-example">Example</h2>
+    <div>
+      <p>
+        Here is an example of a hebrew datepicker which replaces the default datepicker component used in other modules.
+      </p>
+      <o3r-component-replacement-pres></o3r-component-replacement-pres>
+      <p class="mt-4">
+        Do not hesitate to run the application locally. If not installed yet, follow the <a routerLink="/run-app-locally">instructions</a>.
+      </p>
+      <a href="https://github.com/AmadeusITGroup/otter/blob/main/apps/showcase/src/components/showcase/component-replacement" target="_blank" rel="noopener">Source code</a>
+    </div>
+
+    <h2 id="comp-replacement-install">How to install</h2>
+    <p> To use the <i>component replacement mechanism</i> you'll need to install the <i>&#64;o3r/components</i> package.</p>
+    <div>
+      <o3r-copy-text-pres [wrap]="true" language="bash" text="ng add @o3r/components"></o3r-copy-text-pres>
+    </div>
+
+    <h2 id="comp-replacement-references">References</h2>
+    <div>
+      <ul>
+        <li>
+          <a href="https://github.com/AmadeusITGroup/otter/blob/main/docs/components/COMPONENT_REPLACEMENT.md" target="_blank" rel="noopener">Documentation</a>
+        </li>
+      </ul>
+    </div>
+  </div>
+</div>
--- a/apps/showcase/src/components/showcase/component-replacement/component-replacement-pres.component.ts
+++ b/apps/showcase/src/components/showcase/component-replacement/component-replacement-pres.component.ts
@@ -1,0 +1,73 @@
+import { ChangeDetectionStrategy, Component, Input, OnChanges, Optional, SimpleChanges, ViewEncapsulation } from '@angular/core';
+import { Context, O3rComponent } from '@o3r/core';
+import { ConfigObserver } from '@o3r/configuration';
+import { ConfigurationBaseService } from '@o3r/configuration';
+import { ConfigurationObserver } from '@o3r/configuration';
+import { DynamicConfigurable } from '@o3r/configuration';
+import { COMPONENT_REPLACEMENT_PRES_DEFAULT_CONFIG } from './component-replacement-pres.config';
+import { COMPONENT_REPLACEMENT_PRES_CONFIG_ID } from './component-replacement-pres.config';
+import { ComponentReplacementPresConfig } from './component-replacement-pres.config';
+import { Observable } from 'rxjs';
+import { DatePickerInputPresComponent } from '../../utilities/date-picker-input/date-picker-input-pres.component';
+import { Configuration } from '@o3r/core';
+import { DatePickerInputPresContextInput, DatePickerInputPresContextOutput } from '../../utilities/date-picker-input/date-picker-input-pres.context';
+import { C11nModule, C11nService } from '@o3r/components';
+import { CommonModule, formatDate } from '@angular/common';
+import { FormControl, ReactiveFormsModule } from '@angular/forms';
+
+const ONE_DAY_IN_MS = 24 * 60 * 60 * 1000;
+
+@O3rComponent({ componentType: 'ExposedComponent' })
+@Component({
+  selector: 'o3r-component-replacement-pres',
+  standalone: true,
+  imports: [C11nModule, CommonModule, ReactiveFormsModule],
+  templateUrl: './component-replacement-pres.template.html',
+  styleUrl: './component-replacement-pres.style.scss',
+  encapsulation: ViewEncapsulation.None,
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class ComponentReplacementPresComponent implements OnChanges, DynamicConfigurable<ComponentReplacementPresConfig> {
+
+  /** Configuration stream based on the input and the stored configuration*/
+  public config$: Observable<ComponentReplacementPresConfig>;
+
+  @ConfigObserver()
+  private readonly dynamicConfig$: ConfigurationObserver<ComponentReplacementPresConfig>;
+
+  /** Input configuration to override the default configuration of the component*/
+  @Input()
+  public config: Partial<ComponentReplacementPresConfig> | undefined;
+
+  /** Observable of the presenter that we want to use, processed by the c11n directive */
+  public presenter$!: Observable<Context<DatePickerInputPresContextInput, DatePickerInputPresContextOutput> & DynamicConfigurable<Configuration>>;
+
+  public dateFormControl = new FormControl<string | null>(this.formatDate(Date.now() + 7 * ONE_DAY_IN_MS));
+
+  constructor(
+    private readonly c11nService: C11nService,
+    @Optional() configurationService: ConfigurationBaseService) {
+    this.dynamicConfig$ = new ConfigurationObserver<ComponentReplacementPresConfig>(COMPONENT_REPLACEMENT_PRES_CONFIG_ID, COMPONENT_REPLACEMENT_PRES_DEFAULT_CONFIG, configurationService);
+    this.config$ = this.dynamicConfig$.asObservable();
+    this.loadPresenter();
+    this.dynamicConfig$.next({ ...this.config, datePickerCustomKey: 'exampleDatePickerFlavorHebrew' });
+  }
+
+  private formatDate(dateTime: number) {
+    return formatDate(dateTime, 'yyyy-MM-dd', 'en-GB');
+  }
+
+  private loadPresenter() {
+    this.presenter$ = this.config$.pipe(
+      // Compute which presenter to use according to the configuration and the default presenter that we define here
+      this.c11nService.getPresenter(DatePickerInputPresComponent, 'datePickerCustomKey')
+    );
+  }
+
+  public ngOnChanges(changes: SimpleChanges) {
+    if (changes.config) {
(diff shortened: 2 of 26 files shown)


packages/@o3r/components/src/tools/component-replacement/c11n.directive.ts
```diff
@@ -103,14 +101,15 @@ export class C11nDirective <
...
-
-      componentFactory.inputs.forEach((prop) => {
-        this.uninitializedInputs.add(prop.propName);
-      });
 
       this.viewContainerRef.clear();
-      this.componentRef = this.viewContainerRef.createComponent<T>(componentFactory);
+      this.componentRef = this.viewContainerRef.createComponent<T>(changes.component.currentValue);
+      Object.keys(this.componentRef.instance)
+        .filter(prop => !(this.outputs && Object.keys(this.outputs).find(o => o === prop)))
```
kpanot, line 108:
```suggestion
        .filter(prop => !(this.outputs && Object.keys(this.outputs).some(o => o === prop)))
```

mrednic-1A: done, thanks

apps/showcase/src/components/showcase/localization/localization-pres.config.ts
```diff
@@ -0,0 +1,12 @@
+import type {Configuration} from '@o3r/core';
+import {computeItemIdentifier} from '@o3r/core';
+
+export interface LocalizationPresConfig extends Configuration {
+  datePickerCustomKey: string;
```
matthieu-crouzet, line 5: Could we have a dedicated component for component replacement instead of using the localization one ?

apps/showcase/src/components/utilities/date-picker-input-hebrew/date-picker-input-pres.config.ts
```diff
@@ -0,0 +1,8 @@
+import type {Configuration} from '@o3r/core';
+import {computeItemIdentifier} from '@o3r/core';
+
+export interface DatePickerInputPresConfig extends Configuration {}
```
matthieu-crouzet, line 4: We can keep the component simplier if config is not needed

apps/showcase/src/components/utilities/date-picker-input-hebrew/date-picker-input-pres.template.html
```diff
@@ -0,0 +1,40 @@
...
+             let-focused="focused"
+             let-selected="selected"
+             let-currentMonth="currentMonth">
+  <div class="hebrew-day df-datepicker-date"
+       [class.outside]="date.month !== currentMonth"
+       [class.df-datepicker-date-disabled]="disabled"
+       [class.df-datepicker-date-focused]="focused"
+       [class.df-datepicker-date-selected]="selected">
+    <div class="gregorian-num">{{ data.gregorian.day + '/' + (data.gregorian.month) }}</div>
+    <div class="hebrew-num">{{ i18n.getDayNumerals(date) }}</div>
```
matthieu-crouzet, line 38: Maybe we can do a pipe instead of call method in the template

apps/showcase/src/components/utilities/date-picker-input/date-picker-input-pres.config.ts
```diff
@@ -0,0 +1,8 @@
+import type {Configuration} from '@o3r/core';
+import {computeItemIdentifier} from '@o3r/core';
+
+export interface DatePickerInputPresConfig extends Configuration {}
```
matthieu-crouzet, line 4: Is config really needed ?

apps/showcase/src/components/showcase/localization/localization-pres.config.ts
mrednic-1A: I was thinking to create a new chapter for this, but I was thinking to keep it in another feature which will include the update of the doc for component replacement, because the things are a bit simplified with the standalone components (which are quite recent). 
And as a quick win I put this component here to have an example until we decide to have a dedicated chapter in the showcase.

apps/showcase/src/components/utilities/date-picker-input-hebrew/date-picker-input-pres.config.ts
mrednic-1A: Initially the type was needed for c11n directive. I'll double check if I could not pass the Configuration interface directly and remove this interface.

apps/showcase/src/components/utilities/date-picker-input-hebrew/date-picker-input-pres.template.html
mrednic-1A: you are right, but I not insisted on the template at all as I took it from the design-factory doc.

apps/showcase/src/components/utilities/date-picker-input/date-picker-input-pres.config.ts
mrednic-1A: same as previous comment. to be checked

apps/showcase/src/components/utilities/date-picker-input-hebrew/date-picker-input-pres.component.ts
```diff
@@ -0,0 +1,140 @@
...
+import { Observable } from 'rxjs';
+
+@O3rComponent({ componentType: 'ExposedComponent' })
+@Component({
+  selector: 'o3r-date-picker-input-pres-new-design',
+  standalone: true,
+  imports: [FormsModule, CloseInputDatePickerDirective, NgbInputDatepicker, DfInputIconDirective, DfDatePickerModule],
+  templateUrl: './date-picker-input-pres.template.html',
+  styleUrls: ['./date-picker-input-pres.style.scss'],
+  styles: [
```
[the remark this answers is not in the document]
mrednic-1A: Indeed. I followed blindly the doc form design factory :D

apps/showcase/src/components/showcase/component-replacement/component-replacement-pres.component.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  public dateFormControl = new FormControl<string | null>(this.initialDate);
+
+  constructor(
+    private readonly c11nService: C11nService,
+    @Optional() configurationService: ConfigurationBaseService) {
+    this.dynamicConfig$ = new ConfigurationObserver<ComponentReplacementPresConfig>(COMPONENT_REPLACEMENT_PRES_CONFIG_ID, COMPONENT_REPLACEMENT_PRES_DEFAULT_CONFIG, configurationService);
+    this.config$ = this.dynamicConfig$.asObservable();
+    this.loadPresenter();
+    this.dynamicConfig$.next({ ...this.config, datePickerCustomKey: 'exampleDatePickerFlavorHebrew' });
```
matthieu-crouzet, line 58: Maybe we can have a button in the page to trigger this change 
So user can easily see the 2 flavors

matthieu-crouzet: And so it will go through the store as it will work in real use case

apps/showcase/src/components/showcase/component-replacement/component-replacement-pres.template.html
```diff
@@ -0,0 +1,9 @@
+<div class="card p-2">
+  <div class="w-50">
+    <p>Gregorian date: <b>{{(dateFormControl.valueChanges | async) || initialDate}}</b></p>
```
matthieu-crouzet, line 3:
```suggestion
    <p>Gregorian date: <b>{{dateFormControl.value}}</b></p>
```
Isn't it working like this, as you initialize `dateFormControl` with `initialData` ?

apps/showcase/src/components/showcase/component-replacement/component-replacement-pres.template.html
```diff
@@ -0,0 +1,9 @@
...
+  </div>
+  <div class="w-50">
+    <ng-template c11n [component]="presenter$ | async" [inputs]="{id: 'date-outbound'}" [formControl]="dateFormControl"
+      [config]="{}" [outputs]="outputs"></ng-template>
```
matthieu-crouzet, line 7: Do you have to specify `config` and `outputs` if we don't use them in this example ?

apps/showcase/src/components/showcase/component-replacement/component-replacement-pres.component.ts
mrednic-1A: Indeed that's the final goal. But this issue is blocking it. When solved, we'll have the implementation of the switch button. 
https://github.com/AmadeusITGroup/otter/issues/1168

apps/showcase/src/components/showcase/component-replacement/component-replacement-pres.template.html
mrednic-1A: It's working but it;s not fancy :D 
updated :)

apps/showcase/src/components/showcase/component-replacement/component-replacement-pres.template.html
mrednic-1A: good point. They are not mandatory so I removed them.

apps/showcase/src/components/showcase/component-replacement/component-replacement-pres.style.scss
```diff
@@ -0,0 +1,8 @@
+o3r-component-replacement-pres {
+	// Your component custom SCSS
+  .card {
+    display: flex;
```
matthieu-crouzet, line 4: maybe we could use bootstrap classes instead 
(not mandatory)

apps/showcase/src/components/utilities/date-picker-input/date-picker-input-pres.config.ts
matthieu-crouzet: I understand why you must have done that 
but I don't understand why `T` has to extends `DynamicConfigurable`
as configuration for `presKey` is the one of the parent component

mrednic-1A: As params you give the default presenter, which is a component having context and configuration, and a key used to search in presenters map, a custom presenter which has to expend the Type<T>.
Why we force the Context and DynamicConfigurable<Configuration> is because initially we generated all our otter components with context, config, loc etc. Nowadays we don;t force this in the generators, so we might revisit the c11n mechanism to accept more generic components too.

matthieu-crouzet: For the context it could make sense as we are binding inputs and outputs

apps/showcase/src/app/component-replacement/component-replacement.template.html
```diff
@@ -0,0 +1,37 @@
...
+    <o3r-in-page-nav-pres
+      id="component-replacement-nav"
+      [links]="links$ | async"
+    >
+    </o3r-in-page-nav-pres>
+  </div>
+  <div class="order-2 order-lg-1 col-12 col-lg-10">
+    <h2 id="component-replacement-description">Description</h2>
+    <div>
+      <p>This module provides a way to replace the presentation layer of a component inside an application.</p>
```
cpaulve-1A, line 13: which module should the user install via the ng-add?
Shouldn't we provide some code as an example like we do in https://amadeusitgroup.github.io/otter/#/configuration ?

apps/showcase/src/components/showcase/component-replacement/component-replacement-pres.config.ts
```diff
@@ -0,0 +1,12 @@
+import type {Configuration} from '@o3r/core';
+import {computeItemIdentifier} from '@o3r/core';
+
+export interface ComponentReplacementPresConfig extends Configuration {
```
cpaulve-1A, line 4: Missing tsdoc for the exported objects and interfaces

apps/showcase/src/components/showcase/component-replacement/component-replacement-pres.style.scss
mrednic-1A: done :)

apps/showcase/src/app/component-replacement/component-replacement.template.html
mrednic-1A: updated

apps/showcase/src/components/showcase/component-replacement/component-replacement-pres.config.ts
mrednic-1A: update done

apps/showcase/src/components/utilities/date-picker-input/date-picker-input-pres.config.ts
mrednic-1A: Yes. I have created an issue to track this
https://github.com/AmadeusITGroup/otter/issues/1197

apps/showcase/src/components/utilities/date-picker-input-hebrew/date-picker-input-pres.context.ts
```diff
@@ -0,0 +1,23 @@
+import type {Context} from '@o3r/core';
+
+/**
+ * The ContextInput interface describes all the inputs of the component
+ */
+export interface DatePickerInputPresContextInput {
```
matthieu-crouzet, line 6: This interface may extend the original context instead of redefining it no ?

matthieu-crouzet (approved):

apps/showcase/src/components/utilities/date-picker-input-hebrew/date-picker-input-pres.context.ts
mrednic-1A: Indeed it can extend the base one. I prefer to keep it independent to have a clear separation in terms of code.

vscaiceanu-1a (approved):

PR #1151, opened 2023-12-18
feat: remove dependency to @ama-sdk/core

kpanot:
## Proposed change

Remove dependency to @ama-sdk/core to reduce the postinstall time.
This is resulting to a breaking change on `deepFill` (in implementation, not interface)

Alowing script greatly reducing build time (on Mac)

### DIFF
--- a/packages/@o3r/core/src/utils/deep-fill/deep-fill.ts
+++ b/packages/@o3r/core/src/utils/deep-fill/deep-fill.ts
@@ -1,35 +1,49 @@
-import {utils} from '@ama-sdk/core';
+/** Reviver Mapper to detect and create object on deepFill */
+export interface PrimitiveReviverMapper<T = any> {
+  /** Condition to fill to be determine as  */
+  condition: (data: any) => boolean;
 
-/**
- * Check if an object is a date
- * @param obj
- */
-export function isDate(obj: any) {
-  return obj instanceof utils.DateTime || obj instanceof utils.Date;
+  /**
+   * Construct the primitive type if needed
+   * @param data to be constructed
+   * @default {@see defaultConstruct}
+   */
+  construct?: (data: any) => T;
 }
+
+const defaultConstruct = (data: any) => data;
+
+const isDate = (data: any) => data instanceof Date && !isNaN(data as any);
 
 /**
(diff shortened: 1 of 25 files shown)


matthieu-crouzet (dismissed):

packages/@o3r/core/src/utils/deep-fill/deep-fill.ts
```diff
@@ -1,35 +1,49 @@
...
+
 /**
  * Check if an object is not an array or a date
  * @param obj
+ * @param additionalMappers
  */
-export function isObject(obj: any) {
-  return obj instanceof Object && !Array.isArray(obj) && !isDate(obj);
+export function isObject(obj: any, additionalMappers?: PrimitiveReviverMapper[]) {
+  return obj instanceof Object && !Array.isArray(obj) && !additionalMappers?.some(({ condition }) => condition(obj)) && !isDate(obj);
```
fpaul-1A, line 24: since condition is a method of an object, I would call it with the right context
```suggestion
  return obj instanceof Object && !Array.isArray(obj) && !additionalMappers?.some((mapper) => mapper.condition(obj)) && !isDate(obj);
```
in case the method uses `this` for any reason

matthieu-crouzet (approved):

packages/@o3r/core/src/utils/deep-fill/deep-fill.ts
```diff
@@ -1,35 +1,49 @@
...
+export function immutablePrimitive(obj: any, additionalMappers?: PrimitiveReviverMapper[]) {
   if (Array.isArray(obj)) {
     return obj.slice();
-  } else if (obj instanceof utils.DateTime) {
-    return new utils.DateTime(obj);
-  } else if (obj instanceof utils.Date) {
-    return new utils.Date(obj);
+  }
+
+  const matchingPrimitive = additionalMappers?.find(({ condition }) => condition(obj));
```
fpaul-1A, line 37:
```suggestion
  const matchingPrimitive = additionalMappers?.find((mapper) => mapper.condition(obj));
```

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #1152, opened 2023-12-18
feat: captcha support on sdk plugins

cpaulve-1A:
## Proposed change
Stop and restart the timeout fetch plugin in case a specific event interrumpts the fetch request.
Add event emitter to pause the timeout if an imperva captcha has been detected.

### DIFF
--- a/packages/@ama-sdk/core/src/fwk/clients/api-fetch-client.ts
+++ b/packages/@ama-sdk/core/src/fwk/clients/api-fetch-client.ts
@@ -112,7 +112,7 @@ export class ApiFetchClient implements ApiClient {
     // Execute call
     try {
 
-      const controller = typeof AbortController !== 'undefined' ? new AbortController() : undefined;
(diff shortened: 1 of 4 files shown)


matthieu-crouzet (dismissed):

matthieu-crouzet (approved):

fpaul-1A (approved):

packages/@ama-sdk/core/src/fwk/clients/api-fetch-client.ts
```diff
@@ -112,7 +112,7 @@ export class ApiFetchClient implements ApiClient {
     // Execute call
     try {
 
-      const controller = typeof AbortController !== 'undefined' ? new AbortController() : undefined;
+      const controller = new AbortController();
       if (controller) {
```
mrednic-1A, line 116: `if` could be removed

PR #1165, opened 2023-12-19
chore(style-lint): bump stylelint dependency

kpanot:
## Proposed change

feat: mutualize jest silent option
chore(style-lint): bump stylelint dependency

### DIFF
--- a/apps/showcase/src/components/showcase/configuration/configuration-pres.config.ts
+++ b/apps/showcase/src/components/showcase/configuration/configuration-pres.config.ts
@@ -35,7 +35,7 @@ export interface ConfigurationPresConfig extends Configuration {
   destinations: DestinationConfiguration[];
   /**
    * Propose round trip
-   * @o3rCategory globalCategory
+   * @o3rCategory localCategory
    */
   shouldProposeRoundTrip: boolean;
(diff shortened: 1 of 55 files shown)


apps/showcase/src/components/showcase/configuration/configuration-pres.config.ts
```diff
@@ -35,7 +35,7 @@ export interface ConfigurationPresConfig extends Configuration {
   destinations: DestinationConfiguration[];
   /**
    * Propose round trip
-   * @o3rCategory globalCategory
+   * @o3rCategory localCategory
```
matthieu-crouzet, line 38: We want 1 example where we use a globally defined category

matthieu-crouzet (approved):

matthieu-crouzet (approved):

apps/showcase/src/components/showcase/configuration/configuration-pres.config.ts
kpanot: turned because of eslint failure, it may need to be back when fixed (in other PR)

mrednic-1A (approved):

Issue #718, opened 2023-09-04
[Feature]: VSCode intellisense for configuration/styling metadata


kpanot:
### Context

As a developper I would like to have help from my IDE to add metadata to the different modules of Otter.

### Proposal

- [ ] Intellisense for Configuration metadata
  - [x] Support of all configuration tags
  - [ ] Share constants for tag names between packages `extractors`, `eslint-plugin` and vscode extension
  - [x] https://github.com/AmadeusITGroup/otter/issues/625
- [x] Intellisense for Styling metadata https://github.com/AmadeusITGroup/otter/pull/1175
- [x] EsLint plugin rule to validate configuration metadata
  - [x] https://github.com/AmadeusITGroup/otter/issues/1119
  - [x] Support of `@o3rWidget`
    - [x] https://github.com/AmadeusITGroup/otter/issues/893
    - [x] https://github.com/AmadeusITGroup/otter/issues/896
  - [x] Change o3rCategories XML format to JSDoc format
    - [x] Provide migration script

matthieu-crouzet:
https://github.com/AmadeusITGroup/otter/pull/893

matthieu-crouzet:
Intellisense for @o3rWidget and @o3rWidgetParam
https://github.com/AmadeusITGroup/otter/pull/896

PR #1175, opened 2023-12-22
feat(#718): intellisense for styling variable

matthieu-crouzet:
## Proposed change
Intellisense for styling variable

https://github.com/AmadeusITGroup/otter/assets/52541061/4de2350a-6fb8-495b-a569-b7e2e357c822

## Related issues

- :rocket: Feature #718

### DIFF
--- a/apps/vscode-extension/package.json
+++ b/apps/vscode-extension/package.json
@@ -31,6 +31,24 @@
           "type": "string",
           "default": "",
           "description": "If not empty, will add a prefix to all the generated variables"
+        },
+        "otter.styling.variable.categories": {
+          "type": "array",
+          "default": [],
+          "items": {
+            "type": "string"
+          }
+        },
+        "otter.styling.variable.types": {
+          "type": "array",
+          "default": ["color", "string"],
+          "items": {
+            "type": "string"
+          }
+        },
+        "otter.styling.variable.details": {
+          "type": "boolean",
+          "default": true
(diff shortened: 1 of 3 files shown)


apps/vscode-extension/package.json
```diff
@@ -31,6 +31,24 @@
           "type": "string",
           "default": "",
           "description": "If not empty, will add a prefix to all the generated variables"
+        },
+        "otter.styling.variable.categories": {
```
kpanot, line 35: Why not relying on Stylelint as it is done for eslint?

apps/vscode-extension/src/intellisense/styling.ts
```diff
@@ -0,0 +1,48 @@
+import { CompletionItem, CompletionItemKind, CompletionItemProvider, SnippetString, workspace } from 'vscode';
+
+const importStyling = /@use '@o3r\/styling'( as (?<lib>\w+))?;/;
+
+export const stylingCompletionTriggerChar = '$';
```
kpanot, line 5: missing comment on exported var

apps/vscode-extension/package.json
matthieu-crouzet: As discuss we will keep it like that

vscaiceanu-1a (approved):

kpanot (approved):

Issue #1009, opened 2023-11-06
[Feature]: Improve @ama-sdk plugins logging


cpaulve-1A:
### Context

There is no dedicated logging in the @ama-sdk/core. The errors are logged in the console and it is up to the application to catch them and handle them (via the Angular Error Handler for example).

### Proposal

Create a dedicated logger for the API plugins and calls.

kpanot:
This issue is next step of the quick fix from #1008

PR #1184, opened 2024-01-04
feat: improve ama-sdk plugins logging

sdo-1A:
## Proposed change

Possibility to connect a third-party logger in ama-sdk, fallback to console logger if undefined.

## Related issues

- :rocket: Feature #1009


apps/showcase/src/app/app.module.ts
```diff
@@ -40,12 +41,15 @@ const runtimeChecks: Partial<RuntimeChecks> = {
...
 
+const logger = new ConsoleLogger();
+
 function petApiFactory() {
   const apiConfig: ApiClient = new ApiFetchClient(
     {
       basePath: 'https://petstore3.swagger.io/api/v3',
-      requestPlugins: [],
-      fetchPlugins: []
+      requestPlugins: [new SessionIdRequest('Ama-Client-Ref', true, logger)],
```
matthieu-crouzet, line 50: Maybe we could move that to a documentation section 
as it's not needed for this API

packages/@ama-sdk/core/src/plugins/core/fetch-plugin.ts
```diff
@@ -24,6 +25,9 @@ export interface FetchPluginContext {
   // TODO Now supported for all the modern browsers - should become mandatory in @ama-sdk/core@10.0
   /** Abort controller to abort fetch call */
   controller?: AbortController;
+
+  /** Logger (optional, fallback to console logger if undefined) */
+  logger?: Logger;
```
matthieu-crouzet, line 30: Your interface should extend `PluginContext` instead

packages/@ama-sdk/core/src/plugins/core/request-plugin.ts
```diff
@@ -52,11 +53,19 @@ export interface RequestOptions extends RequestInit {
   method: NonNullable<RequestInit['method']>;
 }
 
+/**
+ * Interface of an SDK request plugin context.
+ */
+export interface RequestPluginContext {
```
matthieu-crouzet, line 59: Your interface should extend `PluginContext` instead

packages/@ama-sdk/core/src/plugins/session-id/session-id.request.ts
```diff
@@ -34,22 +35,28 @@ export class SessionIdRequest implements RequestPlugin {
...
+
   /**
    * Constructor.
    * @param sessionIdHeader The request header in which the ID will be added.
    * @param activateRequestId Indicates if the request ID part should be added to the ID.
+   * @param logger Logger (optional, fallback to console logger if undefined)
    */
-  constructor(sessionIdHeader = 'Ama-Client-Ref', activateRequestId = true) {
+  constructor(sessionIdHeader = 'Ama-Client-Ref', activateRequestId = true, logger?: Logger) {
+    this.logger = logger;
```
matthieu-crouzet, line 51: Maybe add a comment to say that it should be done first as `generateSessionId` use the logger
to avoid anyone moving that latter

apps/showcase/src/app/app.module.ts
```diff
@@ -40,12 +41,15 @@ const runtimeChecks: Partial<RuntimeChecks> = {
 registerLocaleData(localeEN, 'en-GB');
 registerLocaleData(localeFR, 'fr-FR');
 
+const logger = new ConsoleLogger();
```
kpanot, line 44: Should not be `LoggerService` injected by Angular instead of being manually instantiated ?

packages/@ama-sdk/core/src/clients/api-angular-client.ts
```diff
@@ -63,7 +63,7 @@ export class ApiAngularClient implements ApiClient {
     let opts = options;
     if (this.options.requestPlugins) {
       for (const plugin of this.options.requestPlugins) {
-        opts = await plugin.load().transform(opts);
+        opts = await plugin.load({logger: this.options.logger}).transform(opts);
```
kpanot, line 66: could be simplified to

```suggestion
        opts = await plugin.load(this.options).transform(opts);
```

packages/@ama-sdk/core/src/logger/index.ts
kpanot, line 1: Could be in `fwk` folder

packages/@ama-sdk/core/src/logger/logger.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * Logger interface.
+ */
+export interface Logger {
```
kpanot, line 4: Why not importing the one from `@o3r/core` (as devDep) ?

packages/@ama-sdk/core/src/fwk/core/base-api-constructor.ts
```diff
@@ -1,3 +1,4 @@
+import {Logger} from '../../logger';
```
kpanot, line 1: could be an `import type`

packages/@ama-sdk/core/src/plugins/bot-protection-fingerprint/bot-protection-fingerprint.request.ts
```diff
@@ -1,9 +1,10 @@
-import {PluginRunner, RequestOptions, RequestPlugin} from '../core';
+import {PluginRunner, RequestOptions, RequestPlugin, RequestPluginContext} from '../core';
+import {Logger} from '../../logger';
```
kpanot, line 2: could be `import type`

packages/@ama-sdk/core/src/plugins/core/fetch-plugin.ts
```diff
@@ -1,4 +1,5 @@
 import type { ApiClient } from '../../fwk/core/api-client';
+import { Logger } from '../../logger';
```
kpanot, line 2: could be `import type`

packages/@ama-sdk/core/src/plugins/pii-tokenizer/pii-tokenizer.request.ts
```diff
@@ -192,21 +193,21 @@ export class PiiTokenizerRequest implements RequestPlugin {
...
-        console.error('Couldn\'t encode the token');
+        (logger || console).error('Couldn\'t encode the token');
       } else {
         throw new Error('Couldn\'t encode the token');
       }
     }
   }
 
-  public load(): PluginRunner<RequestOptions, RequestOptions> {
+  public load(context?: RequestPluginContext): PluginRunner<RequestOptions, RequestOptions> {
```
vscaiceanu-1a, line 210: missing JSDoc

packages/@ama-sdk/core/src/plugins/core/request-plugin.ts
```diff
@@ -52,11 +53,19 @@ export interface RequestOptions extends RequestInit {
...
+}
+
 /**
  * Interface of an SDK request plugin.
  * The plugin will be run on the request of a call
  */
 export interface RequestPlugin extends Plugin<RequestOptions, RequestOptions> {
   /** Load the plugin with the context */
-  load(): PluginRunner<RequestOptions, RequestOptions>;
+  load(context?: RequestPluginContext): PluginRunner<RequestOptions, RequestOptions>;
```
vscaiceanu-1a, line 70: JSDoc

packages/@ama-sdk/core/src/plugins/bot-protection-fingerprint/bot-protection-fingerprint.request.ts
```diff
@@ -248,10 +247,10 @@ export class BotProtectionFingerprintRequest implements RequestPlugin {
     this.hasPolled = true;
   }
 
-  public load(): PluginRunner<RequestOptions, RequestOptions> {
+  public load(context?: RequestPluginContext): PluginRunner<RequestOptions, RequestOptions> {
```
vscaiceanu-1a, line 250: JSDoc

packages/@ama-sdk/core/src/logger/logger.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Log a warning.
+   * @param message Message to log
+   * @param optionalParams Optional parameters to log
+   */
+  warn(message?: any, ...optionalParams: any[]): void;
+
+  /**
+   * Log a message.
```
vscaiceanu-1a, line 20: What's the difference between `info` and `log`?

packages/@ama-sdk/core/README.md
```diff
@@ -42,3 +42,10 @@ A list of API Clients are provided by this package:
 | ApiFetchClient   | @ama-sdk/core                            | Default API Client based on the browser FetchApi                               |
 | ApiBeaconClient  | @ama-sdk/core                            | API Client based on the browser BeaconApi, it is processing synchronous call   |
 | ApiAngularClient | @ama-sdk/core/clients/api-angular-client | API Client using the HttpClient exposed by the `@angular/common` package       |
+
+### Logs
+
+In order to ease the logging in the ama-sdk plugins, it is possible to connect to third-party logging services.
+This can be achieved by adding a `Logger` [implementation](/packages/@ama-sdk/core/src/logger) to the options of an API client (or as a parameter of the plugin for some cases).
```
vscaiceanu-1a, line 49:
```suggestion
This can be achieved by adding a `Logger` [implementation](/packages/@ama-sdk/core/src/logger/logger.ts) to the options of an API client (or as a parameter of the plugin for some cases).
```

packages/@ama-sdk/core/src/plugins/session-id/session-id.request.ts
kpanot: I also wonder if the `this.sessionId = this.generateSessionId();` should not be moved to the `load` function with `this.sessionId ||= this.generateSessionId();` which make more sense and would avoid providing this *additional loader*

packages/@ama-sdk/core/src/logger/logger.ts
sdo-1A: They are aliases, but in some browsers there is a difference in color when displaying the message

packages/@ama-sdk/core/src/plugins/core/request-plugin.ts
sdo-1A: JSDoc is above

apps/showcase/src/app/app.module.ts
cpaulve-1A: LoggerService is provided in @o3r/logger which depends on @o3r/core. Do we really want to add this dependency?

packages/@ama-sdk/core/src/plugins/session-id/session-id.request.ts
```diff
@@ -34,22 +35,28 @@ export class SessionIdRequest implements RequestPlugin {
...
+  protected logger?: Logger;
+
   /**
    * Constructor.
    * @param sessionIdHeader The request header in which the ID will be added.
    * @param activateRequestId Indicates if the request ID part should be added to the ID.
+   * @param logger Logger (optional, fallback to console logger if undefined)
    */
-  constructor(sessionIdHeader = 'Ama-Client-Ref', activateRequestId = true) {
+  constructor(sessionIdHeader = 'Ama-Client-Ref', activateRequestId = true, logger?: Logger) {
```
cpaulve-1A, line 50: Is there a reason why you declare it as a property and not simply scope it in the constructor?
```suggestion
  constructor(sessionIdHeader = 'Ama-Client-Ref', activateRequestId = true, protected logger?: Logger) {
```

sdo-1A: this parameter will be removed with the modifications suggested below

packages/@ama-sdk/core/package.json
```diff
@@ -101,6 +101,7 @@
     "@nx/eslint-plugin": "~17.1.1",
     "@nx/jest": "~17.1.1",
     "@o3r/build-helpers": "workspace:^",
+    "@o3r/core": "workspace:^",
```
kpanot, line 104: I think it should be also in `peer optional` to give hint to the user of the version of @o3r/core used in the definition

packages/@ama-sdk/core/src/plugins/core/reply-plugin.ts
```diff
@@ -1,12 +1,12 @@
 import { ApiTypes } from '../../fwk/api';
 import { ReviverType } from '../../fwk/Reviver';
-import { Plugin, PluginRunner } from './plugin';
+import { Plugin, PluginContext, PluginRunner } from './plugin';
```
kpanot, line 3: They can all be `import type`

packages/@ama-sdk/core/src/plugins/session-id/session-id.request.ts
```diff
@@ -32,29 +33,40 @@ export class SessionIdRequest implements RequestPlugin {
...
 
-  public load(): PluginRunner<RequestOptions, RequestOptions> {
+  /**
+   * Load the plugin with the context
+   * @param context Context of request plugin
+   */
+  public load(context?: RequestPluginContext): PluginRunner<RequestOptions, RequestOptions> {
     return {
       transform: (data: RequestOptions) => {
+        this.sessionId = this.generateSessionId(context?.logger);
```
kpanot, line 69: why it is not in `load` instead of `transform` that would reduce the number of call?
Also you may want to use `||=` instead of `=` for the same purpose

packages/@ama-sdk/core/src/plugins/core/fetch-plugin.ts
```diff
@@ -1,14 +1,14 @@
...
+import type { RequestOptions } from './request-plugin';
 
 export type FetchCall = Promise<Response>;
 
 /**
  * Interface of an SDK reply plugin.
  * The plugin will be run on the reply of a call
  */
-export interface FetchPluginContext {
+export interface FetchPluginContext extends PluginContext{
```
kpanot, line 11:
```suggestion
export interface FetchPluginContext extends PluginContext {
```

packages/@ama-sdk/core/package.json
```diff
@@ -76,6 +76,9 @@
     "@angular/common": {
       "optional": true
     },
+    "@o3r/core": {
```
kpanot, line 79: you should also add it in `peerDependencies`

packages/@ama-sdk/core/.eslintrc.js
```diff
@@ -14,6 +14,20 @@ module.exports = {
     ],
     'sourceType': 'module'
   },
+  'overrides': [
```
kpanot, line 17: may not be needed anymore

packages/@ama-sdk/core/src/plugins/core/request-plugin.ts
kpanot: I think he is referring the parameter.
You may be able also to replace the comment by `/** inheritDoc */`

packages/@ama-sdk/core/src/plugins/session-id/session-id.request.ts
```diff
@@ -32,27 +33,38 @@ export class SessionIdRequest implements RequestPlugin {
   /**
    * The session ID used to track API calls
    */
-  public sessionId: string;
+  public sessionId!: string;
```
kpanot, line 36: It is now really potentially `undefined`:

```suggestion
  public sessionId?: string;
```

packages/@ama-sdk/core/src/plugins/session-id/session-id.request.ts
```diff
@@ -32,27 +33,38 @@ export class SessionIdRequest implements RequestPlugin {
...
-  public load(): PluginRunner<RequestOptions, RequestOptions> {
+  /**
+   * Load the plugin with the context
+   * @param context Context of request plugin
+   */
+  public load(context?: RequestPluginContext): PluginRunner<RequestOptions, RequestOptions> {
+    this.sessionId ||= this.generateSessionId(context?.logger);
     return {
       transform: (data: RequestOptions) => {
         data.headers.append(this.sessionIdHeader, this.requestIdActivated ? `${this.sessionId}:${this.generateRequestId()}` : this.sessionId);
```
kpanot, line 70:
```suggestion
    const sessionId =  this.sessionId ||= this.generateSessionId(context?.logger);
    return {
      transform: (data: RequestOptions) => {
        data.headers.append(this.sessionIdHeader, this.requestIdActivated ? `${sessionId}:${this.generateRequestId()}` : sessionId);
```

packages/@ama-sdk/core/src/plugins/core/request-plugin.ts
sdo-1A: ah I was using the same format as the other core plugins that also have parameters, should I update them all?

sdo-1A: ah I was using the same format as the other core plugins that also have parameters, should I update them all?

kpanot: if `/** inheritDoc */` do the job, I would say "yes". If not, only adding the missing parameter is ok

packages/@ama-sdk/core/src/plugins/session-id/session-id.request.ts
```diff
@@ -32,30 +33,41 @@ export class SessionIdRequest implements RequestPlugin {
...
+  private logSessionId(sessionId: string, date: string, logger?: Logger) {
+    if (logger?.info) {
+      logger.info(`Your debug ID associated to the header "${this.sessionIdHeader}" is: ${sessionId}.`);
+      logger.info(`Generated at: ${date}`);
+    } else if (logger?.log) {
+      logger.log(`Your debug ID associated to the header "${this.sessionIdHeader}" is: ${sessionId}.`);
+      logger.log(`Generated at: ${date}`);
+    } else {
+      console.info(`Your debug ID associated to the header "${this.sessionIdHeader}" is: ${sessionId}.`);
+      console.info(`Generated at: ${date}`);
```
kpanot, line 58:
```suggestion
    (logger?.info || logger?.log || console.info)(`Your debug ID associated to the header "${this.sessionIdHeader}" is: ${sessionId}.`);
    (logger?.info || logger?.log || console.info)(`Generated at: ${date}`);
```

matthieu-crouzet (dismissed):

packages/@ama-sdk/core/package.json
```diff
@@ -76,6 +77,9 @@
...
+      "optional": true
```
fpaul-1A, line 81: is it really optional ?
regardless, `@o3r/core` should also be added in the devDependencies generated by the @ama-sdk/schematics:typescript-shell https://github.com/AmadeusITGroup/otter/blob/release/10.0.0-next/packages/%40ama-sdk/schematics/schematics/typescript/shell/templates/base/package.json.template#L79

sdo-1A: I think it is optional since it is only used to import types (`Logger` from `@o3r/core`)

packages/@ama-sdk/core/src/plugins/session-id/session-id.request.ts
kpanot: not sure to understand why it would not work :D.
Can you give another try, we never now :D

packages/@ama-sdk/core/src/plugins/core/plugin.ts
```diff
@@ -22,12 +24,25 @@ export interface PluginSyncRunner<T, V> {
   transform(data: V): T;
 }
 
+/**
+ * Interface of a plugin context
+ */
+export interface PluginContext {
```
mrednic-1A, line 30: __SUGGESTION__
From what I see the 'Logger' type is used here to export another interface which contains a filed with this type. If I am not mistaken it is the only place where it is used like that. In other places it is used to define directly a type of a property.
If we manage to remove the `logger: Logger` from this interface, we could have the 'o3r/core' as a dev dependency only, because the 'Logger' is not used in the creation of another interface which is exported.
The code will become a bit messy, because where we have the PluginContext we will have PLuginContext & {logger: Logger}
`load(context?: PluginContext & {logger: Logger}):...` but the trade off will be that we will not need `@o3r/core` as a peerDep.

cc: @kpanot @fpaul-1A

packages/@ama-sdk/core/src/plugins/session-id/session-id.request.ts
sdo-1A: I tried again, and it seems only to work when I use the `bind` function:
```
(logger?.info || logger?.log || console.info).bind(logger || console)(`Your debug ID associated to the header "${this.sessionIdHeader}" is: ${sessionId}.`);
(logger?.info || logger?.log || console.info).bind(logger || console)(`Generated at: ${date}`);
```
is this better do you think?

packages/@ama-sdk/core/src/fwk/logger.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Duplicate from [@o3r/core]{@link https://github.com/AmadeusITGroup/otter/blob/main/packages/%40o3r/core/src/log/logger.ts}
+ * All modifications should be made in both interfaces
+ */
+export interface Logger {
+  /**
+   * Log an error.
+   * @param message Message to log
+   * @param optionalParams Optional parameters to log
+   */
+  error(message?: any, ...optionalParams: any[]): void;
```
kpanot, line 13: Maybe it could be valuable to have the `this: Logger` as first param to avoid le context loosing usage

kpanot (approved):

cpaulve-1A (approved):

PR #1188, opened 2024-01-05
feat: add schematic to generate design token specification file

kpanot:
## Proposed change

Provide schematic to generate Design Token from Scss Otter helpers

### DIFF
--- a/packages/@o3r/design/schematics/ extract-token/index.spec.ts
+++ b/packages/@o3r/design/schematics/ extract-token/index.spec.ts
@@ -1,0 +1,106 @@
+import { callRule, Tree } from '@angular-devkit/schematics';
+import { extractToken } from './index';
+import { firstValueFrom } from 'rxjs';
+import type { CssVariable } from '@o3r/styling';
+import { validate } from 'jsonschema';
+import * as fs from 'node:fs';
+import { resolve } from 'node:path';
+
+describe('Extract Token schematic', () => {
+
+  let initialTree: Tree;
+
+  const initialSassFile = `
+@use '@o3r/styling' as o3r;
+
+$breadcrumb-pres-item-icon-size: o3r.variable('breadcrumb-pres-item-icon-size', 3rem);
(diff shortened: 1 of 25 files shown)


packages/@o3r/design/schematics/ extract-token/index.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    const tree = await firstValueFrom(callRule(extractToken({
+      includeTags: true,
+      componentFilePatterns: ['src/component/**.*scss']
+    }), initialTree, { logger } as any));
+
+    expect(tree.readText('src/component/my-comp.theme.scss')).toBe(`
+@use '@o3r/styling' as o3r;
+
+/* --- BEGIN THEME Auto-generated --- */
+$breadcrumb-pres-item-icon-size: o3r.variable('breadcrumb-pres-item-icon-size', 3rem);
```
cpaulve-1A, line 87: Why didn't you pass a different value line 69 to see how it would affect your resulting file ?

packages/@o3r/design/schematics/ extract-token/index.ts
```diff
@@ -0,0 +1,102 @@
...
+ * Extract the token from o3r mixin sass file
+ * @param options
+ */
+export function extractToken(options: ExtractTokenSchematicsSchema): Rule {
+
+  const updateFileContent = (content: string): string => {
+    const start = content.indexOf(patternToDetect);
+    const end = content.lastIndexOf(patternToDetect);
+
+    if (start === -1 || end === -1 || !options.includeTags) {
```
fpaul-1A, line 18: I would say the condition `end === -1` is redundant since it implies `start` would also be `-1`

packages/@o3r/design/schematics/ extract-token/index.spec.ts
kpanot: If you want to confirm that the parsing part does not affect the original file.

fpaul-1A (approved):

cpaulve-1A (approved):

Issue #134, opened 2023-02-23
[Bug]: IT test instability


kpanot:
### Package name

core

### Package version

8.*

### Reproduction steps

Create a PR without code change (can be documentation in docs/)

### Current result

Test_publish job is failing

### Expected result

Test_publish job to pass

fpaul-1A:
Error still reproducing https://github.com/AmadeusITGroup/otter/actions/runs/4572801248/jobs/8072917126

matthieu-crouzet:
maybe fixed by https://github.com/AmadeusITGroup/otter/commit/cccd1eaa37ef187d016fb8e29b4d576677ef0a4a

mrednic-1A:
https://github.com/AmadeusITGroup/otter/actions/runs/5002768186/jobs/8963244512
random error on test-int

kpanot:
Closing this issue thanks to refactor of the IT on Otter v10

PR #1196, opened 2024-01-08
ci: disable it-tests for windows-latest,yarn

fpaul-1A:
## Proposed change

In the current state, the run of it-tests on `[windows-latest, yarn]` is slow and fails most of the time (related to concurrent access to the disk)
This specific combination does not bring a lot of value since we already cover `[linux, yarn]` and `[windows, npm]`, so I propose to disable it, at least until we find a way to make it reliable 

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/.github/workflows/it-tests.yml
+++ b/.github/workflows/it-tests.yml
@@ -52,6 +52,9 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest]
         packageManager: [yarn, npm]
+        exclude:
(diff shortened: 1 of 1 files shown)


kpanot:
I think you will need to add a description on the reason of it (if it's linked to the instability, you can link it to #134 )

kpanot (approved):

matthieu-crouzet (approved):

PR #1206, opened 2024-01-11
fix(ama-sdk-create): add missing dependencies + git init

matthieu-crouzet:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


packages/@ama-sdk/schematics/schematics/typescript/shell/index.ts
```diff
@@ -102,6 +104,13 @@ export function ngGenerateTypescriptSDK(options: NgGenerateTypescriptSDKShellSch
 
   return chain([
     setupRule,
+    async (_, context) => {
+      const workingDirectory = options.directory ? (isAbsolute(options.directory) ? relative(process.cwd(), options.directory) : options.directory) : '.';
+      context.addTask(
+        new RepositoryInitializerTask(workingDirectory)
+      );
+      await lastValueFrom(context.engine.executePostTasks());
```
kpanot, line 112: Do you really need to initialize git before executing the install rule?

What happen when this schematics is used in an existing git repo (when schedule by `ng g @o3r/workspace:sdk`)? Will it be ignored?

matthieu-crouzet: I didn't think of this use case 
I was looking at the create use case
I will double check

mrednic-1A:
I think you can target the 9.6-rc with the fix .

PR #1220, opened 2024-01-16
feat: clean package install in ngAdd schematics

kpanot:
## Proposed change

Refactor of the IT to properly test the generators.
The changes include:
- Migration of install process to task (instead of blocking rule)
- Based the ITs on Npm Create project
- The environment of IT is now defined at pipeline level (as matrix)
- Add NPM Create option to specify the version of yarn used within generated workspace


packages/@o3r/schematics/src/rule-factories/ng-add/dependencies.ts
```diff
@@ -0,0 +1,177 @@
...
+  /** Name of the project */
+  projectName?: string | null;
+
+  /** Skip the run of the linter*/
+  skipLinter?: boolean;
+
+  /** Skip the install process */
+  skipInstall?: boolean;
+
+  [x: string]: any;
```
mrednic-1A, line 22: should not be optional this one too ?

kpanot: It is thanks to `[x: string]` :).
The purpose is just to allow additional fields, they are not mandatory

mrednic-1A: ok. my bad. I was thinking it forces you to have at least one field :)

.vscode/settings.json
```diff
@@ -23,7 +23,8 @@
   "typescript.preferences.quoteStyle": "single",
   "typescript.enablePromptUseWorkspaceTsdk": true,
   "typescript.preferences.preferTypeOnlyAutoImports": true,
-  "jest.jestCommandLine": "yarn jest",
+  "jest.jestCommandLine": "yarn jest --testTimeout=1000000",
```
matthieu-crouzet, line 26: Do we want to keep that ?

package.json
```diff
@@ -14,7 +14,7 @@
     "build:tools": "yarn nx run-many --target=build --projects=eslint-plugin,workspace --parallel $(yarn get:cpus-number)",
     "build:lint": "yarn nx run-many --target=build --projects=eslint-plugin --parallel $(yarn get:cpus-number)",
     "build:swagger-gen": "yarn nx run-many --target=build-swagger --parallel $(yarn get:cpus-number)",
-    "prepare:publish": "prepare-publish $(yarn workspaces:list | shx sed \"s/^(.+)\\$/\\$1\\/dist/\")",
+    "prepare:publish": "prepare-publish $(yarn workspaces:list | yarn node -e 'let data=\"\"; process.openStdin().on(\"data\", (c) => data+=c).on(\"end\", () => process.stdout.write(data.split(/[\\n\\r]+/).slice(0, -1).map((l) => l + \"/dist\").join(require(\"os\").EOL)));')",
```
matthieu-crouzet, line 17: Why do we have to change that ?

.vscode/settings.json
kpanot: I think we do. Without it we cannot run the IT in VSCode :S 
But the value of the timeout can be reviewed of course

packages/@o3r/styling/schematics/index.it.spec.ts
```diff
@@ -4,34 +4,41 @@ import {
...
   beforeAll(async () => {
-    appFolderPath = await prepareTestEnv(appName, 'angular-with-o3r-core');
-    execAppOptions.cwd = appFolderPath;
+    ({ projectPath, workspacePath, projectName, isInWorkspace } = await prepareTestEnv(appFolder));
+    execAppOptions.cwd = workspacePath;
   });
   test('should add styling to existing application', async () => {
-    packageManagerExec(`ng add --skip-confirmation @o3r/styling@${o3rVersion} --enable-metadata-extract`, execAppOptions);
+    const relativeprojectPath = path.relative(workspacePath, projectPath);
+    packageManagerExec({script: 'ng', args: ['add', `@o3r/styling@${o3rVersion}`, '--enable-metadata-extract', '--skip-confirmation']}, execAppOptions);
```
fpaul-1A, line 28: don't you need to pass the `--project-name` ?

package.json
kpanot: I got a duplicate dependency due to shx (which does not seem to be maintenained)

.vscode/settings.json
matthieu-crouzet: 16m40s is maybe a lot

kpanot: Give me a number I will put it :D

packages/@o3r/apis-manager/schematics/index.it.spec.ts
```diff
@@ -3,29 +3,31 @@ import {
...
+
+    expect(() => packageManagerInstall(execAppOptions)).not.toThrow();
+    expect(() => packageManagerRunOnProject(projectName, isInWorkspace, {script: 'build'}, execAppOptions)).not.toThrow();
 
-    const diff = getGitDiff(appFolderPath);
+    const diff = getGitDiff(workspacePath);
     expect(diff.modified).toContain('package.json');
 
-    expect(() => packageManagerInstall(execAppOptions)).not.toThrow();
-    expect(() => packageManagerRun('build', execAppOptions)).not.toThrow();
```
fpaul-1A, line 29: why is this removed ?

packages/@o3r/test-helpers/src/utilities/package-manager.ts
```diff
@@ -187,6 +216,7 @@ export function setPackagerManagerConfig(options: PackageManagerConfig, execAppO
       // Set config to target local registry
       execFileSync('yarn', ['config', 'set', 'checksumBehavior', 'update'], execOptions);
       execFileSync('yarn', ['config', 'set', 'enableImmutableInstalls', 'false'], execOptions);
+      execFileSync('yarn', ['config', 'set', 'enableScripts', 'false'], execOptions);
```
fpaul-1A, line 219: should we add the same for npm ? (`npm config set ignore-scripts true` I suppose)

nx.json
```diff
@@ -73,11 +73,20 @@
...
       "{projectRoot}/jest.config.*",
       "{projectRoot}/testing/**/*",
       "{projectRoot}/mocks/**/*",
       "{workspaceRoot}/jest.config.js"
     ],
+    "integration-test": [
+      "default",
+      "{workspaceRoot}/jest.config.it.js",
+      "{workspaceRoot}/packages/@o3r/create/src/*.ts",
+      "!{workspaceRoot}/packages/@o3r/create/src/*.ts",
```
fpaul-1A, line 86:
```suggestion
      "!{workspaceRoot}/packages/@o3r/create/src/*.spec.ts",
```

packages/@o3r/apis-manager/schematics/index.it.spec.ts
```diff
@@ -3,29 +3,35 @@ import {
...
   test('should add apis-manager to existing application', () => {
-    packageManagerExec(`ng add --skip-confirmation @o3r/apis-manager@${o3rVersion}`, execAppOptions);
+    packageManagerExec({script: 'ng', args: ['add', `@o3r/apis-manager@${o3rVersion}`, '--skip-confirmation', '--project-name', projectName]}, execAppOptions);
 
-    const diff = getGitDiff(appFolderPath);
+    expect(() => packageManagerInstall(execAppOptions)).not.toThrow();
+    expect(() => packageManagerRunOnProject(projectName, isInWorkspace, {script: 'build'}, execAppOptions)).not.toThrow();
+
+    const diff = getGitDiff(workspacePath);
     expect(diff.modified).toContain('package.json');
```
fpaul-1A, line 34: we can add the checks on `dont-modify-me` on every tests
```typescript
expect(diff.all.some((file) => /projects[\\/]dont-modify-me/.test(file))).toBe(false);
```

packages/@o3r/analytics/schematics/index.it.spec.ts
```diff
@@ -2,70 +2,46 @@ import {
...
-      const diff = getGitDiff(execAppOptions.cwd as string);
-      expect(diff.all.some((file) => /projects[\\/]dont-modify-me/.test(file))).toBe(false);
-      expect(diff.modified).toContain('projects/test-app/package.json');
-      expect(diff.added).toContain('projects/test-app/src/components/test-component/test-component.analytics.ts');
-
-      expect(() => packageManagerInstall(execAppOptions)).not.toThrow();
-      expect(() => packageManagerRun('build', execAppOptions)).not.toThrow();
-    });
+  afterAll(async () => {
+    await rm(workspacePath, { recursive: true });
```
fpaul-1A, line 28: are we sure this is not executed in case of failed test ?

kpanot: Not sure but it is not important (as not mandatory).
The reason of it is just because I faced an No Space Left in Windows + NPM VM on the last test.
If it is not clean in case of failure, we would have quick IT to fix before reaching the issue

fpaul-1A: I understand the motive, but my concern is that if we clean the folder in case of failure we won't have access to the files in the artifact published at PR level
not blocking though

fpaul-1A (approved):

packages/@o3r/analytics/schematics/index.it.spec.ts
kpanot: Good point, I will check then

kpanot: It seems that it is not possible in Jest today to avoid executing `afterAll` in case of failure (cf. https://github.com/jestjs/jest/issues/5292)

fpaul-1A: it still possible to work around this like
```typescript
describe('test', () => {
    let wasTestSuccessful = false;

    it('does stuff', () => {
        expect(true).toBe(true);
        wasTestSuccessful = true;
    });

    afterAll(() => {
        if (wasTestSuccessful) {
            try {
                cleanUp();
            } catch (e) {}
        }
    });
});
```
or something like that, the try catch prevents the test to be considered failed if the clean up fails

fpaul-1A (dismissed):

packages/@o3r/analytics/schematics/index.it.spec.ts
kpanot: As discussed, this solution is a workaround working only in case of `--bail 1` (which is the value by default) and if we do not have have series of `it`.

If needed it can be put in place but we need to comment its limit (to not have surprise in the future)

packages/@o3r/core/schematics/ng-add/index.ts
```diff
@@ -68,8 +74,9 @@ function ngAddFn(options: NgAddSchematicsSchema): Rule {
           externalPresetRunner?.rule || noop()
         ])(t, c);
       },
-      displayModuleListRule({ packageName: options.projectName })
-    ])(tree, context);
+      options.projectName ? displayModuleListRule({ packageName: options.projectName }) : noop(),
+      setupDependencies(dependenciesSetupConfig)
```
fpaul-1A, line 78: should we retrofit some of these fixes to v9 ?

packages/@o3r/test-helpers/src/prepare-test-env.ts
```diff
@@ -82,59 +91,68 @@ export async function prepareTestEnv(folderName: string, type: PrepareTestEnvTyp
...
-      angularVersion,
-      materialVersion,
-      replaceExisting: !process.env.CI,
-      generateMonorepo
-    });
+  const prepareFinalApp = (baseApp: string) => {
+    logger.debug?.(`Copying ${baseApp} to ${workspacePath}`);
+    const baseProjectPath = path.join(itTestsFolderPath, baseApp);
+    cpSync(baseProjectPath, workspacePath, { recursive: true, dereference: true, filter: (source) => !/node_modules/.test(source) });
+    packageManagerInstall(execAppOptions);
```
divdavem, line 102:
```suggestion
if (existsSync(path.join(workspacePath, 'package.json'))) {
    packageManagerInstall(execAppOptions);
}
```
Otherwise it can fail if there is no `package.json`. Note that I am wondering why it doesn't fail in CI, whereas it fails on my machine when I only run `yarn nx test-int ama-sdk-create` from a clean state.

fpaul-1A (dismissed):

packages/@o3r/core/schematics/ng-add/index.ts
kpanot: Followed by #1354

packages/@o3r/localization/schematics/localization-base/index.ts
```diff
@@ -440,17 +439,27 @@ export function updateLocalization(options: { projectName?: string | null | unde
    * @param _context
    * @param context
```
vscaiceanu-1a, line 440: no more context *2

packages/@o3r/localization/schematics/localization-base/index.ts
```diff
@@ -22,13 +23,11 @@ import {
...
 import * as path from 'node:path';
+import * as fs from 'node:fs';
 import * as ts from 'typescript';
 
 const packageJsonPath = path.resolve(__dirname, '..', '..', 'package.json');
-const ngxTranslateCoreDep = '@ngx-translate/core';
-const intlMessageFormatDep = 'intl-messageformat';
-const formatjsIntlNumberformatDep = '@formatjs/intl-numberformat';
-const angularCdkDep = '@angular/cdk';
+const ownPackageJson = JSON.parse(fs.readFileSync(packageJsonPath, { encoding: 'utf-8' })) as { peerDependencies: Record<string, string> };
```
vscaiceanu-1a, line 30:
```suggestion
const ownPackageJson = JSON.parse(fs.readFileSync(packageJsonPath, { encoding: 'utf-8' })) as PackageJson;
``` 
(less confusing)

packages/@o3r/localization/schematics/localization-base/index.ts
```diff
@@ -440,17 +439,27 @@ export function updateLocalization(options: { projectName?: string | null | unde
...
+  const addDependencies: Rule = (tree: Tree) => {
+    const ngxTranslateCoreDep = '@ngx-translate/core';
+    const intlMessageFormatDep = 'intl-messageformat';
+    const formatjsIntlNumberformatDep = '@formatjs/intl-numberformat';
+    const angularCdkDep = '@angular/cdk';
     const workspaceProject = options.projectName ? getWorkspaceConfig(tree)?.projects[options.projectName] : undefined;
-    const workingDirectory = workspaceProject?.root || '.';
-    const type = getProjectNewDependenciesType(workspaceProject);
+    const types = getProjectNewDependenciesTypes(workspaceProject);
     const generatorDependencies = [ngxTranslateCoreDep, intlMessageFormatDep, formatjsIntlNumberformatDep, angularCdkDep];
```
vscaiceanu-1a, line 449:
```suggestion
    const generatorDependencies = ['@ngx-translate/core', 'intl-messageformat', '@formatjs/intl-numberformat', '@angular/cdk'];
```

packages/@o3r/schematics/src/rule-factories/ng-add/dependencies.ts
```diff
@@ -0,0 +1,231 @@
...
+export interface SetupDependenciesOptions {
+  /** Map of dependencies to install */
+  dependencies: Record<string, DependencyToAdd>;
+  /**
+   * Pattern of list of the dependency for which the ng-add run process is required
+   */
+  ngAddToRun?: (RegExp | string)[];
+  /**
+   * Will skip install in the end of the package.json update.
+   * if `undefined`, the installation will be process only if a ngAdd run is required).
```
vscaiceanu-1a, line 59:
```suggestion
   * if `undefined`, the installation will be process only if a ngAdd run is required.
```

packages/@o3r/schematics/src/rule-factories/ng-add/dependencies.ts
```diff
@@ -0,0 +1,231 @@
...
+      };
+
+      const finalTaskId = [...ngAddToRun]
+        .map((packageName) => {
+          let schematic: Schematic<any, any> | undefined;
+          try {
+            const collection = context.engine.createCollection(packageName);
+            schematic = collection.createSchematic('ng-add');
+          } catch (e: any) {
+            context.logger.warn(`The packages ${packageName} was not installed, the options check will be skipped`, e);
```
vscaiceanu-1a, line 210:
```suggestion
            context.logger.warn(`The package ${packageName} was not installed, the options check will be skipped`, e);
```

packages/@o3r/schematics/src/rule-factories/ng-add/ng-add.helpers.ts
```diff
@@ -0,0 +1,167 @@
...
+  const installedVersions = packages.map((packageName) => getInstalledVersion(packageName));
+  const packageManager = getPackageManager();
+  const packagesToInstall = packages.filter((packageName, index) => !installedVersions[index] || installedVersions[index] !== versions[packageName]);
+  if (packagesToInstall.length < 1) {
+    return noop;
+  }
+  return chain([
+    // Update package.json in tree
+    (tree) => {
+      const sortDependencies = (packageJson: any, depType: 'dependencies' | 'devDependencies' | 'peerDependencies') => {
```
vscaiceanu-1a, line 85:
```suggestion
      const sortDependencies = (packageJson: PackageJson, depType: 'dependencies' | 'devDependencies' | 'peerDependencies') => {
```

packages/@o3r/schematics/src/rules/install.ts
```diff
@@ -1,17 +1,21 @@
-import type { Rule, SchematicContext, Tree } from '@angular-devkit/schematics';
+import type { Rule } from '@angular-devkit/schematics';
 import { NodePackageInstallTask } from '@angular-devkit/schematics/tasks';
 import { lastValueFrom } from 'rxjs';
-import { getPackageManager } from '../utility/package-manager-runner';
+import { getPackageManager, type SupportedPackageManagers } from '../utility/package-manager-runner';
 
 /**
  * Install the Otter packages
+ * @deprecated use {@link setupDependencies} instead, will be removed in V11
```
vscaiceanu-1a, line 8: removed in V12 I believe

packages/@o3r/schematics/src/rules/install.ts
```diff
@@ -1,17 +1,21 @@
...
  * @param tree
```
vscaiceanu-1a, line 9: no tree nor context.
Also, changing the method interface is a breaking change

packages/@o3r/schematics/src/tasks/ng-add/index.ts
```diff
@@ -31,6 +31,7 @@ export interface NgAddPackageOptions {
   skipNgAddSchematicRun?: boolean;
 }
 
+/** @deprecated use {@link setupDependencies} instead, will be removed in V11 */
```
vscaiceanu-1a, line 34: v12?

packages/@o3r/schematics/src/tasks/package-manager/add-dev-dependency.ts
```diff
@@ -5,6 +5,7 @@ import { NodePackageInstallTaskOptions } from './interfaces';
 
 /**
  * Install new dev dependency on your package
+ * @deprecated use {@link setupDependencies} instead, will be removed in V11
```
vscaiceanu-1a, line 8: v12?

packages/@o3r/schematics/src/utility/loaders.ts
```diff
@@ -76,13 +76,23 @@ export function readPackageJson(tree: Tree, workspaceProject: WorkspaceProject)
...
  * @param tree
  */
 export function getProjectNewDependenciesType(project?: WorkspaceProject) {
   return project?.projectType === 'library' ? NodeDependencyType.Peer : NodeDependencyType.Default;
 }
 
+/**
+ * Return the types of install to run depending on the project type
+ * @param project
+ * @param tree
```
vscaiceanu-1a, line 90: no tree

packages/@o3r/schematics/src/utility/loaders.ts
```diff
@@ -76,13 +76,23 @@ export function readPackageJson(tree: Tree, workspaceProject: WorkspaceProject)
 
 /**
  * Return the type of install to run depending on the project type (Peer or default)
+ * @deprecated use {@link getProjectNewDependenciesTypes instead}, will be removed in V11
  * @param project
  * @param tree
```
vscaiceanu-1a, line 81: no tree and v12?

packages/@o3r/test-helpers/src/prepare-test-env.ts
```diff
@@ -28,36 +28,44 @@ export function getYarnVersionFromRoot(rootFolderPath: string) {
...
-  const globalFolderPath = path.join(rootFolderPath, '.cache', 'test-app');
-  const cacheFolderPath = path.join(globalFolderPath, 'cache');
+  const itTestsFolderPath = path.resolve(rootFolderPath, '..', 'it-tests');
+  const workspacePath = path.resolve(itTestsFolderPath, folderName);
+  const globalFolderPath = path.resolve(rootFolderPath, '.cache', 'test-app');
+  const cacheFolderPath = path.resolve(globalFolderPath, 'cache');
 
-  const o3rCorePackageJson: PackageJson & { generatorDependencies?: Record<string, string> } =
-    JSON.parse(readFileSync(path.join(rootFolderPath, 'packages', '@o3r', 'core', 'package.json')).toString());
+  // const o3rCorePackageJson: PackageJson & { generatorDependencies?: Record<string, string> } =
```
vscaiceanu-1a, line 55: commented lines should be removed

packages/@o3r/test-helpers/src/prepare-test-env.ts
```diff
@@ -28,36 +28,44 @@ export function getYarnVersionFromRoot(rootFolderPath: string) {
...
-    cwd: appFolderPath,
+    cwd: workspacePath,
     stdio: 'inherit',
     // eslint-disable-next-line @typescript-eslint/naming-convention
     env: {...process.env, NODE_OPTIONS: '', CI: 'true'}
   };
 
   // Remove all cache entries relative to local workspaces (@o3r, @ama-sdk, @ama-terasu)
-  if (!process.env.CI && existsSync(cacheFolderPath)) {
+  if (!!process.env.CI && existsSync(cacheFolderPath)) {
```
vscaiceanu-1a, line 68: `!!process.env.CI` or `!process.env.CI`?

packages/@o3r/test-helpers/src/test-environments/create-test-environment-angular-with-o3r-core.ts
```diff
@@ -45,6 +45,7 @@ export interface CreateTestEnvironmentAngularWithO3rCoreOptions extends CreateTe
  * Generate a base angular app with minimal necessary dependencies and @o3r/core installed with basic preset
  * Uses a locker mechanism so this function can be called in parallel
  * The lock will automatically expire after 10 minutes if the creation of the app failed for whatever reason
+ * @deprecated Use {@link createTestEnvironmentOtterProjectWithApp} instead
```
vscaiceanu-1a, line 48: when will it be removed?

packages/@o3r/test-helpers/src/test-environments/create-test-environment-angular.ts
```diff
@@ -133,8 +133,8 @@ export async function createTestEnvironmentAngular(inputOptions: Partial<CreateT
 
     // Run ng-adds
     const project = options.generateMonorepo ? '--project=test-app' : '';
-    packageManagerExec(`ng add @angular/pwa ${project}`, execAppOptions);
-    packageManagerExec(`ng add @angular/material ${project}`, execAppOptions);
+    packageManagerExec({script: 'ng', args: ['add', '@angular/pwa', project]}, execAppOptions);
+    packageManagerExec({script: 'ng', args: ['add', '@angular/material', project]}, execAppOptions);
 
     if (options.generateMonorepo) {
       // TODO remove this when https://github.com/AmadeusITGroup/otter/issues/603 has been resolved
```
vscaiceanu-1a, line 140: isn't it fixed?

packages/@o3r/apis-manager/schematics/ng-add/index.ts
```diff
@@ -1,5 +1,5 @@
 import { chain, noop, Rule, SchematicContext, Tree } from '@angular-devkit/schematics';
-import { createSchematicWithMetricsIfInstalled } from '@o3r/schematics';
+import { createSchematicWithMetricsIfInstalled, getPackageInstallConfig } from '@o3r/schematics';
```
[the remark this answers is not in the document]
kpanot: I don't so, why do you think it should be?

packages/@o3r/test-helpers/src/test-environments/create-test-environment-angular.ts
kpanot: deprecated environement, no fix will be done on it on this PR

packages/@o3r/schematics/src/rules/install.ts
kpanot: nop, we will not wait 1 year to remove a schematic helper

packages/@o3r/schematics/src/tasks/ng-add/index.ts
kpanot: nop, will not wait 1 year to remove a schematic helper

packages/@o3r/schematics/src/tasks/package-manager/add-dev-dependency.ts
kpanot: nop, will not wait 1 year to remove a schematic helper

packages/@o3r/schematics/src/utility/loaders.ts
kpanot: nop, will not wait 1 year to remove a schematic helper

packages/@o3r/test-helpers/src/prepare-test-env.ts
kpanot: `!!process.env.CI`

packages/@o3r/apis-manager/schematics/ng-add/index.ts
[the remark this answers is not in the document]
vscaiceanu-1a: I am asking because in all `ng-add` schematics we are dynamically importing from `@o3r/schematics` to output an error if the package is missing.
```typescript
 try {
      const { ngAddPackages, getO3rPeerDeps, applyEsLintFix, getWorkspaceConfig, getProjectNewDependenciesType } = await import('@o3r/schematics');
      ....
    } catch (e) {
      // o3r apis-manager needs o3r/schematics as peer dep.
      context.logger.error(`[ERROR]: Adding @o3r/apis-manager has failed.
      You need to install '@o3r/schematics' to be able to use the o3r apis-manager package. Please run 'ng add @o3r/schematics' .`);
      throw (e);
    }
  };
```
In this case, I don't see the need of the dynamic import.

fpaul-1A (approved):

fpaul-1A (approved):

packages/@o3r/apis-manager/schematics/ng-add/index.ts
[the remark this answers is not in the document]
vscaiceanu-1a: Since this discussion is not blocking the PR, could you create an issue so that we can tackle it separately?

matthieu-crouzet (approved):

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #1230, opened 2024-01-16
feat(it-tests): test schematics with yarn 1

vscaiceanu-1a:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


packages/@o3r/telemetry/src/environment/index.ts
```diff
@@ -4,22 +4,22 @@ import { execSync } from 'node:child_process';
...
-type SupportedPackageManagers = 'npm' | 'yarn';
+type SupportedPackageManagers = 'npm' | 'yarn' | 'yarn1';
 
 /**
  * Determine if the given packager manager is supported
  * @param name Name of the package manager
  */
 function isSupportedPackageManager(name?: any): name is SupportedPackageManagers {
-  return name === 'yarn' || name === 'npm';
+  return name in ['yarn', 'yarn1', 'npm'];
```
fpaul-1A, line 14:
```suggestion
  return ['yarn', 'yarn1', 'npm'].includes(name);
```

packages/@o3r/create/src/index.it.spec.ts
```diff
@@ -77,7 +77,7 @@ describe('Create new otter project command', () => {
     const appPackageJson = JSON.parse(await fs.readFile(path.join(inAppPath, 'apps', appName, 'package.json'), 'utf-8'));
     // all otter dependencies in both package.json files must be pinned:
     [
-      ...Object.entries(rootPackageJson.dependencies), ...Object.entries(rootPackageJson.devDependencies), ...Object.entries(resolutions),
+      ...Object.entries(rootPackageJson.dependencies), ...Object.entries(rootPackageJson.devDependencies), ...(resolutions ? Object.entries(resolutions) : []),
```
divdavem, line 80: Here as well, I think the test should fail if resolutions are not present.

packages/@ama-sdk/create/src/index.it.spec.ts
```diff
@@ -20,13 +21,14 @@ let sdkFolderPath: string;
 let sdkPackagePath: string;
 const execAppOptions = getDefaultExecSyncOptions();
 const packageManager = getPackageManager();
+const describeSkipIfYarn1 = isYarn1() ? describe.skip : describe;
 
-describe('Create new sdk command', () => {
+describeSkipIfYarn1('Create new sdk command', () => {
```
fpaul-1A, line 26: can't you do something like:
```suggestion
describe('Create new sdk command', () => {
    test.skip(isYarn1());
```
?

vscaiceanu-1a: I don't think this syntax will work.

packages/@o3r/create/src/index.ts
```diff
@@ -201,14 +201,14 @@ const prepareWorkspace = (relativeDirectory = '.', projectPackageManager = 'npm'
     cwd
   }));
 };
-
+const isYarn1 = packageManager === 'yarn' && argv['yarn-version']?.startsWith('1');
```
fpaul-1A, line 204: this will break on Yarn 10 :D

fpaul-1A (dismissed):

packages/@ama-sdk/create/src/index.it.spec.ts
kpanot, line 26: Why do you need to skip this test in case of `yarn 1`?
Is it because of the config? In that case it would maybe make sense to fix it instead

packages/@o3r/create/src/index.ts
kpanot:
```suggestion
const isYarn1 = packageManager === 'yarn' && argv['yarn-version']?.split('.')[0] === '1';
```

packages/@o3r/test-helpers/src/utilities/package-manager.ts
```diff
@@ -41,14 +50,36 @@ type CommandArguments = {
...
+ */
+export function getPackageManager() {
+  return getVersionedPackageManager() === 'npm' ? 'npm' : 'yarn';
+}
+
+
+/**
+ * latest version of yarn 1
+ */
+export const YARN_1_LATEST_VERSION = '1.22.19';
```
kpanot, line 74: should not be hardcoded

packages/@o3r/test-helpers/src/utilities/package-manager.ts
```diff
@@ -41,14 +50,36 @@ type CommandArguments = {
...
+
+/**
+ * is yarn 1
+ */
+export function isYarn1() {
+  return process.env.ENFORCED_PACKAGE_MANAGER === 'yarn1';
```
kpanot, line 80: Not enough to consider the current package manager is yarn 1

packages/@o3r/create/src/index.ts
```diff
@@ -201,14 +201,14 @@ const prepareWorkspace = (relativeDirectory = '.', projectPackageManager = 'npm'
     cwd
   }));
 };
-
+const isYarn1 = packageManager === 'yarn' && argv['yarn-version']?.startsWith('1.');
```
kpanot, line 204:
```suggestion
const isYarn1 = packageManager === 'yarn' && argv['yarn-version']?.split('.')[0] === '1';
```

packages/@o3r/test-helpers/src/prepare-test-env.ts
```diff
@@ -49,7 +48,7 @@ export async function prepareTestEnv(folderName: string, options?: PrepareTestEn
   const cacheFolderPath = path.resolve(globalFolderPath, 'cache');
 
   JSON.parse(readFileSync(path.join(rootFolderPath, 'packages', '@o3r', 'core', 'package.json')).toString());
-  const yarnVersion: string = yarnVersionParam || getYarnVersionFromRoot(rootFolderPath);
+  const yarnVersion: string = yarnVersionParam || (isYarn1() && YARN_1_LATEST_VERSION) || getYarnVersionFromRoot(rootFolderPath);
```
kpanot, line 51: `YARN_1_LATEST_VERSION` is always `truthy`

packages/@o3r/test-helpers/src/utilities/package-manager.ts
```diff
@@ -231,6 +262,23 @@ export function setPackagerManagerConfig(options: PackageManagerConfig, execAppO
...
       break;
     }
+    case 'yarn1': {
+      if (options.yarnVersion) {
+        execFileSync('yarn', ['set', 'version', options.yarnVersion], execOptions);
+      }
+      const ignoreRootCheckConfig = '--add.ignore-workspace-root-check';
+      const yarnRcPath = join(execOptions.cwd as string, '.yarnrc');
+
+      if (existsSync(yarnRcPath)) {
```
kpanot, line 272: why not using a child process to add this config ?

vscaiceanu-1a: because of an issue with yarn 1 https://github.com/yarnpkg/yarn/issues/6499

packages/@o3r/test-helpers/src/utilities/package-manager.ts
```diff
@@ -231,6 +262,34 @@ export function setPackagerManagerConfig(options: PackageManagerConfig, execAppO
       execFileSync('yarn', ['config', 'set', 'unsafeHttpWhitelist', '127.0.0.1'], execOptions);
       break;
     }
+    case 'yarn1': {
+      if (options.yarnVersion) {
+        execFileSync('yarn', ['set', 'version', options.yarnVersion], execOptions);
+        execFileSync('yarn', ['config', 'set', '@ama-sdk:registry', options.registry], execOptions);
+        execFileSync('yarn', ['config', 'set', '@ama-terasu:registry', options.registry], execOptions);
+        execFileSync('yarn', ['config', 'set', '@o3r:registry', options.registry], execOptions);
+        execFileSync('yarn', ['config', 'set', 'unsafeHttpWhitelist', '127.0.0.1'], execOptions);
```
fpaul-1A, line 271: if I remember correctly, these are modifying the `~/.yarnrc` on the machine, not at project level
this probably fine for the CI but not ideal when running locally
it might be better to directly modify the file like you do line 274 instead

packages/@o3r/test-helpers/src/prepare-test-env.ts
```diff
@@ -43,14 +47,15 @@ export async function prepareTestEnv(folderName: string, options?: PrepareTestEn
   const logger = options?.logger || console;
   const yarnVersionParam = options?.yarnVersion;
   const rootFolderPath = process.cwd();
-  const itTestsFolderPath = path.resolve(rootFolderPath, '..', 'it-tests');
+  const itTestsFolderPath = path.resolve(rootFolderPath, '..', 'it-tests', options?.packageScope || '');
```
fpaul-1A, line 50: not linked to this PR but it would be good to keep the `it-tests` folder inside the repository in a .tmp folder
so it can be properly be cleaned-up with git clean

packages/@ama-sdk/create/src/index.it.spec.ts
```diff
@@ -26,10 +27,20 @@ const execAppOptions = getDefaultExecSyncOptions();
 const packageManager = getPackageManager();
 
 describe('Create new sdk command', () => {
-  beforeEach(() => {
```
fpaul-1A, line 29: can't you revert this change ? it seems the beforeEach should be working just fine and you are no longer using `_sdkFolderIndex`

mrednic-1A: of course. It's not needed anymore

packages/@o3r/test-helpers/src/utilities/package-manager.ts
mrednic-1A: true that. I'll do the updates directly on the file in the project

packages/@o3r/test-helpers/src/prepare-test-env.ts
mrednic-1A: we can put it in a **tmp** folder which is already gitignored.

fpaul-1A: unfortunate :(

packages/@o3r/test-helpers/src/utilities/package-manager.ts
```diff
@@ -231,6 +262,41 @@ export function setPackagerManagerConfig(options: PackageManagerConfig, execAppO
...
+        const content = readFileSync(yarnRcPath, { encoding: 'utf8' });
+        const ignoreRootCheckConfig = '--add.ignore-workspace-root-check';
+
+        if (!content.includes(ignoreRootCheckConfig)) {
+          appendFileSync(yarnRcPath, `\n${ignoreRootCheckConfig} true`);
+        }
+
+        if (options.globalFolderPath) {
+          if (!content.includes('prefix')) {
+            appendFileSync(yarnRcPath, `\nprefix "${posix.join(options.globalFolderPath.split(sep).join(posix.sep), `yarn1-global-folder`, options.packageScope || '')}"`);
```
kpanot, line 286:
```suggestion
            appendFileSync(yarnRcPath, `\nprefix "${posix.join(options.globalFolderPath.split(sep).join(posix.sep), 'yarn1-global-folder', options.packageScope || '')}"`);
```

packages/@o3r/test-helpers/src/utilities/package-manager.ts
```diff
@@ -231,6 +262,41 @@ export function setPackagerManagerConfig(options: PackageManagerConfig, execAppO
...
+          }
+          ['cache-folder', 'global-folder'].forEach(folder => {
+            if (!content.includes(folder)) {
+              appendFileSync(yarnRcPath, `\n${folder} "${posix.join(options.globalFolderPath!.split(sep).join(posix.sep), `yarn1-${folder}`, options.packageScope || '')}"`);
+            }
+          });
```
kpanot, line 292:
```suggestion
          ['cache-folder', 'global-folder']
            .filter((folder) => !content.includes(folder))
            .forEach(folder => {
              appendFileSync(yarnRcPath, `\n${folder} "${posix.join(options.globalFolderPath!.split(sep).join(posix.sep), `yarn1-${folder}`, options.packageScope || '')}"`);
            });
```

packages/@o3r/test-helpers/src/utilities/package-manager.ts
```diff
@@ -41,14 +50,30 @@ type CommandArguments = {
...
     process.env.ENFORCED_PACKAGE_MANAGER as keyof typeof PACKAGE_MANAGERS_CMD :
     'yarn';
 }
 
+/**
+ * Get the package manager to be used for the tests based on env ENFORCED_PACKAGE_MANAGER
+ * 'yarn', 'npm'
+ */
+export function getPackageManager() {
+  return getVersionedPackageManager() === 'npm' ? 'npm' : 'yarn';
```
kpanot, line 67:
```suggestion
  const pm = getVersionedPackageManager();
  return pm.startsWith('yarn') ? 'yarn' : pm;
```

packages/@o3r/test-helpers/src/prepare-test-env.ts
mrednic-1A: the code was removed

packages/@ama-sdk/create/src/index.it.spec.ts
mrednic-1A: we are on it. The goal of the PR will be to have yarn 1 IT tests

PR #1247, opened 2024-01-18
fix: ng g application with yarn 1

vscaiceanu-1a:
## Proposed change

Today, with `yarn 1`

```shell
yarn create @o3r my-workspace
cd my-workspace
yarn ng g application my-app
```
we get:
```shell
✔ Packages installed successfully.
error Running this command will add the dependency to the workspace root rather than the workspace itself, which might not be what you want - if you really meant it, make it explicit by running this command again with the -W flag (or --ignore-workspace-root-check).
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
✖ Package install failed, see above.
The Schematic workflow failed. See above.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

### DIFF
--- a/packages/@ama-sdk/core/schematics/ng-add/index.ts
+++ b/packages/@ama-sdk/core/schematics/ng-add/index.ts
@@ -1,14 +1,16 @@
-import {chain, type Rule} from '@angular-devkit/schematics';
+import { chain, type Rule } from '@angular-devkit/schematics';
 import { NodePackageInstallTask } from '@angular-devkit/schematics/tasks';
 import { createSchematicWithMetricsIfInstalled } from '@o3r/schematics';
-import * as ts from 'typescript';
 import * as path from 'node:path';
+import * as ts from 'typescript';
+import type { NgAddSchematicsSchema } from './schema';
 
 /**
  * Rule to import all the necessary dependency to run an @ama-sdk based application
  * Helps to migrate from previous versions with an import replacement
+ * @param options schema options
  */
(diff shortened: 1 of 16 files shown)


packages/@ama-sdk/core/schematics/ng-add/index.ts
```diff
@@ -1,14 +1,16 @@
-import {chain, type Rule} from '@angular-devkit/schematics';
+import { chain, type Rule } from '@angular-devkit/schematics';
 import { NodePackageInstallTask } from '@angular-devkit/schematics/tasks';
 import { createSchematicWithMetricsIfInstalled } from '@o3r/schematics';
-import * as ts from 'typescript';
 import * as path from 'node:path';
+import * as ts from 'typescript';
+import { NgAddSchematicsSchema } from './schema';
```
kpanot, line 6:
```suggestion
import type { NgAddSchematicsSchema } from './schema';
```

kpanot:
missing a pipeline based on `yarn 1` to test the content of this :S.

vscaiceanu-1a:
> missing a pipeline based on `yarn 1` to test the content of this :S.

We are far from having a working CI for end-to-end test with yarn 1. That is the goal of https://github.com/AmadeusITGroup/otter/pull/1230

The goal of this PR is to allow users at least to be able to generate a first application within a monorepo while using yarn 1.

kpanot (approved):

PR #1263, opened 2024-01-24
fix: turn markdown links check optional for publish

mrednic-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/.github/workflows/code-check.yml
+++ b/.github/workflows/code-check.yml
@@ -24,6 +24,7 @@ env:
 jobs:
   markdown-link-check:
     runs-on: ubuntu-latest
+    continue-on-error: true
     steps:
     - uses: actions/checkout@master
     - uses: gaurav-nelson/github-action-markdown-link-check@v1
--- a/.gitignore
+++ b/.gitignore
@@ -84,3 +84,6 @@ coverage
 
 # OS
 *.DS_Store
+
+# NX
+.nx/cache


vscaiceanu-1a (approved):

fpaul-1A (dismissed):

.github/workflows/code-check.yml
```diff
@@ -24,6 +24,7 @@ env:
 jobs:
   markdown-link-check:
     runs-on: ubuntu-latest
+    continue-on-error: true
```
kpanot, line 27: Not sure it is the good way to go.
Does other job depending of this one?

fpaul-1A (approved):

.github/workflows/code-check.yml
mrednic-1A: A test has been done, in order to simulate the behaviors.
if `markdown-link-check` job fails the jobs depending on this one will run.
if other job from from the same group ('code-check.yml') fails, this will block the execution of the depending jobs

In this commit I have tested that 'it-tests' runs even if _markdown_ fails, and it was run
https://github.com/AmadeusITGroup/otter/compare/e8b335a3f924fd7cfa906bc08f9bb13509abd682..be4f6d038895fc74eca610be891b754f43c08f0f
Then I have tested with a unit test failed on purpose, and the _it-tests_ job was not run

kpanot (approved):

PR #1267, opened 2024-01-24
feat: weekly check external links 

vscaiceanu-1a:
## Proposed change

The goal is to have a weekly (or manually triggered) Github action that will check external links for all our `.md` files.

### DIFF
--- a/.github/markdown-external-links.config.json
+++ b/.github/markdown-external-links.config.json
@@ -1,0 +1,11 @@
+{
+  "ignorePatterns": [
+    {
+      "pattern": "^(?!http(s)?:\/\/).*"
+    },
+    {
+      "pattern": "https://pkgs.dev.azure.com/AmadeusDigitalAirline/Otter/_packaging/otter-pr/npm/registry/"
+    }
+  ],
+  "aliveStatusCodes": [200, 203]
+}
--- a/.github/workflows/md-external-links-check.yml
+++ b/.github/workflows/md-external-links-check.yml
@@ -1,0 +1,33 @@
+name: Check external links
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "20 4 * * 1"
+permissions:
+  contents: read
+  issues: write
+jobs:
+  markdown-link-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: gaurav-nelson/github-action-markdown-link-check@v1
+        with:
+          use-quiet-mode: 'yes'
+          base-branch: 'main'
+          config-file: '.github/markdown-external-links.config.json'
+  open-issue:
+    runs-on: ubuntu-latest
+    needs: [markdown-link-check]
+    if: failure()
+    steps:
+      - uses: actions/checkout@v4
+      - name: Create an issue
+        uses: JasonEtco/create-an-issue@v2
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RUN_URL: ${{ format('https://github.com/{0}/actions/runs/{1}/attempts/{2}', github.repository, github.run_id, github.run_attempt || 1) }}
+        with:
+          filename: .github/markdown-check-issue-template.md
+          update_existing: true
+
--- a/.github/markdown-check-issue-template.md
+++ b/.github/markdown-check-issue-template.md
@@ -1,0 +1,5 @@
+---
+title: Links are broken in the documentation
+labels: bug, documentation
+---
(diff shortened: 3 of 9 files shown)


fpaul-1A (dismissed):

sdo-1A (dismissed):

kpanot (dismissed):

.github/workflows/md-external-links-check.yml
```diff
@@ -0,0 +1,15 @@
+name: Check external links
+on:
+  workflow_dispatch:
+  schedule:
```
kpanot, line 4: Are us sure we will see the result of this task?
I understand you are not doing it in PR to not break a PR on something which does not depend of its content but I am afraid that we never see the status of the PR.

Maybe we should also provide a more visible result

.github/markdown-external-links.config.json
```diff
@@ -0,0 +1,11 @@
+{
```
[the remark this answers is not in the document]
vscaiceanu-1a: Nope but we have an issue for it. https://github.com/tcort/markdown-link-check/issues/84

.github/workflows/md-external-links-check.yml
vscaiceanu-1a: > we can:
1. open an issue on failure using something like https://github.com/JasonEtco/create-an-issue
2. send an email https://github.com/marketplace/actions/send-email

I would go for 1. WDYT?

.github/markdown-external-links.config.json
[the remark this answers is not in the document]
matthieu-crouzet: The repo have no activities since 9 months

.github/workflows/md-external-links-check.yml
matthieu-crouzet: I would go for 1 as well

kpanot: I would go for one (but you do not need a github action for that, a simple script would do the job)

vscaiceanu-1a: @kpanot I have added a simple script that will create the issue. However, this will always create a new issue, it won't check if an issue on this topic is already open.
If we want this, I would use the Github action (which provides this option).

kpanot: Why not doing something like this:
- create an issue with a specific title (if it does not exist another opened issue with the same title)
- Replace the body of the issue with the specific title (new issue or previous one) with the list of the broken links as check box
?

vscaiceanu-1a: > * Replace the body of the issue with the specific title (new issue or previous one) with the list of the broken links as check box

This can be cumbersome because `github-action-markdown-link-check` does not provide any output.

matthieu-crouzet (dismissed):

.github/workflows/md-external-links-check.yml
```diff
@@ -0,0 +1,38 @@
...
+      - uses: actions/checkout@v4
+      - name: Compute Run URL
+        uses: actions/github-script@v7
+        id: compute-run-url
+        with:
+          script: return `https://github.com/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}/attempts/${context.runAttempt || 1}`;
+          result-encoding: string
+
+      - name: Create an issue
+        uses: JasonEtco/create-an-issue@v2
```
fpaul-1A, line 31:
```suggestion
        uses: JasonEtco/create-an-issue@v2
        permissions:
          contents: read
          issues: write 
```

.github/workflows/md-external-links-check.yml
```diff
@@ -0,0 +1,40 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RUN_URL: ${{ steps.compute-run-url.outputs.result }}
```
kpanot, line 36:
```suggestion
          RUN_URL: ${{ format('https://github.com/{0}/{1}/actions/runs/{3}/attempts/{4}', context.repo.owner, context.repo.repo, context.runId, context.runAttempt || 1) }}
```
Would not do the job?

fpaul-1A (approved):

.github/workflows/md-external-links-check.yml
vscaiceanu-1a: unfortunately, no.
To access `context` we need to use `actions/github-script`

kpanot: but you do have access to `GITHUB_REPOSITORY_OWNER`, `GITHUB_RUN_ATTEMPT` and `GITHUB_RUN_ID` `GITHUB_REPOSITORY` .
If I remember well, `context` variable is actually populated with the `github` envvar

kpanot (approved):

fpaul-1A (approved):

PR #1273, opened 2024-01-24
feat: support missing design token references

kpanot:
## Proposed change

Support the case where a Design Token value refer an not existing variable


fpaul-1A (approved):

fpaul-1A:
should it be retrofitted into v9 ?

Issue #1214, opened 2024-01-15
[Bug]: Otter pipes extending angular pipes generate warnings.


guillaumerochelle:
### Package name

localization

### Package version

9.5.0

### Reproduction steps

Add the LocalizationModule from @o3r/localization in a component.

### Current result

During unit tests, the following warning is thrown:
```
NG0313: Multiple pipes match the name `translate` in the 'MyComponent' component. check the imports of this module
```

### Expected result

No warnings about multiple pipes with same name.

### Additional comments

The warnings started to be generated since angular 16.2
https://github.com/angular/angular/commit/3dafc14e84ecfc21ffed2b064eac2ce789c854a4

Pawel Kozlowski suggests that libraries should prefix their pipes:
https://github.com/angular/angular/issues/13569#issuecomment-268019834

All the pipes with the same name as their angular counterpart are affected:
- decimal
- currency
- date
- translate

Angular may decide to throw an error in the future.

PR #1280, opened 2024-01-25
feat(pipes): prefix pipes with o3r

matthieu-crouzet:
## Proposed change
Prefix pipes with o3r to avoid conflicts with others libraries
IMO, it's not possible to provide ng-update because we cannot know easily from where the pipe is imported.

## Related issues

- :rocket: Feature #1214

### DIFF
--- a/packages/@o3r/components/src/tools/pipes/replace-with-bold/replace-with-bold.pipe.ts
+++ b/packages/@o3r/components/src/tools/pipes/replace-with-bold/replace-with-bold.pipe.ts
@@ -1,20 +1,30 @@
 import {Pipe, PipeTransform} from '@angular/core';
 
+const escapeRegExp = (str: string) => str.replace(/[-[\]/{}()*+?.\\^$|]/g, '\\$&');
+
 @Pipe({
-  name: 'replaceWithBold',
-  pure: true
+  name: 'o3rReplaceWithBold',
+  standalone: true
 })
-export class ReplaceWithBoldPipe implements PipeTransform {
+export class O3rReplaceWithBoldPipe implements PipeTransform {
   public transform(value: string, inputText: string): string {
     if (inputText) {
-      const regexp = new RegExp(this.escapeRegExp(inputText.trim()), 'gi');
+      const regexp = new RegExp(escapeRegExp(inputText.trim()), 'gi');
       return value.replace(regexp, (match) => `<strong>${match}</strong>`);
     } else {
       return value;
     }
   }
+}
 
+/**
+ * @deprecated please use O3rReplaceWithBoldPipe, will be removed in v12.
+ */
+@Pipe({
+  name: 'replaceWithBold'
+})
+export class ReplaceWithBoldPipe extends O3rReplaceWithBoldPipe implements PipeTransform {
   public escapeRegExp(str: string) {
-    return str.replace(/[-[\]/{}()*+?.\\^$|]/g, '\\$&');
+    return escapeRegExp(str);
   }
 }
--- a/packages/@o3r/schematics/src/utility/update-pipes.ts
+++ b/packages/@o3r/schematics/src/utility/update-pipes.ts
@@ -1,0 +1,116 @@
+import { Rule, SchematicContext, Tree } from '@angular-devkit/schematics';
+import { basename, dirname, join } from 'node:path';
+import * as ts from 'typescript';
+import { DecoratorWithArg, isDecoratorWithArg } from './ast';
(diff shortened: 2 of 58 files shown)


kpanot:
I think it is a too important breaking change to not provide `ng-update`.

For its implementation, I think that just checking the import of the different Otter modules in the module of the component to update is enough to consider the renaming of the otter's pipes in the template.

packages/@o3r/schematics/src/utility/update-pipes.ts
```diff
@@ -0,0 +1,119 @@
...
+    );
+    if (importsProp) {
+      matchers.forEach((matcher) => {
+        const pipeName = matcher[1];
+        const importModuleRegexp = new RegExp(`\\b${pipeReplacementInfo[pipeName].import}\\b`, 'g');
+        if (
+          importsProp.initializer.elements.find((element) =>
+            element.getText().match(importModuleRegexp)
+          )
+        ) {
```
fpaul-1A, line 35:
```suggestion
        if (
          importsProp.initializer.elements.some((element) =>
            importModuleRegexp.test(element.getText())
          )
        ) {
```

packages/@o3r/schematics/src/utility/update-pipes.ts
```diff
@@ -0,0 +1,115 @@
...
+    ).filter((matcher) => !!pipeReplacementInfo[matcher[1]]);
+    if (matchers.length) {
+      const directory = dirname(file.path);
+      const baseFileName = basename(file.path, '.template.html');
+      const componentFile = join(directory, `${baseFileName}.component.ts`);
+      const moduleFile = join(directory, `${baseFileName}.module.ts`);
+      let standalone = false;
+      if (tree.exists(componentFile)) {
+        try {
+          const info = getO3rComponentInfoOrThrowIfNotFound(tree, componentFile);
```
fpaul-1A, line 78: it would be better to avoid unnecessary try/catch if possible
maybe we can create the same function that doesn't throw

fpaul-1A (dismissed):

packages/@o3r/schematics/src/utility/update-pipes.ts
```diff
@@ -0,0 +1,115 @@
...
+          standalone = info.standalone;
+        } catch {}
+      }
```
kpanot, line 81: can be valuable to add a `else` case with a info message that we have detected the usage of pipes in this template but, because we didn't find the module, the update will be skipped.
Something like:
```typescript
logger.info(`Pipe renaming will be skipped for "${file.path}" because the associated module has not been detected.`);
```

packages/@o3r/schematics/src/utility/update-pipes.ts
```diff
@@ -0,0 +1,117 @@
...
+          }
+          tree.overwrite(
+            templateFile,
+            tree.readText(templateFile).replaceAll(new RegExp(`\\|\\s*${pipeName}`, 'g'), `| ${pipeReplacementInfo[pipeName].new.name}`)
+          );
+        } else {
+          context.logger.warn(
+            `Deprecated pipe ${pipeName} was found in ${templateFile} `
+            + `but no associated import in ${fileWithImports}`
+            + `to be able to migrate to the new pipe ${pipeReplacementInfo[pipeName].new.name}.${EOL}`
```
kpanot, line 47: Not sure `${EOL}` is needed in the end of the message, it will already be handled by `logger.warm`

packages/@o3r/components/src/tools/pipes/replace-with-bold/replace-with-bold.pipe.ts
```diff
@@ -1,20 +1,26 @@
...
     if (inputText) {
-      const regexp = new RegExp(this.escapeRegExp(inputText.trim()), 'gi');
+      const regexp = new RegExp(escapeRegExp(inputText.trim()), 'gi');
       return value.replace(regexp, (match) => `<strong>${match}</strong>`);
     } else {
       return value;
     }
   }
-
-  public escapeRegExp(str: string) {
```
fpaul-1A, line 17: this is theoretically a breaking change

packages/@o3r/schematics/src/utility/update-pipes.ts
fpaul-1A: not a big deal if this is only used for ng-updates

fpaul-1A (approved):

kpanot (approved):

PR #1285, opened 2024-01-26
fix(metric): avoid awaiting the metric report call

### LINKED ISSUE #1201: [Feature]: Otter cli metrics
### Context

Based on the same idea than https://angular.io/cli/analytics
As an Otter developer, I would be able to collect data to see what are the generators/builders used by our users, how they are used, if they encounter some issues, errors etc so I can provide a better user experience (bugfixes) and prioritize new features for the future as well.

### Proposal

A wrapper method to create a generator/builder that will retrieve all needed data and send those to a server.

- [x] Wrapper methods https://github.com/AmadeusITGroup/otter/pull/1183
- [x] Service to call the server
   - [ ] Do not throw error if the call fail

kpanot:
## Proposed change

Avoid awaiting the metric report call

## Related issues

- :bug: Fixes #1201

### DIFF
--- a/packages/@o3r/telemetry/src/builders/index.ts
+++ b/packages/@o3r/telemetry/src/builders/index.ts
@@ -4,12 +4,10 @@ import type {
 } from '@angular-devkit/architect';
 import { getEnvironmentInfo } from '../environment/index';
 import { performance } from 'node:perf_hooks';
-import { BuilderMetricData, sendData as defaultSendData } from '../sender';
+import { BuilderMetricData, sendData as defaultSendData, type SendDataFn } from '../sender';
 
 type BuilderWrapperFn<S, O extends BuilderOutput = BuilderOutput> =
   (opts: S, ctx: BuilderContext) => O | Promise<O>;
-
-type SendDataFn = (data: BuilderMetricData, logger?: { error: (msg: string) => void } | undefined) => Promise<void>;
 
 /**
  * Type of a function that wraps a builder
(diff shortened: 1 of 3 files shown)


packages/@o3r/telemetry/src/builders/index.ts
```diff
@@ -35,38 +34,36 @@ export const createBuilderWithMetrics: BuilderWrapper = (builderFn, sendData = d
...
+                name: context.target.target,
+                projectName: context.target.project,
+                configuration: context.target.configuration
+              }
+            } : {}
+          )
+        },
+        error
+      };
+      context.logger.debug(JSON.stringify(data, null, 2));
```
fpaul-1A, line 61: we should probably remove this

packages/@o3r/telemetry/src/schematics/index.ts
```diff
@@ -31,28 +33,26 @@ export const createSchematicWithMetrics: SchematicWrapper =
...
+        options,
+        interactive: context.interactive
+      };
+      const data: SchematicMetricData = {
+        environment,
+        schematic,
+        duration,
+        error
+      };
+      context.logger.debug(JSON.stringify(data, null, 2));
```
fpaul-1A, line 50: we should probably remove this (or add a message)

fpaul-1A (approved):

packages/@o3r/telemetry/src/builders/index.ts
matthieu-crouzet: It was more for us when we run the command with `--verbose`

PR #1276, opened 2024-01-25
test: rework unit tests to separate dom and node tests

fpaul-1A:
## Proposed change

Currently, we only have 1 jest config for all unit-tests
Unfortunately, the tests for `src` usually don't require the same environment as the tests for `schematics` and `builders`

The idea is to have a different project for each (check `packages/@o3r/core/jest.config.ts`)

This will fix the tests issue on https://github.com/AmadeusITGroup/otter/pull/1255 and https://github.com/AmadeusITGroup/otter/pull/1243

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

PR #1291, opened 2024-01-29
fix: increase test time limit for schematics

kpanot:


packages/@ama-sdk/schematics/schematics/typescript/shell/index.spec.ts
```diff
@@ -78,7 +78,7 @@ describe('Typescript Shell Generator', () => {
       skipInstall: true,
       packageManager: 'npm'
     }, Tree.empty());
-  }, 15000);
+  }, 25000);
```
fpaul-1A, line 81: In my PR about reworking the unit tests, I've increased the global testTimeout to 30s https://github.com/AmadeusITGroup/otter/pull/1276/files#diff-cacec94ff84332c24eff579dc017461bb1cd2caf48ef3a408a70ecbcff09738eR74

fpaul-1A (approved):

packages/@ama-sdk/schematics/schematics/typescript/shell/index.spec.ts
kpanot: then i flag it duplicate of #1276

PR #1301, opened 2024-01-31
fix(c11n): change the return type of the operator getpresenter

matthieu-crouzet:
## Proposed change
The return type of `getPresenter` was incorrect

### DIFF
--- a/packages/@o3r/components/schematics/ng-update/v10-0/index.spec.ts
+++ b/packages/@o3r/components/schematics/ng-update/v10-0/index.spec.ts
@@ -7,16 +7,19 @@ const migrationPath = path.join(__dirname, '..', '..', '..', 'migration.json');
 
 
 describe('Update v10', () => {
+  let initialTree: Tree;
+  let runner: SchematicTestRunner;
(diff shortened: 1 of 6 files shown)


kpanot (dismissed):

fpaul-1A (approved):

packages/@o3r/components/schematics/ng-update/v10-0/index.spec.ts
```diff
@@ -74,4 +77,29 @@ describe('Update v10', () => {
...
+        import { Observable } from 'rxjs';
+
+        @O3rComponent({ componentType: 'Component' })
+        @Component({
+          selector: 'o3r-example',
+          standalone: true,
+          templateUrl: './example.template.html'
+        })
+        export class ExampleComponent {
+          public presenter$!: Observable<MyComp>;
```
cpaulve-1A, line 96: could be nice to add another example that should not be impacted by the ng-update

cpaulve-1A (approved):

fpaul-1A (approved):

PR #1306, opened 2024-01-31
chore: reverting removed lint dependency on build in nx.json

divdavem:
## Proposed change

Commit eb8aec15798664fa475bbe13bc574d1456b3ed30 broke the `yarn nx run ama-sdk-showcase-sdk:lint` command
(when run from an empty cache).
This PR reverts the specific change that caused the issue.
The issue was hidden until now in CI because of the reused cache.

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -98,7 +98,8 @@
       },
       "executor": "@nx/eslint:lint",
       "inputs": ["lint"],
-      "cache": true
+      "cache": true,
+      "dependsOn": ["^build"]
     },
     "test": {
       "inputs": [


fpaul-1A (approved):

matthieu-crouzet (approved):

divdavem:
@fpaul-1A @matthieu-crouzet Thank you for your review

kpanot (approved):

nx.json
```diff
@@ -98,7 +98,8 @@
       },
       "executor": "@nx/eslint:lint",
       "inputs": ["lint"],
-      "cache": true
+      "cache": true,
+      "dependsOn": ["^build"]
```
kpanot, line 102: Not sure this is the correct to requesting the build of the dependencies.
It seems to be a typing issue, the fact to build de depending package may do the job but does not seems to be the correct fix for it.
The typescript path mapping can be the cause of the issue you are facing

divdavem: Thank you for your comment. I had opened this PR to restore the previous working behavior but indeed, it is better to have less dependencies for the lint task and I have opened #1309 to directly address the problem of types.

divdavem:
I am closing this PR because it is superseded by #1309

PR #1309, opened 2024-02-01
fix: missing types when linting showcase-sdk

divdavem:
## Proposed change

When linting showcase-sdk, if the build of dependent packages was not done (which is the case in CI), types of dependent packages are interpreted by typescript as `any` because the tsconfig in showcase-sdk overrides paths defined in the root of the repository and, as a consequence, typescript does not find the source files defining the types.

## Related issues

This PR supersedes #1306

### DIFF
--- a/packages/@ama-sdk/showcase-sdk/tsconfig.build.json
+++ b/packages/@ama-sdk/showcase-sdk/tsconfig.build.json
@@ -24,24 +24,6 @@
     "composite": true,
     "declaration": true,
     "declarationMap": false,
-    "rootDir": "./src",
-    "baseUrl": ".",
-    "paths": {
-      "@ama-sdk/showcase-sdk/helpers/*": [
-        "./src/helpers/*"
-      ],
-      "@ama-sdk/showcase-sdk/helpers": [
-        "./src/helpers/index"
-      ],
-      "@ama-sdk/showcase-sdk": [
-        "./src/index"
-      ],
-      "@ama-sdk/showcase-sdk/*": [
-        "./src/*"
-      ],
-      "@ama-sdk/showcase-sdk/spec": [
-        "./src/spec/index"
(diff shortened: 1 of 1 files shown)


packages/@ama-sdk/showcase-sdk/tsconfig.build.json
```diff
@@ -24,24 +24,6 @@
     "composite": true,
     "declaration": true,
     "declarationMap": false,
-    "rootDir": "./src",
-    "baseUrl": ".",
```
fpaul-1A, line 28: should we remove it from the generators as well ? https://github.com/AmadeusITGroup/otter/blob/main/packages/%40ama-sdk/schematics/schematics/typescript/shell/templates/base/tsconfig.build.json#L26

fpaul-1A (approved):

packages/@ama-sdk/showcase-sdk/tsconfig.build.json
divdavem: That's a good question.
What was the purpose of having this in the generators? Do people use those paths?
In the generators, by default, they do not cause any issue in the same way as they caused an issue for showcase-sdk, because the generated tsconfig does not extend anything (so it does not override any other paths defined somewhere else).
But people can easily have this problem in case they extend another tsconfig.json and want to define paths there.

cpaulve-1A (approved):

packages/@ama-sdk/showcase-sdk/tsconfig.build.json
divdavem: I have added a commit to remove paths from the tsconfig in the generator.

fpaul-1A (approved):

kpanot (approved):

PR #1310, opened 2024-02-01
chore: upgrade angular

fpaul-1A:
## Proposed change

Investigating why the build fails when upgrading angular to 17.1
I've noticed that the build succeeds when the yarn cache is not existing prior to the run
It fails when re-using the yarn cache

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


kpanot (approved):

apps/chrome-devtools/project.json
```diff
@@ -21,8 +21,8 @@
         "^build",
         "build-panel",
         "build-components",
-        "build-extension",
-        "copy-assets"
+        "copy-assets",
```
kpanot, line 24: I hope this change does not have effect.
If you want to apply an order, you should me one depending of the other

matthieu-crouzet (approved):

PR #1316, opened 2024-02-05
chore: upgrade nx 18

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


packages/@o3r/create/.eslintrc.js
```diff
@@ -16,14 +16,7 @@ module.exports = {
     {
```
kpanot, line 16: So the whole `override` section can be removed?

package.json
```diff
@@ -230,7 +230,7 @@
     "ng-packagr": "~17.0.0",
     "ngx-highlightjs": "^10.0.0",
     "npm-run-all2": "^6.0.0",
-    "nx": "~17.3.0",
+    "nx": "~18.0.1",
```
kpanot, line 233: Should be set to the resolution of `@angular-eslint/builder/nx` as well

kpanot (approved):

kpanot (approved):

Issue #1319, opened 2024-02-06
[Feature]: Add the possibility to specify action types to invoke the reducer in the storage sync helper


Abhirocks889:
### Context

The requirement is for syncing stores in feature modules instead of the root app module. This requires certain additional action types to be taken into account for invoking the reducer.

### Proposal

The idea is to add a new optional property ** actionTypes** to the **SyncStorageConfig** that will specify the action types for which the reducer should be invoked in case the state is undefined. This is required for syncing feature stores for which the action type **UPDATE** needs to be taken into account.

kpanot:
Can you explain in which case you have a state that is `undefined` after the store initialization?

PR #1320, opened 2024-02-06
feat: Add the possibility to specify action types to invoke the reducer in the storage sync helper

Abhirocks889:
## Proposed change
This PR adds a new optional property **actionTypes** to the **SyncStorageConfig** that will specify the action types for which the reducer should be invoked in case the state is undefined. This is required for syncing feature stores for which the action type UPDATE needs to be taken into account.

## Related issues

- :rocket: Feature [#(1319)](https://github.com/AmadeusITGroup/otter/issues/1319)


packages/@o3r/store-sync/src/sync-storage/interfaces.ts
```diff
@@ -61,4 +61,6 @@ export interface SyncStorageConfig {
   logger?: Logger;
   /** Post process after sync storage execution */
   postProcess?: (state: any) => void;
+  /** List of action types to invoke the reducer */
```
kpanot, line 64: This is not really clear, you should specify what is the reducer you are talking about

packages/@o3r/store-sync/src/sync-storage/storage-sync.ts
```diff
@@ -328,7 +332,7 @@ export const syncStorage = (config: SyncStorageConfig) => (reducer: any) => {
 
     // If state arrives undefined, we need to let it through the supplied reducer
     // in order to get a complete state as defined by user
-    if (action.type === INIT && !state) {
+    if (config.actionTypes?.includes(action.type) && !state) {
```
kpanot, line 335:
```suggestion
    if (config.actionTypes!.includes(action.type) && !state) {
```

Because it can bot be null

Abhirocks889: Done

packages/@o3r/store-sync/src/sync-storage/interfaces.ts
Abhirocks889: Added more description

Abhirocks889:
Essentially, when a store is imported inside a lazy loaded module and the store sync meta reducer is used, the state is undefined as can be seen in the attached screenshot:

Since the action type INIT is specified, the reducer is not invoked resulting in an empty state.

So, for such scenarios we need to specify the UPDATE action type as well for invoking the reducer.

We do not have the issue when the sync is done in the root app module.

image

kpanot:
> Essentially, when a store is imported inside a lazy loaded module and the store sync meta reducer is used, the state is undefined as can be seen in the attached screenshot:
> 
> Since the action type INIT is specified, the reducer is not invoked resulting in an empty state.
> 
> So, for such scenarios we need to specify the UPDATE action type as well for invoking the reducer.
> 
> We do not have the issue when the sync is done in the root app module.
> 
> image

Thank you for the explanation.
So would not make more sense to run the reducer on `init` or `change` if `init` has not be trigger for the store previously?

Abhirocks889:
> > Essentially, when a store is imported inside a lazy loaded module and the store sync meta reducer is used, the state is undefined as can be seen in the attached screenshot:
> > Since the action type INIT is specified, the reducer is not invoked resulting in an empty state.
> > So, for such scenarios we need to specify the UPDATE action type as well for invoking the reducer.
> > We do not have the issue when the sync is done in the root app module.
> > image
> 
> Thank you for the explanation. So would not make more sense to run the reducer on `init` or `change` if `init` has not be trigger for the store previously?

Indeed, for now it is INIT and UPDATE for which we need to invoke the reducer. However, there could possibly be more in the future. Hence, I thought of keeping it configurable to specify the action types from the application level instead of repeatedly making changes here whenever a new action is identified.

kpanot:
> > > Essentially, when a store is imported inside a lazy loaded module and the store sync meta reducer is used, the state is undefined as can be seen in the attached screenshot:
> > > Since the action type INIT is specified, the reducer is not invoked resulting in an empty state.
> > > So, for such scenarios we need to specify the UPDATE action type as well for invoking the reducer.
> > > We do not have the issue when the sync is done in the root app module.
> > > image
> > 
> > 
> > Thank you for the explanation. So would not make more sense to run the reducer on `init` or `change` if `init` has not be trigger for the store previously?
> 
> Indeed, for now it is INIT and UPDATE for which we need to invoke the reducer. However, there could possibly be more in the future. Hence, I thought of keeping it configurable to specify the action types from the application level instead of repeatedly making changes here whenever a new action is identified.

I don't really see cases where you will need more than these events.
I would suggest to do the suggested change (fallback to `change` event of no `init` has been called) instead of exposing this mechanism

Abhirocks889:
> > > > Essentially, when a store is imported inside a lazy loaded module and the store sync meta reducer is used, the state is undefined as can be seen in the attached screenshot:
> > > > Since the action type INIT is specified, the reducer is not invoked resulting in an empty state.
> > > > So, for such scenarios we need to specify the UPDATE action type as well for invoking the reducer.
> > > > We do not have the issue when the sync is done in the root app module.
> > > > image
> > > 
> > > 
> > > Thank you for the explanation. So would not make more sense to run the reducer on `init` or `change` if `init` has not be trigger for the store previously?
> > 
> > 
> > Indeed, for now it is INIT and UPDATE for which we need to invoke the reducer. However, there could possibly be more in the future. Hence, I thought of keeping it configurable to specify the action types from the application level instead of repeatedly making changes here whenever a new action is identified.
> 
> I don't really see cases where you will need more than these events. I would suggest to do the suggested change (fallback to `change` event of no `init` has been called) instead of exposing this mechanism

Well, I found one where the store dev tools recompute action is accompanied by an undefined state:

image

packages/@o3r/store-sync/src/sync-storage/storage-sync.ts
```diff
@@ -328,7 +331,7 @@ export const syncStorage = (config: SyncStorageConfig) => (reducer: any) => {
 
     // If state arrives undefined, we need to let it through the supplied reducer
     // in order to get a complete state as defined by user
-    if (action.type === INIT && !state) {
+    if ((action.type === INIT || action.type === UPDATE || action.type === RECOMPUTE) && !state) {
```
kpanot, line 334: `UPDATE` and `RECOMPUTE` should trigger the reducer only if no `INIT` has been emitted before for this store, isn't it?

matthieu-crouzet: I'm not sure if I well understood
I think he need to call the reducer to have the initialState of a lazyloaded store 
or maybe the change should be done a bit above in this file [L269](https://github.com/AmadeusITGroup/otter/blob/main/packages/%40o3r/store-sync/src/sync-storage/storage-sync.ts#L269)

packages/@o3r/store-sync/src/sync-storage/storage-sync.ts
```diff
@@ -3,6 +3,9 @@ import { deepFill } from '@o3r/core';
 import type { Logger } from '@o3r/core';
 import type { StorageKeyConfiguration, StorageKeys, SyncStorageConfig, SyncStorageSyncOptions } from './interfaces';
 
+/** The recompute action type of ngrx store devtools */
+const RECOMPUTE = '@ngrx/store-devtools/recompute';
```
matthieu-crouzet, line 7: you could use `RECOMPUTE` from `@ngrx/store-devtools` instead of redefining the variable

Abhirocks889: Would have to add dependency of that package on `o3r-store-sync`. Also, this adds the entire dev tools chunk to the main bundle if we use this method for syncing stores in the app module.

Abhirocks889:
Also, @kpanot , @matthieu-crouzet I have created a repo for minimal reproduction of the issue:
https://github.com/Abhirocks889/store-sync-error
You can run the app to reproduce the same issue. Check the readme for more details

packages/@o3r/store-sync/src/sync-storage/storage-sync.ts
kpanot: A compromise can be to import the type only (to ensure alignment):

```typescript
import type { RECOMPUTE } from '@ngrx/store-devtools';

const RECOMPUTE: typeof RECOMPUTE = '@ngrx/store-devtools/recompute';
```

packages/@o3r/store-sync/src/sync-storage/storage-sync.ts
kpanot: *Turned this comment as unresolved, please do not resolve a question without the question author approval*

If we agree on the fact that the `reducer` should be called on `RECOMPUTE` and `INIT` actions *only* if the `INIT` event has not be emitted for this store, this is not the implementation on the code.
So:
- either you should consider checking that the `INIT` action has not been emitted to this store before checking for the 2 others actions,
- either it s a case it never happen (then an explanation as response of the question would be appreciated), 
- either executing the `reducer` a second time on a store that has already got an `INIT` action is not an issue (an explanation would also be needed).

packages/@o3r/store-sync/src/sync-storage/storage-sync.ts
Abhirocks889: Updated

packages/@o3r/store-sync/src/sync-storage/storage-sync.ts
Abhirocks889: As mentioned above, when syncing the state using the `syncStorage` method within a lazily loaded feature module, the `INIT` action is not received and therefore, the reducer is not invoked resulting in an undefined state.

So, point no 2 from your comment. `INIT` is never called.

Please check the minimal reproduction of the same: https://github.com/Abhirocks889/store-sync-error

packages/@o3r/store-sync/src/sync-storage/storage-sync.ts
```diff
@@ -328,7 +332,7 @@ export const syncStorage = (config: SyncStorageConfig) => (reducer: any) => {
 
     // If state arrives undefined, we need to let it through the supplied reducer
     // in order to get a complete state as defined by user
-    if (action.type === INIT && !state) {
```
matthieu-crouzet, line 331: Maybe we should check `syncForFeature` when `action.type` is not `INIT`

I don't understand why we need `RECOMPUTE`
because `UPDATE` will be called when loading the lazy loaded store, and it should be enough no ?
`RECOMPUTE` seems to be for devtools so in production mode it will not be called no ?

packages/@o3r/mobile/package.json
```diff
@@ -39,6 +39,7 @@
     "@capacitor/device": "^5.0.0",
     "@capacitor/preferences": "^5.0.0",
     "@ngrx/store": "~16.3.0",
+    "@ngrx/store-devtools": "~16.3.0",
```
matthieu-crouzet, line 42: it has to be optional if it's only for typing

packages/@o3r/store-sync/package.json
```diff
@@ -23,6 +23,7 @@
     "@angular-devkit/schematics": "~16.2.0",
     "@ngrx/entity": "~16.3.0",
     "@ngrx/store": "~16.3.0",
+    "@ngrx/store-devtools": "~16.3.0",
```
matthieu-crouzet, line 26: it has to be optional peer if it's only for typing

Abhirocks889: Noted

packages/@o3r/store-sync/src/sync-storage/storage-sync.ts
Abhirocks889: Added the check. Also, for the second question, we need  `RECOMPUTE` because in dev mode, we get an undefined state for both `UPDATE` and `RECOMPUTE` action types even if the reducer is invoked for `UPDATE`, the state arrives as undefined for `RECOMPUTE`. Hence, the app breaks and is not usable.

Please check the reproduction: https://github.com/Abhirocks889/store-sync-error

matthieu-crouzet:
It seems that your issue could be fixed by using `StoreSync` from `@o3r/store-sync` 
https://github.com/Abhirocks889/store-sync-error/pull/1

kpanot:
@Abhirocks889 , is this PR still required or did the fix mentioned by @matthieu-crouzet fix your issue?

PR #1332, opened 2024-02-09
feat: regen sdk

matthieu-crouzet:
## Proposed change
Regenerate SDK


apps/showcase/tsconfig.spec.json
```diff
@@ -2,7 +2,8 @@
   "extends": "../../tsconfig.jest",
   "compilerOptions": {
     "outDir": "test",
-    "rootDir": "."
+    "rootDir": ".",
+    "noUnusedParameters": false
```
fpaul-1A, line 6:
```suggestion
    "rootDir": "."
```

fpaul-1A, line 6: should this be removed ?

matthieu-crouzet: Added on purpose to have the test running else it raise the error that "dictionnaries" is not used in the reviver
I added this rule only for test I think it's okay

kpanot (approved):

sdo-1A (approved):

fpaul-1A (approved):

PR #1343, opened 2024-02-14
fix(create): fail and show error if sub-process exits with non-zero code

divdavem:
## Proposed change

This PR fixes the following problem: in case a sub process started by `npm create @ama-sdk` or `npm create @o3r` fails (i.e. exiting with a non-zero code, for example, if the spec file given to the sdk generator does not exist, or if there is a dependency conflict when doing `npm install` as it was the case recently), the `npm create` command itself was hiding the error and exiting with 0.


cpaulve-1A (approved):

kpanot (dismissed):

divdavem:
@fpaul-1A Thank you for your review!

matthieu-crouzet (approved):

divdavem:
@matthieu-crouzet Thank you for your review!

vscaiceanu-1a (approved):

divdavem:
@vscaiceanu-1a  Thank you for your review!

PR #1344, opened 2024-02-14
fix(sdk-core): microgateway token generation

kpanot:
## Proposed change

Fix the way the microgateway token is calculated

### DIFF
--- a/packages/@ama-sdk/core/src/plugins/mgw-mdw-auth/mgw-mdw-auth.request.ts
+++ b/packages/@ama-sdk/core/src/plugins/mgw-mdw-auth/mgw-mdw-auth.request.ts
@@ -1,25 +1,21 @@
 import { v4 } from 'uuid';
-import { ab2str, str2ab } from '../../utils/crypto';
-import { createBase64Encoder, createBase64UrlEncoder } from '../../utils/json-token';
+import { base64EncodeUrl, createBase64Encoder, createBase64UrlEncoder } from '../../utils/json-token';
 import { PluginRunner, RequestOptions, RequestPlugin } from '../core';
+import type { createHmac as createHmacType, webcrypto } from 'node:crypto';
+
 
 /**
  * Computes the SHA256 digest of the given string
  * @param value Value to hash
  */
 export async function sha256(value: string) {
-  // Web browsers
-  if (typeof window !== 'undefined' && typeof window.crypto !== 'undefined') {
-    return ab2str(await crypto.subtle.digest('SHA-256', str2ab(value)));
-  }
-  // NodeJS
-  try {
-    return ab2str(require('node:crypto').createHash('sha256')
-      .update(value)
-      .digest('latin1'));
-  } catch (err) {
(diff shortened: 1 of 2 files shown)


packages/@ama-sdk/core/src/plugins/mgw-mdw-auth/mgw-mdw-auth.request.ts
```diff
@@ -1,22 +1,45 @@
...
 
 /**
  * Computes the SHA256 digest of the given string
  * @param value Value to hash
  */
 export async function sha256(value: string) {
-  // Web browsers
   if (typeof window !== 'undefined' && typeof window.crypto !== 'undefined') {
-    return ab2str(await crypto.subtle.digest('SHA-256', str2ab(value)));
+    return sha256Browser(value);
```
divdavem, line 38: Depending on which node.js versions you target, maybe you do not really need two different versions for node.js and the browser. It looks like the browser version also works in node.js (v20.11.0).

packages/@ama-sdk/core/src/plugins/mgw-mdw-auth/mgw-mdw-auth.request.ts
```diff
@@ -1,22 +1,45 @@
...
+}
+
+/**
+ *
+ * @param value
+ */
+export function sha256Node(value: string) {
+  const { createHash }: { createHash: typeof createHashType } = require('node:crypto');
+  return createHash('sha256')
+    .update(value, 'latin1')
```
divdavem, line 28: This does not match the browser version that uses utf8.

packages/@ama-sdk/core/src/plugins/mgw-mdw-auth/mgw-mdw-auth.request.ts
kpanot: indeed, seems the case from Node 15 (cf. https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest)

divdavem:
Actually, is this plugin intended to run in a browser environment?
I don't see how it can be useful without using the `createHmac` function from node:crypto.
I don't think it would be hard to make it work in a browser as well, though with the [crypto.subtle.sign function](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign#hmac).

kpanot:
> Actually, is this plugin intended to run in a browser environment? I don't see how it can be useful without using the `createHmac` function from node:crypto. However, I don't think it would be hard to make it work in a browser as well, with the [crypto.subtle.sign function](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign#hmac).

No, it is not (their is a throw in the constructor if not used on not env) but the function `sha256` is exported and its previous implementation supported (theorically) both Browser and Node envs.
(I said "theorically" because it was actually not working correcly since few node versions)

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #1346, opened 2024-02-15
fix: playwright fixtures issues

cpaulve-1A:
## Proposed change

Resolve the element in throwOnUndefined
Remove user simulated behaviour and replace it with playwright's standard methods

## Related issues

- :bug: throwOnUndefined does not resolve with element
- :bug:  setValue inconsistent behaviour

### DIFF
--- a/packages/@o3r/testing/src/core/playwright/component-fixture.ts
+++ b/packages/@o3r/testing/src/core/playwright/component-fixture.ts
@@ -36,10 +36,7 @@ export class O3rComponentFixture<V extends O3rElement = O3rElement> implements C
     if (!element) {
       throw new Error('Element not found in ' + this.constructor.name);
     }
-    const count = await withTimeout(element.sourceElement.element.count(), timeout);
-    if (!count) {
-      throw new Error('Element not found in ' + this.constructor.name);
-    }
+    await element.sourceElement.element.first().waitFor({state: 'attached', timeout});
     return element;
   }
 
@@ -49,16 +46,13 @@ export class O3rComponentFixture<V extends O3rElement = O3rElement> implements C
    * @param element ElementProfile to test
    * @param timeout specific timeout that will throw when reach
    */
-  protected async throwOnUndefined<T extends O3rElement>(element: Promise<T | undefined>, timeout?: number): Promise<T> {
-    return withTimeout(element, timeout)
-      .then((el) => el?.sourceElement.element.count())
-      .then((count) => (count || 0) > 0)
-      .then((isPresent) => {
-        if (!isPresent) {
-          throw new Error('Element not found in ' + this.constructor.name);
-        }
-      })
-      .then(() => element as Promise<T>);
+  protected async throwOnUndefined<T extends O3rElement>(element: Promise<T>, timeout?: number): Promise<T> {
+    return withTimeout(
+      element.then(async (el) => {
+        await el.sourceElement.element.first().waitFor({state: 'attached'});
+        return el;
+      }),
+      timeout);
   }
 
   /**
@@ -78,8 +72,8 @@ export class O3rComponentFixture<V extends O3rElement = O3rElement> implements C
       shouldThrowIfNotPresent?: boolean;
       timeout?: number;
     } = {}
-  ): Promise<O3rElement | undefined> {
-    let element: O3rElement | undefined;
+  ): Promise<O3rElement> {
+    let element: O3rElement;
     if (options.index !== undefined) {
       element = await this.queryNth(selector, options.index, elementConstructor as any);
     } else {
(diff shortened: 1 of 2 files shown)


packages/@o3r/testing/src/core/playwright/component-fixture.ts
```diff
@@ -50,15 +50,12 @@ export class O3rComponentFixture<V extends O3rElement = O3rElement> implements C
...
-      .then((isPresent) => {
-        if (!isPresent) {
-          throw new Error('Element not found in ' + this.constructor.name);
-        }
-      })
-      .then(() => element as Promise<T>);
+    return withTimeout(
+      element.then(async (el) => {
+        await el?.sourceElement.element.first().waitFor({state: 'attached'});
+        return el!;
```
kpanot, line 56: But if `el == undefined` you will return `el!`.
It sounds like a lie :D

cpaulve-1A: In a playwright context when will you ever have el undefined if you do not pass it explicitely with a Promise.resolve(undefined)?
Locator always are defined, you create it from a query which has no chance to fail, the typing is only there for compatibility with the other frameworks.

The actual throwOnUndefined is handled line 55 where the timeout will throw if the element is not present in the page.

But yeah I will fix it to stay consistent with the signature

fpaul-1A (approved):

packages/@o3r/testing/src/core/playwright/component-fixture.ts
cpaulve-1A: As discussed, there is no impact in making the signature more restrictive and will go for this solution

fpaul-1A (approved):

packages/@o3r/testing/src/core/playwright/element.ts
```diff
@@ -68,26 +68,14 @@ export class O3rElement implements ElementProfile {
...
-    // Clears the value in the field
-    await this.clearValue();
-
     await this.sourceElement.element.fill(input);
 
     await this.sourceElement.element.press('Tab');
   }
 
   /** @inheritdoc */
   public async clearValue() {
```
pginoux-1A, line 77: Wasn't it working before ? Does the simulation of a user deleting character by character had an utility compare to the fill with blank space ?

fpaul-1A (approved):

packages/@o3r/testing/src/core/playwright/element.ts
cpaulve-1A: It is now pretty buggy, there seems to be some timing issues and if I deactivate the headless mode it is really obvious the behaviour is not quite deterministic.

packages/@o3r/testing/src/core/playwright/component-fixture.ts
```diff
@@ -49,16 +46,13 @@ export class O3rComponentFixture<V extends O3rElement = O3rElement> implements C
...
-          throw new Error('Element not found in ' + this.constructor.name);
-        }
-      })
-      .then(() => element as Promise<T>);
+  protected async throwOnUndefined<T extends O3rElement>(element: Promise<T>, timeout?: number): Promise<T> {
+    return withTimeout(
+      element.then(async (el) => {
+        await el.sourceElement.element.first().waitFor({state: 'attached'});
+        return el;
+      }),
```
kpanot, line 54:
```suggestion
      (async () => {
        const el = await element;
        await el.sourceElement.element.first().waitFor({state: 'attached'});
        return el;
      })(),
```

packages/@o3r/testing/src/core/playwright/component-fixture.ts
```diff
@@ -148,38 +142,29 @@ export class O3rComponentFixture<V extends O3rElement = O3rElement> implements C
...
-    try {
-      const elements = this.rootElement.sourceElement.element.locator(selector);
-      const element = elements.first();
-      const selectedElement: PlaywrightSourceElement = {element: element, page: this.rootElement.sourceElement.page};
-      return Promise.resolve(new (returnType || O3rElement)(selectedElement));
-    } catch (err) {
-      console.warn(`Failed to query ${selector}`, err);
-      return Promise.resolve(undefined);
-    }
+  public async query(selector: string, returnType?: undefined): Promise<O3rElement>;
```
kpanot, line 151: is it OK for the linter to have a `async` function without `await`?

packages/@o3r/testing/src/core/playwright/component-fixture.ts
```diff
@@ -148,38 +142,29 @@ export class O3rComponentFixture<V extends O3rElement = O3rElement> implements C
...
+  public async query<T extends O3rElement>(selector: string, returnType: O3rElementConstructor<T>): Promise<T>;
+  public async query<T extends O3rElement>(selector: string, returnType: O3rElementConstructor<T> | undefined): Promise<T | O3rElement> {
+    const elements = this.rootElement.sourceElement.element.locator(selector);
+    const element = elements.first();
+    const selectedElement: PlaywrightSourceElement = {element: element, page: this.rootElement.sourceElement.page};
+    return Promise.resolve(new (returnType || O3rElement)(selectedElement));
```
kpanot, line 157: in async function this is not required:

```suggestion
     return new (returnType || O3rElement)(selectedElement);
```

(same below)

PR #1350, opened 2024-02-15
fix: ng updates for v10

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


vscaiceanu-1a (dismissed):

vscaiceanu-1a (approved):

.verdaccio/conf/config.yaml
```diff
@@ -29,14 +29,21 @@ packages:
     access: $all
     publish: $authenticated
     proxy: no-proxy
+#    proxy: npmjs
```
matthieu-crouzet, line 32: Do we want to keep these comments ?

matthieu-crouzet (approved):

.verdaccio/conf/config.yaml
fpaul-1A: I figured it might be helpful since I needed that to run the ng-updates

PR #1355, opened 2024-02-16
feat(design): allow to flatten the design token when extracting

kpanot:
## Proposed change

The purpose of this feature is to allow to extract design token in a flatten way.

For example it can mow be extracted as :
```json
{
  "myComponent": {
    "item": {
      "color": {
        "$type": "color"
        "$value": "#ffffff"
      }
    }
  }
}
```

Or :
```json
{
  "myComponent": {
    "item.color": {
      "$type": "color"
      "$value": "#ffffff"
    }
  }
}
```


packages/@o3r/design/schematics/extract-token/index.spec.ts
```diff
@@ -66,6 +66,67 @@ $breadcrumb-pres-item-other-color: o3r.variable('breadcrumb-pres-item-other-colo
...
+        public extractFileContent = jest.fn<CssVariable[], any>().mockReturnValue([
+          {
+            defaultValue: '3rem',
+            name: 'breadcrumb-pres-item-icon-size',
+            label: 'breadcrumb pres item icon size',
+            type: 'string'
+          },
+          {
+            defaultValue: '#fff',
+            name: 'breadcrumb-pres-item-icon-color',
```
fpaul-1A, line 83: won't there be a problem if you don't have the same length for variables names ?
say `breadcrumb-pres-item-icon-color` and `breadcrumb-pres-item-icon-color-bg` for example

fpaul-1A (approved):

packages/@o3r/design/schematics/extract-token/index.spec.ts
fpaul-1A: I guess I just misunderstood how the flattenLevel was working, I thought it was flattening from right to left for some reason

matthieu-crouzet (approved):

PR #1361, opened 2024-02-19
feat(sdk-schematics): initialize git repository

divdavem:
## Proposed change

Initialize a git repository when using either `npm create @ama-sdk` or the sdk schematics.


packages/@ama-sdk/schematics/schematics/typescript/shell/schema.json
```diff
@@ -28,6 +28,11 @@
       "description": "Skip NPM install",
       "default": false
     },
+    "skipGit": {
```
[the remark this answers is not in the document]
divdavem: Thank you @matthieu-crouzet for your comment! I have added the `commit` option in the shell generator schematics. However, it is currently not used anywhere.

I have some questions:
- Should the `commit` option of the shell generator schematics be `true` by default?
- Should the `npm create @ama-sdk` command create a commit?
- Should the `npm create @ama-sdk` command use the `commit` option of the schematics to create a commit?
- If the answer to the previous question is yes, the commit would only contain the shell files and neither the result of `yarn set version` nor the output of sdk generation... so for those extra changes, we could either create another commit or amend the previous one?

I personally prefer not creating a commit in `npm create` (as it is done in the current state of this PR), but if it is decided to create one, then I think it would be better to create it directly with the right files (calling the shell generator schematics with the `commit` option set to `false`) and creating the commit at the end, rather than amending a commit. Note that if a repository already exists in a parent folder, the `RepositoryInitializerTask` does nothing, so we should be especially careful not to amend a previous commit in this case.

kpanot (dismissed):

packages/@ama-sdk/schematics/schematics/typescript/shell/schema.json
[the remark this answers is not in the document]
matthieu-crouzet: Indeed it should not be part of the `shell` schematic but more of the `create` script at the end with all the files.
It's the behavior that we have by default when run `npm create @o3r` if I'm not mistaken.

packages/@ama-sdk/schematics/schematics/typescript/shell/index.ts
```diff
@@ -108,7 +116,8 @@ function ngGenerateTypescriptSDKFn(options: NgGenerateTypescriptSDKShellSchemati
 
   return chain([
     setupRule,
-    ...(options.skipInstall ? [] : [installRule])
+    ...(options.skipInstall ? [] : [installRule]),
+    ...(options.skipGit ? [] : [initGitRule])
```
matthieu-crouzet, line 120: What happen if we create a sdk as a package in a monorepo ?
Because in this case I think we should not create a git repo as we are not at the root folder of the monorepo.

divdavem: If there is already a git repository, the [RepositoryInitializerTask](https://github.com/angular/angular-cli/blob/main/packages/angular_devkit/schematics/tasks/repo-init/executor.ts#L62) does nothing.

packages/@ama-sdk/schematics/schematics/typescript/shell/schema.json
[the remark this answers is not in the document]
divdavem: `npm create @o3r` creates a commit, but does not include all changes in it.

packages/@ama-sdk/schematics/schematics/typescript/shell/index.ts
matthieu-crouzet: But what happens if there is not (I don't know why but it could happen)

divdavem: If there is no git repository, then a new git repository is created where the sdk is generated, there is no detection of the monorepo. I think this is an acceptable behavior.

packages/@ama-sdk/schematics/schematics/typescript/shell/schema.json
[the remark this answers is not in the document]
matthieu-crouzet: In your opinion should we create an issue to handle it later
I think it could be great if the repo is "clean" when you finish the creation

matthieu-crouzet (dismissed):

packages/@o3r/test-helpers/src/prepare-test-env.ts
```diff
@@ -149,8 +149,10 @@ export async function prepareTestEnv(folderName: string, options?: PrepareTestEn
 
   prepareFinalApp(appDirectory);
 
-  // Setup git and initial commit to easily make checks on the diff inside the tests
-  setupGit(workspacePath);
+  if (type !== 'blank') {
```
kpanot, line 152: Why this change?
Actually this git init is for comparing the code generated post env setup and it would make sense in case of blank env as well :S.

kpanot: I see, so maybe it would make more sens to add an option to the env setup function to explicitly not initialize a git env when it is not required.
The purpose is to still allow the git diff functions in this env for future tests that may not explicitly initialize a new git repository

fpaul-1A (approved):

kpanot (approved):

PR #1366, opened 2024-02-20
feat: align otter versions on ng update

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/core/schematics/ng-update/index.ts
+++ b/packages/@o3r/core/schematics/ng-update/index.ts
@@ -2,7 +2,7 @@
 /* eslint-disable camelcase */
 
 import { chain, Rule, SchematicContext, Tree } from '@angular-devkit/schematics';
-import { createSchematicWithMetricsIfInstalled, updateImports } from '@o3r/schematics';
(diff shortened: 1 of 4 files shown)


packages/@o3r/core/schematics/ng-update/index.ts
```diff
@@ -34,7 +34,8 @@ function updateV10_0Fn(): Rule {
     const updateRules: Rule[] = [
       // Some of these imports were missed in the generators of v9, so it's easier to just run the update again
       updateImports(mapImportAsyncStore),
-      updateConfiguration
+      updateConfiguration,
+      alignVersions([/@o3r\/.*/, /@ama-sdk\/.*/])
```
fpaul-1A, line 38: aligning only otter dependencies won't be enough because of the transitive dependencies (angular, ngrx, ...)
we'll need a full version-harmonize

fpaul-1A:
The migrations scripts won't update the projects `package.json`, only workspace root
We'll recommend using our version-harmonize mechanism to do that

PR #1377, opened 2024-02-21
test: remove yarn-related command from npm it-tests

divdavem:
## Proposed change

Small PR to not call `yarn set version` from it-tests that use npm.
Note that `setPackagerManagerConfig` already takes care of setting yarn version in case yarn is used.


fpaul-1A (approved):

matthieu-crouzet (approved):

divdavem:
@fpaul-1A @matthieu-crouzet Thank you for your review!

PR #1381, opened 2024-02-21
Exact workspace dependencies

divdavem:
## Proposed change

I am suggesting to use exact dependencies rather than `^` dependencies for all workspace dependencies. This way, when using `npm create @ama-sdk@x.y.z` or `npm create @o3r@x.y.z`, I will get the exact version requested, and not a newer version.
This is especially important if a new version breaks something and I want to use an earlier version.


fpaul-1A (dismissed):

packages/@o3r/core/package.json
```diff
@@ -74,7 +74,7 @@
     }
   },
   "dependencies": {
-    "@o3r/schematics": "workspace:^",
+    "@o3r/schematics": "workspace:*",
```
kpanot, line 77: This dependency of this package and `@o3r/workspace` may cause issue if they are pined.
I could result to conflict as they will not be aligned with the other modules of the workspace.

For example `@o3r/* -> @o3r/core -> @o3r/schematics` will not referred the same package of `@o3r/* -> @o3r/schematics`

(Which is not the case of the other packages which CLI and this repo's apps)

divdavem: Should "@o3r/schematics" be a peer dependency of "@o3r/core" rather than a dependency then?
As part of #1379, I had to add "@o3r/schematics" as a direct (dev) dependency of the workspace created by `npm create @o3r` to solve the issue with npm, so this could go in the same direction. What do you think?

kpanot: I think it is a bit complicated.
In the scenario using `@o3r/create`, it could be added by the CLI before executing the `ngAdd` schematics but in the scenario (not properly tested yet) to run `ng add @o3r/core` on an existing app/lib (or `ng add @o3r/workspace` at workspace level) it will fail because of missing dep.
It can be an acceptable failure but, as these 2 packages supposed to be the entrypoints to Otter, it can be a bit complicated to justify.
It is a question that need to be debated within the team I think.

PR #1382, opened 2024-02-21
fix: lib should work with node-modules

vscaiceanu-1a:
## Proposed change

The goal is to make the library work with `node-modules` also, on top of `pnp`.

To do that it seems that we should get rid of the old `@typescript-eslint/experimental-utils` and use `@typescript-eslint/utils` instead.
This means jumping from v5 to v7 and we face breaking changes.


fpaul-1A:
This is handled in PR https://github.com/AmadeusITGroup/otter/pull/1500

vscaiceanu-1a:
done in #1507

PR #1407, opened 2024-02-26
fix(telemetry): use the new setup method

matthieu-crouzet:


packages/@o3r/extractors/src/core/wrapper.ts
```diff
@@ -20,17 +20,16 @@ export const createBuilderWithMetricsIfInstalled: BuilderWrapper = (builderFn) =
   let wrapper: BuilderWrapper = noopBuilderWrapper;
   try {
     const { createBuilderWithMetrics } = await import('@o3r/telemetry');
-    if (packageJson.config?.o3rMetrics) {
-      wrapper = createBuilderWithMetrics;
-    }
+    wrapper = createBuilderWithMetrics;
```
kpanot, line 23: This is not exactly correct: in `npm` context, this dep can be accessible because part of a dep dependency but you may have disable the metrics in your project.
It is also enabling telemetry per default if no `o3rMetrics` is specified :S.
I think the previous was actually more correct in that case

packages/@o3r/schematics/src/utility/wrapper.ts
```diff
@@ -47,16 +48,20 @@ For more details and instructions on how to change these settings, see https://g
 
       if (isReplyPositive) {
         const version = JSON.parse(readFileSync(path.join(__dirname, '..', '..', 'package.json'), 'utf-8')).version;
-        context.addTask(
-          new NodePackageNgAddTask(
-            '@o3r/telemetry',
-            {
-              dependencyType: NodeDependencyType.Dev,
-              version
+        const installTelemetryRule = setupDependencies({
```
kpanot, line 51: To reduce the number of install, this function should be called only once per schematics.
Maybe you can align with what it is done in the @o3r/core schematics' helpers

packages/@o3r/schematics/src/utility/wrapper.ts
```diff
@@ -47,16 +48,20 @@ For more details and instructions on how to change these settings, see https://g
...
+          dependencies: {
+            // eslint-disable-next-line @typescript-eslint/naming-convention
+            '@o3r/telemetry': {
+              inManifest: [{
+                range: `~${version}`,
+                types: [NodeDependencyType.Dev]
+              }],
+              requireInstall: true
```
kpanot, line 59: Not sure the installation is required explicitly.
This function is mainly in case of `ngAdd` chain to make sure the installation is run for this package before running the following `ngAdd` schematic. I think the installation is not really mandatory here for following `ngAdd`

packages/@o3r/extractors/src/core/wrapper.ts
matthieu-crouzet: The wrapper handle all the logic to know if data should be sent or not 
with the environment variable etc...

packages/@o3r/schematics/src/utility/wrapper.ts
```diff
@@ -41,43 +42,46 @@ export const createSchematicWithMetricsIfInstalled: SchematicWrapper = (schemati
...
-          const { createSchematicWithMetrics } = await import('@o3r/telemetry');
-          wrapper = createSchematicWithMetrics;
-        } catch {
-          // If pnp context package installed in the same process will not be available
-        }
-      } else {
-        context.logger.info('You can activate it at any time by running `ng add @o3r/telemetry`.');
+          },
+          ngAddToRun: ['@o3r/telemetry'],
+          runAfterTasks: taskIdFromContext // next PR from Kilian
```
kpanot, line 69: comment can be removed

fpaul-1A (approved):

fpaul-1A (approved):

fpaul-1A (approved):

fpaul-1A (dismissed):

vscaiceanu-1a (dismissed):

fpaul-1A (approved):

kpanot (approved):

PR #1434, opened 2024-03-01
fix: file pack incorrect subentries

cpaulve-1A:
File-pack copies the root and subentries package.json file and re-export subentries on the root package.json It failed to expose the subentry as it was not skipping the src folder during the copy while d.ts files were directly set at the root of the dist folder.

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@ama-sdk/core/cli/files-pack.cts
+++ b/packages/@ama-sdk/core/cli/files-pack.cts
@@ -14,19 +14,20 @@ import type { PackageJson } from 'type-fest';
 const argv = minimist(process.argv.slice(2));
 const distFolder = argv.dist || 'dist';
 const baseDir = argv.cwd && path.resolve(process.cwd(), argv.cwd) || process.cwd();
+const posixBaseDir = path.posix.join(...baseDir.split(path.sep));
 const {watch, noExports} = argv;
 
 const files = [
-  'README.md',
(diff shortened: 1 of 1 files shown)


matthieu-crouzet (approved):

sdo-1A (approved):

vscaiceanu-1a (approved):

packages/@ama-sdk/core/cli/files-pack.cts
```diff
@@ -67,18 +68,18 @@ void (async () => {
   };
 
   // Move files into the dist folder
-  const copies = files.map(async (glob) => {
+  const copies = files.map(async ({glob, cwdForCopy}) => {
     return watch ?
       import('chokidar')
-        .then((chokidar) => chokidar.watch(path.join(baseDir, glob)))
+        .then((chokidar) => chokidar.watch(path.posix.join(posixBaseDir, glob)))
```
kpanot, line 74: Actually `path.posix.join('c:\something', '*')` will result to `c:\something/*`. Not sure it is what you want

cpaulve-1A: Sorry did not see that before merging, but this has already been transformed to posix

PR #1443, opened 2024-03-04
fix: incorrect glob for globby for linux and mac

cpaulve-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


packages/@ama-sdk/core/cli/files-pack.cts
```diff
@@ -78,7 +78,7 @@ void (async () => {
             return updateExports();
           }
         })) :
-      globby.sync(path.posix.join(posixBaseDir, glob))
+      globby.sync(glob)
```
kpanot, line 81: With this change you will not have issue on Windows because of globby does not support `\`? (or `files` contains always only `/` paths?)

packages/@ama-sdk/showcase-sdk/project.json
```diff
@@ -7,10 +7,12 @@
   "targets": {
     "build": {
       "executor": "nx:noop",
+      "inputs": ["default", "^default", "^cli"],
```
kpanot, line 10: not needed

cpaulve-1A: It was a tentative to fix the cache because the showcase-sdk build was not invalidated and thus not rebuild in the CI.
Anyway it does not work so it should be handled in another task

packages/@ama-sdk/core/cli/files-pack.cts
cpaulve-1A: glob is not the file just the pattern because mac did not seem to work correctly with the absolute path
image

matthieu-crouzet (approved):

matthieu-crouzet (approved):

packages/@ama-sdk/showcase-sdk/project.json
cpaulve-1A: image
image

packages/@ama-sdk/core/cli/files-pack.cts
```diff
@@ -27,7 +26,7 @@ const files = [
 /**  Update package.json exports */
 const updateExports = async () => {
   const packageJson = JSON.parse(await fs.readFile(path.join(baseDir, 'package.json'), { encoding: 'utf8' }));
-  const packageJsonFiles = globby.sync(path.posix.join(posixBaseDir, distFolder, '*', '**', 'package.json'), {absolute: true});
```
fpaul-1A, line 30: don't we need this is `argv.cmd` is passed ?

fpaul-1A (approved):

packages/@ama-sdk/core/cli/files-pack.cts
cpaulve-1A: Let's open a ticket to follow up

PR #1460, opened 2024-03-06
fix(ama-sdk-core): removing useless abort controller in angular client

divdavem:
## Proposed change

The angular client in `@ama-sdk/core` creates an abort controller, but there is no way to abort the request, so that part of the code is useless and should be removed. Also the correct way to abort a request with [httpClient.request](https://angular.io/api/common/http/HttpClient#request) is to unsubscribe from the observable (and not to pass a signal to the request method).

### DIFF
--- a/packages/@ama-sdk/core/src/clients/api-angular-client.ts
+++ b/packages/@ama-sdk/core/src/clients/api-angular-client.ts
@@ -102,10 +102,6 @@ export class ApiAngularClient implements ApiClient {
     try {
       const headers = Object.fromEntries(options.headers.entries());
 
-      const controller = typeof AbortController !== 'undefined' ? new AbortController() : undefined;
-      if (controller) {
-        options.signal = controller.signal;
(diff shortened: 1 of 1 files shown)


kpanot (approved):

fpaul-1A (approved):

kpanot:
Would be interested to have an equivalent unsubscribing the observable

divdavem:
@kpanot @fpaul-1A Thank you for your review!

divdavem:
> Would be interested to have an equivalent unsubscribing the observable

Thank you @kpanot for your comment.
Currently, the abort feature of the fetch client is only usable from fetch plugins.
Do you think we should do exactly the same thing for the angular client? With specific angular plugins, or reusing fetch plugins?
But more generally, I think it would be good to expose the abort feature directly (without having to pass through a plugin) for both clients (fetch and angular). What do you think?

PR #1470, opened 2024-03-08
feat(ama-sdk-core): abort requests by passing an abort signal in metadata

divdavem:
## Proposed change

Adds the ability to abort requests in `@ama-sdk/core` (and any request of an SDK generated by the Otter SDK generator) by passing an [abort signal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) in metadata. It is implemented both for the fetch and the angular client.


kpanot (approved):

matthieu-crouzet (approved):

divdavem:
@kpanot @matthieu-crouzet Thank you for your review!

PR #1474, opened 2024-03-08
feat(ama-sdk-create): allow creation without a scope

divdavem:
## Proposed change

When running `npm create @ama-sdk typescript my-package`, I would intuitively expect a `my-package` package to be created in `my-package` folder.
What currently happens is the creation of the `@sdk/my-package` package inside the `sdk/my-package` folder. Why is this `sdk` scope added? I suggest in this PR to remove the `sdk` default scope.

### DIFF
--- a/packages/@ama-sdk/create/src/index.ts
+++ b/packages/@ama-sdk/create/src/index.ts
@@ -7,7 +7,6 @@ import { dirname, join, relative, resolve } from 'node:path';
 import * as minimist from 'minimist';
 
 const packageManagerEnv = process.env.npm_config_user_agent?.split('/')[0];
-const defaultScope = 'sdk';
 const binPath = resolve(require.resolve('@angular-devkit/schematics-cli/package.json'), '../bin/schematics.js');
 const args = process.argv.slice(2);
 const argv = minimist(args);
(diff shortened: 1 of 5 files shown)


matthieu-crouzet (dismissed):

divdavem:
@matthieu-crouzet Thank you for your review!

packages/@ama-sdk/create/src/index.ts
```diff
@@ -26,20 +25,21 @@ if (argv._.length < 2) {
...
   console.error('Only the generation of "typescript" SDK is available');
   process.exit(-2);
 }
 
-if (!pck) {
-  console.warn(`The given package name, is not a scoped package. a default "@${defaultScope}" scope will be used`);
-  pck = name;
-  name = defaultScope;
+const fullPackage = argv._[1];
+const packageMatch = /^(?:@?([^@/]+)\/)?([^@/]+)$/.exec(fullPackage);
```
kpanot, line 35: Not sure to understand the first `?`, do you want to support path before the package name? In that case it is supported only for packages without scope :S
Maybe the simplest way would be to remove it:

```suggestion
const packageMatch = /^(?:@([^@/]+)\/)?([^@/]+)$/.exec(fullPackage);
```

divdavem: Indeed, it makes more sense to make it mandatory to write a correct package name, either with a scope (including the @ character) or without a scope. I had kept the previous behavior (which allowed to specify a scope without the @ character, and the create command was adding it), but I prefer what you suggest.

kpanot (approved):

fpaul-1A (approved):

divdavem:
@kpanot @fpaul-1A Thank you for your review!

Issue #1420, opened 2024-02-28
[Feature]: IntelliSense for CSS Variables based on Design Token


kpanot:
### Context

As a UI Developper, I would like to have auto-completion to help using CSS Variable in component CSS.

### Proposal

The auto-completion should be provided according to the following criteria:
- The auto-completion should be triggered when cursor in `var()` function parameter
- The suggestion should be based on Design Token defined in the specified JSON files
- The suggestion should contains only the Design Token defined at component level, global level and not private
- The Design Token files should be determined via a pattern provided to VsCode extension input and defaulting to `**/*.token.json`

PR #1479, opened 2024-03-12
feat(vscode-ext): intellisense for design tokens

matthieu-crouzet:
## Related issues

- :rocket: Feature #1420

### DIFF
--- a/apps/vscode-extension/package.json
+++ b/apps/vscode-extension/package.json
@@ -21,6 +21,13 @@
     "configuration": {
       "title": "Otter",
       "properties": {
+        "otter.design.filesPatterns": {
+          "type": "array",
+          "default": [
+            "**/*.token.json"
+          ],
+          "description": "List of pattern to find files that define design tokens"
+        },
         "otter.extract.styling.forbiddenWords": {
           "type": "array",
           "default": [
@@ -332,6 +339,7 @@
     "prepare:publish": "prepare-publish ./dist"
   },
   "dependencies": {
+    "@o3r/design": "workspace:^",
     "@o3r/schematics": "workspace:^",
     "eslint": "^8.42.0"
   },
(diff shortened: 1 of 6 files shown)


apps/vscode-extension/src/intellisense/design-token.ts
```diff
@@ -0,0 +1,72 @@
+import { CompletionItem, CompletionItemKind, CompletionItemProvider, HoverProvider } from 'vscode';
+import { DesignTokenVariableSet, DesignTokenVariableStructure, getCssTokenValueRenderer, parseDesignTokenFile } from '@o3r/design';
+import * as vscode from 'vscode';
+import * as path from 'node:path';
+
+const readFile = async (filePat: string) => (await vscode.workspace.fs.readFile(vscode.Uri.parse(filePat))).toString();
+
+const getDesignTokens = async (currentFile?: string) => {
```
kpanot, line 8: I think you should add a cache mechanism for the token calculation for the case we will have a large amount of files.
A simple way it to recalculate the Design Token only if one of the file of the glob has changed

apps/vscode-extension/package.json
```diff
@@ -21,6 +21,13 @@
     "configuration": {
       "title": "Otter",
       "properties": {
+        "otter.design.filesPatterns": {
+          "type": "array",
+          "default": [
+            "**/*.token.json"
```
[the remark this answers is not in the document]
matthieu-crouzet: Could be done in another task I think

apps/vscode-extension/src/intellisense/design-token.ts
matthieu-crouzet: Will be done in another [task](https://github.com/AmadeusITGroup/otter/issues/1490)

kpanot (approved):

vscaiceanu-1a (approved):
Documentation can be improved via https://github.com/AmadeusITGroup/otter/issues/1505

PR #1485, opened 2024-03-13
fix(devtools): activate all devtools on showcase

matthieu-crouzet:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


fpaul-1A (approved):

packages/@o3r/localization/src/devkit/localization-devtools.message.service.ts
```diff
@@ -40,6 +42,7 @@ export class LocalizationDevtoolsMessageService implements OnDestroy {
       }
       case 'displayLocalizationKeys': {
         this.localizationDevTools.showLocalizationKeys(message.toggle);
+        this.appRef.tick();
```
kpanot, line 45: Why is it needed here, the localization observable should do the job itself

PR #1486, opened 2024-03-13
fix(chrome-ext): display a specific message when component selection is not available

matthieu-crouzet:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


fpaul-1A (approved):

packages/@o3r/components/src/devkit/components-devkit.interface.ts
```diff
@@ -12,7 +12,12 @@ export interface ToggleInspectorMessage extends OtterMessageContent<'toggleInspe
   isRunning: boolean;
 }
 
+export interface IsComponentSelectionAvailableMessage extends OtterMessageContent<'isComponentSelectionAvailable'> {
```
kpanot, line 15: missing comment (en if it is not the only one in this file)

kpanot (dismissed):

fpaul-1A (approved):

kpanot (approved):

PR #1507, opened 2024-03-15
chore: support node-modules linker

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


matthieu-crouzet (dismissed):

vscaiceanu-1a (dismissed):

packages/@o3r/components/package.json
```diff
@@ -12,6 +12,9 @@
     "otter-cms"
   ],
   "exports": {
+    "./package.json": {
```
kpanot, line 15: It is not conflicting with the ngPackagr process adding it? It does not print warning/error anymore?

kpanot (approved):

packages/@o3r/components/package.json
fpaul-1A: I do get the following warning:
> WARNING: Found a conflicting export condition for "./package.json". The "default" condition would be overridden by ng-packagr. Please unset it.

But if I don't add this export here the build fails using `nodeLinker: node-modules`

kpanot (approved):

packages/@o3r/components/package.json
fpaul-1A: Changed `default` export to `node` export

matthieu-crouzet (approved):

PR #1522, opened 2024-03-20
fix: duplicates of operators in rules-engine

sdo-1A:
## Proposed change

An issue was raised regarding duplicated operators in the rules.operators.metadata.json file, so the proposed fix is to filter the operators before adding them in the file

### DIFF
--- a/packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.ts
+++ b/packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.ts
@@ -316,7 +316,7 @@ export class RulesEngineExtractor {
         });
       }
     });
-    return operators;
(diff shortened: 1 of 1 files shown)


fpaul-1A (approved):

packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.ts
```diff
@@ -316,7 +316,7 @@ export class RulesEngineExtractor {
         });
       }
     });
-    return operators;
+    return Array.from((new Map(operators.map((operator) => [operator.id, operator]))).values());
```
kpanot, line 319: at this level I wonder if a simple filter would not be faster and more readable:
```typescript
return operators
  .fitler(({id}, index, arr) => !arr.slice(index).some((o) => o.id === id));
```
but I am fine with it :)

kpanot (approved):

PR #1528, opened 2024-03-21
fix(vscode): add caching to the parsed design token files

### LINKED ISSUE #1490: [Feature]: Improve vscode design token intellisense performance
I think you should add a cache mechanism for the token calculation for the case we will have a large amount of files.
A simple way it to recalculate the Design Token only if one of the file of the glob has changed

_Originally posted by @kpanot in https://github.com/AmadeusITGroup/otter/pull/1479#discussion_r1522411335_

kpanot:
## Proposed change

Add caching to the parsed design token files mechanism

## Related issues

- :rocket: Feature resolves #1490

### DIFF
--- a/apps/vscode-extension/src/intellisense/design-token.ts
+++ b/apps/vscode-extension/src/intellisense/design-token.ts
@@ -1,16 +1,34 @@
 import { CompletionItem, CompletionItemKind, CompletionItemProvider, HoverProvider } from 'vscode';
-import { DesignTokenVariableSet, DesignTokenVariableStructure, getCssTokenValueRenderer, parseDesignTokenFile } from '@o3r/design';
+import { type DesignTokenVariableSet, type DesignTokenVariableStructure, getCssTokenValueRenderer, parseDesignTokenFile } from '@o3r/design';
 import * as vscode from 'vscode';
 import * as path from 'node:path';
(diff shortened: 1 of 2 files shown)


apps/vscode-extension/src/intellisense/design-token.ts
```diff
@@ -1,16 +1,33 @@
...
+type DesignTokenCache = {
+  lastExtractionMtime: number;
+  tokenStructure: DesignTokenVariableSet;
+};
+
 const readFile = async (filePat: string) => (await vscode.workspace.fs.readFile(vscode.Uri.parse(filePat))).toString();
 
-const getDesignTokens = async (currentFile?: string) => {
+const getDesignTokens = async (currentFile: string, cache: Map<string, DesignTokenCache>) => {
+  const lastExtractionMtime = (new Date()).getTime();
```
matthieu-crouzet, line 14:
```suggestion
  const lastExtractionMtime = Date.now()
```

matthieu-crouzet (approved):
Maybe a small comment for `mtime` 
can be hard to understand that m stands for modification

matthieu-crouzet (approved):

fpaul-1A (approved):

PR #1538, opened 2024-03-22
test: use jest custom environment for it-tests

fpaul-1A:
## Proposed change

Use a jest custom environment to be able to use the jest lifecycle events to clean the workspace only if the test succeeds

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@ama-sdk/create/src/index.it.spec.ts
+++ b/packages/@ama-sdk/create/src/index.it.spec.ts
@@ -1,3 +1,11 @@
+/**
+ * Test environment exported by O3rEnvironment, must be first line of the file
+ * @jest-environment @o3r/test-helpers/jest-environment
+ * @jest-environment-o3r-app-folder test-create-sdk
+ * @jest-environment-o3r-type blank
+ */
+const o3rEnvironment = globalThis.o3rEnvironment;
+
 import {
   getDefaultExecSyncOptions,
   getPackageManager,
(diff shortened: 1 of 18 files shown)


mrednic-1A (approved):

packages/@o3r/test-helpers/src/jest-environment/index.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  /**
+   * Catch Jest lifecycle events
+   * @param event
+   * @param _state
+   */
+  public async handleTestEvent(event: Circus.AsyncEvent, _state: Circus.State) {
+    // Create test environment before test starts
+    if (event.name === 'test_start') {
+      const appFolder = `${this.appFolder}${this.appIndex++ || ''}`;
```
cpaulve-1A, line 90: Is the use of the increment in a function always evaluated before its use in a function for node?
Is this a reliable behaviour?

fpaul-1A: `++a` returns the value after increment
`a++` returns the value before increment

packages/@ama-sdk/create/src/index.it.spec.ts
```diff
@@ -1,32 +1,32 @@
+/**
```
kpanot, line 1: Should not be `\*` instead?

packages/@o3r/test-helpers/src/jest-environment/index.ts
```diff
@@ -0,0 +1,118 @@
+import type { EnvironmentContext, JestEnvironmentConfig } from '@jest/environment';
+import type { Circus } from '@jest/types';
+import { TestEnvironment as NodeTestEnvironment } from 'jest-environment-node';
+import { execSync } from 'node:child_process';
+import pidFromPort from 'pid-from-port';
+import { rm } from 'node:fs/promises';
+import path from 'node:path';
+import { prepareTestEnv, type PrepareTestEnvType } from '../prepare-test-env';
+
+export type TestEnvironment = Awaited<ReturnType<typeof prepareTestEnv>>;
```
kpanot, line 10: Missing comments

packages/@o3r/test-helpers/src/jest-environment/index.ts
```diff
@@ -0,0 +1,118 @@
...
+
+const rootFolder = path.join(__dirname, '..', '..', '..', '..');
+
+/**
+ * Custom Jest environment used to manage test environments with Verdaccio setup
+ */
+export class JestEnvironmentO3r extends NodeTestEnvironment {
+  /**
+   * Stop Verdaccio server at the end of the test
+   * @private
```
kpanot, line 24: Duplicate information
(Same for others)

packages/@o3r/test-helpers/src/jest-environment/index.ts
```diff
@@ -0,0 +1,118 @@
...
+   * Type of test environment to be created
+   * @private
+   */
+  private readonly prepareTestEnvType: PrepareTestEnvType | undefined;
+
+  /**
+   * Map test name with test environment
+   * @private
+   */
+  private testEnvironments: Record<string, TestEnvironment> = {};
```
kpanot, line 50: Readonly as well?

packages/@ama-sdk/create/src/index.it.spec.ts
fpaul-1A: I don't think so
https://jestjs.io/docs/configuration#testenvironment-string

kpanot: Does it work with simple comment then? Because a meanful comment to an `import` statement it is a bit strange

fpaul-1A: No, I've changed the code to associate the comment with a `const` representing the model exported by the custom Jest Environment

kpanot (approved):

mrednic-1A (approved):

PR #1539, opened 2024-03-22
feat: only create schema file for facts of type object

sdo-1A:
## Proposed change

Currently schema files are being created for all facts that have non-standard types or are of type enum. The goal is to only create schema files for facts of type object.

### DIFF
--- a/packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.interfaces.ts
+++ b/packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.interfaces.ts
@@ -1,18 +1,49 @@
 /** Supported types for a fact */
-export type FactSupportedTypes = 'string' | 'number' | 'date' | 'object' | 'boolean';
+export type FactSupportedTypes = 'string' | 'number' | 'date' | 'object' | 'boolean' | 'array';
 
-/** Fact as description in the metadata file */
-export interface MetadataFact {
+/** Base metadata fact */
+interface BaseMetadataFact {
   /** Name of the Fact */
   name: string;
   /** Fact description */
   description?: string;
   /** Type of the fact */
   type: FactSupportedTypes;
+}
+
+/** Metadata fact of type enum */
+export interface EnumMetadataFact extends BaseMetadataFact {
+  /** @inheritdoc */
+  type: 'string';
+  /** Set of values of enum of type string */
+  enum: string[];
+}
+
+/** Metadata fact of type object */
+export interface ObjectMetadataFact extends BaseMetadataFact {
+  /** @inheritdoc */
+  type: 'object';
   /** Path to schema file describing the complex file */
   schemaFile?: string;
 }
 
+/** Metadata fact of type array */
+export interface ArrayMetadataFact extends BaseMetadataFact {
+  /** @inheritdoc */
+  type: 'array';
+  /** Items in array */
+  // eslint-disable-next-line no-use-before-define
+  items: MetadataFact;
+}
+
+/** Metadata fact of type string, number, date, or boolean */
+export interface OtherMetadataFact extends BaseMetadataFact {
+  /** @inheritdoc */
+  type: 'string' | 'number' | 'date' | 'boolean';
+}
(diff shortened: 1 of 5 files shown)


packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.interfaces.ts
```diff
@@ -1,18 +1,49 @@
...
   description?: string;
   /** Type of the fact */
   type: FactSupportedTypes;
+}
+
+/** Metadata fact of type enum */
+export interface EnumMetadataFact extends BaseMetadataFact {
+  /** @inheritdoc */
+  type: 'string';
+  /** Path to schema file describing the complex file */
```
fpaul-1A, line 18: copy pasted

packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.interfaces.ts
```diff
@@ -1,18 +1,49 @@
 /** Supported types for a fact */
-export type FactSupportedTypes = 'string' | 'number' | 'date' | 'object' | 'boolean';
+export type FactSupportedTypes = 'string' | 'number' | 'date' | 'object' | 'boolean' | 'array';
 
-/** Fact as description in the metadata file */
-export interface MetadataFact {
+/** Base metadata fact */
+export interface BaseMetadataFact {
```
vscaiceanu-1a, line 5: if you don't need `BaseMetadataFact` outside this file I would advise not to export it.

packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.ts
```diff
@@ -4,12 +4,25 @@ import { ConfigDocParser } from '@o3r/extractors';
 import { O3rCliError } from '@o3r/schematics';
 import { promises as fs } from 'node:fs';
 import globby from 'globby';
+import { JSONSchema7Type } from 'json-schema';
```
vscaiceanu-1a, line 7:
```suggestion
import type { JSONSchema7Type } from 'json-schema';
```

packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.interfaces.ts
```diff
@@ -1,18 +1,49 @@
 /** Supported types for a fact */
-export type FactSupportedTypes = 'string' | 'number' | 'date' | 'object' | 'boolean';
+export type FactSupportedTypes = 'string' | 'number' | 'date' | 'object' | 'boolean' | 'array';
 
-/** Fact as description in the metadata file */
-export interface MetadataFact {
+/** Base metadata fact */
+interface BaseMetadataFact {
```
kpanot, line 5: There is a reason to not export this interface?

kpanot (approved):

packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.interfaces.ts
sdo-1A: It is not being used outside of this file so I removed the export

cpaulve-1A: Are you sure it is not used by our users?

sdo-1A: I don't think so since I just created this interface but I can export it in case it needs to be used

cpaulve-1A: My bad, I was just looking at the comment and did not realize you were not talking about the MetadataFact which is still exposed as an union type. It is fine for me.

PR #1548, opened 2024-03-25
chore: enhance testing on sdk generation with or without reviver

cpaulve-1A:
## Proposed change

Add a new use case to make sure non light sdk will not throw errors after generation

## Related issues

- :bug: Add test cases for https://github.com/AmadeusITGroup/otter/pull/1537

### DIFF
--- a/packages/@ama-sdk/create/src/index.it.spec.ts
+++ b/packages/@ama-sdk/create/src/index.it.spec.ts
@@ -10,7 +10,7 @@ import {
   setupLocalRegistry
 } from '@o3r/test-helpers';
 import * as fs from 'node:fs';
-import { cpSync, mkdirSync, renameSync } from 'node:fs';
+import { cpSync, existsSync, mkdirSync, renameSync } from 'node:fs';
 import * as path from 'node:path';
 
 const projectName = 'test-sdk';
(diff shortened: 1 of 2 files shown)


packages/@ama-sdk/create/src/index.it.spec.ts
```diff
@@ -84,6 +96,7 @@ describe('Create new sdk command', () => {
     ).not.toThrow();
     expect(() => packageManagerRun({script: 'build'}, { ...execAppOptions, cwd: sdkPackagePath })).not.toThrow();
     expect(() => packageManagerRun({script: 'doc:generate'}, { ...execAppOptions, cwd: sdkPackagePath })).not.toThrow();
+    expect(existsSync(path.join(sdkPackagePath, 'src', 'models', 'base', 'pet', 'pet.reviver.ts'))).toBeFalsy();
```
sdo-1A, line 99: this expect statement could be added to the test of light SDK generation also ?

sdo-1A (approved):

kpanot (approved):

packages/@ama-sdk/create/src/index.it.spec.ts
```diff
@@ -48,16 +48,29 @@ describe('Create new sdk command', () => {
...
       packageManagerCreate({
         script: '@ama-sdk',
         args: ['typescript', sdkPackageName, '--package-manager', packageManager, '--spec-path', path.join(sdkFolderPath, 'swagger-spec.yml')]
       }, execAppOptions)
     ).not.toThrow();
     expect(() => packageManagerRun({script: 'build'}, { ...execAppOptions, cwd: sdkPackagePath })).not.toThrow();
+    expect(existsSync(path.join(sdkPackagePath, 'src', 'models', 'base', 'pet', 'pet.reviver.ts'))).toBeFalsy();
+  });
+
+  test('should generate a full SDK when the specification is provided', () => {
```
pginoux-1A, line 65: Is that the kind of tests where we want to add generated file comparison on top of the "build" step ?

cpaulve-1A: for this test, the idea was to see if we could make sure the projects actually build and some expected files are here. I don't  want to over complexify it as it would be unmaintainable. Making sure that at least one reviver exists is easier to maintain

PR #1559, opened 2024-03-27
fix(eslint-config): use jest rule only if jest dependency is found

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/eslint-config-otter/typescript.cjs
+++ b/packages/@o3r/eslint-config-otter/typescript.cjs
@@ -1,3 +1,8 @@
+const hasJestDependency = (() => {
+  try { return !!require.resolve('jest'); }
+  catch { return false; }
+})();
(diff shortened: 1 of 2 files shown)


packages/@o3r/eslint-config-otter/typescript.cjs
```diff
@@ -1,3 +1,8 @@
+const hasJestDependency = (() => {
+  try { return !!require.resolve('jest'); }
+  catch (ex) { return false; }
```
kpanot, line 3:
```suggestion
  catch { return false; }
```

mrednic-1A (dismissed):

kpanot (approved):

mrednic-1A (approved):

PR #1565, opened 2024-03-28
fix: github action NX build inputs/outputs

kpanot:


mrednic-1A (dismissed):

nx.json
```diff
@@ -198,6 +198,12 @@
       "outputs": ["{projectRoot}/dist", "{projectRoot}/build/.tsbuildinfo"],
       "cache": true
     },
+    "compile-github-action": {
+      "dependsOn": ["^build"],
+      "inputs": ["source", "^source"],
+      "outputs": ["{projectRoot}/packaged-action", "{projectRoot}/tmp"],
```
fpaul-1A, line 204: is `"{projectRoot}/packaged-action"` really an output of the `build` script ? It seems to be only the output of the `package` script

kpanot: indeed

fpaul-1A (approved):

cpaulve-1A (approved):

PR #1588, opened 2024-04-04
feat: support linked components in rules engine

mrednic-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/apps/chrome-devtools/src/app-components/app.component.ts
+++ b/apps/chrome-devtools/src/app-components/app.component.ts
@@ -101,7 +101,12 @@ export class AppComponent {
       this.selectedComponentInfo$.pipe(filter((info): info is OtterLikeComponentInfo => !!info)),
       rulesetHistoryService.rulesetExecutions$.pipe(startWith([]))
     ]).pipe(
-      map(([info, executions]) => executions.filter((execution) => execution.rulesetInformation?.linkedComponent?.name === info.componentName))
+      map(([info, executions]) =>
+        executions.filter((execution) =>
+          (execution.rulesetInformation?.linkedComponent?.name === info.componentName) ||
+          (execution.rulesetInformation?.linkedComponents?.or.some(linkedComp => linkedComp.name === info.componentName))
+        )
+      )
     );
   }
 
--- a/docs/ab-testing/AB_TESTING.md
+++ b/docs/ab-testing/AB_TESTING.md
@@ -1,7 +1,7 @@
 # A/B Testing
 ## Introduction
 In this page, we will help you integrate an A/B testing solution to the rest of your application.
-Let's first define a core concept of A/B testing: an ``experiment`` is a behavior specific to a group of users 
+Let's first define a core concept of A/B testing: an ``experiment`` is a behavior specific to a group of users
 -- for instance the display of a banner or a configuration to drive a UI component.
 
 In its [@o3r/third-party](https://www.npmjs.com/package/@o3r/third-party package, the Otter framework exposes a technical
@@ -9,10 +9,10 @@ bridge to allow third party scripts to drive experiments on your application.
 
 ## Mechanism
 The ``AbTestBridge`` is a generic class that can support multiple experiment types to fit your
-needs (string, object etc.). The ``AbTestBridge`` exposes two methods to manage the list of experiments to apply on the 
+needs (string, object etc.). The ``AbTestBridge`` exposes two methods to manage the list of experiments to apply on the
 application: ``window.start`` and ``window.stop``.
 
-The ``start(experiment: T | T[])`` method requests the application to run one or several experiments, the ``stop(experiment: T | T[])`` method will 
+The ``start(experiment: T | T[])`` method requests the application to run one or several experiments, the ``stop(experiment: T | T[])`` method will
 remove one or several experiments from the list of experiments while ``stop()`` will stop all the ongoing experiments.
 
 The list of the requested experiments is exposed via ``experiments$`` a public ``Observable`` matching your type.
@@ -90,7 +90,7 @@ export class ExperimentFactsService extends FactsService<ExperimentFacts> {
 
   /**
    * Emit a new list of experiments to re-evaluate the rulesets
-   * 
+   *
    * @param experiments: new list of experiments to apply. It will override the previous one
    */
   public setExperiments(experiments: {id: string; variationId: string}[]) {
@@ -100,7 +100,7 @@ export class ExperimentFactsService extends FactsService<ExperimentFacts> {
   }
 }
 ```
-**Warning** Your service must be imported only once in the application. A good way to do it is to provide it in root as 
+**Warning** Your service must be imported only once in the application. A good way to do it is to provide it in root as
 a singleton.
 
 #### The A/B Testing service
@@ -166,7 +166,7 @@ export class AbTestService {
   /**
    * Update experiment fact in rule engine
    *
-   * @param experiments 
+   * @param experiments
    */
   protected applyExperimentRules(experiments: ABTestingExperiment[]) {
     this.experimentFacts.setExperiments(experiments);
@@ -183,7 +183,7 @@ export class AbTestService {
 
   /**
    * Use logger service to log different informations
-   * 
+   *
    * @param args
    */
   private log(...args: any[]) {
@@ -201,11 +201,15 @@ You can now create A/B testing-driven rulesets .
     {
       "id": "1",
       "name": "AB testing rule",
-      "linkedComponent": {
-        "library": "@my/lib",
-        "name": "ABTestingComponent"
+      "linkedComponents": {
+        "or": [
+          {
+            "library": "@my/lib",
+            "name": "ABTestingComponent"
+          }
+        ]
       },
-       "rules": [
+      "rules": [
         {
           "name": "A/B Test - Update configuration ",
           "inputRuntimeFacts": [],
@@ -257,4 +261,5 @@ You can now create A/B testing-driven rulesets .
 }
 
 ```
+> Note: In v10 and previously, we used `linkedComponent` property to activate a ruleset on demand. This becomes deprecated and will be removed in v12. Use `linkedComponents` instead;
 
--- a/docs/rules-engine/how-to-use/cms-json-imports.md
+++ b/docs/rules-engine/how-to-use/cms-json-imports.md
@@ -16,16 +16,20 @@ If the application decides to migrate to the CMS, it will be able to import the 
 To help you to write the ruleset, we have provided a JSON schema that you can use to get indication in your IDE.
 This can be used as following:
 
-```json
+```json5
 {
   "$schema": "./node_module/@o3r/rules-engine/schemas/rulesets.schema.json",
   "ruleSets": [
     {
       "id": "e5th46e84-5e4th-54eth65seth46se8th2",
       "name": "the second ruleset",
-      "linkedComponent": {
-        "library": "@scope/components",
-        "name": "TestComponent"
+      "linkedComponents": {
+        "or": [
+          {
+            "library": "@scope/components",
+            "name": "TestComponent"
+          }
+        ]
       },
       "rules": [
         ...
@@ -34,3 +38,4 @@ This can be used as following:
   ]
 }
 ```
+> Note: In v10 and previously, we used `linkedComponent` property to activate a ruleset on demand. This becomes deprecated and will be removed in v12. Use `linkedComponents` instead;
--- a/docs/rules-engine/how-to-use/dedicated-component-ruleset.md
+++ b/docs/rules-engine/how-to-use/dedicated-component-ruleset.md
@@ -9,17 +9,25 @@ To do that we will subscribe in the onInit of the component, to the dedicated ru
 
 ## How to add it to your application
 
-First, you need to have a ruleset defined in your json (created via the CMS interface) that will be identified by a unique id, and you need to specify that the ruleSet is linked to a component, meaning that it's not executed per default. The json will look as follows :
+First, you need to have a ruleset defined in your json (created via the CMS interface) that will be identified by a unique id, and you need to specify that the ruleSet is linked to at least one component, meaning that it's not executed per default. The json will look as follows :
 
-```json
+```json5
 {
   "ruleSets": [
     {
       "id": "e5th46e84-5e4th-54eth65seth46se8th2",
       "name": "the second ruleset",
-      "linkedComponent": {
-        "library": "@scope/components",
-        "name": "TestComponent"
+      "linkedComponents": {
+        "or": [ // 'or' keyword means: if at least one component from the list is registered, the ruleset is active
+          {
+            "library": "@scope/components",
+            "name": "TestComponent"
+          },
+          {
+            "library": "@scope/components",
+            "name": "TestComponent2"
+          }
+        ]
       },
       "rules": [
         ...
@@ -28,25 +36,27 @@ First, you need to have a ruleset defined in your json (created via the CMS inte
   ]
 }
 ```
+> Note: In v10 and previously, we used `linkedComponent` property to activate a ruleset on demand. This becomes deprecated and will be removed in v12. Use `linkedComponents` instead;
 
 Now that you have the id, you will need to subscribe to the ruleset in your component :
 
 ```typescript
 import {LinkableToRuleset, RulesEngineRunnerService} from '@o3r/rules-engine';
-import {SIMPLE_HEADER_PRES_CONFIG_ID} from 'my-component.config';
-import {ExposedComponent} from '@o3r/core';
+import {computeItemIdentifier, type ExposedComponent} from '@o3r/core';
 
 export class MyComponent implements LinkableToRuleset, ExposedComponent {
-  constructor(@Optional() private service: RulesEngineRunnerService) {
+
+  private readonly componentIdentifier = computeItemIdentifier('@scope/components', 'TestComponent');
+
+  constructor(@Optional() private readonly service: RulesEngineRunnerService) {
   }
 
   public ngOnInit() {
-    // You can also call the computeConfigurationName from @o3r/core to compute SIMPLE_HEADER_PRES_CONFIG_ID
-    this.service?.enableRuleSetFor(SIMPLE_HEADER_PRES_CONFIG_ID);
+    this.service?.enableRuleSetFor(componentIdentifier);
   }
 
   public ngOnDestroy() {
-    this.service?.disableRuleSetFor(SIMPLE_HEADER_PRES_CONFIG_ID);
+    this.service?.disableRuleSetFor(componentIdentifier);
   }
 }
 ```
@@ -60,7 +70,7 @@ At the initialization of the application, the ruleset store will provide a selec
 Then this selector will be used to subscribe to those rulesets in the service as an initial behavior.
 Our onDemand ruleset will not be part of this initial subscription, that's why as long as no one subscribes to it, it will not be evaluated.
 As soon as our component is instantiated on the page (ngOnInit called), it will explicitly subscribe to it, resulting in its evaluation (and refresh if any input facts changes).
-Once the subscription is removed (ngOnDestroy) called, the ruleset will go back to its previous 'disabled' mode.
+Once the subscription is removed (ngOnDestroy) called, and no other component is linked to it, the ruleset will go back to its previous 'disabled' mode.
 
 ## Possible use cases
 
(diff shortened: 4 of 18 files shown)


packages/@o3r/rules-engine/package.json
```diff
@@ -111,10 +114,12 @@
     "@nx/jest": "~18.2.0",
     "@nx/js": "~18.2.0",
     "@o3r/build-helpers": "workspace:^",
+    "@o3r/configuration": "workspace:^",
```
kpanot, line 117: The rules engine should not have dependency to this packages

mrednic-1A: the new added deps are used in the integration test. They are added as dev-deps only

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -317,6 +317,7 @@ export class RulesetExecutor {
       id: this.ruleset.id,
       validityRange: this.ruleset.validityRange,
       linkedComponent: this.ruleset.linkedComponent,
+      linkedComponents: this.ruleset.linkedComponents,
```
kpanot, line 320: Could plan the future migration by using array only
```suggestion
      linkedComponents: this.ruleset.linkedComponents || this.ruleset.linkedComponent && [this.ruleset.linkedComponent],
```

kpanot (dismissed):

packages/@o3r/rules-engine/.eslintrc.js
```diff
@@ -15,6 +15,24 @@ module.exports = {
...
+      ],
+      'plugins': [
+        '@nx',
+      ],
+      'rules': {
+        '@nx/dependency-checks': ['error', {
+          'buildTargets': ['build', 'build-builders', 'compile', 'test'],
+          'checkObsoleteDependencies': false,
+          'checkVersionMismatches': false,
+          'ignoredDependencies': ['ora', '@o3r/test-helpers', '@o3r/configuration', '@o3r/localization', '@o3r/testing']
```
kpanot, line 31: not sure you need `ora`

mrednic-1A: indeed. I'll check that

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
mrednic-1A: But I support both usecases. I can have a linkedComponent and linkedComponents too.
Eventually I could do an array:
`Array.from(new Set([this.ruleset.linkedComponent, ...this.ruleset.linkedComponents])).filter(c => !!c)`
?

apps/chrome-devtools/src/app-components/app.component.ts
```diff
@@ -101,7 +101,12 @@ export class AppComponent {
       this.selectedComponentInfo$.pipe(filter((info): info is OtterLikeComponentInfo => !!info)),
       rulesetHistoryService.rulesetExecutions$.pipe(startWith([]))
     ]).pipe(
-      map(([info, executions]) => executions.filter((execution) => execution.rulesetInformation?.linkedComponent?.name === info.componentName))
+      map(([info, executions]) =>
+        executions.filter((execution) =>
+          (execution.rulesetInformation?.linkedComponent?.name === info.componentName) ||
+          (execution.rulesetInformation?.linkedComponents?.find(linkedComp => linkedComp.name === info.componentName))
```
fpaul-1A, line 107:
```suggestion
          (execution.rulesetInformation?.linkedComponents?.some(linkedComp => linkedComp.name === info.componentName))
```

docs/rules-engine/how-to-use/dedicated-component-ruleset.md
```diff
@@ -33,20 +35,22 @@ Now that you have the id, you will need to subscribe to the ruleset in your comp
 
 ```typescript
 import {LinkableToRuleset, RulesEngineRunnerService} from '@o3r/rules-engine';
-import {SIMPLE_HEADER_PRES_CONFIG_ID} from 'my-component.config';
-import {ExposedComponent} from '@o3r/core';
+import { computeItemIdentifier } from '@o3r/core';
+import type { ExposedComponent } from '@o3r/core';
```
fpaul-1A, line 39:
```suggestion
import { computeItemIdentifier, type ExposedComponent } from '@o3r/core';
```

packages/@o3r/rules-engine/.eslintrc.js
```diff
@@ -15,6 +15,24 @@ module.exports = {
...
+      ],
+      'plugins': [
+        '@nx',
+      ],
+      'rules': {
+        '@nx/dependency-checks': ['error', {
+          'buildTargets': ['build', 'build-builders', 'compile', 'test'],
+          'checkObsoleteDependencies': false,
+          'checkVersionMismatches': false,
+          'ignoredDependencies': ['@o3r/test-helpers', '@o3r/configuration', '@o3r/localization', '@o3r/testing']
```
fpaul-1A, line 31: is that not fixable using Nx `implicitDependencies` ?

packages/@o3r/rules-engine/schemas/rulesets.schema.json
```diff
@@ -51,23 +51,16 @@
...
-          "additionalItems": false,
-          "required": [
-            "library",
-            "name"
-          ]
+          "deprecated": true
+        },
+        "linkedComponents": {
+          "type": "array",
+          "description": "Components linked to the ruleset. If at least one present, it will disable the ruleset execution by default, waiting to a subscription",
```
fpaul-1A, line 60:
```suggestion
          "description": "Components linked to the ruleset. If present, it will disable the ruleset execution by default, waiting for at least one subscription",
```

packages/@o3r/rules-engine/testing/jest.config.it.js
```diff
@@ -1,11 +1,23 @@
 const { dirname } = require('node:path');
-const getJestConfig = require('../../../../jest.config.it').getJestConfig;
+const getJestProjectConfig = require('../../../../jest.config.ut').getJestProjectConfig;
```
fpaul-1A, line 2: this doesn't feel right
maybe we can add a comment to explain why this would be needed ?

packages/@o3r/rules-engine/.eslintrc.js
mrednic-1A: I tried it and it manually and does not do the fix.  
Anyway, I would expect to propose an autofix if it would be the case.

packages/@o3r/rules-engine/testing/jest.config.it.js
mrednic-1A: good point. I parametrized the one from the root so I can pass the angularSetup boolean as true.

packages/@o3r/rules-engine/src/engine/engine.interface.ts
```diff
@@ -74,7 +74,12 @@ export interface EngineRule extends Rule {
 export interface EngineRuleset {
   /** Optional date range where the ruleset will be executed, it supports a dateString or a timestamp as number, more info on javascript Date() documentation */
   validityRange?: { from?: string | number; to?: string | number };
-  /** Component linked to the ruleset, if set it will disable the ruleset execution per default, waiting to a subscription */
+  /** Components linked to the ruleset, if one set, it will disable the ruleset execution per default, waiting to a subscription */
+  linkedComponents?: { library: string; name: string }[];
```
pginoux-1A, line 78: So we settle for a solution with a list and support only the OR use case (execution happens whether the component A OR B OR C etc is active). I think @kpanot  was concerned about having to support a AND use case in the future (Execute the rule only if component A AND B are displayed), we have currently no such requirements, but we were discussing the possibility to have an object as a new structure supporting "and" at the root of the object to be able to add a "or" later if the use case comes up without having to do a breaking change in the structure. On my side I have no strong opinion on this, just want to see if everyone is aligned.

mrednic-1A: update done with the new type which includes 'or'

docs/rules-engine/how-to-use/cms-json-imports.md
```diff
@@ -23,9 +23,13 @@ This can be used as following:
...
+      "linkedComponents": {
+        "or": [
+          {
+            "library": "@scope/components",
+            "name": "TestComponent"
+          }
+        ]
       },
       "rules": [
         ...
```
kpanot, line 35: The language can be switch to `json5` to avoid the red color

docs/rules-engine/how-to-use/dedicated-component-ruleset.md
```diff
@@ -33,20 +41,21 @@ Now that you have the id, you will need to subscribe to the ruleset in your comp
...
 import {LinkableToRuleset, RulesEngineRunnerService} from '@o3r/rules-engine';
-import {SIMPLE_HEADER_PRES_CONFIG_ID} from 'my-component.config';
-import {ExposedComponent} from '@o3r/core';
+import {computeItemIdentifier, type ExposedComponent} from '@o3r/core';
 
 export class MyComponent implements LinkableToRuleset, ExposedComponent {
+
+  private readonly componentIdentifier = computeItemIdentifier('@scope/components', 'TestComponent');
+
   constructor(@Optional() private service: RulesEngineRunnerService) {
```
kpanot, line 50:
```suggestion
  constructor(@Optional() private readonly service: RulesEngineRunnerService) {
```

packages/@o3r/rules-engine/src/components/rules-engine/ruleset-history/ruleset-history-pres.component.ts
```diff
@@ -11,6 +11,32 @@ export type RulesetExecutionDebug = (RulesetExecutionEvent | RulesetExecutionErr
...
+ * The only usecase when we can have duplicates is when 'linkedComponent' and 'linkedComponents' are used in the same time for a ruleset
+ * @deprecated It will be removed in v12 in the same time with 'linkedComponent' property
+ */
+@Pipe({
+  name: 'unique',
+  pure: true
+})
+export class UniquePipe implements PipeTransform {
+
+  public transform(value: ({ library: string; name: string } | undefined)[], _args?: any): any {
```
kpanot, line 25:
```suggestion
  public transform(value: ({ library: string; name: string } | undefined)[]): { library: string; name: string }[] {
```

packages/@o3r/rules-engine/src/components/rules-engine/ruleset-history/ruleset-history-pres.template.html
```diff
@@ -11,8 +11,10 @@ <h4 class="mb-4">Ruleset Execution History</h4>
         [class.error]="execution.type === 'RulesetExecutionError'"
         (click)="toggleExpansion(execution.executionId, 'ruleset')">
         <div><span [title]="'This ruleset has been evaluated ' + execution.executionCounter + ' time(s)'">{{execution.executionCounter}}</span> - {{execution.rulesetName | titlecase }}
-          <div class="ruleset-panel-subtitle" *ngIf="execution.rulesetInformation?.linkedComponent as linkedComponent">
-            {{linkedComponent.name}} {{linkedComponent.library}}
+          <div class="ruleset-panel-subtitle" *ngIf="execution.rulesetInformation?.linkedComponents.or || execution.rulesetInformation?.linkedComponent">
+            <ng-container *ngFor="let lc of (execution.rulesetInformation?.linkedComponents.or || []).concat(execution.rulesetInformation?.linkedComponent) | unique; last as isLast">
```
kpanot, line 15:
```suggestion
            <ng-container *ngFor="let lc of (execution.rulesetInformation.linkedComponents.or || [execution.rulesetInformation.linkedComponent]) | unique; last as isLast">
```

packages/@o3r/rules-engine/src/engine/engine.interface.ts
```diff
@@ -74,7 +74,15 @@ export interface EngineRule extends Rule {
...
   validityRange?: { from?: string | number; to?: string | number };
-  /** Component linked to the ruleset, if set it will disable the ruleset execution per default, waiting to a subscription */
+  /**
+   * Components linked to the ruleset. If present the ruleset will not be active by default.
+   * 'or' condition: If at least one component has subscribed, the ruleset will become active.
+   */
+  linkedComponents?: Record<'or', { library: string; name: string }[]>;
+  /**
+   * Component linked to the ruleset, if set it will disable the ruleset execution per default, waiting to a subscription
+   * @deprecated It will be removed in v12, use linkedComponents instead
```
kpanot, line 84:
```suggestion
   * @deprecated It will be removed in v12, use {@link linkedComponents} instead
```

packages/@o3r/rules-engine/src/stores/rulesets/rulesets.selectors.ts
```diff
@@ -39,7 +39,7 @@ export const selectActiveRuleSets = createSelector(
   selectAllRulesets,
   (ruleSets) => ruleSets.filter((ruleSet: Ruleset) => {
 
-    if (ruleSet.linkedComponent) {
+    if (ruleSet.linkedComponents?.or?.length || ruleSet.linkedComponent) {
```
cpaulve-1A, line 42: why should we return false if there are linkedComponents?
I know this was the behaviour before, but I fail to understand why we would only consider only ruleset applied without this condition and call this selector "selectActiveRulesets"

cpaulve-1A (approved):

packages/@o3r/rules-engine/src/stores/rulesets/rulesets.selectors.ts
mrednic-1A: ruleset which have defined linkedComponents ar not considered active by default. 
That's why they are not considered in this selector and they have their own selector. You are right about the naming. It should be 'selectDefaultRulesets' or 'selectDefaultActiveRulesets'. 
Do you think I should create a second one and deprecate the existing one ?

packages/@o3r/rules-engine/src/components/rules-engine/ruleset-history/ruleset-history-pres.template.html
mrednic-1A: with this change I do not support both linkedComponent and linkedComponents in the same time.

packages/@o3r/rules-engine/src/engine/structure.ts
```diff
@@ -136,6 +136,14 @@ export interface Ruleset {
   rules: Rule[];
   /** Optional date range where the ruleset will be executed*/
   validityRange?: {from?: string; to?: string};
-  /** Component linked to the ruleset, if set it will disable the ruleset execution per default, waiting to a subscription */
+  /**
+   * Components linked to the ruleset. If present the ruleset will not be active by default.
+   * 'or' condition: If at least one component has subscribed, the ruleset will become active.
+   */
+  linkedComponents?: Record<'or', {library: string; name: string}[]>;
```
matthieu-crouzet, line 143: I know we already had some discussion on the format but what will happened when we will add `and`
People will be able to have an object like this 
```json5
{
  linkedComponents: {
    or: [...],
    and: [...]
}
```
And I don't see how we can handle `or` and `and` at the same time
I think we should have something like this
```typescript
type ComponentList = {library: string; name: string}[];
type GroupOperator = 'or';
type GroupComponents = 
   | { operator: GroupOperator, components: ComponentList } 
   | { operator: GroupOperator, groups: GroupComponents[] }
linkedComponents?: GroupComponents;
```
So like this we will be able to support use case like this
```json5
linkedComponents: {
  operator: 'or',
  groups: [
    {
      operator: 'and',
      components: [
        {library: '@o3r/lib'; name: 'AndExample1Component'}, 
        {library: '@o3r/lib'; name: 'AndExample2Component'}
      ]
    },
    {
       operator: 'or',
       components: [
         {library: '@o3r/lib'; name: 'OrExample1Component'}, 
         {library: '@o3r/lib'; name: 'OrExample2Component'}
       ]
    }
  ]
}
```
and would lead to this logic
`(AndExample1Component && AndExample2Component) || (OrExample1Component || OrExample2Component)`

mrednic-1A: I think we are going to far here with the nested groups.
We don't even have the usecase of a simple 'AND'.
With the actual implementation we support a list of linked components driven by 'OR' operator and if we want we'll have another list with 'AND' operator. 
If in some day we'll want to support the 'AND' we will need one selector for 'OR' components and one selector for 'AND'. Nowadays the 'OR' selector is in place (the default one we have).
I let the others give their opinion too :)

packages/@o3r/rules-engine/src/components/rules-engine/ruleset-history/ruleset-history-pres.template.html
mrednic-1A: let's support _linkedComponents_ first and fallback on _linkedComponent_.

docs/ab-testing/AB_TESTING.md
```diff
@@ -201,9 +201,13 @@ You can now create A/B testing-driven rulesets .
     {
       "id": "1",
       "name": "AB testing rule",
-      "linkedComponent": {
```
vscaiceanu-1a, line 204: open question: is it worth leaving the `linkedComponent` in the documentation and mention that it is deprecated and will be removed in v12? And adding an example with `linkedComponents` as well.

packages/@o3r/rules-engine/schemas/rulesets.schema.json
```diff
@@ -51,22 +51,25 @@
...
+              "items": {
+                "$ref": "#/definitions/LinkedComponent"
+              }
             }
           },
           "additionalItems": false,
           "required": [
-            "library",
-            "name"
+            "or"
```
matthieu-crouzet, line 72: Should it be `required` if we want to be backward compatible at some point and introduce the `and` use case ?

packages/@o3r/rules-engine/src/engine/engine.interface.ts
```diff
@@ -74,7 +74,16 @@ export interface EngineRule extends Rule {
 export interface EngineRuleset {
   /** Optional date range where the ruleset will be executed, it supports a dateString or a timestamp as number, more info on javascript Date() documentation */
   validityRange?: { from?: string | number; to?: string | number };
-  /** Component linked to the ruleset, if set it will disable the ruleset execution per default, waiting to a subscription */
+  /**
+   * Components linked to the ruleset. If present the ruleset will not be active by default.
+   * 'or' condition: If at least one component has subscribed, the ruleset will become active.
+   * If present, the {@link linkedComponent} property will not be taken into consideration
+   */
+  linkedComponents?: {or: { library: string; name: string }[]};
```
matthieu-crouzet, line 82: We have this type `{ library: string; name: string }` at a lot of place in Otter Framework 
We may open an issue to create a reusable type no ?

packages/@o3r/rules-engine/schemas/rulesets.schema.json
mrednic-1A: If I have 'linkedComponents' I want to have the 'or' always, but when adding the 'and' we'll have either one or the other :)) 
I think if we remove the required from the 'or' we stay backward compatible, no ?

packages/@o3r/rules-engine/src/engine/engine.interface.ts
mrednic-1A: Good point . Either an issue or I could do it in this PR. It depends in how many places and packages it is used.

docs/ab-testing/AB_TESTING.md
mrednic-1A: We should inform somehow that 'linkedComponent' is deprecated and 'linkedComponents' should be used instead. 
I agree that just changing the examples is not enough, mostly for the users which have already used 'linkedComponent' will be misleading.

kpanot (dismissed):

vscaiceanu-1a (approved):

packages/@o3r/rules-engine/schemas/rulesets.schema.json
matthieu-crouzet: For our devtools I would agree, but we are also providing our metadata to CMS. If `or` is required in `linkedComponents` and we set it to optional later, we will introduce a breaking change.
We can make it optional directly I think. It should not matter on our side.

mrednic-1A: indeed. Let's do that :)
This is a nice catch :) thanks

fpaul-1A (approved):

kpanot (dismissed):

matthieu-crouzet (approved):

fpaul-1A (approved):

kpanot (approved):

Issue #1606, opened 2024-04-08
[Bug]: tests not running on new otter webapp


vscaiceanu-1a:
### Package name

testing

### Package version

10.1.1

### Reproduction steps

```shell
npm create @o3r otter-project
cd otter-project
ng g application webapp
# choose preset BASIC, Jest and Playwright when prompted
npm run test -w webapp
```

### Current result

```shell
$ npm run test -w webapp

> webapp@0.0.0-placeholder test
> jest

Error: Cannot find module '../../tsconfig.base'
Require stack:
- /Users/vscaiceanu/workspace/temp/quickstart-otter/jest.config.ut.js
- /Users/vscaiceanu/workspace/temp/quickstart-otter/apps/webapp/jest.config.js
...
```

### Expected result

```shell
$ npm run test -w webapp

> webapp@0.0.0-placeholder test
> jest

 PASS   webapp  src/app/app.component.spec.ts
  AppComponent
    ✓ should create the app (51 ms)
    ✓ should have the 'webapp' title (6 ms)
    ✓ should render title (10 ms)

Test Suites: 1 passed, 1 total
Tests:       3 passed, 3 total
Snapshots:   0 total
Time:        2.395 s
Ran all test suites.
```

vscaiceanu-1a:
fixed by #1607

PR #1607, opened 2024-04-08
fix: jest test not running on new otter webapp

vscaiceanu-1a:
## Related issues

- :bug: #1606

### DIFF
--- a/packages/@o3r/testing/schematics/ng-add/fixture/index.ts
+++ b/packages/@o3r/testing/schematics/ng-add/fixture/index.ts
@@ -8,10 +8,9 @@ import * as ts from 'typescript';
  * Add fixture configuration
  * @param options @see RuleFactory.options
  * @param options.projectName
- * @param registerUnitTestFixturePaths should add path mapping for unit test fixtures
  * @param options.testingFramework
  */
-export function updateFixtureConfig(options: { projectName?: string | null | undefined; testingFramework?: string | null | undefined }, registerUnitTestFixturePaths: boolean): Rule {
+export function updateFixtureConfig(options: { projectName?: string | null | undefined; testingFramework?: string | null | undefined }): Rule {
 
   const oldPaths = ['@otter/testing/core', '@otter/testing/core/*'];
   /**
@@ -90,12 +89,6 @@ export function updateFixtureConfig(options: { projectName?: string | null | und
       }
       return acc;
     }, {});
-
-    if (registerUnitTestFixturePaths) {
-      configWithPath.content.compilerOptions.baseUrl ||= '.';
-      configWithPath.content.compilerOptions.paths['@o3r/testing/core'] = ['node_modules/@o3r/testing/core/angular'];
-      configWithPath.content.compilerOptions.paths['@o3r/testing/core/*'] = ['node_modules/@o3r/testing/core/angular/*'];
-    }
 
     tree.overwrite(configWithPath.tsconfig, JSON.stringify(configWithPath.content, null, 2));
 
--- a/packages/@o3r/testing/schematics/ng-add/templates/workspace/jest.config.ut.js.template
+++ b/packages/@o3r/testing/schematics/ng-add/templates/workspace/jest.config.ut.js.template
@@ -1,8 +1,8 @@
-const { relative } = require('node:path');
+const { join, relative } = require('node:path');
 const { pathsToModuleNameMapper } = require('ts-jest');
-// In the following statement, replace `<%= rootRelativePath %>/tsconfig.base` with the path to your `tsconfig` file
+// In the following statement, replace `./tsconfig.json` with the path to your `tsconfig` file
 // which contains the path mapping (ie the `compilerOptions.paths` option):
-const { compilerOptions } = require('<%= rootRelativePath %>/tsconfig.base');
+const { compilerOptions } = require('./tsconfig.json');
 
 globalThis.ngJest = {
   skipNgcc: true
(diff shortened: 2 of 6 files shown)


packages/@o3r/testing/schematics/ng-add/templates/workspace/jest.config.ut.js.template
```diff
@@ -1,8 +1,8 @@
 const { relative } = require('node:path');
 const { pathsToModuleNameMapper } = require('ts-jest');
-// In the following statement, replace `<%= rootRelativePath %>/tsconfig.base` with the path to your `tsconfig` file
+// In the following statement, replace `./tsconfig` with the path to your `tsconfig` file
 // which contains the path mapping (ie the `compilerOptions.paths` option):
-const { compilerOptions } = require('<%= rootRelativePath %>/tsconfig.base');
+const { compilerOptions } = require('./tsconfig');
```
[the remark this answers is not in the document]
vscaiceanu-1a: for an Otter app, the root `tsconfig.json` isn't only used for IDE only.
It looks something like this:
```json
{
  "compileOnSave": false,
  "compilerOptions": {
    "outDir": "./dist/out-tsc",
    "forceConsistentCasingInFileNames": true,
    "strict": true,
    "noImplicitOverride": true,
    "noImplicitReturns": true,
    "noFallthroughCasesInSwitch": true,
    "skipLibCheck": true,
    "esModuleInterop": true,
    "sourceMap": true,
    "declaration": false,
    "experimentalDecorators": true,
    "moduleResolution": "node",
    "importHelpers": true,
    "target": "ES2022",
    "module": "ES2022",
    "useDefineForClassFields": false,
    "lib": [
      "ES2022",
      "dom"
    ],
    "paths": {
      "@o3r/testing/core": [
        "node_modules/@o3r/testing/core/angular"
      ],
      "@o3r/testing/core/*": [
        "node_modules/@o3r/testing/core/angular/*"
      ]
    },
    "baseUrl": "."
  },
  "angularCompilerOptions": {
    "enableI18nLegacyMessageIdFormat": false,
    "strictInjectionParameters": true,
    "strictInputAccessModifiers": true,
    "strictTemplates": true
  }
}
```
Most important, it contains the path mapping

packages/@o3r/testing/schematics/ng-add/templates/workspace/jest.config.ut.js.template
```diff
@@ -14,11 +14,12 @@ globalThis.ngJest = {
  * @returns {import('ts-jest/dist/types').JestConfigWithTsJest}
  */
 module.exports.getJestProjectConfig = (rootDir, isAngularSetup) => {
-  const relativePath = relative(rootDir, '.');
-  const moduleNameMapper = Object.entries(pathsToModuleNameMapper(compilerOptions.paths)).reduce((acc, [moduleName, path]) => {
+  const relativePath = relative(rootDir, __dirname);
+  const moduleNameMapper = Object.entries(pathsToModuleNameMapper(compilerOptions.paths || {})).reduce((acc, [moduleName, path]) => {
```
vscaiceanu-1a, line 18: This is also handling the deep imports. No need for extra mappers like `moduleNameMapper['^@o3r/testing/core/(.*)'] = [join(require.resolve('@o3r/testing/core/angular'), '$1')];`

packages/@o3r/testing/schematics/ng-add/templates/workspace/jest.config.ut.js.template
```diff
@@ -14,11 +14,12 @@ globalThis.ngJest = {
...
     acc[moduleName] = `<rootDir>/${relativePath}/${path}`;
     return acc;
   }, {});
+  moduleNameMapper['^@o3r/testing/core$'] = [require.resolve('@o3r/testing/core/angular')];
```
vscaiceanu-1a, line 22: This is also handling the deep imports. No need for extra mappers like `moduleNameMapper['^@o3r/testing/core/(.*)'] = [join(require.resolve('@o3r/testing/core/angular'), '$1')];`

packages/@o3r/testing/schematics/ng-add/fixture/index.ts
```diff
@@ -8,10 +8,9 @@ import * as ts from 'typescript';
  * Add fixture configuration
  * @param options @see RuleFactory.options
  * @param options.projectName
- * @param registerUnitTestFixturePaths should add path mapping for unit test fixtures
  * @param options.testingFramework
  */
-export function updateFixtureConfig(options: { projectName?: string | null | undefined; testingFramework?: string | null | undefined }, registerUnitTestFixturePaths: boolean): Rule {
+export function updateFixtureConfig(options: { projectName?: string | null | undefined; testingFramework?: string | null | undefined }): Rule {
```
matthieu-crouzet, line 13: we can't do breaking change

packages/@o3r/testing/schematics/ng-add/templates/workspace/jest.config.ut.js.template
matthieu-crouzet: Why not doing the change in the tsconfig ?

vscaiceanu-1a: very good question. Because in `tsconfig` we should point to the implementation in `node_modules` but this does not work with `pnp`

packages/@o3r/testing/schematics/ng-add/fixture/index.ts
vscaiceanu-1a: we cannot do a breaking change if the method is exposed from the package and here it is not the case. 
image

packages/@o3r/testing/schematics/ng-add/templates/workspace/jest.config.ut.js.template
kpanot: The regexp does explicitly not match deep import, are you sure it is supported?

vscaiceanu-1a: I don't know how jest works under the hood with `moduleNameMapper` but I did some tests and I am pretty sure deep imports are supported.

mrednic-1A (approved):

fpaul-1A (approved):

kpanot (approved):

PR #1610, opened 2024-04-09
fix(chrome-ext): patch manifest.json with the right version number

### LINKED ISSUE #1517: [Bug]: Manifest.json does not contain the right version number
### Application name

Chrome Devtools

### Version

10

### Reproduction steps

Generate a new version of the Otter Chrome Devtools

### Current result

The `manifest.json` generated contains the version `0.0.0` instead of the right version

### Expected result

Generate the `manifest.json` with the version aligned with the package.json`

### Additional comments

_No response_

matthieu-crouzet:
## Related issues

- :bug: Fixes #1517

### DIFF
--- a/apps/chrome-devtools/package.json
+++ b/apps/chrome-devtools/package.json
@@ -13,18 +13,17 @@
   ],
   "scripts": {
     "build": "yarn nx build chrome-devtools",
-    "postbuild:patch": "yarn patch:manifest && yarn patch:extension && yarn patch:package",
+    "postbuild:patch": "yarn patch:package && yarn patch:manifest && yarn patch:extension",
     "patch:package": "cpy 'package.json' 'dist' && patch-package-json-main",
     "copy:assets": "cpy 'src/manifest.json' 'src/devtools.html' dist --flat && cpy './src/assets/**' dist",
     "patch:manifest": "node scripts/sanitize-manifest.cjs",
     "build:extension": "tsc -b tsconfig.extension.json",
-    "patch:extension": "node scripts/sanitize-extension.cjs",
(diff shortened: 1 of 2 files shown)


vscaiceanu-1a (dismissed):

apps/chrome-devtools/package.json
```diff
@@ -13,18 +13,17 @@
...
     "patch:manifest": "node scripts/sanitize-manifest.cjs",
     "build:extension": "tsc -b tsconfig.extension.json",
-    "patch:extension": "node scripts/sanitize-extension.cjs",
+    "patch:extension": "node scripts/sanitize-extension.cjs && node scripts/set-manifest-version.cjs",
     "start": "concurrently 'yarn:watch:*'",
     "watch:panel": "yarn build:panel --watch",
     "watch:components": "yarn build:components --watch",
     "watch:manifest": "yarn patch:manifest --watch",
     "watch:extension": "tsc-watch -b tsconfig.extension.json --noClear --onSuccess 'yarn patch:extension'",
-    "prepare:publish": "prepare-publish ./dist && node scripts/set-manifest-version.cjs",
```
kpanot, line 27: This is used by the `prepare-publish` task. You may need to keep `prepare-publish ./dist` or remove the task

matthieu-crouzet: I removed the task it was not called in this package

kpanot: It's called at the workspace level for all the packages.
If I remember well, it is updating the package.json to add the common information (such as the contributors).
What I don't know it is if this added information is meaningful in the context of Chrome Extension of if this information should be set in the manifest.json instead.
But not sure it should be removed without be replaced by equivalent :S.

vscaiceanu-1a (approved):

PR #1611, opened 2024-04-09
[draft] fix(stylelint-plugin): support stylelint 16 with esm implementation

fpaul-1A:
## Proposed change

`@o3r/stylelint-plugin` is currently broken when using `stylelint@16` with the following error:
> Error: stylelint requires plugins to expose a ruleName

To fix this, we need to build the files as ESM modules instead of CommonJS

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


jest.config.ut.js
```diff
@@ -35,7 +35,7 @@ module.exports.getJestProjectConfig = (rootDir, isAngularSetup) => {
     },
     transform: {
       // eslint-disable-next-line @typescript-eslint/naming-convention
-      '^.+\\.tsx?$': [
+      '^.+\\.m?tsx?$': [
```
kpanot, line 38:
```suggestion
      '^.+\\.[mc]?tsx?$': [
```
To get everything

packages/@o3r/stylelint-plugin/package.json
```diff
@@ -16,8 +16,8 @@
   "scripts": {
     "nx": "nx",
     "ng": "yarn nx",
-    "build": "yarn nx build stylelint-plugin",
-    "postbuild": "patch-package-json-main",
+    "build": "tsc -b tsconfig.build.json",
```
kpanot, line 19: Generally we keep the `build` to run the nx task and do something like

```suggestion
    "build:source": "tsc -b tsconfig.build.json",
```

packages/@o3r/stylelint-plugin/testing/mjs-resolver.js
```diff
@@ -0,0 +1,21 @@
...
+ * Copied from https://github.com/kulshekhar/ts-jest/blob/main/e2e/native-esm-ts/mjs-resolver.ts
+ * @param path
+ * @param options
+ */
+const mjsResolver = (path, options) => {
+  const mjsExtRegex = /\.mjs$/i
+  const resolver = options.defaultResolver
+  if (mjsExtRegex.test(path)) {
+    try {
+      return resolver(path.replace(mjsExtRegex, '.mts'), options)
```
kpanot, line 12:
```suggestion
      return resolver(path.replace(mjsExtRegex, '.mts'), options);
```
Missing linting on this file

packages/@o3r/stylelint-plugin/testing/mjs-resolver.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Jest resolver to replace .mjs imports with .mts for ESM compatibility
+ * Copied from https://github.com/kulshekhar/ts-jest/blob/main/e2e/native-esm-ts/mjs-resolver.ts
+ * @param path
+ * @param options
+ */
+const mjsResolver = (path, options) => {
+  const mjsExtRegex = /\.mjs$/i
```
kpanot, line 8: Should be declared outside of the function

matthieu-crouzet (approved):

packages/@o3r/stylelint-plugin/package.json
```diff
@@ -26,9 +27,16 @@
     "tslib": "^2.5.3"
   },
   "peerDependencies": {
+    "@schematics/angular": "~17.2.0",
     "postcss": "^8.4.5",
+    "postcss-scss": "^4.0.9",
```
mrednic-1A, line 32: I could not find where this dep is used . Is it something we have forgotten to add before ?

fpaul-1A: I originally added it because I wanted the ng-add to install it, this might not be needed
then again, it doesn't really make sense to use this plugin without `postcss-scss` in the stylelint config

vscaiceanu-1a (approved):

mrednic-1A (approved):

Issue #1547, opened 2024-03-25
[Feature]: Visualize Placeholder through a Toggle


ntrotner:
### Context

When running an app I'd like to have the option to visualize all available placeholders. This would allow users to easily get a picture of where placeholders exists and what their name is, instead of looking it up through the HTML DOM.

Example:
![Uploading placeholder.png…]()

### Proposal

- [ ] Add a toggle to the debug panel for the visualization
- [ ] Fill placeholder with content
  - required: 
    - id
  - optional
    - which fragments are included in the placeholder already and what is their priority
    - which rules target this placeholder

PR #1615, opened 2024-04-10
feat(chrome-ext): toggle button for placeholder

matthieu-crouzet:
## Proposed change
image

## Related issues

- :rocket: Feature #1547


apps/showcase/src/style/dark-theme/dark-theme.scss
fpaul-1A, line 1: we'll fix this one day

apps/showcase/update-screenshots.cjs
fpaul-1A, line 1: should be reverted, probably

apps/showcase/src/app/placeholder/placeholder.style.scss
```diff
@@ -0,0 +1,3 @@
+o3r-placeholder {
```
cpaulve-1A, line 1: Should we keep this file?

apps/showcase/src/app/placeholder/placeholder.component.ts
```diff
@@ -0,0 +1,67 @@
...
+    rulesEngine: RulesEngineRunnerService,
+    placeholderRulesEngineActionHandler: PlaceholderRulesEngineActionHandler
+  ) {
+    this.tripFactsService.register();
+    rulesEngine.actionHandlers.add(placeholderRulesEngineActionHandler);
+    void this.loadRuleSet();
+  }
+
+  private async loadRuleSet() {
+    const path = await firstValueFrom(
```
mrednic-1A, line 52: __THOUGHT__
We have to places where we load the same 'rulesets.json' file. 
Would not be better to load it once at app level?
I am just trying to see how it could be interpreted by someone looking for an example of rules engine usage. They will load it as we do, at route level.
From what I remember from the guidelines, the rulesets are loaded once at app level and linked to components if we don;t want them to be active from the beginning. 
Opinions ?

matthieu-crouzet: I think the idea behind was to keep the app component as simple as possible and to add the logic in each page to understand what is needed for this feature.
I agree that it could be confusing we can discuss that with the team :)

apps/showcase/src/app/placeholder/placeholder.style.scss
matthieu-crouzet: As discussed we kept it for the others pages, we can removed them in another PR if you think it can be better without

apps/showcase/src/app/placeholder/placeholder.component.ts
matthieu-crouzet: As discussed we will add a comment to explain that for the sake of example we put it in the page logic but usually we recommend to set it in the app component

cpaulve-1A (approved):

fpaul-1A (approved):

PR #1616, opened 2024-04-10
fix(sdk-create): cannot create an sdk with debug mode

fpaul-1A:
## Proposed change

Currently running `npm create @ama-sdk typescript test --debug` triggers an exception on spawnSync ENOENT
This is because --debug enables --dry-run by default on angular schematics
We are not currently supporting dry-run mode for generators because we are running shell commands during generation so the files need to be present on disk

To use debug mode, we need to be able to override the `--dry-run` parameter

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@ama-sdk/create/src/index.ts
+++ b/packages/@ama-sdk/create/src/index.ts
@@ -66,7 +66,8 @@ const schematicArgs = [
   '--name', name,
   '--package', pck,
   '--package-manager', packageManager,
-  ...(typeof argv['o3r-metrics'] !== 'undefined' ? [`--${!argv['o3r-metrics'] ? 'no-' : ''}o3r-metrics`] : [])
+  ...(typeof argv['dry-run'] !== 'undefined' ? [`--${!argv['dry-run'] || argv['dry-run'] === 'false' ? 'no-' : ''}dry-run`] : []),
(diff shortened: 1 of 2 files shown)


packages/@ama-sdk/create/src/index.ts
```diff
@@ -66,7 +66,8 @@ const schematicArgs = [
   '--name', name,
   '--package', pck,
   '--package-manager', packageManager,
-  ...(typeof argv['o3r-metrics'] !== 'undefined' ? [`--${!argv['o3r-metrics'] ? 'no-' : ''}o3r-metrics`] : [])
+  ...(typeof argv['dry-run'] !== 'undefined' ? [`--${!argv['dry-run'] || argv['dry-run'] === 'false' ? 'no-' : ''}dry-run`] : []),
+  ...(typeof argv['o3r-metrics'] !== 'undefined' ? [`--${!argv['o3r-metrics'] || argv['o3r-metrics'] === 'false' ? 'no-' : ''}o3r-metrics`] : [])
```
cpaulve-1A, line 70: Could be nice to throw if user calls the schematics with dry-run

fpaul-1A: I think this check should be done at schematics level, for all of them

kpanot (approved):

Issue #1617, opened 2024-04-10
[Feature]: Add the capability to use the placeholder without the Rules Engine


kpanot:
### Context

The purpose of this task is to be able to use the placeholder feature without requiring the Effect from `@o3r/components/rules-engine`.

### Proposal

Today the `@o3r/components/rules-engine` is bringing the dependencies to the different features to regen the template of the Rules Engine, the purpose of able to split the Effects to turn smaller part and be able to turn the dependency optionals.

PR #1621, opened 2024-04-10
feat(components): expose the placeholder service in @o3r/components

kpanot:
## Proposed change

Expose the service to provide Placeholder Template URLs in the `@o3r/components` package instead on the rules engine.
The next (followed by #1617) is let the Effect also accessible via `@o3r/components`

## Related issues

- :rocket: Feature #1617


docs/rules-engine/how-to-use/placeholders.md
```diff
@@ -135,15 +148,52 @@ You need to reference one or several CSS files from your application in the `cms
...
+## Manual usage of the Placeholder
+
+The Placeholder does not requires the Rules Engine to be used and can be integrated to your application independently.
+
+To do so you will need to import the `PlaceholderModule` in your application (as described in the [preview section](#/inside-an-application)) and describe the template to set to your application placeholders in the following manner:
+
+```typescript
+import { EffectsModule } from '@ngrx/effects';
+import { PlaceholderService, PlaceholderTemplateResponseEffect } from '@o3r/components';
+import { PlaceholderTemplateResponseEffect } from '@o3r/components/rules-engine';
```
mrednic-1A, line 175: imported 2 times. You can remove it from the previous line

packages/@o3r/components/src/stores/placeholder-effect/placeholder.effect.ts
```diff
@@ -34,18 +31,26 @@ export class PlaceholderTemplateResponseEffect {
...
-                // eslint-disable-next-line new-cap
-                factValue: (fact.path && factValue) ? JSONPath({ wrap: false, json: factValue, path: fact.path }) : factValue
-              })),
-              distinctUntilChanged((previous, current) => previous.factValue === current.factValue)
-            )) : [];
-
-          const factsStreamsList$ = factsStreamsList.length ? combineLatest(factsStreamsList) : of([]);
+          const factsStreamsList$ = from(import('@o3r/rules-engine')).pipe(
+            // eslint-disable-next-line @typescript-eslint/naming-convention
+            switchMap(({ RulesEngineRunnerService }) => {
```
mrednic-1A, line 36: With the dynamic import you'll always have a value for the service even if it is not provided in the app, but the module is installed. So I guess we are losing the 'optional' feature here.
Correct me if I am wrong.

docs/rules-engine/how-to-use/placeholders.md
kpanot: Actually I need to remove this line :).
But thanks for the notice

packages/@o3r/components/src/stores/placeholder-effect/placeholder.effect.ts
kpanot: If "a value" is "a class reference" yes, but it does not mean that we have the class instance.
According where it is injected (if not injected inRoot), it can be present only if the module it is attached to is loaded.

mrednic-1A: I see, thanks for the explanation :)

PR #1641, opened 2024-04-12
fix(e2e): update screenshots scripts

matthieu-crouzet:


fpaul-1A (dismissed):

apps/showcase/scripts/update-screenshots/index.cjs
```diff
@@ -0,0 +1,36 @@
...
+const ipAddresses = [
+  'http://host.docker.internal:8080',
+  ...Object.values(os.networkInterfaces())
+    .flat()
+    .filter((net) => net.family === 'IPv4' && net.address !== '127.0.0.1')
+    .map((net) => `http://${net.address}:8080`)
+];
+
+// Command to create the docker container and run the script
+const script = `podman run -it --rm --ipc=host -v ${absolutePathFromRoot}/:${mountPath} mcr.microsoft.com/playwright:v${playwrightVersion}-jammy /bin/bash ${pathToPodmanScript} ${ipAddresses.join(',')}`;
```
fpaul-1A, line 32: I think we should keep docker here, as it's more likely people using podman have an alias to docker than the other way around
```suggestion
const script = `docker run -it --rm --ipc=host -v ${absolutePathFromRoot}/:${mountPath} mcr.microsoft.com/playwright:v${playwrightVersion}-jammy /bin/bash ${pathToPodmanScript} ${ipAddresses.join(',')}`;
```

apps/showcase/scripts/update-screenshots/index.cjs
```diff
@@ -0,0 +1,36 @@
...
+    .filter((net) => net.family === 'IPv4' && net.address !== '127.0.0.1')
+    .map((net) => `http://${net.address}:8080`)
+];
+
+// Command to create the docker container and run the script
+const script = `docker run -it --rm --ipc=host -v ${relativeRoot}/:${mountPath} mcr.microsoft.com/playwright:v${playwrightVersion}-jammy /bin/bash ${pathToPodmanScript} ${ipAddresses.join(',')}`;
+console.log(`Executing: ${script}`);
+
+// Execute the script
+childProcess.execSync(script, {stdio: 'inherit'});
```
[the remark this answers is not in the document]
cpaulve-1A: As discussed, the issue raised here is less about security and more avoiding bugs if the user has unexpected characters in its path (space for instance). Since the alternative might not work with windows, let's keep the alert and investigate a solution working both for windows and space in directories

fpaul-1A (approved):

apps/showcase/scripts/update-screenshots/podman-script.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# Move to the folder
+cd tests-clone
+
+# Delete the existing screenshots
+echo 'Delete previous screenshots'
+rm -rdf apps/showcase/e2e-playwright/sanity/screenshots
+
+# Configure the global cache to target a folder outside the container to have it available for the following runs
+yarn config set enableGlobalCache true
```
kpanot, line 26: Would be safer to do a `corepack enable` before (if not already done in the docker config)

fpaul-1A (approved):

kpanot (approved):

PR #1677, opened 2024-04-17
fix: generate angular app with module

### LINKED ISSUE #1676: [Bug]: prefersReducedMotion is missing from new Otter webapp
### Package name

application

### Package version

10.1.6

### Reproduction steps

```shell
npm create @o3r otter-project
cd otter-project
ng g application webapp
```

### Current result

`prefersReducedMotion` is imported and unused in `app.config.ts`. It is also causing linter issues.

### Expected result

`prefersReducedMotion` is imported and used in the new Otter app

### Additional comments

Might be because Angular application schematics generates standalone app by default (no app.module) and our schematics is adding `prefersReducedMotion` in the app.module file

vscaiceanu-1a:
## Proposed change

Our schematics do not currently support standalone Angular apps.

## Related issues

- :bug: Fixes #1676

### DIFF
--- a/packages/@o3r/workspace/schematics/application/index.ts
+++ b/packages/@o3r/workspace/schematics/application/index.ts
@@ -83,6 +83,7 @@ function generateApplicationFn(options: NgGenerateApplicationSchema): Rule {
         ...Object.entries(extendedOptions).reduce((acc, [key, value]) => (angularOptions.includes(key) ? {...acc, [key]: value} : acc), {}),
         name: cleanName,
         projectRoot,
+        standalone: false,
         style: Style.Scss}),
       addProjectSpecificFiles(targetPath, rootDependencies),
       updateProjectTsConfig(targetPath, 'tsconfig.app.json'),


packages/@o3r/workspace/schematics/application/index.ts
```diff
@@ -83,6 +83,7 @@ function generateApplicationFn(options: NgGenerateApplicationSchema): Rule {
         ...Object.entries(extendedOptions).reduce((acc, [key, value]) => (angularOptions.includes(key) ? {...acc, [key]: value} : acc), {}),
         name: cleanName,
         projectRoot,
+        standalone: false,
```
matthieu-crouzet, line 86: IMO the fix should be on the schematic to add `prefersReducedMotion` if it's the only part that does not work with standalone application.

Else we should not expose a `standalone` option in the schema.

vscaiceanu-1a: I was wrong when I thought we do not support standalone apps. We should fix the generator and add a test to make sure the content is correct.
Closing this PR.

vscaiceanu-1a:
I was wrong when I thought we do not support standalone apps. We should fix the generator and add a test to make sure the content is correct.
Closing this PR.

PR #1678, opened 2024-04-17
feat(sdk): generate an sdk with spec from npm

fpaul-1A:
## Proposed change

The goal is to be able to generate an SDK with the necessary setup to fetch the spec file form an npm registry

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


packages/@ama-sdk/core/cli/update-spec-from-npm.cts
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env node
```
kpanot, line 1: would be maybe simpler to generate this file in `.cjs` within the sdk repo

fpaul-1A: I don't see how this would make things simpler, at least from the user point of view

kpanot: For the user point of view it will have access to the script that it can trick if needed. It will also reduce the number of required inputs.

For our point of view we do not need to expose a new CLI.

fpaul-1A: I personally see this as unnecessary boilerplate in the generated SDK, and more difficult to update with migration scripts

packages/@ama-sdk/core/package.json
```diff
@@ -168,6 +168,7 @@
   "schematics": "./collection.json",
   "bin": {
     "amasdk-clear-index": "./dist/cli/clear-index.cjs",
-    "amasdk-files-pack": "./dist/cli/files-pack.cjs"
+    "amasdk-files-pack": "./dist/cli/files-pack.cjs",
+    "amasdk-update-spec-from-npm": "./dist/cli/update-spec-from-npm.cjs"
```
vscaiceanu-1a, line 172: Is this new script documented anywhere?

packages/@ama-sdk/core/cli/update-spec-from-npm.cts
kpanot: (I am still not convince but) Ok then you will need to provide documentation (and probably UTs) to the CLI because SDK user will not have access to it directly.

packages/@ama-sdk/core/src/fwk/open-api-tools-configuration.ts
```diff
@@ -0,0 +1,27 @@
...
+  output: string;
+  /** Generator to use */
+  generatorName: string;
+  /** Path to configuration file. It can be JSON or YAML */
+  config?: string;
+  /** Sets specified global properties */
+  globalProperty?: string | Record<string, any>;
+}
+
+export interface OpenApiToolsGeneratorObject {
```
kpanot, line 14: missing comments to export interfaces and fields

packages/@ama-sdk/create/README.md
```diff
@@ -43,5 +43,10 @@ npm create @ama-sdk typescript <project-name> -- --package-manager=yarn [...opti
 - `--o3r-metrics`: Enable or disable the collection of anonymous data for Otter
 - `--exact-o3r-version` : use a pinned version for [otter packages](https://github.com/AmadeusITGroup/otter/blob/main/docs/README.md).
```
kpanot, line 45: You may need to regroup `--spec-path` as well

packages/@ama-sdk/schematics/schematics/ng-add/index.ts
```diff
@@ -79,7 +80,7 @@ const createOpenApiToolsConfig: Rule = (tree) => {
   const openApiDefaultStorageDir = '.openapi-generator';
   if (tree.exists(openApiConfigPath)) {
     // eslint-disable-next-line @typescript-eslint/naming-convention
-    const openapitoolsConfig = tree.readJson(openApiConfigPath) as { 'generator-cli'?: { storageDir?: string; version?: string } } || {};
+    const openapitoolsConfig = (tree.readJson(openApiConfigPath) || {}) as any as OpenApiToolsConfiguration;
```
kpanot, line 83: `{}` is not a value `OpenApiToolsConfiguration`

packages/@ama-sdk/schematics/schematics/typescript/core/index.spec.ts
```diff
@@ -32,7 +33,7 @@ describe('Typescript Core Generator', () => {
     const tree = await runner.runSchematic('typescript-core', {
       specPath: path.join(__dirname, '..', '..', '..', 'testing', 'MOCK_swagger.yaml')
     }, baseTree);
-    const content: any = tree.readJson('/openapitools.json');
+    const content = tree.readJson('/openapitools.json') as any as OpenApiToolsConfiguration;
```
kpanot, line 36: not sure why `as any` is needed here :S

packages/@ama-sdk/core/cli/update-spec-from-npm.cts
```diff
@@ -0,0 +1,30 @@
...
+  const packageName = process.argv[2];
+  const packagePath = process.argv[3];
+
+  if (!packageName || !packagePath) {
+    console.error('Need to provide packageName and packagePath\nUsage: amasdk-update-spec-from-npm <package-name> <package-path>');
+    process.exit(-1);
+  }
+
+  const specSourcePath = require.resolve(`${packageName}/${packagePath}`);
+  const openApiConfigPath = './openapitools.json';
```
cpaulve-1A, line 21: this can technically be an input on the openapi tools script

packages/@ama-sdk/schematics/schematics/typescript/core/index.spec.ts
fpaul-1A: the output of `readJson` doesn't seem to like `undefined` fields, I had to change them to `null` to make the interface compatible without the use of `any`

packages/@ama-sdk/schematics/schematics/ng-add/index.ts
fpaul-1A: yes, also, probably unnecessary

packages/@ama-sdk/core/package.json
fpaul-1A: no, same as the old scripts
I can add them to the readme

packages/@ama-sdk/create/README.md
```diff
@@ -37,11 +37,16 @@ npm create @ama-sdk typescript <project-name> -- --package-manager=yarn [...opti
...
 
-- `--spec-path`: Path to the swagger/open-api specification used to generate the SDK
 - `--package-manager`: Node package manager to be used (`npm` and `yarn` are available).
 - `--debug --no-dry-run`: Enable schematics debug mode (dry-run is not currently supported).
 - `--o3r-metrics`: Enable or disable the collection of anonymous data for Otter
 - `--exact-o3r-version` : use a pinned version for [otter packages](https://github.com/AmadeusITGroup/otter/blob/main/docs/README.md).
 
+- `--spec-path`: Path to the swagger/open-api specification used to generate the SDK
+- `--spec-package-name`: The npm package name where the spec file can be fetched
+- `--spec-package-registry`: The npm registry where the spec file can be fetched
```
mrednic-1A, line 47: I am not sure if I understood correctly but I think it would be interesting to add  a note at the end of the list of args where to explain that a token is needed to be set at user level for the specs package (if this is private). Maybe a link to the doc where Corinne explains how to handle the tokens, if this doc is public.

packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/package.json.template
```diff
@@ -41,6 +41,7 @@
     "resolve": "node -e 'process.stdout.write(require.resolve(process.argv[1]));'",
     "generate": "schematics @ama-sdk/schematics:typescript-core",
     "spec:regen": "<%=packageManager%> run generate <% if (packageManager === 'npm') {%>-- <%}%>--generator-key <%=projectName%>-<%=projectPackageName%> && amasdk-clear-index",
+    "spec:upgrade": "<% if (specPackageName) { %>amasdk-update-spec-from-npm <%=specPackageName%> <%=specPackagePath%> && <% } %><%=packageManager%> run spec:regen",
```
kpanot, line 44:
```suggestion
    "spec:upgrade": "<% if (specPackageName) { %>amasdk-update-spec-from-npm <%=specPackageName%> \"<%=specPackagePath%>\" && <% } %><%=packageManager%> run spec:regen",
```

packages/@ama-sdk/create/README.md
fpaul-1A: you got it right, I added a note to explain that this option requires a global setup for the token

packages/@ama-sdk/core/cli/update-spec-from-npm.cts
fpaul-1A: I added an option to specify the output path

packages/@ama-sdk/core/cli/update-spec-from-npm.cts
```diff
@@ -0,0 +1,51 @@
...
+const noop = () => undefined;
+const logger = quiet ? {error: noop, warn: noop, log: noop, info: noop, debug: noop} : console;
+
+if (help) {
+  console.log(`This script can be used to update your local spec file from a given locally installed npm package.
+ Usage: amasdk-update-spec-from-npm <package-name> [--package-path] [--output]
+
+    package-name      The full identifier of the npm package (eg. @my-scope/my-package)
+    --package-path    The relative path inside the npm package where to find the spec file (default: './openapi.yml')
+    --output          The path where the spec file should be copied (default: './openapi.yml')
```
kpanot, line 25: This is depending of the format. If we receive a spec in JSON, we want to create a `openapi.json` instead.

packages/@ama-sdk/core/cli/update-spec-from-npm.cts
kpanot: May need to be referred in the readme.md as well :)

packages/@ama-sdk/core/cli/update-spec-from-npm.cts
fpaul-1A: it's already possible to use JSON file for these parameters, do you mean we should handle `openapi.yml` and `openapi.json` by default when unspecified ?

kpanot: Sorry for the unclear message, yes for the default.
I think to the case I target a JSON file within the `package-path` without providing `output`. I would be quite surprise to retrieve my JSON file content in the `openapi.yml` file :S.

vscaiceanu-1a (dismissed):

mrednic-1A:
I think we need to update the 'sdk' generator inside a monorepo with this feature too.
https://github.com/AmadeusITGroup/otter/blob/main/packages/%40o3r/workspace/schematics/sdk/schema.json
You can create a feature request to track that .

mrednic-1A (approved):

packages/@ama-sdk/core/cli/update-spec-from-npm.cts
```diff
@@ -0,0 +1,71 @@
...
+  }
+  if (!specSourcePath || !existsSync(specSourcePath)) {
+    logger.error(`Unable to find source spec from ${packageName}, please make sure it is correctly exported in package.json`);
+    process.exit(-2);
+  }
+
+  let specDestinationPath = output;
+  if (!specDestinationPath) {
+    const specSourceExtension = specSourcePath.split('.').pop();
+    specDestinationPath = `./openapi.${specSourceExtension}`;
```
kpanot, line 59: or
```suggestion
    const specSourceExtension = path.extname(specSourcePath); 
    specDestinationPath = `./openapi${specSourceExtension}`;
```

kpanot (dismissed):

packages/@o3r/workspace/schematics/sdk/index.ts
```diff
@@ -49,26 +58,28 @@ function generateSdkFn(options: NgGenerateSdkSchema): Rule {
         packageManager,
         skipInstall: !!options.specPath || options.skipInstall
       }),
-      packageManager === 'yarn' ? (t) => {
```
mrednic-1A, line 52: Nice that you could remove this :)

mrednic-1A (approved):

packages/@o3r/schematics/src/tasks/package-manager/npm-exec.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  constructor(private readonly script: string, private readonly args: string[] = [], private readonly workingDirectory?: string) {}
+
+  public toConfiguration(): TaskConfiguration<NodePackageTaskOptions> {
+    const packageManager = getPackageManager();
+    return {
+      name: NodePackageName,
+      options: {
+        command: 'exec',
+        packageName: `exec ${this.script} ${packageManager === 'npm' ? '-- ' : ''}${this.args.join(' ')}`,
```
kpanot, line 19:
```suggestion
        packageName: `exec ${this.script} ${packageManager === 'npm' ? '-- ' : ''}${this.args.map((arg) => `"${arg}"`).join(' ')}`,
```

mrednic-1A: updated

kpanot (approved):

PR #1688, opened 2024-04-19
test(e2e): fix visual testing

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


mrednic-1A (approved):

apps/showcase/scripts/update-screenshots/podman-script.sh
```diff
@@ -22,9 +22,6 @@ cd tests-clone
 echo 'Delete previous screenshots'
 rm -rdf apps/showcase/e2e-playwright/sanity/screenshots
 
-# Ensure yarn is installed
-corepack enable
```
kpanot, line 26: why is it removed, is it already part of the docker?

fpaul-1A: yes https://github.com/microsoft/playwright/blob/main/utils/docker/Dockerfile.focal#L22, and the corepack setup requires to perform additional setup to work with VPN/Proxy

PR #1696, opened 2024-04-22
feat: ng-update script for migration to openapitools 7.4

cpaulve-1A:
## Proposed change

Ng update script to help the migration to @ama-sdk/schematics 10.3
As projects generated via the @ama-sdk/create command, the update is exposed as a schematics.
Unfortunately, it will not allow the update of several version in one command and needs the 10.3 version of ama-sdk/schematics to already be installed to be run.

### DIFF
--- a/packages/@ama-sdk/schematics/migration.json
+++ b/packages/@ama-sdk/schematics/migration.json
@@ -5,6 +5,11 @@
       "version": "10.0.0-alpha.0",
       "description": "Updates of @ama-sdk/schematics to v10.0.*",
       "factory": "./schematics/ng-update/index#updateV10_0"
+    },
+    "migration-v10_3": {
+      "version": "10.3.0-alpha.0",
+      "description": "Updates of @ama-sdk/schematics to v10.3.*",
+      "factory": "./schematics/ng-update/index#updateV10_3"
     }
   }
 }
--- a/packages/@ama-sdk/schematics/schematics/migrate/index.ts
+++ b/packages/@ama-sdk/schematics/schematics/migrate/index.ts
@@ -4,17 +4,21 @@ import { getMigrationRuleRunner, getWorkspaceConfig, type MigrationRulesMap } fr
 import { resolve } from 'node:path';
 import { gt, minVersion } from 'semver';
 import { isTypescriptSdk } from '../helpers/is-typescript-project';
+import {updateOpenApiVersionInProject} from '../ng-update/typescript/v10.3/update-openapiversion';
 
+/* eslint-disable @typescript-eslint/naming-convention */
 const tsMigrationMap: MigrationRulesMap = {
-
+  '~10.3.2': updateOpenApiVersionInProject()
 };
+/* eslint-enable @typescript-eslint/naming-convention */
+
 /**
  * Facilitate the migration of a version to another by the run of migration rules
  * @param options
  */
 function migrateFn(options: MigrateSchematicsSchemaOptions): Rule {
 
-  const currentVersion = JSON.parse(require(resolve(__dirname, '..', '..', 'package.json'))).version;
+  const currentVersion = require(resolve(__dirname, '..', '..', 'package.json')).version;
   const to: string = options.to || currentVersion;
   const minimumVersion = minVersion(to);
 
--- a/packages/@ama-sdk/schematics/schematics/ng-update/typescript/v10.3/update-openapiversion.ts
+++ b/packages/@ama-sdk/schematics/schematics/ng-update/typescript/v10.3/update-openapiversion.ts
@@ -1,0 +1,74 @@
+import type {Rule} from '@angular-devkit/schematics';
+import * as semver from 'semver';
+import {getWorkspaceConfig, globInTree} from '@o3r/schematics';
+import {chain} from '@angular-devkit/schematics';
+
+/**
+ * Update open api version used in the project to match with the one used in @ama-sdk/schematics:typescript-core
+ */
+export const updateOpenApiVersionInProject = (): Rule => {
+  const overwriteOpenApiVersion = (pathOfWorkspace: string): Rule => {
+    return (tree, context) => {
+      const packageJson = tree.readText(`${pathOfWorkspace}/package.json`);
+      const amaSdkTypescriptUsage = packageJson.match(/@ama-sdk\/schematics:typescript-core.*",?$/mg);
+      if (!amaSdkTypescriptUsage) {
+        context.logger.info(`Skipping ng-update for ${pathOfWorkspace} as it does not use @ama-sdk/schematics:typescript-core`);
+        return tree;
+      }
+      const openapitoolsPaths: string[] = [];
+      amaSdkTypescriptUsage.forEach((script) => {
+        const openapiToolsPath = new RegExp(/@ama-sdk\/schematics:typescript-core.* --spec-config-path(?:=|\s+)["']?(?:\.\/)?([a-zA-Z-/.0-9_]*\.json)/).exec(script);
+        if (openapiToolsPath?.length === 2) {
+          openapitoolsPaths.push(`${pathOfWorkspace}/${openapiToolsPath[1]}`);
+        }
+      });
+      if (amaSdkTypescriptUsage.length > openapitoolsPaths.length) {
+        openapitoolsPaths.push(`${pathOfWorkspace}/openapitools.json`);
+      }
+      (openapitoolsPaths || []).forEach((path) => {
+        if (!tree.exists(path)) {
+          context.logger.warn(`Missing ${path} file described in the package json.`);
+          return;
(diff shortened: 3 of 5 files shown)


packages/@ama-sdk/schematics/collection.json
```diff
@@ -31,6 +31,10 @@
       "description": "Generate a Java SDK source code based on swagger specification",
       "factory": "./schematics/java/client-core/index#ngGenerateJavaClientCore",
       "schema": "./schematics/java/client-core/schema.json"
+    },
+    "update": {
```
matthieu-crouzet, line 35:
```suggestion
    "update-openapi-version": {
```

packages/@ama-sdk/schematics/schematics/ng-update/typescript/v10.3/update-openapiversion.ts
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Update open api version used in the project to match with the one used in @ama-sdk/schematics:typescript-core
+ */
+export const updateOpenApiVersionInProject = (): Rule => {
+  const overwriteOpenApiVersion = (pathOfWorkspace: string): Rule => {
+    return (tree, context) => {
+      const packageJson = tree.readText(`${pathOfWorkspace}/package.json`);
+      const amaSdkTypescriptUsage = packageJson.match(/@ama-sdk\/schematics:typescript-core.*",/mg);
+      if (!amaSdkTypescriptUsage) {
+        context.logger.info(`Skipping ng-update for ${pathOfWorkspace} as it does not use `);
```
matthieu-crouzet, line 15: You forgot the end of your sentence no ?

packages/@ama-sdk/schematics/schematics/ng-update/typescript/v10.3/update-openapiversion.ts
```diff
@@ -0,0 +1,76 @@
...
+import {chain} from '@angular-devkit/schematics';
+
+/**
+ * Update open api version used in the project to match with the one used in @ama-sdk/schematics:typescript-core
+ */
+export const updateOpenApiVersionInProject = (): Rule => {
+  const overwriteOpenApiVersion = (pathOfWorkspace: string): Rule => {
+    return (tree, context) => {
+      const packageJson = tree.readText(`${pathOfWorkspace}/package.json`);
+      const amaSdkTypescriptUsage = packageJson.match(/@ama-sdk\/schematics:typescript-core.*",/mg);
```
matthieu-crouzet, line 13: why adding `.*` at the end ?

matthieu-crouzet, line 13: maybe the `,` could be optional if it's the last script

packages/@ama-sdk/schematics/collection.json
matthieu-crouzet, line 35: As we are not in the `migration.json` I would expect a name that explain what it does

packages/@ama-sdk/schematics/migration.json
```diff
@@ -5,6 +5,11 @@
       "version": "10.0.0-alpha.0",
       "description": "Updates of @ama-sdk/schematics to v10.0.*",
       "factory": "./schematics/ng-update/index#updateV10_0"
+    },
+    "migration-v10_3": {
+      "version": "10.0.0-alpha.0",
```
kpanot, line 10: The version is not matching the description

packages/@ama-sdk/schematics/schematics/ng-update/index.ts
```diff
@@ -2,14 +2,15 @@
...
+import { updateV10_0 as tsUpdateV10_0, updateV10_3 as tsUpdateV10_3 } from './typescript';
+import { globInTree } from '@o3r/schematics';
 
 /**
  * Determine if the script is run in a Typescript SDK
  * @param tree
  */
 const isTypescriptSdk = (tree: Tree) => {
-  return tree.exists('/tsconfig.json');
+  return globInTree(tree, ['/tsconfig.*json']);
```
kpanot, line 13:
```suggestion
  return globInTree(tree, ['/tsconfig.*.json']);
```
(same thing but the `*json` is not usual)

packages/@ama-sdk/schematics/schematics/ng-update/typescript/v10.3/update-openapiversion.ts
```diff
@@ -0,0 +1,76 @@
...
+        }
+        tree.overwrite(path, JSON.stringify(openApiToolsConfig, null, 2));
+      }
+      );
+      return tree;
+    };
+  };
+
+  return (tree, context) => {
+    let pathsPackageJson: string[];
```
kpanot, line 48:
```suggestion
    let pathsPackageJson: string[] = ['.'];
```

to be aligned with the type

cpaulve-1A: Do we want to run it at the root if we detect workspaces?

matthieu-crouzet: Then maybe `[]`

cpaulve-1A: If you prefer, but you will still need to specify the string[] and you can't do the ||= later on, but push the '.' inside a if(pathsPackageJson.length === 0). Either way, I don't mind

cpaulve-1A: In fact, after more thinking and discussion, it seems it is best to actually test the root folder as the openapitools config could be set at the root for a monorepo with different setup for each workspace.

packages/@ama-sdk/schematics/schematics/ng-update/typescript/v10.3/update-openapiversion.ts
```diff
@@ -0,0 +1,75 @@
+import type {Rule} from '@angular-devkit/schematics';
+import * as semver from 'semver';
+import {getWorkspaceConfig, globInTree} from '@o3r/schematics';
+import {chain} from '@angular-devkit/schematics';
+
+/**
+ * Update open api version used in the project to match with the one used in @ama-sdk/schematics:typescript-core
+ */
+export const updateOpenApiVersionInProject = (): Rule => {
```
kpanot, line 9: The schematic seems a bit tricky and based on complex regexp, maybe it would help to provide UT for it :S

packages/@ama-sdk/schematics/collection.json
```diff
@@ -31,6 +31,10 @@
       "description": "Generate a Java SDK source code based on swagger specification",
       "factory": "./schematics/java/client-core/index#ngGenerateJavaClientCore",
       "schema": "./schematics/java/client-core/schema.json"
+    },
+    "migration-v10_3": {
```
kpanot, line 35: I wonder if, only for the manual updates, it would not make more sense to expose a single `migrate` schematics taking the 2 `--from` `--to` arguments to determine which rules to apply

kpanot: I thinking in user and renovate case.
How you will document for the user and indicate to renovate to run this schematic?
I case you have 1 schematic that takes the version as input you will need to just add in the doc this single schematic and the run of it can be added within the shared Renovate preset

cpaulve-1A: This

cpaulve-1A: After discussion, we agreed to exclude this script for now as we do not have a proper schematic to manage the migration for non angular project. We should first implement a schematic

Note: meanwhile we will need to document the manual migration step

sdo-1A (approved):

packages/@ama-sdk/schematics/schematics/migrate/index.ts
```diff
@@ -4,17 +4,21 @@ import { getMigrationRuleRunner, getWorkspaceConfig, type MigrationRulesMap } fr
...
+/* eslint-enable @typescript-eslint/naming-convention */
+
 /**
  * Facilitate the migration of a version to another by the run of migration rules
  * @param options
  */
 function migrateFn(options: MigrateSchematicsSchemaOptions): Rule {
 
-  const currentVersion = JSON.parse(require(resolve(__dirname, '..', '..', 'package.json'))).version;
+  const currentVersion = require(resolve(__dirname, '..', '..', 'package.json')).version;
```
kpanot, line 21: Shame on me 😅

kpanot (approved):

vscaiceanu-1a (approved):

PR #1713, opened 2024-04-25
feat(localization): add schema for localization.json files

### LINKED ISSUE #1702: [Bug]: @o3r/localization  extractor skips objects
### Package name

extractors

### Package version

9.3.10

### Reproduction steps

If i pass an object inside the transaltions.json as 
"a": {
  "b": "my message"
} 

I wait for it to be "a.b": "my message" as it is working fine in plain ngx-translate

### Current result

If i pass an object inside the transaltions.json as 
"a": {
  "b": "my message"
} 

I get "a": {}

### Expected result

I wait for it to be "a.b": "my message" as it is working fine in plain ngx-translate

### Additional comments

_No response_

kpanot:
## Proposed change

Add localization.json files JSON Schema

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature resolves #1702

### DIFF
--- a/packages/@o3r/localization/builders/helpers/localization.generator.ts
+++ b/packages/@o3r/localization/builders/helpers/localization.generator.ts
@@ -18,11 +18,11 @@ export interface LocalizationMetadataAsMap {
 
 /** Localization structure */
 export interface LocalizationJsonValue {
-  /** Localtion Description */
+  /** Localization Description */
   description: string;
   /** Determine if the item can have multiple extensions */
   dictionary?: boolean;
-  /** Determine if the value has to be overriden */
+  /** Determine if the value has to be overridden */
   referenceData?: boolean;
   /** Localization default value */
   defaultValue?: string;
@@ -36,6 +36,10 @@ export interface LocalizationJsonValue {
 export interface LocalizationJsonFile {
   [key: string]: LocalizationJsonValue;
 }
+
+type LocalizationJsonFileContent = LocalizationJsonFile & {
+  '$schema'?: string;
+};
 
 /** Localization file mapping */
 export interface LocalizationFileMap {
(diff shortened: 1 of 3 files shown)


packages/@o3r/localization/schemas/localization.schema.json
```diff
@@ -0,0 +1,83 @@
...
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "LocalizationSchema",
+  "description": "Schema of Localization file",
+  "type": "object",
+  "additionalProperties": false,
+  "patternProperties": {
+    "\\$schema": true,
+    "^[a-zA-Z0-9.-]+": {
+      "type": "object",
+      "oneOf": [
```
matthieu-crouzet, line 11: shouldn't we have a oneOf for ref / dictionary / basic ?
because for basic I would expect to have `defaultValue` required

packages/@o3r/localization/builders/helpers/localization.generator.ts
```diff
@@ -143,7 +147,17 @@ export class LocalizationExtractor {
    * @param locFile Path to the localization file
    */
   private readLocalizationFile(locFile: string) {
-    return new Promise<LocalizationJsonFile>((resolve, reject) => fs.readFile(locFile, 'utf8', (err, data) => err ? reject(err) : resolve(JSON.parse(data))));
+    return new Promise<LocalizationJsonFile>((resolve, reject) => fs.readFile(locFile, 'utf8', (err, data) => {
```
fpaul-1A, line 150: this would be more readable using `fs.promises.readFile`

packages/@o3r/localization/schemas/localization.schema.json
```diff
@@ -0,0 +1,83 @@
...
+              "type": "string",
+              "description": "Localization default value"
+            },
+            "description": {
+              "type": "string",
+              "description": "Localization Description"
+            },
+            "dictionary": {
+              "type": "boolean",
+              "description": "Determine if the item can have multiple extensions",
```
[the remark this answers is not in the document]
kpanot: Just used the description from TS interface, it is not perfect but not wrong.
Basically it determines that the key can have a `.<somthing>` suffix.
If you have better, I take it :)

packages/@o3r/localization/schemas/localization.schema.json
matthieu-crouzet: There you have set defaultValue mandatory even when dictionary is true
I would expect something like 
```json5
{
...
  "oneOf": [
    { required: ["description", "defaultValue"], properties: { ..., dictionary: { value: false } }, ... },
    { required: ["description", "dictionary"], properties: { ..., dictionary: { value: true } }, ... },
    { required: ["description", "$ref"], properties: { ..., dictionary: { value: false } }, ... },
  ]
}
```
maybe I'm mistaken

kpanot: you're right, I did the fix quickly between 2 calls and forgot it.
I will add a 3rd case.

mrednic-1A (approved):

fpaul-1A (approved):

PR #1736, opened 2024-04-30
chore: enhance documentation of the rules engine

cpaulve-1A:
## Proposed change

Improve the documentation of the rules engine


apps/showcase/src/assets/rules/rulesets.json
```diff
@@ -5,13 +5,25 @@
...
+          "name": "RulesEnginePresConfig"
+        }]
+      },
       "rules": [
         {
           "id": "5467e501-b9ff-414f-8026-56885d0d7a4b",
           "name": "New-York availability",
           "disabled": false,
-          "outputRuntimeFacts": [],
+          "outputRuntimeFacts": ["MY_FACT_UPDATE"],
```
cpaulve-1A, line 24:
```suggestion
          "outputRuntimeFacts": [],
```

docs/rules-engine/how-to-use/industrialize-ruleset-generation.md
```diff
@@ -0,0 +1,297 @@
...
+  // ...
+  "extraOptions": {
+    "otterSubModuleRefs": ["./**/my-sub-entry/index.ts"], // Relative path to the index.ts where the interfaces of your sub-entry are exported
+  },
+  // ...
+}
+```
+
+This will allow the extractor to find the definition of your sub-entry interface and generate their schema files.
+For instance, let's consider the MyFacts object defined below.
```
sdo-1A, line 253:
```suggestion
For instance, let's consider the `MyFacts` object defined below.
```

docs/rules-engine/how-to-use/industrialize-ruleset-generation.md
```diff
@@ -0,0 +1,297 @@
...
+  ]
+}
+```
+
+> [!WARNING]
+> 
+> The fact extractor is not able to make references to sub-entry schema in other model schema files. 
+> That means that if the ``SubEntryClass`` refers to another sub-entry it will be ignored.
+> 
+> In the following example, there will be a schema generated for the SubEntryClass, and it will describe ``SameSubEntryClass``
```
sdo-1A, line 283:
```suggestion
> In the following example, there will be a schema generated for the ``SubEntryClass``, and it will describe the definitions of ``SameSubEntryClass`` and ``SomeNodeModulesRef``, but will ignore the ``OtherSubEntryClass`` definition.
```

docs/rules-engine/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+> [!NOTE]
+> If no condition is defined in the ``IfElseBlock``, its ``successElements`` will always be executed.
+
+A ``Rule`` also contains information needed to optimize the reevaluation of its conditions:
+* ``inputRuntimeFacts`` : the input runtime facts the rule is based on
+* ``inputFacts`` : the application facts that are used by the rule
+* ``outputRuntimeFacts`` : the runtime facts that are updated/set by the rule
+
+Thanks to these parameters, the rule engines can make sure to only trigger rules evaluation if the input and output of
```
matthieu-crouzet, line 49:
```suggestion
Thanks to these parameters, the rule engines can make sure to only trigger rules evaluation if the `inputFacts` of
```
The `outputRuntimeFacts` cannot be the trigger of the rule evaluation, as they depends on `inputFacts` no ?

docs/rules-engine/README.md
```diff
@@ -0,0 +1,169 @@
...
+(provided by [@o3r/dynamic-content](https://www.npmjs.com/package/@o3r/dynamic-content))
+* ``UPDATE_LOCALISATION`` substitutes the translation of a key with the one from another key.
+([@o3r/localization](https://www.npmjs.com/package/@o3r/localization))
+* ``UPDATE_PLACEHOLDER`` injects html code in a specific placeholder
+([@o3r/components](https://www.npmjs.com/package/@o3r/components))
+
+You can provide your own custom action following [this guide](./how-to-use/custom-action.md).
+
+> [!NOTE]
+> The actions handled by the rules engine need to be [registered](./how-to-use/README.md#step-2-register-the-action-handlers-in-the-runner) in your application to be recognized by the service.
```
matthieu-crouzet, line 103:
```suggestion
> The actions handled by the rules engine need to be [registered](./how-to-use/README.md#step-2-register-the-action-handlers-in-the-runner) in your application to be recognized by the engine.
```

docs/rules-engine/README.md
```diff
@@ -0,0 +1,169 @@
...
+> This ensures you only import packages you actually needs.
+
+### Fact
```
matthieu-crouzet, line 106: Maybe we should have the Fact paragraph before the Rule one because we use several time the "fact" concept in the Rule section

docs/rules-engine/README.md
```diff
@@ -0,0 +1,169 @@
...
+During a rule execution, its conditions will be evaluated using the latest fact value(s).
+Every fact can be identified by its ``id`` property and exposes a ``value$`` stream that will trigger a new evaluation
+of the Ruleset it has been linked for every value it emits.
+This allows a big optimization since the rules engine will react to a fact change in order to execute only the subset
+of rules that relies on this fact.
+
+Facts can be defined inside the code of the application as the example below, or can be defined by the ruleset itself.
+In that case, we call it a [runtime fact](#runtime-fact).
+
+```typescript
```
matthieu-crouzet, line 119: I'm not fan of this kind of example as it will not be extracted like this

docs/rules-engine/examples/basic-rule.md
```diff
@@ -1,46 +1,67 @@
...
+                "actionType": "UPDATE_CONFIG",
+                "component": "o3r-dummy-component",
+                "library": "@o3r/dummy-lib",
+                "property": "numberItemPerPage",
+                "value": 10
+              }
             ],
-            'failureElements': [
+            "failureElements": [
               // Any action
```
matthieu-crouzet, line 59:
```suggestion
              {
                "elementType": "ACTION",
                "actionType": "UPDATE_CONFIG",
                "component": "o3r-dummy-component",
                "library": "@o3r/dummy-lib",
                "property": "numberItemsPerPage",
                "value": 10
              }
```

docs/rules-engine/examples/complex-fact.md
```diff
@@ -1,47 +1,80 @@
...
-import {Ruleset} from '@o3r/rules-engine';
+// complexFact interface
+interface ComplexFactInterface {
+  arrayPropName: { propName: string }[]
+}
+```
+
+In this Ruleset, we need to evaluate the value of the ``propName`` of the first item in ``arrayPropName`` and test
+whether
+it equals to ``SUCCESS_VALUE``.
```
matthieu-crouzet, line 17: I would not use the word `success` to not confuse people with `successElements`
```suggestion
it equals to ``MY_SEARCHED_VALUE``.
```

docs/rules-engine/examples/portal-facts.md
```diff
@@ -0,0 +1,123 @@
...
+## Implementation
+
+### PortalFacts
+
+The deeplink parameters are exposed to the rule engine via a ``portalFacts`` fact which will emit a map of `<string, string>` with
+the different portal parameters name associated to their values.
+
+> [!WARNING]
+>
+> At the moment, `portalFacts` only supports string and no complex value.
```
matthieu-crouzet, line 27: Do we plan to support it later ?
if no 
```suggestion
> `portalFacts` only supports string and no complex value.
```

docs/rules-engine/examples/portal-facts.md
```diff
@@ -0,0 +1,123 @@
...
+
+Note that, as its content is dynamic, ``portalFacts`` should not be part of the [metadata of the rules engine](../how-to-use/industrialize-ruleset-generation.md).
+To avoid its extraction, it will be part of a dedicated folder. 
+
+You will need to generate a ``PortalFactService`` in this folder and register it in the application rules engine.
+For this purpose, please follow the [custom fact documentation](../how-to-use/custom-fact.md).
+
+This service will expose a fact with the ``portalFacts`` object sent by the deeplink.
+
+For this use case, we will take the assumption that ``portalFacts`` is sent as a POST parameter and injected in the index.html
```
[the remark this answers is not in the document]
matthieu-crouzet:
```suggestion
For this use case, we will take the assumption that ``portalFacts`` is sent as a POST parameter and injected in the `data-portalFacts` property of the body tag.
```

docs/rules-engine/examples/portal-facts.md
```diff
@@ -0,0 +1,123 @@
...
+``data-post`` property.
+
+This means that the ``https://url-of-my-app?portalFacts=[{"key":"portalFactName1", "value":"portalFactValue2"}]&`` request will return this html :
```
matthieu-crouzet, line 42: Here you give an example of GET params not POST no ?

docs/rules-engine/examples/portal-facts.md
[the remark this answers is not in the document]
matthieu-crouzet:
```suggestion
For this use case, we will take the assumption that ``portalFacts`` is sent as a POST parameter and injected in the ``data-post`` property of the `body` tag.
```

docs/rules-engine/examples/portal-facts.md
```diff
@@ -0,0 +1,123 @@
...
+import {RequestParametersService} from '@o3r/dynamic-content';
+import {of} from 'rxjs';
+
+@Injectable()
+export class PortalFactsService extends FactsService<PortalFacts> {
+
+  public facts: FactSet<PortalFacts>;
+
+  constructor(rulesEngine: RulesEngineRunnerService, requestParametersService: RequestParametersService) {
+    super(rulesEngine);
```
matthieu-crouzet, line 67: maybe we can do the example with the `inject` method to follow the new Angular guideline

docs/rules-engine/examples/runtime-facts.md
```diff
@@ -1,92 +1,431 @@
...
+            }
+          }
+        }
+      ]
+    }
+  ]
+}
+```
+
+In this Ruleset, there is some duplication over the logic of "the customer is a minor". This makes it more difficult to
```
matthieu-crouzet, line 221: Instead can't we avoid providing wrong example and directly say 
"to avoid duplication and maintenance complexity we can go this way"
and show only the way to go

docs/rules-engine/examples/runtime-facts.md
```diff
@@ -1,92 +1,431 @@
...
+                      "rhs": {
+                        "type": "LITERAL",
+                        "value": "New York"
+                      }
+                    }
+                  ]
+                }
+              ]
+            },
+            "successElements": [
```
matthieu-crouzet, line 320: shoudn't we have a `failureElements`
```
"failureElements": [
            {
                "actionType": "SET_FACT",
                "key": "IS_MINOR",
                "value": "TRUE"
              }
]

File: docs/rules-engine/how-to-use/README.md [hunk 16]
```diff
@@ -0,0 +1,410 @@
+# Rules engine - Integration
```
matthieu-crouzet, line 1: Isn't this page what we should have in the showcase application ?

docs/rules-engine/how-to-use/build-in-facts.md
```diff
@@ -0,0 +1,155 @@
...
+        "value": "otter-late.svg"
+      }
+    ],
+    "failureElements": []
+  }
+}
+```
+
+This does not mean that the fact itself should not be registered.
+If you fail to register the fact the same way as the non-build-in facts, the whole ruleset will fail.
```
matthieu-crouzet, line 57:
```suggestion
If you fail to register the fact the same way as the non-built-in facts, the whole ruleset will fail.
```

docs/rules-engine/how-to-use/build-in-facts.md
```diff
@@ -0,0 +1,155 @@
...
+```typescript
+import {Component, OnInit} from '@angular/core';
+import {CurrentTimeFactsService} from '@o3r/rules-engine';
+
+@Component({
+  selector: 'app-root',
+  templateUrl: './app.component.html',
+})
+export class AppComponent implements OnInit {
+  constructor(private currentTimeFactsService: CurrentTimeFactsService) {}
```
matthieu-crouzet, line 68: we could use the `inject` method

docs/rules-engine/how-to-use/build-in-facts.md
```diff
@@ -0,0 +1,155 @@
...
+  }
+}
+```
+
+This does not mean that the fact itself should not be registered.
+If you fail to register the fact the same way as the non-build-in facts, the whole ruleset will fail.
+
+```typescript
+import {Component, OnInit} from '@angular/core';
+import {CurrentTimeFactsService} from '@o3r/rules-engine';
```
matthieu-crouzet, line 61: This service does not come from '@o3r/rules-engine' but should be implemented on each application

docs/rules-engine/how-to-use/build-in-facts.md
```diff
@@ -0,0 +1,155 @@
...
+  ngOnInit() {
+    this.currentTimeFactsService.register();
+  }
+}
+```
+
+The ``CurrentTimeFactsService`` is based on a simple implementation based on a refresh mechanism. It is up to the 
+application to choose when to refresh the fact via its ``tick`` method (on page navigation, on a given time interval etc. ).
+
+```typescript
```
matthieu-crouzet, line 79: Why don't we provide a default implementation of this service ?
instead of duplicate that on all application ?

docs/rules-engine/how-to-use/build-in-facts.md
```diff
@@ -0,0 +1,155 @@
...
+import {Injectable} from '@angular/core';
+import {CurrentTimeFacts, FactsService, RulesEngineService} from '@o3r/rules-engine';
+import {BehaviorSubject} from 'rxjs';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class CurrentTimeFactsService extends FactsService<CurrentTimeFacts> {
+
+  private currentTimeSubject$ = new BehaviorSubject(new Date('2023-11-2').getTime()); // define a date in the past instead of the current time
```
matthieu-crouzet, line 89: why  a date in the past ?

docs/rules-engine/how-to-use/build-in-facts.md
```diff
@@ -0,0 +1,155 @@
...
+  }
+}
+```
+
+> [!INFO]
+> 
+> The reason why the fact is not registered automatically, even though it is provided by the module, is to allow the 
+> override of the fact.
+
+## Override the default build-in fact
```
matthieu-crouzet, line 111:
```suggestion
## Override the default built-in fact
```

docs/rules-engine/how-to-use/build-in-facts.md
cpaulve-1A: Indeed, I was sure we did provide it

docs/rules-engine/how-to-use/chrome-extension.md
```diff
@@ -0,0 +1,111 @@
...
+
+```typescript
+import {NgModule} from '@angular/core';
+import {RulesEngineRunnerModule, RulesEngineDevtoolsModule} from '@o3r/rules-engine';
+import {AppComponent} from './app.component';
+
+@NgModule({
+    imports: [
+      RulesEngineDevtoolsModule,
+      RulesEngineRunnerModule.forRoot({debug: true}) // Activate rule engine debug mode
```
matthieu-crouzet, line 18: It's not enough you need to have the communication service between the app and the extension as well
`inject(RulesEngineDevtoolsMessageService)` in the main.ts
and `{provide: OTTER_RULES_ENGINE_DEVTOOLS_OPTIONS, useValue: {isActivatedOnBootstrap: true}}`

docs/rules-engine/how-to-use/chrome-extension.md
```diff
@@ -0,0 +1,111 @@
...
+    ],
+    declarations: [AppComponent],
+    bootstrap: [AppComponent]
+})
+export class AppModule {}
```
matthieu-crouzet, line 23: Angular doesn't generate app with module by default now 
we may update the documentation with the standalone use case

docs/rules-engine/how-to-use/README.md
```diff
@@ -0,0 +1,410 @@
...
+    ruleEngine.actionHandlers.add(configurationRulesEngineActionHandler);
+    this.deviceDetectionFacts.register();
+  }
+}
+```
+
+Thanks to this, the application is now ready to retrieve and evaluate the rules and execute the resulting actions.
+
+> [!NOTE]
+> There is no constraints on whether to register facts before setting up the Rulesets. The ruleset using them will just
```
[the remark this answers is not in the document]
cpaulve-1A: I think it is `there is no constraint` instead

docs/rules-engine/how-to-use/dedicated-component-ruleset.md
```diff
@@ -1,15 +1,60 @@
...
 
-## How to add it to your application
+You can make a component ``linkedComponents`` compatible thanks the following Otter CLI command:
+``ng g rules-engine-to-component --path=[path-to-my-component-file]``
 
-First, you need to have a ruleset defined in your json (created via the CMS interface) that will be identified by a unique id, and you need to specify that the ruleSet is linked to at least one component, meaning that it's not executed per default. The json will look as follows :
+This command will compute the component's identifier and add the methods to enable or disable the ruleset 
+matching the component name.
+
+```typescript
```
matthieu-crouzet, line 18: instead of typescript we could use `diff`
to easily understand what will be modified
```diff
import {ChangeDetectionStrategy, Component, ViewEncapsulation, inject, OnInit, OnDestroy} from '@angular/core';
+ import {LinkableToRuleset, RulesEngineRunnerService} from '@o3r/rules-engine';
+ import {computeItemIdentifier} from '@o3r/core';

@O3rComponent({ componentType: 'Component' })
@Component({
  selector: 'app-my-component',
  templateUrl: './my-component.template.html',
  styleUrl: './my-component.style.scss',
  encapsulation: ViewEncapsulation.None,
  changeDetection: ChangeDetectionStrategy.OnPush
})
- export class MyComponent {
+ export class MyComponent implements OnInit, OnDestroy {

+  private readonly rulesEngineService = inject(RulesEngineRunnerService, { optional: true });
+  private readonly componentIdentifier = computeItemIdentifier('MyComponent', '@scope/components');

+  public ngOnInit() {
+    if (this.rulesEngineService) {
+      this.rulesEngineService.enableRuleSetFor(this.componentName);
+    }
+  }
+  public ngOnDestroy() {
+    if (this.rulesEngineService) {
+      this.rulesEngineService.disableRuleSetFor(this.componentName);
+   }
+  }
}
```

docs/rules-engine/how-to-use/dedicated-component-ruleset.md
```diff
@@ -1,15 +1,60 @@
...
+import {ChangeDetectionStrategy, Component, ViewEncapsulation, inject, OnInit, OnDestroy} from '@angular/core';
+import {LinkableToRuleset, RulesEngineRunnerService} from '@o3r/rules-engine';
+import {computeItemIdentifier, type ExposedComponent} from '@o3r/core';
```
matthieu-crouzet, line 21: type `ExposedCompoent` is not used

docs/rules-engine/how-to-use/dedicated-component-ruleset.md
```diff
@@ -18,60 +63,40 @@ First, you need to have a ruleset defined in your json (created via the CMS inte
       "id": "e5th46e84-5e4th-54eth65seth46se8th2",
       "name": "the second ruleset",
       "linkedComponents": {
-        "or": [ // 'or' keyword means: if at least one component from the list is registered, the ruleset is active
+        // 'or' keyword means: if at least one component from the list is registered, the ruleset is active
+        "or": [
```
matthieu-crouzet, line 67: Maybe add a note to say that the `and` is not supported but if needed they can open an issue (link to open the issue) and we will add the support

docs/rules-engine/how-to-use/industrialize-ruleset-generation.md
```diff
@@ -0,0 +1,297 @@
...
+  /**
+   * Description of property
+   */
+  property: number
+}
+interface ComplexInterface  {
+  /**
+   * Description of property1
+   */
+  property1: []
```
matthieu-crouzet, line 160: array of what ?

docs/rules-engine/how-to-use/industrialize-ruleset-generation.md
```diff
@@ -0,0 +1,297 @@
...
+}
+
+interface CustomFacts extends FactDefinitions {
+  /** Description of fact1 */
+  fact1: string,
+  /** Description of fact2*/
+  fact2: ComplexInterface,
+  /** Description of fact3*/
```
matthieu-crouzet, line 168:
```suggestion
  /** Description of fact3 */
```

docs/rules-engine/how-to-use/industrialize-ruleset-generation.md
```diff
@@ -0,0 +1,297 @@
...
+}
+```
+
+> [!NOTE]
+>
+> The models for the OperatorsMetadata is available in the [rules engine metadata interface file](https://github.com/AmadeusITGroup/otter/blob/main/packages/%40o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.interfaces.ts).
+
+### Project with module subentries
+
+As mentioned in the previous section, the extractor looks into your project and its node_modules files to identify the
```
matthieu-crouzet, line 231: not necessary node_modules if you use PnP

docs/rules-engine/how-to-use/custom-fact.md
```diff
@@ -0,0 +1,125 @@
...
+import {Component} from '@angular/core';
+import {CustomFactsService} from '../facts/custom-facts.service';
+
+@Component({
+  selector: 'app-root',
+  templateUrl: './app.component.html',
+})
+export class AppComponent {
+  constructor(customFactsService: CustomFactsService) {
+    customFactsService.register();
```
matthieu-crouzet, line 116: Do we still want to recommand to pollute the AppComponent with the registration of facts etc...
in the `main.ts`
you could do something like 
```typescript
bootstrapApplication(AppComponent, appConfig)
   .then((m) => {
      runInInjectionContext(m.injector, () => {
        const customFactsService = inject(CustomFactsService);
        customFactsService.register();
      });
   })
```

docs/rules-engine/how-to-use/custom-fact.md
```diff
@@ -0,0 +1,125 @@
...
+  } 
+}
+```
+The ``FactService`` extended by ``CustomFactsService`` will handle the registration on the ``RulesEngineRunnerService``.
+
+> [!IMPORTANT]
+> 
+> If you administrate your rules in a dedicated UI, make sure to extract your fact with the rest of your metadata.
```
matthieu-crouzet, line 124:
```suggestion
> If you administrate your rules in a dedicated UI, make sure to extract your fact with the rest of your metadata.
```
```suggestion
> If you administrate your rules in a dedicated UI, make sure to extract your fact with the rest of your metadata.
```
```suggestion
> If you administrate your rules in a dedicated UI, make sure to extract your facts with the rest of your metadata.
```

docs/rules-engine/how-to-use/custom-action.md
```diff
@@ -0,0 +1,153 @@
...
+/**
+ * Service to handle async Asset actions
+ */
+@Injectable()
+export class PopupActionHandler implements RulesEngineActionHandler<RulesEngineAction> {
+
+  /** @inheritdoc */
+  public readonly supportingActions = [RULES_ENGINE_POPUP_ACTION_TYPE] as const;
+
+  constructor() {
```
matthieu-crouzet, line 46: if not needed we can remove it

docs/rules-engine/how-to-use/custom-action.md
```diff
@@ -0,0 +1,153 @@
...
+/**
+ * Service to handle async Asset actions
+ */
+@Injectable()
+export class PopupActionHandler implements RulesEngineActionHandler<ActionDisplayPopup> {
+
+  /** @inheritdoc */
+  public readonly supportingActions = [RULES_ENGINE_POPUP_ACTION_TYPE] as const;
+
+  constructor(private readonly popupManagerService: PopupManagerService) {
```
matthieu-crouzet, line 117:
```suggestion
  private readonly popupManagerService = inject(PopupManagerService)
```

docs/rules-engine/how-to-use/custom-action.md
```diff
@@ -0,0 +1,153 @@
...
+import {Component} from '@angular/core';
+import {RulesEngineRunnerService} from '@o3r/rules-engine';
+
+@Component({
+  selector: 'app-root',
+  templateUrl: './app.component.html',
+})
+export class AppComponent {
+
+  constructor(rulesEngine: RulesEngineRunnerService, popupActionHandler: PopupActionHandler) {
```
matthieu-crouzet, line 149: Same comment as earlier we can maybe recommend to do it in the `main.ts`

docs/rules-engine/examples/portal-facts.md
cpaulve-1A: I got lost writing this

docs/rules-engine/how-to-use/build-in-facts.md
cpaulve-1A: As discussed, we cannot do it in this case

docs/rules-engine/how-to-use/build-in-facts.md
cpaulve-1A: I agree with you on this point.

docs/rules-engine/how-to-use/chrome-extension.md
```diff
@@ -0,0 +1,111 @@
...
+![rule-applied.png](../../../.attachments/screenshots/rules-engine-debug/rule-applied.png)
+* Ruleset with no effect: A ruleset execution that did not fail but that did not result in any actions after execution.
+![rule-no-effect.png](../../../.attachments/screenshots/rules-engine-debug/rule-no-effect.png)
+* Failed ruleset: A ruleset execution that resulted in an error. The output actions will not be applied.
+![rule-error.png](../../../.attachments/screenshots/rules-engine-debug/rule-error.png)
+
+> [!WARNING]
+> 
+> If a Ruleset has never been executed because its linkedComponents and its validity range conditions has never been met,
+> the Ruleset will not be part of the list. It will not be deactivated.
```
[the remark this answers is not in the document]
cpaulve-1A: shown as deactivated: it does not appear in the list

docs/rules-engine/how-to-use/industrialize-ruleset-generation.md
cpaulve-1A: carrots :)

docs/rules-engine/how-to-use/industrialize-ruleset-generation.md
```diff
@@ -0,0 +1,297 @@
...
+export const notArrayContains: Operator<SupportedSimpleTypes[], SupportedSimpleTypes> = {
+  name: 'notArrayContains',
+  evaluator: (array, value) => array.indexOf(value) === -1,
+  validateLhs: (inputArray) => Array.isArray(inputArray)
+};
+```
+
+### Facts Metadata
+
+The facts file are identified thanks to ``factFilePatterns`` parameters of the ``@o3r/rules-engine:extractor`` builder.
```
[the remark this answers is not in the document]
cpaulve-1A: Actually, it is the same as the operator, maybe I should either use `fact files` or `fact's files`?
What do you think @sdo-1A

docs/rules-engine/how-to-use/build-in-facts.md
cpaulve-1A: As discussed, I will keep this documentation but we will expose the CurrentTimeFactsService in the rules-engine package

docs/components/PLACEHOLDERS.md
```diff
@@ -346,3 +396,53 @@ depend on the event itself (Easter or next Summer Holidays for example).
 This means that ``increment`` might have a different value depending on the context of the page which might be tricky to
 maintain and to debug.
 Try to keep it as simple as possible.
+
+[//]: # (These sections come from the placeholder.md and should be reworked in a dedicated PR)
```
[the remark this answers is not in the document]
cpaulve-1A: I don't think so, it should be part of the placeholder refacto 
Should it be part of the component documentation refacto?

docs/rules-engine/how-to-use/README.md
cpaulve-1A: Well we should probably rework the showcase so that it matches our recommendation and then make reference to the code in the documentation. 
This PR is now getting a bit long, I would rather do the move of the showcase in another one.

docs/components/PLACEHOLDERS.md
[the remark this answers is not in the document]
cpaulve-1A: https://github.com/AmadeusITGroup/otter/issues/1861

docs/rules-engine/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+In that case, we call it a [runtime fact](#runtime-fact).
+
+> [!WARNING]
+> A rule depending on a fact will be re-evaluated whenever this fact is updated. 
+> If you use a fact that will be updated a lot in a short time frame, you may encounter performance issues.
+> 
+> You can avoid this by:
+> * disabling the ruleset if the component impacted by the rule is missing from the page (see [linkedComponents](./how-to-use/dedicated-component-ruleset.md))
+> * avoiding using [JSONPath](./examples/complex-fact.md) on facts that are updated a lot in a short time frame and
```
pginoux-1A, line 124: should now be less a problem with the distinctUntilChanged : 
image

pginoux-1A: But still a good practice, and if a subset is used several times it's better to create a dedicated fact instead of applying the same jsonPath several times on the same fact.

docs/rules-engine/how-to-use/README.md
```diff
@@ -0,0 +1,218 @@
...
+
+This Ruleset only makes reference to an action already provided by the Otter framework.
+For this example, let's consider the `outboundDate` fact in the [Otter showcase fact folder](https://github.com/AmadeusITGroup/otter/blob/main/apps/showcase/src/facts/trip/trip.facts.ts#L23) and the 
+`UPDATE_CONFIG` action provided in the `@o3r/configuration` package and the `duringSummer` operator defined in the [Otter showcase operator folder](https://github.com/AmadeusITGroup/otter/blob/main/apps/showcase/src/operators/during-summer/during-summer.operator.ts#L6).
+
+Look into the [Going Further](#going-further) section to look for more complex and optimized use cases.
+
+### Step 1: Integrate the service
+
+First, the `RulesEngineRunnerModule` should be imported your application module (see [showcase application configuration](https://github.com/AmadeusITGroup/otter/blob/main/apps/showcase/src/app/app.module.ts#L102)).
```
pginoux-1A, line 38: in your*

docs/rules-engine/how-to-use/README.md
```diff
@@ -0,0 +1,218 @@
...
+run on pages where this component is missing. You can restrict the evaluation of your Ruleset to only pages with your
+component thanks to the `linkedComponents` property and improve the performances of your application.
+
+Check out the [linkedComponents documentation](./dedicated-component-ruleset.md).
+
+#### How to use query facts
+
+Let's consider an application that can be directly targeted via a deeplink. The parameters of the deeplink will be used 
+as the inputs of your rules, and they will help deliver a different user experience based on the deeplink.
+This is something you may need for A/B testing or to apply a customization based on the website used as a portal to your
```
pginoux-1A, line 209: For example, you can leverage this to do A/B testing in your application, sending a parameter with 2 different values and creating rules based on this fact's value different parameters.

docs/rules-engine/how-to-use/built-in-facts.md
```diff
@@ -0,0 +1,114 @@
...
+export class AppComponent implements OnInit {
+  constructor(private currentTimeFactsService: CurrentTimeFactsService) {}
+
+  ngOnInit() {
+    this.currentTimeFactsService.register();
+  }
+}
+```
+
+The ``CurrentTimeFactsService`` is based on a simple implementation of a refresh mechanism. It is up to the
```
pginoux-1A, line 76: and we do not provide a default refresh timer to avoid any performance issue. Instead, it is up to...

docs/rules-engine/how-to-use/custom-fact.md
```diff
@@ -0,0 +1,125 @@
...
+
+```
+
+> [!INFO]
+> There are two interesting things to notice about `fact1`:
+> * There is a pipe with a filter operator to ensure that the state is defined, before using it. This permits to register 
+> the fact even if the store is lazy loaded later in the flow. This observable will not emit until the store is loaded, but
+> rules engine will create the first emission (undefined) while waiting for the fact to emit its first value.
+> * Facts based on observable should always contain a `distinctUntilChange()` to avoid the re-evaluation of all the rules
+that depend on them.
```
pginoux-1A, line 82: formatting

docs/rules-engine/how-to-use/industrialize-ruleset-generation.md
```diff
@@ -0,0 +1,295 @@
...
+}
+```
+
+> [!NOTE]
+> The models for the `OperatorsMetadata` are available in the [rules engine metadata interface file](https://github.com/AmadeusITGroup/otter/blob/main/packages/%40o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.interfaces.ts).
+
+### Project with module sub-entries
+
+As mentioned in the previous section, the extractor looks into your project and the files of its dependencies to identify
+the schema of your complex facts.
```
pginoux-1A, line 230: formatting

cpaulve-1A: This is markdown, the back to line is just to avoid too long lines on the file. This won't appear in the documentation.

docs/rules-engine/how-to-use/custom-fact.md
cpaulve-1A: This is markdown, the back to line is just to avoid too long lines on the file. This won't appear in the documentation.

docs/rules-engine/README.md
cpaulve-1A: As discussed, I will remove it as it is sensibly the same (JSONPath versus map / pick operator that will be evaluated as many times)

docs/components/PLACEHOLDERS.md
```diff
@@ -346,3 +397,52 @@ depend on the event itself (Easter or next Summer Holidays for example).
...
+  "defaultLanguage": "en-GB",
+  "placeholderCssFiles": [
+    "dist/assets/placeholders/placeholders.css"
+  ]
+}
+```
+
+Those files will be loaded by the CMS to show the placeholder preview.
+Note that you could provide an empty file and update it with the dynamic content mechanism from AEM, to be able to
+reference the new classes afterward.
```
fpaul-1A, line 433: I am getting a bit confused by this part. Also, should we mention AEM at all ?

docs/rules-engine/README.md
```diff
@@ -0,0 +1,147 @@
+# Otter Rules Engine
+
+## Introduction
+
+This document describes the various concepts and models that are part of the [Otter rules engine module](https://www.npmjs.com/package/@o3r/rules-engine).
+It serves as an introduction course before you can integrate it within your application following the
+[rules engine implementation guide](./how-to-use/README.md).
+
+As a reminder, the rules engine is a client-based mechanism that evaluates a list of [Rulesets](#ruleset) to define [actions](#action) to
```
fpaul-1A, line 9: The `As a reminder` looks weird to me in the document that is supposed to introduce the concept
```suggestion
In a nutshell, the rules engine is a client-based mechanism that evaluates a list of [Rulesets](#ruleset) to define [actions](#action) to
```

cpaulve-1A: It is a reminder on concepts covered before in other documentation (the npmjs), so in a nutshell would not fit.
Maybe, 
```suggestion
As covered in the [package documentation](https://www.npmjs.com/package/@o3r/rules-engine), the rules engine is a client-based mechanism that evaluates a list of [Rulesets](#ruleset) to define [actions](#action) to
```
What do you think?

docs/components/PLACEHOLDERS.md
cpaulve-1A: This part is just moved from the rules engine documentation and should be covered in the placeholder (component) documentation (refacto).
See #1861

fpaul-1A (dismissed):

docs/rules-engine/examples/runtime-facts.md
```diff
@@ -1,92 +1,238 @@
...
+                "actionType": "UPDATE_PLACEHOLDER",
+                "key": "teaser-placeholder",
+                "value": "assets/placeholders/teaser-activities.json"
+              }
             ]
           }
         }
       ]
     },
+    {
```
mrednic-1A, line 190: I think this rule has to be in the ruleset above. As the indentation is done now, I think it is at the same level as the ruleset definition.

docs/rules-engine/how-to-use/README.md
```diff
@@ -0,0 +1,220 @@
...
+Note that in the 'New-York availability in summer rule', we only need to register the `UPDATE_CONFIG` action.
+
+> [!WARNING]
+> In case the Rules Engine results in an unregistered action, a warning will be raised. You will be able to access it
+> thanks to the [rules engine debug tools](#how-to-debug-your-application).
+
+### Step 3: Integrate your facts
+
+Now that the rules engine recognizes the `UPDATE_CONFIG` action, it needs to recognize the `outboundDate` [fact](../README.md#fact)
+which will emit values according to the window size of your user's browser.
```
mrednic-1A, line 72: this is not from the 'isMobileDeviceFact'?
I think it should be `which should emit values when the outboundDate changes`

docs/rules-engine/how-to-use/built-in-facts.md
```diff
@@ -0,0 +1,116 @@
...
+this is not the case for `dateInNextMinutes` nor `dateNotInNextMinutes`.
+
+While this fact is provided in the rules engine module, you should be able to override its implementation to fit your needs.
+
+This section explains how to use operators with an implicit fact and how to override the fact implementation.
+
+## Usage
+The operators using a built-in implicit fact are used in the same way as any other operator, the implicit fact will just
+not be referred to in the Ruleset object:
+```json55
```
mrednic-1A, line 18: json5 ?

docs/rules-engine/how-to-use/dedicated-component-ruleset.md
```diff
@@ -1,15 +1,60 @@
...
+})
+- export class MyComponent {
++ export class MyComponent implements OnInit, OnDestroy {
+  
++   private readonly rulesEngineService = inject(RulesEngineRunnerService, { optional: true });
++   private readonly componentIdentifier = computeItemIdentifier('MyComponent', '@scope/components');
+
++   public ngOnInit() {
++     if (this.rulesEngineService) {
++       this.rulesEngineService.enableRuleSetFor(this.componentName);
```
mrednic-1A, line 39: `this.componentIdentifier` instead of `this.componentName`

docs/rules-engine/how-to-use/dedicated-component-ruleset.md
```diff
@@ -18,60 +63,41 @@ First, you need to have a ruleset defined in your json (created via the CMS inte
...
 
-The RulesEngineRunnerService is the one that will handle the default subscriptions.
-At the initialization of the application, the ruleset store will provide a selector to get all Active ruleset.
-Then this selector will be used to subscribe to those rulesets in the service as an initial behavior.
-Our onDemand ruleset will not be part of this initial subscription, that's why as long as no one subscribes to it, it will not be evaluated.
-As soon as our component is instantiated on the page (ngOnInit called), it will explicitly subscribe to it, resulting in its evaluation (and refresh if any input facts changes).
-Once the subscription is removed (ngOnDestroy) called, and no other component is linked to it, the ruleset will go back to its previous 'disabled' mode.
+`RulesEngineRunnerService` handles the subscriptions for all the rulesets.
+At the initialization of the application, the service listens to a selector to get all the active rulesets that are not linked
+to any components -- note that they may have a validity date range.
```
mrednic-1A, line 95: could we add a small section with some words about this feature. If there is one, you can ignore this comment :)

packages/@o3r/rules-engine/README.md
```diff
@@ -59,6 +74,13 @@ Otter framework provides a set of code generators based on [angular schematics](
 | facts-service             | Generate a facts service                                      | `ng g facts-service`             |
 | operators                 | Generate an operator                                          | `ng g operator`                  |
 
-## Details
+## Debug
+A whole section of the [Otter Chrome Devtool extension](https://chromewebstore.google.com/detail/otter-devtools/aejabgendbpckkdnjaphhlifbhepmbne) is dedicated to the debug of the rules engine with
+a visual representation of your Rulesets and a history of the run, conditions met and actions applied.
+
+Find more information on the tool in the [Otter Chrome Devtool extension documentation](https://github.com/AmadeusITGroup/otter/blob/main/docs/dev-tools/chrome-devtools.md).
```
mrednic-1A, line 81: I think we could add a link to the `how-it-works/chrome-extension.md` here too, because it is explaining very well the dedicated section of rules engine inside the chrome extension

docs/rules-engine/how-to-use/README.md
cpaulve-1A: I must have been tired :D

mrednic-1A (approved):

mrednic-1A (approved):

fpaul-1A (approved):

PR #1739, opened 2024-05-01
fix(schematic-sdk): to support the multi tag API

### LINKED ISSUE #855: [Bug]: Multi tagged APIs not supported
### Package name

generator-sdk

### Package version

9.*.*

### Reproduction steps

API to use:

```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
servers:
  - url: http://petstore.swagger.io/v1
paths:
  /pets:
    get:
      summary: List all pets
      operationId: listPets
      tags:
        - pets
        - other
      parameters:
        - name: limit
          in: query
          description: How many items to return at one time (max 100)
          required: false
          schema:
            type: integer
            maximum: 100
            format: int32
      responses:
        '200':
          description: A paged array of pets
          headers:
            x-next:
              description: A link to the next page of responses
              schema:
                type: string
          content:
            application/json:    
              schema:
                $ref: "#/components/schemas/Pets"
 components:
  schemas:
    Pet:
      type: object
      required:
        - id
        - name
      properties:
        id:
          type: integer
          format: int64
        name:
          type: string
        tag:
          type: string
    Pets:
      type: array
      maxItems: 100
      items:
        $ref: "#/components/schemas/Pet"
    Error:
      type: object
      required:
        - code
        - message
      properties:
        code:
          type: integer
          format: int32
        message:
          type: string
```

### Current result

Build failure because of double export of `listPetsRequestData` from `other.ts` and `pets.ts`

### Expected result

- Best solution: Export once the interface used in both Apis
- possible solution: Export both interface with different name

### Additional comments

The error has been reported for `RequestData` interface only but may be also on the Apis exported functions as well.

kpanot:
## Proposed change

Add the support of multi Api Tag via the duplication of the Api Class.

## Related issues

- :bug: Fixes resolves #855

### DIFF
--- a/migration-guides/11.0.md
+++ b/migration-guides/11.0.md
@@ -6,6 +6,9 @@
 ## @ama-sdk/core
 
 - `prepareOptions` method has been removed from `ApiClient`. `getRequestOptions` should be used instead.
+- To support `JSON` specification format and to remove the reference to the outdated `Swagger` format name, the file containing the current specification used to generate the SDK will be stored in a file `open-api.yaml` or `open-api.json` instead of `swagger-api.yaml`.
+- To avoid interface name conflict in case of API with multiple tags (See issue [#855](https://github.com/AmadeusITGroup/otter/issues/855)), the API parameter object interface named with the pattern `<api-function-name>RequestData` will now be named with the `<api-function-name><api-name>RequestData`.
+  - The same renaming is applied to the `enum` types part of request parameters. It changes from `<api-function-name><parameter-name>Enum` to `<api-name><api-function-name><parameter-name>Enum`.
 
 ## @o3r/schematics
 
--- a/packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/resources/typescriptFetch/api/api.mustache
+++ b/packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/resources/typescriptFetch/api/api.mustache
@@ -11,24 +11,20 @@ import { Api, ApiClient, ApiTypes, computePiiParameterTokens, isJsonMimeType, Re
   {{#operation}}
     {{#allParams}}
       {{#isEnum}}
-export type {{#uppercaseFirst}}{{nickname}}{{/uppercaseFirst}}{{#uppercaseFirst}}{{paramName}}{{/uppercaseFirst}}Enum = {{#trimPipe}}{{#allowableValues}}{{#values}}'{{.}}' | {{/values}}{{/allowableValues}}{{/trimPipe}};
+/** Enum {{paramName}} used in the {{classname}}'s {{nickname}} function parameter */
+export type {{classname}}{{#uppercaseFirst}}{{nickname}}{{/uppercaseFirst}}{{#uppercaseFirst}}{{paramName}}{{/uppercaseFirst}}Enum = {{#trimPipe}}{{#allowableValues}}{{#values}}'{{.}}' | {{/values}}{{/allowableValues}}{{/trimPipe}};
+
       {{/isEnum}}
     {{/allParams}}
   {{/operation}}
   {{#operation}}
-export interface {{#uppercaseFirst}}{{nickname}}{{/uppercaseFirst}}RequestData {
+/** Parameters object to {{classname}}'s {{nickname}} function */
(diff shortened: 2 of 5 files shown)


packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/resources/typescriptFetch/api/api.mustache
```diff
@@ -11,23 +11,29 @@ import { Api, ApiClient, ApiTypes, computePiiParameterTokens, isJsonMimeType, Re
   {{#operation}}
     {{#allParams}}
       {{#isEnum}}
-export type {{#uppercaseFirst}}{{nickname}}{{/uppercaseFirst}}{{#uppercaseFirst}}{{paramName}}{{/uppercaseFirst}}Enum = {{#trimPipe}}{{#allowableValues}}{{#values}}'{{.}}' | {{/values}}{{/allowableValues}}{{/trimPipe}};
+/** Enum {{paramName}} used in the {{classname}}'s {{nickname}} function parameter */
+export type {{classname}}{{#uppercaseFirst}}{{nickname}}{{/uppercaseFirst}}{{#uppercaseFirst}}{{paramName}}{{/uppercaseFirst}}Enum = {{#trimPipe}}{{#allowableValues}}{{#values}}'{{.}}' | {{/values}}{{/allowableValues}}{{/trimPipe}};
+
+/** @deprecated please use {{classname}}{{#uppercaseFirst}}{{nickname}}{{/uppercaseFirst}}{{#uppercaseFirst}}{{paramName}}{{/uppercaseFirst}}Enum instead, will be remove in Otter v12 */
```
sdo-1A, line 17:
```suggestion
/** @deprecated please use {{classname}}{{#uppercaseFirst}}{{nickname}}{{/uppercaseFirst}}{{#uppercaseFirst}}{{paramName}}{{/uppercaseFirst}}Enum instead, will be removed in Otter v12 */
```
(same for 2 other comments below)

packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/resources/typescriptFetch/api/api.mustache
```diff
@@ -11,23 +11,29 @@ import { Api, ApiClient, ApiTypes, computePiiParameterTokens, isJsonMimeType, Re
...
     {{/allParams}}
 }
   {{/operation}}
   {{#operation}}
-/**
- * @Deprecated, please use {{#uppercaseFirst}}{{nickname}}{{/uppercaseFirst}}RequestData
- */
-export interface {{#uppercaseFirst}}{{nickname}}{{/uppercaseFirst}} extends {{#uppercaseFirst}}{{nickname}}{{/uppercaseFirst}}RequestData {}
+/** @deprecated please use {{classname}}{{#uppercaseFirst}}{{nickname}}{{/uppercaseFirst}}RequestData instead, will be remove in Otter v12 */
+export interface {{#uppercaseFirst}}{{nickname}}{{/uppercaseFirst}}RequestData extends {{classname}}{{#uppercaseFirst}}{{nickname}}{{/uppercaseFirst}}RequestData {}
```
mrednic-1A, line 33: even if the interface is deprecated it is still exported, and I am afraid that we will still have duplications on the barrel file, no ?  
`{{#uppercaseFirst}}{{nickname}}{{/uppercaseFirst}}RequestData` from one `my.api.ts` file and the same maybe from another file `your.api.ts`

kpanot: Yes, unfortunately the multi tag support cannot be achieved without breaking change.
So this PR is an enabler but the support will be effective only after breaking change.

cpaulve-1A (dismissed):

vscaiceanu-1a (approved):

migration-guides/11.0.md
```diff
@@ -0,0 +1,10 @@
+# Major 11.0 Migration guide
```
cpaulve-1A, line 1: Do we allow this kind of breaking in v11 where we should only allow dependency breaking changes?

kpanot: We discussed about this change and agreed that bringing this breaking change on the major would bring more benefit than waiting to v12 to allow multi tag

PR #1776, opened 2024-05-13
feat(sdk): add abort plugin SDK

kpanot:
## Proposed change

Add plugin to simplify request canceling


packages/@ama-sdk/core/src/plugins/abort/abort.fetch.ts
```diff
@@ -0,0 +1,54 @@
+import type { FetchCall, FetchPlugin, FetchPluginContext, RequestOptions } from '../core';
+
+interface AbortCallbackParameters {
+  /** URL targeted */
+  url: string;
+
+  /** Fetch call options */
+  options: RequestInit | RequestOptions;
+
+  // TODO Now supported for all the modern browsers - should become mandatory in @ama-sdk/core@10.0
```
matthieu-crouzet, line 10: we are in Otter v10 already :)

kpanot: then we forgot to apply this breaking change, will be for v11 then

packages/@ama-sdk/core/src/plugins/abort/abort.fetch.ts
```diff
@@ -0,0 +1,54 @@
...
+  /** Abort controller to abort fetch call */
+  controller?: AbortController;
+}
+
+const isPromise = (result: boolean | void | Promise<void> | Promise<boolean>): result is (Promise<void> | Promise<boolean>) => {
+  if (typeof result !== 'object') {
```
cpaulve-1A, line 16: Is this check enough and future proof? Could we check something like an exisiting catch or then property?

matthieu-crouzet (approved):

mrednic-1A (approved):

PR #1783, opened 2024-05-14
fix: schematics params should be added only for one project

vscaiceanu-1a:
## Proposed change

Today, during the `ng add`, schematic parameters are added on all projects present in the workspace. 
It should be added only for one given project.

### DIFF
--- a/packages/@o3r/schematics/src/rule-factories/ng-add/schematics-register.ts
+++ b/packages/@o3r/schematics/src/rule-factories/ng-add/schematics-register.ts
@@ -24,7 +24,51 @@ export function registerPackageCollectionSchematics(packageJson: PackageJson, an
 }
 
 /**
+ * Setup schematics default params in angular.json and per project
+ * @param schematicsDefaultParams default params to setup by schematic
+ * @param projectName The name of the project on which to add the schematics
+ * @param overrideValue Define if the given value should override the one already defined in the workspace
+ */
+export function setupSchematicsParamsForProject(schematicsDefaultParams: Record<string, SchematicOptionObject>, projectName?: string, overrideValue = false): Rule {
+  return (tree, context) => {
+    const workspace = getWorkspaceConfig(tree);
+    if (!workspace) {
+      context.logger.error('No workspace found');
+      return tree;
+    }
+    Object.entries(schematicsDefaultParams).forEach(([schematicName, defaultParams]) => {
+      workspace.schematics ||= {};
+      workspace.schematics[schematicName] = overrideValue ? {
+        ...workspace.schematics[schematicName],
+        ...defaultParams
+      } : {
+        ...defaultParams,
+        ...workspace.schematics[schematicName]
+      };
+    });
+
+    if (projectName && workspace.projects[projectName]) {
+      const project = workspace.projects[projectName];
+      Object.entries(schematicsDefaultParams).forEach(([schematicName, defaultParams]) => {
+        project.schematics ||= {};
+        if (project.schematics[schematicName]) {
+          project.schematics[schematicName] = overrideValue ? {
+            ...project.schematics[schematicName],
+            ...defaultParams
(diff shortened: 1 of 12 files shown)


packages/@o3r/schematics/src/rule-factories/ng-add/schematics-register.ts
```diff
@@ -23,8 +23,51 @@ export function registerPackageCollectionSchematics(packageJson: PackageJson, an
   };
 }
 
+/**
+ * Setup schematics default params in angular.json and per project
+ * @param schematicsDefaultParams default params to setup by schematic
+ * @param projectName The name of the project on which to add the schematics
+ * @param overrideValue Define if the given value should override the one already defined in the workspace
+ */
+export function setupSchematicsParamsForProject(schematicsDefaultParams: Record<string, SchematicOptionObject>, projectName: string | undefined, overrideValue = false): Rule {
```
matthieu-crouzet, line 32:
```suggestion
export function setupSchematicsParamsForProject(schematicsDefaultParams: Record<string, SchematicOptionObject>, projectName?: string, overrideValue = false): Rule {
```
so we don't need to specify `undefined` for the `projectName` when it's not needed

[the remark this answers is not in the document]
vscaiceanu-1a: Historically, it was considered to be useful but it is no longer the case.

mrednic-1A (dismissed):

packages/@o3r/schematics/src/rule-factories/ng-add/schematics-register.ts
```diff
@@ -23,8 +23,51 @@ export function registerPackageCollectionSchematics(packageJson: PackageJson, an
...
+        }
+      });
+    }
+    return writeAngularJson(tree, workspace);
+  };
+}
+
 /**
  * Setup schematics default params in angular.json
+ * @deprecated will be removed in v12. Use `setupSchematicsParamsForProject`
```
kpanot, line 70:
```suggestion
 * @deprecated will be removed in v12. Use {@link setupSchematicsParamsForProject}
```

packages/@o3r/schematics/src/rule-factories/ng-add/schematics-register.ts
```diff
@@ -23,8 +23,51 @@ export function registerPackageCollectionSchematics(packageJson: PackageJson, an
...
+        ...workspace.schematics![schematicName],
+        ...defaultParams
+      } : {
+        ...defaultParams,
+        ...workspace.schematics![schematicName]
+      };
+    });
+
+    if (projectName && workspace.projects[projectName]) {
+      const project = workspace.projects[projectName];
```
kpanot, line 51:
```suggestion
      const project = workspace.projects[projectName];
      project.schematics ||= {};
```

packages/@o3r/schematics/src/rule-factories/ng-add/schematics-register.ts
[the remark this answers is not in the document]
matthieu-crouzet: Will it not be the case if we want to support nx.json later ?

packages/@o3r/schematics/src/rule-factories/ng-add/schematics-register.ts
```diff
@@ -23,8 +23,52 @@ export function registerPackageCollectionSchematics(packageJson: PackageJson, an
...
+      project.schematics ||= {};
+      Object.entries(schematicsDefaultParams).forEach(([schematicName, defaultParams]) => {
+        if (project.schematics?.[schematicName]) {
```
kpanot, line 54:
```suggestion
        if (project.schematics[schematicName]) {
```

mrednic-1A (approved):

PR #1794, opened 2024-05-16
fix(create): add config registry when create is called with --registry option

matthieu-crouzet:


packages/@o3r/create/src/index.ts
```diff
@@ -198,6 +198,19 @@ const prepareWorkspace = (relativeDirectory = '.', projectPackageManager = 'npm'
     }));
   }
 
+  if (argv.registry) {
```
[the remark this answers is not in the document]
matthieu-crouzet: That option is available on all npm create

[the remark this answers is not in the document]
kpanot: Then do we receive it?
Because what you describe seems to be the difference between `npm create <package> --registry` and `npm create <package> -- --registry`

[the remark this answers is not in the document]
matthieu-crouzet: As discussed for npm we need to relate to the env variable `npm_config_registry`

sdo-1A (approved):

packages/@o3r/create/src/index.ts
```diff
@@ -191,18 +203,33 @@ const prepareWorkspace = (relativeDirectory = '.', projectPackageManager = 'npm'
...
-    cwd
-  }));
+  const registry = process.env.npm_config_registry || argv.registry;
+
+  if (registry) {
+    // Need to add this even for yarn because `ng add` only reads registry from .npmrc
+    // Not using `npm config set registry ...` to be sure to set a local config and not a user one
+    exitProcessIfErrorInSpawnSync(CREATE_NPMRC_ERROR_CODE, spawnSync(
+      'echo',
+      [`registry=${registry}`, '>', '.npmrc'],
```
kpanot, line 220: This override the complete `.npmrc` file content, maybe not what we want if there is already one.
Not sure to append the registry is a good solution neither that can lead to a double registry line.
May you can use a `npm set config -L=project`

packages/@o3r/create/src/index.ts
```diff
@@ -191,18 +203,38 @@ const prepareWorkspace = (relativeDirectory = '.', projectPackageManager = 'npm'
...
 
   if (projectPackageManager === 'yarn') {
-    exitProcessIfErrorInSpawnSync(2, spawnSync(runner, ['set', 'version', argv['yarn-version'] || 'stable'], {
-      stdio: 'inherit',
-      shell: true,
-      cwd
-    }));
+    exitProcessIfErrorInSpawnSync(YARN_SET_VERSION_ERROR_CODE, spawnSync(
+      runner,
+      ['set', 'version', argv['yarn-version'] || 'stable'],
```
cpaulve-1A, line 208: NB as not part of your code but it could be nice to add some checks over the yarn-version argument

packages/@o3r/create/src/index.ts
```diff
@@ -191,18 +203,38 @@ const prepareWorkspace = (relativeDirectory = '.', projectPackageManager = 'npm'
...
+
+  if (registry) {
+    // Need to add this even for yarn because `ng add` only reads registry from .npmrc
+    // Not using `npm config set registry ...` to be sure to set a local config and not a user one
+    exitProcessIfErrorInSpawnSync(CREATE_NPMRC_ERROR_CODE, spawnSync(
+      'echo',
+      [`registry=${registry}`, '>', '.npmrc'],
+      spawnSyncOpts
+    ));
+    exitProcessIfErrorInSpawnSync(YARN_CONFIG_REGISTRY_ERROR_CODE, spawnSync(
```
kpanot, line 223: I think this is the wrong contant.
This code is also a duplicate of the previous block, maybe you can remove the `echo` spawn above

packages/@o3r/create/src/index.ts
```diff
@@ -191,18 +205,33 @@ const prepareWorkspace = (relativeDirectory = '.', projectPackageManager = 'npm'
   writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, 2));
 
   if (projectPackageManager === 'yarn') {
-    exitProcessIfErrorInSpawnSync(2, spawnSync(runner, ['set', 'version', argv['yarn-version'] || 'stable'], {
-      stdio: 'inherit',
-      shell: true,
-      cwd
-    }));
+    const yarnVersion = quote(argv['yarn-version']);
```
kpanot, line 208: why `quote` is required here?

kpanot (approved):

packages/@o3r/create/src/index.ts
matthieu-crouzet: It was to fix the security issue raised by @cpaulve-1A https://github.com/AmadeusITGroup/otter/pull/1794#discussion_r1616131180

sdo-1A (dismissed):

packages/@o3r/create/README.md
```diff
@@ -20,25 +20,32 @@ This package is simplifying the start of an [Otter Framework](https://github.com
...
 
 ```shell
-yarn create @o3r <project-name> [...options]
+npm create @o3r <project-name> -- --package-manager=yarn [...options]
 ```
 
-> [!WARNING]
-> Please notice that the command `yarn create` is **not** available for versions *>= 2.0.0* (see [Yarn cli commands](https://yarnpkg.com/cli)).
+> [!NOTE]
+> At the moment, we support `yarn` and `npm`.
```
cpaulve-1A, line 32: It may be misleading, I thought you mean for the command not the parameter

packages/@o3r/create/src/index.ts
```diff
@@ -146,14 +155,19 @@ const createNgProject = () => {
   const options = schematicsCliOptions
     .filter(([key]) => isNgNewOptions(key))
     .flat();
-  exitProcessIfErrorInSpawnSync(1, spawnSync(`"${process.execPath}"`, [binPath, 'new', ...argv._, ...options], {
+  exitProcessIfErrorInSpawnSync(NG_NEW_ERROR_CODE, spawnSync(`"${process.execPath}"`, [binPath, 'new', ...argv._, ...options], {
```
cpaulve-1A, line 158: you could use quote for these parameters as well

packages/@o3r/create/src/index.ts
```diff
@@ -100,7 +101,8 @@ let defaultPackageManager = 'npm';
 if (packageManagerEnv && ['npm', 'yarn'].includes(packageManagerEnv)) {
   defaultPackageManager = packageManagerEnv;
 }
-const packageManager = argv['package-manager'] || defaultPackageManager;
+const argvPackageManager = argv['package-manager'];
+const packageManager = /(npm|yarn)/.test(argvPackageManager) ? argv['package-manager'] : defaultPackageManager;
```
kpanot, line 105: Maybe you want to throw with an error if an invalid (or unsupported) package manager is explicitly given

packages/@o3r/create/src/index.ts
```diff
@@ -211,16 +262,20 @@ const addOtterFramework = (relativeDirectory = '.', projectPackageManager = 'npm
...
-    stdio: 'inherit',
-    cwd,
-    shell: true,
-    env: exactO3rVersion && projectPackageManager === 'npm' ? {
-      ...process.env,
-      // eslint-disable-next-line @typescript-eslint/naming-convention, camelcase
-      NPM_CONFIG_SAVE_EXACT: 'true'
-    } : undefined
-  }));
+  exitProcessIfErrorInSpawnSync(ADD_O3R_CORE_ERROR_CODE, spawnSync(
```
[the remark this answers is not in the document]
matthieu-crouzet: I rollback the error code number 
I only change the INSTALL_PROCESS_ERROR_CODE from 2 to 6 because we were having two errors with the same code

sdo-1A (dismissed):

kpanot (approved):

sdo-1A (approved):

cpaulve-1A (approved):

vscaiceanu-1a (approved):

PR #1807, opened 2024-05-21
fix: default specs filename on sdk generation

### LINKED ISSUE #1790: [Bug]: Create an sdk from specs from an npm module fails when specs file extension is not '.yml'
### Package name

create

### Package version

10.4.0-prerelease.14

### Reproduction steps

Run:
`npm create '@ama-sdk@10.4.0-prerelease.14' typescript @my-repo/sdk -- --spec-package-name=@my-spec/core --spec-package-registry=https://my-registry/npm/registry/ --spec-package-path=dist/my-specs.yaml 

### Current result

after creating the repository  (with the shell) it will fail at sdk generation (core) because it does not find the `openapi.yml` file from which it should generate the content.

### Expected result

the sdk to be generated and not only the shell files.

### Additional comments

the file extension given as specs is hardcoded in `create` command.
the extension should be the same as the file found in the npm package containing the specs
https://github.com/AmadeusITGroup/otter/blob/5fbacc558c0b6b9d8d35f89d453ba270656b8f45/packages/%40ama-sdk/create/src/index.ts#L91

mrednic-1A:
## Related issues

- :bug: Fixes #1790
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@ama-sdk/core/cli/update-spec-from-npm.cts
+++ b/packages/@ama-sdk/core/cli/update-spec-from-npm.cts
@@ -7,18 +7,22 @@
 import * as minimist from 'minimist';
 import { existsSync } from 'node:fs';
 import { createRequire } from 'node:module';
-import { extname, join } from 'node:path';
+import { extname, posix } from 'node:path';
 import { copyFile, readFile } from 'node:fs/promises';
 import type { PackageJson } from 'type-fest';
 import type { OpenApiToolsConfiguration, OpenApiToolsGenerator } from '@ama-sdk/core';
+import { LOCAL_SPEC_FILENAME, SPEC_JSON_EXTENSION, SPEC_YAML_EXTENSION } from '@ama-sdk/core';
 
 const argv = minimist(process.argv.slice(2));
 const packageName = argv._[0];
 const { help, output, 'package-path': packagePath, quiet } = argv;
 const openApiConfigDefaultPath = './openapitools.json';
-const supportedExtensions = ['json', 'yaml', 'yml'];
 const noop = () => undefined;
 const logger = quiet ? {error: noop, warn: noop, log: noop, info: noop, debug: noop} : console;
+const SPEC_YML_EXTENSION = 'yml';
+const DEFAULT_SPEC_EXPORT_PATH_IN_NPM_MODULE = 'openapi';
+
+const supportedExtensions = [SPEC_JSON_EXTENSION, SPEC_YAML_EXTENSION, SPEC_YML_EXTENSION];
 
 if (help) {
   // eslint-disable-next-line no-console
@@ -27,7 +31,7 @@ if (help) {
 
     package-name      The full identifier of the npm package (e.g. @my-scope/my-package)
     --package-path    The relative path inside the npm package where to find the spec file (default: './openapi.yml')
-    --output          The path where the spec file should be copied (default: './openapi.yml')
+    --output          The path where the spec file should be copied (default: './${LOCAL_SPEC_FILENAME}.${SPEC_YAML_EXTENSION}')
     --quiet           Don't log anything
   `);
   process.exit(0);
@@ -40,14 +44,14 @@ if (!packageName) {
 
 void (async () => {
   let specSourcePath;
-  const appRequire = createRequire(join(process.cwd(), 'package.json'));
+  const appRequire = createRequire(posix.join(process.cwd(), 'package.json'));
   const packageJsonPath = appRequire.resolve(`${packageName}/package.json`);
   if (!packagePath) {
     const packageJson = JSON.parse(await readFile(packageJsonPath, {encoding: 'utf8'})) as PackageJson;
-    const exportMatcher = new RegExp(`openapi\\.(?:${supportedExtensions.join('|')})$`);
+    const exportMatcher = new RegExp(`^\\./${DEFAULT_SPEC_EXPORT_PATH_IN_NPM_MODULE}\\.(?:${supportedExtensions.join('|')})$`);
     const matchingExport = packageJson.exports && Object.keys(packageJson.exports).find((exportPath) => exportMatcher.test(exportPath));
     if (matchingExport) {
-      specSourcePath = appRequire.resolve(`${packageName}/${matchingExport}`);
+      specSourcePath = appRequire.resolve(posix.join(packageName, matchingExport));
     }
   } else {
     specSourcePath = packageJsonPath.replace(/package.json$/, packagePath);
@@ -59,8 +63,11 @@ void (async () => {
 
   let specDestinationPath = output;
   if (!specDestinationPath) {
-    const specSourceExtension = extname(specSourcePath);
-    specDestinationPath = `./openapi${specSourceExtension}`;
+    let specSourceExtension = extname(specSourcePath);
+    if (specSourceExtension === `.${SPEC_YML_EXTENSION}`) {
+      specSourceExtension = `.${SPEC_YAML_EXTENSION}`;
+    }
+    specDestinationPath = `./${LOCAL_SPEC_FILENAME}${specSourceExtension}`;
     if (existsSync(openApiConfigDefaultPath)) {
       const openApiConfig = JSON.parse(await readFile(openApiConfigDefaultPath, {encoding: 'utf8'})) as OpenApiToolsConfiguration;
       const generators: OpenApiToolsGenerator[] = Object.values(openApiConfig['generator-cli']?.generators ?? {});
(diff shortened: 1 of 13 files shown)


vscaiceanu-1a (dismissed):

packages/@ama-sdk/create/src/index.ts
```diff
@@ -65,6 +65,11 @@ if (argv['spec-path'] && argv['spec-package-name']) {
   process.exit(-4);
 }
 
+if (argv['spec-package-name'] && !argv['spec-package-path']) {
```
kpanot, line 68: Not sure it is a good idea to enforce `spec-package-path` in every case mainly because we generate the artifact containing the specification so we should be able to retrieve it in most of the cases.
Forcing the user to provide the path to a yarm within a package containing only one yaml file and for which the user does not the structure seems a bit too much.

For me a fix can be:
- To align the structure of the artifact (generated by the swagger-builder today may that we should define in doc) with the this part of the code determining the default spec file
- Displaying a warning with the default spec file used
- If no spec file available, throwing with this error

mrednic-1A: I think it's a good idea to default on a path. I am not so sure about the structure, because I don't know how much the swagger-builder will be used. But as I see no other options, let's use that one.

packages/@ama-sdk/core/cli/update-spec-from-npm.cts
```diff
@@ -58,8 +61,11 @@ void (async () => {
 
   let specDestinationPath = output;
   if (!specDestinationPath) {
-    const specSourceExtension = extname(specSourcePath);
-    specDestinationPath = `./openapi${specSourceExtension}`;
+    let specSourceExtension = extname(specSourcePath);
+    if (specSourceExtension === `.${SPEC_YAML_EXTENSION}` || specSourceExtension === `.${SPEC_YML_EXTENSION}`) {
```
sdo-1A, line 65: do you really need to specify the second condition since the variable will remain unchanged?
```suggestion
    if (specSourceExtension === `.${SPEC_YAML_EXTENSION}`) {
```

packages/@ama-sdk/schematics/README.md
```diff
@@ -101,6 +101,7 @@ The generated package comes with the following script in the package.json:
 
 > [!TIP]
 > The `--spec-path` parameter supports YAML and JSON file formats based on the file system path or remote URL.
+> `--spec-package-name` parameter can be used as an alternative to `--spec-path` if you want to generate an sdk with specs from an npm package.  It comes with optional parameters `--spec-package-registry`, `--spec-package-path` to be able to retrive the npm package and the specs file from it. By default the generator expects the `package.json` file inside the npm module containing the specs, to have an export with the key `./openapi.[yaml|yml|json]` and the value will be the relative path to the spec file inside the package.
```
sdo-1A, line 104:
```suggestion
> The `--spec-package-name` parameter can be used as an alternative to `--spec-path` if you want to generate an SDK with specs from an npm package.  It comes with the optional parameters `--spec-package-registry` and `--spec-package-path` to be able to retrieve the npm package and the specs file from it. By default, the generator expects the `package.json` file inside the npm module containing the specs to have an export with the key `./openapi.[yaml|yml|json]` and the value will be the relative path to the spec file inside the package.
```

packages/@ama-sdk/schematics/schematics/typescript/core/index.ts
```diff
@@ -27,15 +28,6 @@ import { generateOperationFinderFromSingleFile } from './helpers/path-extractor'
 const JAVA_OPTIONS = ['specPath', 'specConfigPath', 'globalProperty', 'outputPath'];
 const OPEN_API_TOOLS_OPTIONS = ['generatorName', 'output', 'inputSpec', 'config', 'globalProperty'];
 
-// TODO: Change to `open-api` when #1735 is done
```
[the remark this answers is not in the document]
mrednic-1A: I need to double check if we do not change it already. 
Anyway as we are still in prerelease version, I could revert the name to 'swagger-spec.yaml' and report the breaking change to v11.

packages/@ama-sdk/core/src/utils/generators.ts
```diff
@@ -0,0 +1,10 @@
+/** Name of the specification file copied locally (without extension) */
+export const LOCAL_SPEC_FILENAME = 'openapi';
+/** Extension of the Specification file in YAML format */
+export const SPEC_YAML_EXTENSION = 'yaml';
+/** Extension of the Specification file in YAML format */
+export const SPEC_YML_EXTENSION = 'yml';
```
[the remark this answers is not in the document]
mrednic-1A: I think we will stay with SPEC_YML_EXTENSION for the default even if the file ends with SPEC_YAML_EXTENSION.
So in the constants we can keep only the SPEC_YML_EXTENSION .

packages/@ama-sdk/core/cli/update-spec-from-npm.cts
mrednic-1A: true :)

packages/@ama-sdk/core/src/utils/generators.ts
```diff
@@ -0,0 +1,10 @@
+/** Name of the specification file copied locally (without extension) */
+export const LOCAL_SPEC_FILENAME = 'openapi';
```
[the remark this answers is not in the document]
mrednic-1A: kept as _swagger-spec_ for v10 and a TODO to change it in v11

packages/@ama-sdk/create/src/index.ts
```diff
@@ -64,6 +64,17 @@ if (argv['spec-path'] && argv['spec-package-name']) {
   console.error('--spec-path cannot be set with --spec-package-name');
   process.exit(-4);
 }
+/* Default path of the specification file inside npm module */
+const DEFAUL_SPEC_FILEPATH_IN_NPM_MODULE = './openapi.yml';
```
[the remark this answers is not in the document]
mrednic-1A: I removed the constant and took the logic from the `sdk` workspace generator

packages/@o3r/workspace/schematics/sdk/index.ts
```diff
@@ -47,8 +47,9 @@ function generateSdkFn(options: NgGenerateSdkSchema): Rule {
     ]), MergeStrategy.Overwrite);
 
     const packageManager = getPackageManager({ workspaceConfig });
-    const specExtension = options.specPackagePath ? path.extname(options.specPackagePath) : '.yml';
-    const specPath = options.specPackageName ? `openapi${specExtension}` : options.specPath;
+    const specExtension = options.specPackagePath ? path.extname(options.specPackagePath) : '.yaml';
+    // TODO: Change `swagger-spec` to `openapi` in v11
```
vscaiceanu-1a, line 51: can we create an issue so that we don't forget about it?

mrednic-1A: updated

packages/@ama-sdk/create/src/index.ts
```diff
@@ -65,6 +65,22 @@ if (argv['spec-path'] && argv['spec-package-name']) {
   process.exit(-4);
 }
 
+// TODO: Change to `openapi` in v11 (ref: #1745)
+/* Name of the specification file copied locally (without extension) */
+const LOCAL_SPEC_FILENAME = 'swagger-spec';
+/** Extension of the Specification file in YAML format */
+export const SPEC_YAML_EXTENSION = 'yaml';
+/** Extension of the Specification file in JSON format */
+export const SPEC_JSON_EXTENSION = 'json';
```
sdo-1A, line 74: should be `yaml` for now and updated to `json` in v11

packages/@ama-sdk/create/src/index.ts
```diff
@@ -65,6 +65,23 @@ if (argv['spec-path'] && argv['spec-package-name']) {
   process.exit(-4);
 }
 
+// TODO: Change to `openapi` in v11 (ref: #1745)
+/* Name of the specification file copied locally (without extension) */
+const LOCAL_SPEC_FILENAME = 'swagger-spec';
```
kpanot, line 70: why is it duplicating the constants here, they are already exposed by schematics and core packages which are in dep on create?

mrednic-1A: indeed. ama-sdk/core is a dep here so I could import the constants from there

sdo-1A (approved):

fpaul-1A (approved):

packages/@ama-sdk/core/cli/update-spec-from-npm.cts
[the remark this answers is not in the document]
mrednic-1A: Agree with this. Here is the issue opened for next major https://github.com/AmadeusITGroup/otter/issues/1843.

kpanot (approved):

PR #1817, opened 2024-05-23
feat: update store-sync documentation

sdo-1A:
## Proposed change

Update documentation on store-sync


matthieu-crouzet (approved):

packages/@o3r/store-sync/README.md
```diff
@@ -12,32 +12,21 @@ This package is an [Otter Framework Module](https://github.com/AmadeusITGroup/ot
...
 [![Bundle Size](https://img.shields.io/bundlephobia/min/@o3r/store-sync?color=green&style=for-the-badge)](https://www.npmjs.com/package/@o3r/store-sync)
 
-This module is exposing an NgRx store synchronization solution (synchronous and asynchronous) via the class `StorageSync`.
+This module exposes an NgRx store synchronization solution (synchronous and asynchronous) via the class `StorageSync`.
 
-To facilitate the synchronization, the `StorageSync` class is based on a fork of [ngrx-store-localstorage](https://github.com/btroncone/ngrx-store-localstorage) that is also exposed and includes the following changes on top of the original version:
+To facilitate the synchronization, the `StorageSync` class is based on a fork of [ngrx-store-localstorage](https://github.com/btroncone/ngrx-store-localstorage), an exposed package used to sync an NgRx store and the local or session storage. 
+Compared to the original version of the fork, the **@o3r/store-sync** module includes changes that improve overall synchronization performance.
+You can read about these changes in the module's [documentation](https://github.com/AmadeusITGroup/otter/blob/main/docs/store-sync/STORE_SYNC.md).
```
cpaulve-1A, line 19: I think we should discuss alltogether and decide whether we should keep the npmjs doc this vague and only specifies the feature in our documentation.

The npmjs is the entry point for our packages. I fear that if we keep it minimalistic, people will not know if it fits their needs and won't bother going more into details. 

I would be happy to have a discussion with the rest of the team on this point as they might not share this opinion.

cpaulve-1A: I would at least mention the smartSync and asyncSynchronization as features of the module.

PR #1821, opened 2024-05-23
fix: highlight test-only message

kpanot:


packages/@ama-sdk/showcase-sdk/readme.md
```diff
@@ -1,7 +1,7 @@
 # Showcase SDK
 
->Note:
-This package is intended for **testing purposes only**. It is not suitable for production use.
+> [!CAUTION]
+> This package is intended for **testing purposes only**. It is not suitable for production use.
```
mrednic-1A, line 4: As we added the 'deprecate' in package.json, the output looks like:
image

Should not be too much to have the [Caution], too ?
Or maybe remove the **deprecate** and keep only the **caution** ?

PR #1840, opened 2024-05-31
feat: use inject

matthieu-crouzet:
## Proposed change
Use `inject` to reduce the boilerplate code, and simplify the DI


sdo-1A (approved):

apps/showcase/src/app/rules-engine/rules-engine.component.ts
```diff
@@ -34,6 +34,19 @@ import { TripFactsService } from '../../facts/index';
 import { duringSummer } from '../../operators/index';
 import { CurrentTimeFactsService } from '../../services/current-time-facts.service';
 
+const prepareRulesEngine = () => {
```
mrednic-1A, line 37: why not a method in the class ?

apps/showcase/src/app/placeholder/placeholder.component.ts
```diff
@@ -11,6 +11,21 @@ import { IN_PAGE_NAV_PRES_DIRECTIVES, type InPageNavLink, InPageNavLinkDirective
...
+  const path = await firstValueFrom(
+    inject(DynamicContentService).getContentPathStream(
+      `${!environment.production ? 'assets/' : ''}rules/rulesets.json`
+    )
+  );
+
+  const resultCall = await fetch(path);
+  const result = await resultCall.json();
+
+  inject(Store<RulesetsStore>).dispatch(setRulesetsEntities({ entities: result.rulesets }));
```
fpaul-1A, line 26: is this really working ? I would assume that after the first `await` we are outside the injection context of Angular

apps/showcase/src/components/showcase/design-token/design-token-pres.component.ts
```diff
@@ -21,33 +22,40 @@ import { DatePickerInputPresComponent } from '../../utilities';
...
-    this.form.valueChanges.subscribe((value) => {
-      cleanUpStyle();
-      if (value.theme === 'dark') {
-        style = styleLoader.loadStyleFromURL({href: 'dark-theme.css'});
-      } else if (value.theme === 'horizon') {
-        style = styleLoader.loadStyleFromURL({href: 'horizon-theme.css'});
-      }
-    });
+    this.subscription.add(
+      this.form.valueChanges.subscribe((value) => {
```
fpaul-1A, line 46: should we also use `takeUntilDestroy()` here to get rid of the subscriptions array ?

matthieu-crouzet: I can be done in another PR I will open an issue :)

fpaul-1A (approved):

apps/showcase/src/components/showcase/design-token/design-token-pres.component.ts
matthieu-crouzet: https://github.com/AmadeusITGroup/otter/issues/1874

fpaul-1A (approved):

apps/showcase/src/components/showcase/component-replacement/component-replacement-pres.component.ts
```diff
@@ -22,48 +22,31 @@ const ONE_DAY_IN_MS = 24 * 60 * 60 * 1000;
...
 
-  private formatDate(dateTime: number) {
-    return formatDate(dateTime, 'yyyy-MM-dd', 'en-GB');
-  }
+  public dateFormControl = new FormControl<string | null>(this.formatDate(Date.now() + 7 * ONE_DAY_IN_MS));
 
-  private loadPresenter() {
-    this.presenter$ = this.config$.pipe(
+  constructor() {
+    this.datePickerComponent = toSignal(toObservable(this.configSignal).pipe(
```
cpaulve-1A, line 43: can't you do it with a compute() ?

apps/showcase/src/components/showcase/rules-engine/rules-engine-pres.component.ts
```diff
@@ -46,50 +50,43 @@ export class RulesEnginePresComponent implements OnDestroy, OnChanges, DynamicCo
...
+      RULES_ENGINE_PRES_CONFIG_ID,
+      RULES_ENGINE_PRES_DEFAULT_CONFIG
+    );
+
+  constructor() {
+    this.subscription.add(this.form.controls.destination.valueChanges.subscribe((destination) => this.tripService.updateDestination(destination)));
+    this.subscription.add(this.form.controls.outboundDate.valueChanges.subscribe((outboundDate) => this.tripService.updateOutboundDate(outboundDate)));
+    const inXDays$ = toObservable(computed(() => this.configSignal().inXDays));
+    this.subscription.add(
+      inXDays$.subscribe((inXDays) => {
```
cpaulve-1A, line 74: Why don't we use an effect here?

cpaulve-1A: This is also true for the `destination$` object

apps/showcase/src/components/showcase/component-replacement/component-replacement-pres.component.ts
matthieu-crouzet: As discussed `getPresenter` returns a rxjs operator so I need to convert the configSignal into an observable

apps/showcase/src/components/showcase/rules-engine/rules-engine-pres.component.ts
matthieu-crouzet: It can be done in another PR to keep it simple as it's not linked to the usage of `inject`

cpaulve-1A (approved):

PR #1857, opened 2024-06-04
feat: expose current time fact service

cpaulve-1A:
## Proposed change
Expose in library the inner o3rCurrentTime fact and delete unexpose fact factory.

Rename facts folder into inner-facts not to expose o3rCurrentTime in the fact metadata

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


packages/@o3r/rules-engine/src/public_api.ts
```diff
@@ -2,7 +2,7 @@ export * from './components/index';
 export * from './devkit/index';
 export * from './engine/index';
 export * from './fact/index';
-export * from './facts/index';
+export * from './inner-facts/index';
```
kpanot, line 5: So you do not export the `currentTimeSubject$` const anymore, it sounds like a breaking change.
I think you should still export it and flagged it as deprecated until v12

fpaul-1A (approved):

kpanot (approved):

pginoux-1A (approved):

PR #1862, opened 2024-06-06
fix: remove breaking in request-id

cpaulve-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


packages/@ama-sdk/core/src/plugins/session-id/session-id.request.ts
```diff
@@ -66,7 +65,7 @@ export class SessionIdRequest implements RequestPlugin {
   /**
    * Generates a session ID and stores it in session / backup storage.
    */
-  public generateSessionId(logger?: Logger) {
```
kpanot, line 69: Then you do a breaking change to fix a breaking change few months after the release :S.
Not sure it is the good approach.

If I understand the issue, this is linked to `public sessionId: string;` that became potentially `undefined`. If so, only this field (and its initialization in the constructor) is needed

matthieu-crouzet (approved):

PR #1863, opened 2024-06-06
test: use git clone instead of cpsync to copy it-tests app folder

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


fpaul-1A:
doesn't seem to help much

PR #1865, opened 2024-06-07
fix: publish new package for specs

mrednic-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


fpaul-1A:
I don't think we should include the publication of the showcase-sdk specs as part of the CI/CD of this repository
If it's used only for manual tests, it should be done in an other repo, a fork, or even just in local

mrednic-1A:
> I don't think we should include the publication of the showcase-sdk specs as part of the CI/CD of this repository If it's used only for manual tests, it should be done in an other repo, a fork, or even just in local

yeah, it's a bit too much code pollution just for testing  :)

PR #1899, opened 2024-06-17
test: fix it-tests not running in local

fpaul-1A:
## Proposed change

- No need to create yarn config when doing the create since we always use npm, this prevent some race condition issues 
- No need to copy the `.git` folder of the base app, this prevent some files being incorrectly versioned if committed before the setup of `.gitignore`

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/test-helpers/src/prepare-test-env.ts
+++ b/packages/@o3r/test-helpers/src/prepare-test-env.ts
@@ -69,7 +69,7 @@ export async function prepareTestEnv(folderName: string, options?: PrepareTestEn
     logger.debug?.(`Creating it-tests folder`);
     await createWithLock(() => {
       mkdirSync(itTestsFolderPath);
-      setPackagerManagerConfig(packageManagerConfig, {...execAppOptions, cwd: itTestsFolderPath});
+      setPackagerManagerConfig(packageManagerConfig, {...execAppOptions, cwd: itTestsFolderPath}, 'npm');
       return Promise.resolve();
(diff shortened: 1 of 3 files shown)


packages/@o3r/test-helpers/src/prepare-test-env.ts
```diff
@@ -89,7 +89,11 @@ export async function prepareTestEnv(folderName: string, options?: PrepareTestEn
   const prepareFinalApp = (baseApp: string) => {
     logger.debug?.(`Copying ${baseApp} to ${workspacePath}`);
     const baseProjectPath = path.join(itTestsFolderPath, baseApp);
-    cpSync(baseProjectPath, workspacePath, { recursive: true, dereference: true, filter: (source) => !/node_modules/.test(source) });
+    cpSync(baseProjectPath, workspacePath, {
+      recursive: true,
+      dereference: true,
+      filter: (source) => !/node_modules/.test(source) && !/.git[\\/]/.test(source)
```
kpanot, line 95:
```suggestion
      filter: (source) => !/node_modules/.test(source) && !/\\.git[\\/]/.test(source)
```
to avoid le path like `digit/` :D

fpaul-1A: legit

mrednic-1A (approved):

kpanot (approved):

PR #1903, opened 2024-06-18
fix: retrieve dep versions also from devDeps

### LINKED ISSUE #1902: [Bug]: `ng add @o3r/testing` is adding the `jest` dependencies with the version `latest`
### Package name

testing

### Package version

10.0.0+

### Reproduction steps

Create a new application with

```shell
npm create @o3r my-project
cd my-project
npm exec ng g application my-app
```

Add the `@o3r/testing` package:
```shell
npm exec ng add @o3r/testing --project-name my-app
```

### Current result

The dependencies to Jest have been added with the versions `latest`.

### Expected result

The versions of the Jest dependencies should be added aligned with the versions specified in the `@o3r/testing` dependencies.

### Additional comments

_No response_

vscaiceanu-1a:
## Related issues

- :bug: Fixes #1902

### DIFF
--- a/packages/@o3r/schematics/src/utility/dependencies.ts
+++ b/packages/@o3r/schematics/src/utility/dependencies.ts
@@ -1,16 +1,28 @@
 import * as fs from 'node:fs';
 import { NodeDependency, NodeDependencyType } from '@schematics/angular/utility/dependencies';
+import { logging } from '@angular-devkit/core';
+import type { PackageJson } from 'type-fest';
 
 /**
- *  Method to extract the provided package version range from a package.json file
- *  @param packageNames list of package we want to retrieve the version
- *  @param packageJsonPath Path to the package.json to refer to
- *  @returns The version range value retrieved from the provided package.json file
+ * TODO make logger parameter mandatory in v11 and remove the console fallback.
+ * Method to extract the provided package version range from a package.json file
+ * @param packageNames list of package we want to retrieve the version
+ * @param packageJsonPath Path to the package.json to refer to
+ * @param logger logger
+ * @returns The version range value retrieved from the provided package.json file
  */
-export function getExternalDependenciesVersionRange<T extends string>(packageNames: T[], packageJsonPath: string): Record<T, string> {
-  const packageJsonContent = JSON.parse(fs.readFileSync(packageJsonPath, {encoding: 'utf-8'}));
+export function getExternalDependenciesVersionRange<T extends string>(packageNames: T[], packageJsonPath: string, logger?: logging.LoggerApi): Record<T, string> {
+  const packageJsonContent = JSON.parse(fs.readFileSync(packageJsonPath, {encoding: 'utf-8'})) as PackageJson & { generatorDependencies: Record<string, string> };
   return packageNames.reduce((acc: Partial<Record<T, string>>, packageName) => {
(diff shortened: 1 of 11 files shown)


packages/@o3r/schematics/src/utility/dependencies.ts
```diff
@@ -8,9 +9,14 @@ import { NodeDependency, NodeDependencyType } from '@schematics/angular/utility/
...
-  const packageJsonContent = JSON.parse(fs.readFileSync(packageJsonPath, {encoding: 'utf-8'}));
+  const packageJsonContent = JSON.parse(fs.readFileSync(packageJsonPath, {encoding: 'utf-8'})) as PackageJson & { generatorDependencies: Record<string, string> };
   return packageNames.reduce((acc: Partial<Record<T, string>>, packageName) => {
-    acc[packageName] = packageJsonContent.peerDependencies?.[packageName] || packageJsonContent.generatorDependencies?.[packageName] || packageJsonContent.dependencies?.[packageName] || 'latest';
+    acc[packageName] =
+      packageJsonContent.peerDependencies?.[packageName] ||
+      packageJsonContent.generatorDependencies?.[packageName] ||
+      packageJsonContent.dependencies?.[packageName] ||
+      packageJsonContent.devDependencies?.[packageName] ||
+      'latest';
```
fpaul-1A, line 19: should we keep that or throw an error ?

vscaiceanu-1a: That's a good question. We might consider logging a warning and fallbacking on `latest` so that we don't break the flow.

packages/@o3r/schematics/src/utility/dependencies.ts
```diff
@@ -8,9 +9,14 @@ import { NodeDependency, NodeDependencyType } from '@schematics/angular/utility/
  *  @returns The version range value retrieved from the provided package.json file
  */
 export function getExternalDependenciesVersionRange<T extends string>(packageNames: T[], packageJsonPath: string): Record<T, string> {
-  const packageJsonContent = JSON.parse(fs.readFileSync(packageJsonPath, {encoding: 'utf-8'}));
+  const packageJsonContent = JSON.parse(fs.readFileSync(packageJsonPath, {encoding: 'utf-8'})) as PackageJson & { generatorDependencies: Record<string, string> };
   return packageNames.reduce((acc: Partial<Record<T, string>>, packageName) => {
-    acc[packageName] = packageJsonContent.peerDependencies?.[packageName] || packageJsonContent.generatorDependencies?.[packageName] || packageJsonContent.dependencies?.[packageName] || 'latest';
+    acc[packageName] =
+      packageJsonContent.peerDependencies?.[packageName] ||
+      packageJsonContent.generatorDependencies?.[packageName] ||
```
kpanot, line 16: the `generatorDependencies` should be the first in the priority order actually (to handle the case we do not want it to be aligned with devDeps for example)

kpanot (approved):

fpaul-1A (approved):

packages/@o3r/schematics/src/utility/dependencies.ts
fpaul-1A: I was thinking an error would be better to catch the exception directly in the it-tests, but I'm ok with the warning

mrednic-1A (approved):

PR #1908, opened 2024-06-20
feat(ama-sdk): enhance performance plugin to report performance mark

kpanot:
## Proposed change

Enhance performance plugin

### DIFF
--- a/packages/@ama-sdk/core/src/plugins/perf-metric/perf-metric.fetch.ts
+++ b/packages/@ama-sdk/core/src/plugins/perf-metric/perf-metric.fetch.ts
@@ -40,6 +40,14 @@ export interface Mark {
    */
   endTime?: number;
 }
+/** Performance object supporting NodeJs Performance and Web Performance reporting  */
+type CrossPlatformPerformance = {
+  /** @see Performance.mark */
+  mark: (...x: Parameters<Performance['mark']>) => ReturnType<Performance['mark']> | void;
+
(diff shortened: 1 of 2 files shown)


matthieu-crouzet (dismissed):

fpaul-1A (approved):

matthieu-crouzet (approved):

packages/@ama-sdk/core/src/plugins/perf-metric/perf-metric.fetch.ts
```diff
@@ -115,15 +162,18 @@ export class PerformanceMetricPlugin implements FetchPlugin {
    * @param response Response of the call associated to the mark to close
    */
   public closeMark(markId: string, response: Response) {
+    const perfMark = this.performance?.mark(this.getPerformanceTag('end', markId)) || undefined;
+    const endTime = perfMark?.startTime ?? this.getTime();
```
mrednic-1A, line 166: Dummy question. It is not 'perfMark?.endtime ?? ...'

kpanot: Actually it's a mark so there no "start" and "end", it's just a date when the mark has been created.
The create date of a mark is returned by the `startTime` field.
(cf. https://developer.mozilla.org/en-US/docs/Web/API/PerformanceMark)

PR #1912, opened 2024-06-21
chore: angular 18

sdo-1A:
## Proposed change

Migration to Angular 18


packages/@o3r/workspace/package.json
```diff
@@ -106,13 +107,13 @@
...
-    "@ngrx/effects": "~17.2.0",
-    "@ngrx/store-devtools": "~17.2.0"
+    "@angular/material": "~18.0.0",
+    "@ngrx/router-store": "~18.0.0",
+    "@ngrx/effects": "~18.0.0",
+    "@ngrx/store-devtools": "~18.0.0"
   },
   "engines": {
-    "node": ">=18.0.0"
+    "node": "^18.19.1 || ^20.11.1 || >=22.0.0"
```
vscaiceanu-1a, line 116: `^18.19.1 \|\| ^20.11.1 \|\| ^22.0.0`

packages/@o3r/components/src/rules-engine/placeholder.rules-engine.effect.spec.ts
```diff
@@ -1,6 +1,6 @@
 import {TestBed} from '@angular/core/testing';
 import {provideMockActions} from '@ngrx/effects/testing';
-import type {TypedAction} from '@ngrx/store/src/models';
+import type {Action} from '@ngrx/store/src/models';
```
matthieu-crouzet, line 3:
```suggestion
import type {Action} from '@ngrx/store';
```

packages/@o3r/workspace/package.json
fpaul-1A: is that needed ?
we are using the same as in angular repo https://github.com/angular/angular/blob/main/packages/core/package.json#L8

vscaiceanu-1a: According to their docs: https://angular.dev/reference/versions#actively-supported-versions
But maybe for some reason they do not want to block future majors. So not blocking.

vscaiceanu-1a (approved):

packages/@o3r/workspace/package.json
fpaul-1A: Oh sorry, I didn't noticed the `^22` vs `>=22` I was too focused on the `\|\|` :D
I'm not sure why the mismatch, but that means we are pretending to support the odd versions of Node after 22
I suggest we keep it aligned with the package.json of angular for now but keep an eye on it for future releases of Node and Angular

matthieu-crouzet (approved):

mrednic-1A (approved):

PR #1930, opened 2024-06-28
ci: upload chrome extension artifact even if vscode fails

fpaul-1A:
## Proposed change

Today, if one of the tasks in `publish:extension` fails, the upload of chrome-extension is skipped

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -140,7 +140,7 @@ jobs:
           CHROME_EXT_ID: ${{ secrets.CHROME_EXT_ID }}
           CHROME_REFRESH_TOKEN: ${{ secrets.CHROME_REFRESH_TOKEN }}
       - name: Expose Chrome extension artifact
-        if: '!inputs.prerelease'
+        if: (success() || failure()) && !inputs.prerelease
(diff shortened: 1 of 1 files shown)


kpanot (approved):

.github/workflows/publish.yml
```diff
@@ -140,7 +140,7 @@ jobs:
           CHROME_EXT_ID: ${{ secrets.CHROME_EXT_ID }}
           CHROME_REFRESH_TOKEN: ${{ secrets.CHROME_REFRESH_TOKEN }}
       - name: Expose Chrome extension artifact
-        if: '!inputs.prerelease'
+        if: (success() || failure()) && !inputs.prerelease
```
kpanot, line 143: May also need to fix (or create the issue) for the VsCode extension publishing failure :)

matthieu-crouzet (approved):

.github/workflows/publish.yml
fpaul-1A: https://github.com/AmadeusITGroup/otter/issues/1931

PR #1939, opened 2024-07-01
fix: do not log error for undefined reviver for string response

cpaulve-1A:
## Proposed change

In the case of simple response for a valid status code (for example number or string), the reviver is undefined.
We should not log this scenario

### DIFF
--- a/packages/@ama-sdk/core/src/fwk/api.helpers.ts
+++ b/packages/@ama-sdk/core/src/fwk/api.helpers.ts
@@ -132,7 +132,7 @@ export function getResponseReviver<T>(revivers: { [statusCode: number]: ReviverT
   if (typeof revivers === 'function' || typeof revivers === 'undefined') {
     return revivers;
   }
-  if (response.status && revivers[response.status]) {
+  if (response.status && Object.keys(revivers).indexOf(`${response.status}`) > -1) {
     return revivers[response.status];
   }
   if (options?.disableFallback) {


packages/@ama-sdk/core/src/fwk/api.helpers.ts
```diff
@@ -132,7 +132,7 @@ export function getResponseReviver<T>(revivers: { [statusCode: number]: ReviverT
   if (typeof revivers === 'function' || typeof revivers === 'undefined') {
     return revivers;
   }
-  if (response.status && revivers[response.status]) {
+  if (response.status && Object.keys(revivers).indexOf(`${response.status}`) > -1) {
```
kpanot, line 135: Why do you need to enforce the string conversion of the index when it was not needed before?

kpanot (approved):

packages/@ama-sdk/core/src/fwk/api.helpers.ts
cpaulve-1A: before we would try to find a fallback for 201 if we had "201: undefined".
This does not make sense since we may not need to revive the response from 201 if this is string for example.

kpanot: Sorry, I may not have been clear :S.
The purpose of my question is: why ```indexOf(`${response.status}`)``` instead of `indexOf(response.status)` ?

cpaulve-1A: Object.keys returns string and the response status is a number, these two types are not compatible.
I could use the Object.hasOwn which is experimental but seems to be supported by the recent browser (https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn#compatibilit%C3%A9_des_navigateurs). What do you think?

PR #1955, opened 2024-07-08
fix(showcase): prevent arbitrary url evaluation

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)


cpaulve-1A (approved):

apps/showcase/src/components/utilities/otter-icon/otter-icon-pres.component.ts
```diff
@@ -0,0 +1,32 @@
...
+@O3rComponent({ componentType: 'Component' })
+@Component({
+  selector: 'o3r-otter-icon-pres',
+  standalone: true,
+  templateUrl: './otter-icon-pres.template.html',
+  styleUrls: ['./otter-icon-pres.style.scss'],
+  encapsulation: ViewEncapsulation.None,
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class OtterIconPresComponent {
```
kpanot, line 13: Why a component more that a pipe?
Here you just want to calculate the `realUrl` of an icon (I also think that the `alt` should not be the url of the icon but a meaningful short description)

fpaul-1A: I agree in general that the `alt` should be a meaningful description, but in that specific case we are displaying the image that has been uploaded to create a pet. There is no description for that image and I thought in that specific case the url was relevant
As for component vs pipe, I used a component with the idea that we could want to change some html/css on these images but at the end it might not be the case, so I'm ok to only use a pipe

apps/showcase/src/components/utilities/otter-picker/otter-icon-path.pipe.ts
```diff
@@ -0,0 +1,16 @@
...
+@Pipe({
+  name: 'otterIconPath',
+  standalone: true
+})
+export class OtterIconPathPipe implements PipeTransform {
+  private readonly BASE_URL = location.href.split('/#', 1)[0];
+  private readonly ICON_MATCHER = new RegExp(OTTER_ICONS.join('|'));
+
+  public transform(value: string) {
+    const iconPath = this.ICON_MATCHER.test(value) ? value : OTTER_ICONS[0];
```
cpaulve-1A, line 13: you could control completely the value and try to look for the value in the list of supported icons ?

apps/showcase/src/components/utilities/otter-picker/otter-icon-path.pipe.ts
```diff
@@ -0,0 +1,16 @@
...
+
+@Pipe({
+  name: 'otterIconPath',
+  standalone: true
+})
+export class OtterIconPathPipe implements PipeTransform {
+  private readonly BASE_URL = location.href.split('/#', 1)[0];
+  private readonly ICON_MATCHER = new RegExp(OTTER_ICONS.join('|'));
+
+  public transform(value: string) {
```
kpanot, line 12: missing `/** @inheritdoc */`

apps/showcase/src/components/utilities/otter-picker/otter-icons.ts
```diff
@@ -0,0 +1,17 @@
+export const OTTER_ICONS = [
```
kpanot, line 1: missing comments

kpanot (approved):

vscaiceanu-1a (approved):

PR #1956, opened 2024-07-08
feat: add lerna at the root of a new otter project

vscaiceanu-1a:
## Proposed change

The goal is to add lerna at the root when creating an Otter project so that we can easily run the commands for all projects locally and on the CI with both npm and yarn.

### DIFF
--- a/packages/@o3r/workspace/package.json
+++ b/packages/@o3r/workspace/package.json
@@ -110,7 +110,8 @@
     "@angular/material": "~18.0.0",
     "@ngrx/router-store": "~18.0.0",
     "@ngrx/effects": "~18.0.0",
-    "@ngrx/store-devtools": "~18.0.0"
+    "@ngrx/store-devtools": "~18.0.0",
+    "lerna": "^8.1.7"
   },
   "engines": {
     "node": "^18.19.1 || ^20.11.1 || >=22.0.0"
--- a/packages/@o3r/workspace/schematics/index.it.spec.ts
+++ b/packages/@o3r/workspace/schematics/index.it.spec.ts
@@ -9,10 +9,12 @@ import {
   getDefaultExecSyncOptions, getGitDiff,
   packageManagerExec,
   packageManagerInstall,
+  packageManagerRun,
   packageManagerRunOnProject
 } from '@o3r/test-helpers';
-import { existsSync } from 'node:fs';
+import { existsSync, promises as fs } from 'node:fs';
 import * as path from 'node:path';
+import type { PackageJson } from 'type-fest';
 
 describe('new otter workspace', () => {
   test('should add sdk to an existing workspace', () => {
@@ -77,5 +79,22 @@ describe('new otter workspace', () => {
     expect(existsSync(path.join(workspacePath, 'project'))).toBe(false);
     generatedLibFiles.forEach(file => expect(existsSync(path.join(inLibraryPath, file))).toBe(true));
     expect(() => packageManagerRunOnProject(libName, true, { script: 'build' }, execAppOptions)).not.toThrow();
+  });
+
+  test('should generate a monorepo setup', async () => {
+    const { workspacePath } = o3rEnvironment.testEnvironment;
+    const defaultOptions = getDefaultExecSyncOptions();
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    const execAppOptions = {...defaultOptions, cwd: workspacePath, env: {...defaultOptions.env, NX_CLOUD_ACCESS_TOKEN: ''}};
+    expect(() => packageManagerInstall(execAppOptions)).not.toThrow();
+    const rootPackageJson = JSON.parse(await fs.readFile(path.join(workspacePath, 'package.json'), 'utf-8')) as PackageJson;
+    expect(rootPackageJson.scripts).toHaveProperty('build', 'lerna run build');
+    expect(rootPackageJson.scripts).toHaveProperty('test', 'lerna run test');
+    expect(rootPackageJson.scripts).toHaveProperty('lint', 'lerna run lint');
+    expect(() => packageManagerRun({script: 'build'}, execAppOptions)).not.toThrow();
+    expect(() => packageManagerRun({script: 'test'}, execAppOptions)).not.toThrow();
+    expect(() => packageManagerRun({script: 'lint'}, execAppOptions)).not.toThrow();
+    expect(rootPackageJson.workspaces).toContain('libs/*');
+    expect(rootPackageJson.workspaces).toContain('apps/*');
   });
 });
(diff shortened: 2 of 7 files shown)


packages/@o3r/workspace/schematics/ng-add/project-setup.ts
```diff
@@ -67,6 +68,7 @@ export const prepareProject = (options: NgAddSchematicsSchema): Rule => {
       generateRenovateConfig(ownSchematicsFolder),
       addVsCodeRecommendations(vsCodeExtensions),
       updateGitIgnore(workspaceConfig),
+      addLerna(ownPackageJsonContent),
```
kpanot, line 71: Maybe you can add a config to select the Monorepo manager (and so probably retrain the choice to `lerna` today)

packages/@o3r/workspace/schematics/ng-add/helpers/npm-workspace.ts
```diff
@@ -75,3 +75,33 @@ export function filterPackageJsonScripts(tree: Tree, _context: SchematicContext)
...
+      ...rootPackageJsonObject.devDependencies,
+      'lerna': o3rWorkspacePackageJson.generatorDependencies.lerna
+    };
+    rootPackageJsonObject.scripts = {
+      ...rootPackageJsonObject.scripts,
+      'build': 'lerna run build',
+      'test': 'lerna run test',
+      'lint': 'lerna run lint'
+    };
+    tree.create('/lerna.json', JSON.stringify({
```
kpanot, line 100: would not make sense to run `npx lerna init` instead?

packages/@o3r/workspace/schematics/ng-add/helpers/npm-workspace.ts
```diff
@@ -75,3 +75,33 @@ export function filterPackageJsonScripts(tree: Tree, _context: SchematicContext)
...
+      '$schema': 'node_modules/lerna/schemas/lerna-schema.json',
```
kpanot, line 101: Prefer public URL instead

packages/@o3r/workspace/schematics/ng-add/helpers/npm-workspace.ts
```diff
@@ -75,3 +75,33 @@ export function filterPackageJsonScripts(tree: Tree, _context: SchematicContext)
...
+      'version': rootPackageJsonObject.version || '0.0.0'
```
kpanot, line 102:
```suggestion
      'version': rootPackageJsonObject.version || '0.0.0-placeholder'
```

packages/@o3r/workspace/schematics/ng-add/helpers/npm-workspace.ts
kpanot, line 1: The `lerna.json` may need to be added to `set-version.cts` default files

packages/@o3r/workspace/schematics/ng-add/helpers/npm-workspace.ts
vscaiceanu-1a: I tried doing a lerna init but since the tree is in memory, and the `addWorkspacesToProject` task is not executed on the drive, `lerna init` throws an error:
`lerna ERR! Cannot initialize lerna because your package manager has not been configured to use `workspaces`, and you have not explicitly specified any packages to operate on`

Let's stick with adding the dev dep and creating the `lerna.json` by hand since we modify both properties inside ($schema and version).

packages/@o3r/workspace/schematics/ng-add/schema.json
```diff
@@ -50,6 +50,14 @@
       "type": "boolean",
       "description": "Use a pinned version for otter packages",
       "default": false
+    },
+    "monorepoManager": {
+      "description": "Which monorepo manager to use",
+      "type": "string",
+      "default": "lerna",
+      "enum": [
+        "lerna"
```
fpaul-1A, line 59: could we propose an option to select `none` ?

packages/@o3r/workspace/schematics/ng-add/helpers/npm-workspace.ts
```diff
@@ -75,3 +76,39 @@ export function filterPackageJsonScripts(tree: Tree, _context: SchematicContext)
...
+      const rootPackageJsonObject = tree.readJson(rootPackageJsonPath) as PackageJson;
+      rootPackageJsonObject.devDependencies = {
+        ...rootPackageJsonObject.devDependencies,
+        'lerna': o3rWorkspacePackageJson.generatorDependencies.lerna
+      };
+      rootPackageJsonObject.scripts = {
+        ...rootPackageJsonObject.scripts,
+        'build': 'lerna run build',
+        'test': 'lerna run test',
+        'lint': 'lerna run lint'
```
fpaul-1A, line 102: might be worth to add some test on those scripts in the it-tests

packages/@o3r/workspace/schematics/index.it.spec.ts
```diff
@@ -78,4 +79,16 @@ describe('new otter workspace', () => {
     generatedLibFiles.forEach(file => expect(existsSync(path.join(inLibraryPath, file))).toBe(true));
     expect(() => packageManagerRunOnProject(libName, true, { script: 'build' }, execAppOptions)).not.toThrow();
   });
+
+  test('should generate a monorepo setup', async () => {
+    const { workspacePath } = o3rEnvironment.testEnvironment;
+    const execAppOptions = {...getDefaultExecSyncOptions(), cwd: workspacePath};
+    expect(() => packageManagerInstall(execAppOptions)).not.toThrow();
+    const rootPackageJson = JSON.parse(await fs.readFile(path.join(workspacePath, 'package.json'), 'utf-8')) as PackageJson;
+    expect(rootPackageJson.scripts).toHaveProperty('build', 'lerna run build');
```
fpaul-1A, line 88: I was thinking about actually running the script in the test

packages/@o3r/test-helpers/src/utilities/package-manager.ts
```diff
@@ -226,7 +226,7 @@ export function setPackagerManagerConfig(options: PackageManagerConfig, execAppO
         execFileSync('yarn', ['config', 'set', 'enableGlobalCache', 'true'], execOptions);
         execFileSync('yarn', ['config', 'set', 'globalFolder', options.globalFolderPath], execOptions);
       }
-      execFileSync('yarn', ['config', 'set', 'nodeLinker', 'pnp'], execOptions);
+      execFileSync('yarn', ['config', 'set', 'nodeLinker', 'node-modules'], execOptions);
```
vscaiceanu-1a, line 229: it's a test

packages/@o3r/workspace/schematics/ng-add/schema.ts
```diff
@@ -1,5 +1,6 @@
 import type { SchematicOptionObject } from '@o3r/schematics';
 
+export type MonorepoManager = 'lerna' | 'none';
```
kpanot, line 3: missing comment to an exported type

packages/@o3r/workspace/package.json
```diff
@@ -110,7 +110,8 @@
     "@angular/material": "~18.0.0",
     "@ngrx/router-store": "~18.0.0",
     "@ngrx/effects": "~18.0.0",
-    "@ngrx/store-devtools": "~18.0.0"
+    "@ngrx/store-devtools": "~18.0.0",
+    "lerna": "8.1.4"
```
kpanot, line 114: should not be pin

mrednic-1A (approved):

fpaul-1A (approved):

PR #1967, opened 2024-07-15
fix: missing css variables in chrome dev-tool

cpaulve-1A:
## Proposed change

Do not remove css variable name from metadata extracted as it is used in the devtools
Do not fail on third party css scripts
Do not create variable for css var details

### DIFF
--- a/packages/@o3r/styling/src/devkit/styling-devtools.message.service.ts
+++ b/packages/@o3r/styling/src/devkit/styling-devtools.message.service.ts
@@ -21,8 +21,8 @@ const getCSSRulesAppliedOnRoot = () => Array.from(document.styleSheets)
     let rules;
     try {
       rules = styleSheet.cssRules || styleSheet.rules;
-    } catch {
-      console.debug(`Could not access to stylesheet ${styleSheet.href}. This might be due to CORS issues.`);
+    } catch (err) {
+      console.debug(`Could not access to stylesheet ${styleSheet.href}. This might be due to CORS issues.`, err);
     }
 
     return acc.concat(
--- a/packages/@o3r/styling/src/devkit/styling-devtools.message.service.ts
+++ b/packages/@o3r/styling/src/devkit/styling-devtools.message.service.ts
@@ -17,11 +17,20 @@ const isStylingMessage = (message: any): message is AvailableStylingMessageConte
 
 const getCSSRulesAppliedOnRoot = () => Array.from(document.styleSheets)
   .reverse()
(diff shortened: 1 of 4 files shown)


packages/@o3r/styling/src/devkit/styling-devtools.message.service.ts
```diff
@@ -17,11 +17,20 @@ const isStylingMessage = (message: any): message is AvailableStylingMessageConte
...
-    Array.from(styleSheet.cssRules || styleSheet.rules)
-      .reverse()
-      .filter((rule): rule is CSSStyleRule => rule instanceof CSSStyleRule && /\b:root\b/.test(rule.selectorText))
-  ), []);
+  .reduce((acc: CSSStyleRule[], styleSheet) => {
+    let rules;
+    try {
+      rules = styleSheet.cssRules || styleSheet.rules;
+    } catch {
+      // Could not read the rules probably due to CORS issue
```
vscaiceanu-1a, line 25: should we log something in this case?

vscaiceanu-1a (dismissed):

fpaul-1A (dismissed):

vscaiceanu-1a (approved):

fpaul-1A (dismissed):

packages/@o3r/styling/src/devkit/styling-devtools.message.service.ts
```diff
@@ -17,11 +17,20 @@ const isStylingMessage = (message: any): message is AvailableStylingMessageConte
...
-    Array.from(styleSheet.cssRules || styleSheet.rules)
-      .reverse()
-      .filter((rule): rule is CSSStyleRule => rule instanceof CSSStyleRule && /\b:root\b/.test(rule.selectorText))
-  ), []);
+  .reduce((acc: CSSStyleRule[], styleSheet) => {
+    let rules;
+    try {
+      rules = styleSheet.cssRules || styleSheet.rules;
+    } catch (err) {
+      console.debug(`Could not access to stylesheet ${styleSheet.href}. This might be due to CORS issues.`, err);
```
kpanot, line 25: Or any other network issue.
I think it's a good idea to provide hints but maybe we should be more generic.
I am thinking to something like:
```suggestion
      console.debug(`Could not access to stylesheet ${styleSheet.href}. This might be due to network issues, please check:
- network connectivity
- CORS setup
- granted access to the stylesheet`, err);
```

kpanot (dismissed):

fpaul-1A (approved):

vscaiceanu-1a (approved):

PR #1969, opened 2024-07-15
fix: wait for language to be loaded before update

cpaulve-1A:
## Proposed change

Devtool state reload fails because the update localization message is sent before the reload of the language.


packages/@o3r/localization/src/devkit/localization-devtools.service.ts
```diff
@@ -65,9 +65,10 @@ export class OtterLocalizationDevtools {
    * @param keyValues key/values to update
    * @param language if not provided, the current language value
    */
-  public updateLocalizationKeys(keyValues: { [key: string]: string }, language?: string): void | Promise<void> {
+  public async updateLocalizationKeys(keyValues: { [key: string]: string }, language?: string) {
```
kpanot, line 68: Not really a big deal but a small breaking change here :D

kpanot (approved):

fpaul-1A (approved):

PR #1985, opened 2024-07-22
fix(design): update Shadow and Gradient rendering according to specifications

kpanot:
## Proposed change

update Shadow and Gradient rendering according to specifications
cf. https://design-tokens.github.io/community-group/format/


mrednic-1A (approved):

packages/@o3r/design/src/core/design-token/parsers/design-token.parser.ts
fpaul-1A, line 1: No need to update the unit tests as well ?

fpaul-1A (dismissed):

fpaul-1A (approved):

mrednic-1A (approved):

PR #1989, opened 2024-07-24
Feature/design token jsonschema

kpanot:
## Proposed change

Add JSON Schema render to Design Token

- [x] requires #1985


packages/@o3r/design/builders/generate-jsonschema/index.ts
```diff
@@ -0,0 +1,83 @@
...
+        determineFileToUpdate: () => resolve(context.workspaceRoot, options.output),
+        tokenDefinitionRenderer: getJsonSchemaTokenDefinitionRenderer(),
+        styleContentUpdater: getJsonSchemaStyleContentUpdater({
+          id: options.schemaId,
+          description: options.schemaDescription
+        })
+      });
+      return { success: true };
+    } catch (err) {
+      return { success: false, error: `${err as any}` };
```
fpaul-1A, line 61:
```suggestion
      return { success: false, error: String(err) };
```
?

packages/@o3r/design/builders/generate-jsonschema/index.ts
```diff
@@ -0,0 +1,83 @@
...
+        .then((watcher) => watcher.on('all', async () => {
+          const res = await execute();
+
+          if (res.error) {
+            context.logger.error(res.error);
+          }
+        }));
+      return { success: true };
+    } catch (err) {
+      return { success: false, error: `${err as any}` };
```
fpaul-1A, line 80:
```suggestion
      return { success: false, error: String(err) };
```

packages/@o3r/design/builders/generate-css/index.ts
```diff
@@ -101,7 +101,7 @@ export default createBuilder<GenerateCssSchematicsSchema>(async (options, contex
       await renderDesignTokens(tokens, renderDesignTokenOptions);
       return { success: true };
     } catch (err) {
-      return { success: false, error: `${err as any}` };
```
matthieu-crouzet, line 104: Why do we do this change ?

packages/@o3r/design/README.md
```diff
@@ -49,6 +50,20 @@ The following configurations are available:
...
 
+### generate-jsonschema
+
+The `generate-jsonschema` builder can generate a [JSON Schema](https://json-schema.org/) validating and providing auto-completion to a third party Design Token file implementing a theme for the current Design system.
+The following configurations are available:
+
+| Options                     | Default Value | Description                                                                                                                          |
+| --------------------------- | ------------- | ------------------------------------------------------------------------------------------------------------------------------------ |
+| **designTokenFilePatterns** | [] *Require*  | Path patterns to the Design Token JSON files. <br /> Files in dependencies are supported and resolved with Node Resolver.            |
+| **output**                  | *null*        | Output file where the CSS will be generated. <br /> The path specified in `o3rTargetFile` will be ignore if this option is specified |
```
fpaul-1A, line 61:
```suggestion
| **output**                  | *null*        | Output file where the CSS will be generated. <br /> The path specified in `o3rTargetFile` will be ignored if this option is specified |
```

packages/@o3r/design/builders/generate-css/index.ts
fpaul-1A: I suggested this change in a previous comment.
For me it makes more sense to use `String(err)` since we are essentially just casting the Error to string.
Using template string to do a string cast seems misleading, it also requires a `as any` which is not pretty

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #1993, opened 2024-07-24
fix(localization): do not declare pipes that override another one

### LINKED ISSUE #1960: [Bug]: Warning Multiple pipes match the name `date` while using a date + translate pipe in a component for the unit tests
### Package name

localization

### Package version

10.2.16

### Reproduction steps

Create a component and use both the date and the translate pipes in the template of this component
Implement the UT file for this component with the proper imports, providers and declaration
Write whatever test and try to run it

NOTE:
we are using jest on which we have added https://www.npmjs.com/package/jest-fail-on-console in the setup-jest.ts. This will make the UT fails as soon as something is logged in the console.

### Current result

image

A warning is raised, and due to the use of jest-fail-on-console then the tests are failing. We could deactivate it (or ignore for console.warn) but we would rather not.

### Expected result

No multiple date pipe warning.

### Additional comments

This seems coming from the use of mockTranslationModules (https://github.com/AmadeusITGroup/otter/blob/9b7c439f0d4c098f18919d1264dff7d7db88cb66/packages/%40o3r/testing/src/localization/localization-mock.ts#L56) which imports the LocalizationModule (https://github.com/AmadeusITGroup/otter/blob/9b7c439f0d4c098f18919d1264dff7d7db88cb66/packages/%40o3r/localization/src/tools/localization.module.ts) and the LocalizationModule is overriding the DatePipe.
image

matthieu-crouzet:
## Proposed change
Do not declare pipes that override another one
Else we have 2 declaration of pipes with the same name and it causes `NG0313`

## Related issues

- :bug: Fixes #1960

### DIFF
--- a/packages/@o3r/testing/src/localization/localization-mock.ts
+++ b/packages/@o3r/testing/src/localization/localization-mock.ts
@@ -17,10 +17,7 @@ export class TranslatePipeMock implements PipeTransform {
   }
 }
 
-@NgModule({
-  declarations: [TranslatePipeMock],
-  exports: [TranslatePipeMock]
-})
+@NgModule()
 export class LocalizationDependencyMocks {
   public static forTest(): ModuleWithProviders<LocalizationDependencyMocks> {
     return {
--- a/packages/@o3r/localization/src/tools/localization.module.ts
+++ b/packages/@o3r/localization/src/tools/localization.module.ts
@@ -39,9 +39,9 @@ export function localeIdNgBridge(localizationService: LocalizationService) {
 export const CUSTOM_LOCALIZATION_CONFIGURATION_TOKEN = new InjectionToken<Partial<LocalizationConfiguration>>('Partial Localization configuration');
 
 @NgModule({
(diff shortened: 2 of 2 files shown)


fpaul-1A (approved):

kpanot (approved):

packages/@o3r/testing/src/localization/localization-mock.ts
```diff
@@ -17,10 +17,7 @@ export class TranslatePipeMock implements PipeTransform {
   }
 }
 
-@NgModule({
-  declarations: [TranslatePipeMock],
-  exports: [TranslatePipeMock]
-})
+@NgModule()
```
kpanot, line 20: On this version of Angular (pre-module compilation) was it possible to declare a pipe coming from a built library?
If not you may need to provide an additional module declaring these removed pipes

matthieu-crouzet: In fact outside the module you will have the pipe from the common module and not the overrided one without the export

PR #1998, opened 2024-07-25
fix(sdk-core): remove the API serialization from the mock plugin

kpanot:


fpaul-1A (approved):

matthieu-crouzet (approved):

packages/@ama-sdk/core/src/plugins/mock-intercept/mock-intercept.request.ts
```diff
@@ -37,6 +37,8 @@ export class MockInterceptRequest implements RequestPlugin {
           ...data,
           method: data.method || 'GET'
         };
+        delete requestOption.api;
```
cpaulve-1A, line 40: Do we really want to delete it as it will modify the request option for all the plugins? Is there a possibility something else relying on this property could crash (since it is mandatory on v11),
Shouldn't we make a copy or override the api toJSON method?

cpaulve-1A (approved):

PR #1999, opened 2024-07-25
fix(metadata-check): aggregate migration scripts of libraries

fpaul-1A:
## Proposed change

With the current implementation of migration-scripts, all the changes need to be documented at project level.
This is not very convenient when the changes happen in a library that the project depends on.

The proposal is to add a script to aggregate all the changes from the project + those of the libraries in the migration scripts that will be used for metadata-check and CMS updates

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/docs/cms-adapters/CMS_ADAPTERS.md
+++ b/docs/cms-adapters/CMS_ADAPTERS.md
@@ -339,3 +339,58 @@ Make sure to expose them in the bundled application by adding them in the `files
 ```
 
 Also make sure to place them in a folder name `migration-scripts` in your packaged app or to set the `migrationScriptFolder` in your `cms.json`.
+
+### Case of dependencies on libraries
+
+If the libraries that you use provide migration scripts, you need to aggregate them with your own to make the metadata-checks pass.
+
+You will need to specify in your migration script those libraries with their current version.
+
+```json5
+{
+  "$schema": "https://raw.githubusercontent.com/AmadeusITGroup/otter/main/packages/@o3r/extractors/schemas/migration.metadata.schema.json",
+  "version":  "10.0.0",
+  // List of libraries with migration scripts that your project depend on
+  "libraries": {
+    "@mylib/lib": "1.0.0"
+  },
+  "changes": [
+    // The changes specific to your project
+  ]
+}
+```
+
+Then you can automate the aggregation of migration scripts by adding the `@o3r/extractors:aggregate-migration-scripts` builder in your `angular.json` file as follows:
+```json5
+{
+  // ...,
+  "projects": {
+    // ...,
+    "<project-name>": {
+      // ...,
+      "architect": {
+        "aggregate-migration-scripts": {
+          "builder": "@o3r/extractors:aggregate-migration-scripts",
+          "options": {
+            "migrationDataPath": "./migration-scripts/src/MIGRATION-*.json",
+            "outputDirectory": "./migration-scripts/dist"
+          }
+        },
+        "check-localization-migration-metadata": {
+          "builder": "@o3r/localization:check-localization-migration-metadata",
+          "options": {
+            "migrationDataPath": "./migration-scripts/dist/MIGRATION-*.json"
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+Calling the `aggregate-migration-scripts` builder will generate the full migration-scripts including the ones from the libraries.
+
+In the previous example, you should include the output of the aggregate in the packaged application instead of the original migration scripts (`./migration-scripts/dist` instead of `./mìgration-scripts/src`).
+
+> [!WARNING]
+> The migration scripts of the libraries need be placed in the `./migration-scripts/` folder at the root the library package.
--- a/packages/@o3r/extractors/builders.json
+++ b/packages/@o3r/extractors/builders.json
@@ -1,0 +1,10 @@
+{
+  "$schema": "https://raw.githubusercontent.com/angular/angular-cli/master/packages/angular_devkit/architect/src/builders-schema.json",
+  "builders": {
+    "aggregate-migration-scripts": {
+      "implementation": "./builders/aggregate-migration-scripts/",
+      "schema": "./builders/aggregate-migration-scripts/schema.json",
+      "description": "Aggregate migration scripts"
+    }
+  }
+}
--- a/packages/@o3r/extractors/builders/aggregate-migration-scripts/index.ts
+++ b/packages/@o3r/extractors/builders/aggregate-migration-scripts/index.ts
@@ -1,0 +1,175 @@
+import { BuilderOutput, createBuilder } from '@angular-devkit/architect';
+import { createBuilderWithMetricsIfInstalled, type MigrationFile } from '@o3r/extractors';
+import * as fs from 'node:fs';
+import { mkdir, readFile, rm, writeFile } from 'node:fs/promises';
+import { basename, dirname, join, posix, resolve } from 'node:path';
+import * as globby from 'globby';
+import * as semver from 'semver';
+import { AggregateMigrationScriptsSchema } from './schema';
+
+const STEPS = [
+  'Find all migration files',
+  'Fetch migration files from libraries and aggregate the changes',
+  'Write the output'
+];
+
+/** The content of the migration file plus the filename */
+interface MigrationFileEntry<T = any> {
+  fileName: string;
+  migrationObject: MigrationFile<T>;
+}
+
+/** A map that defines for each library the history of versions defined in all the original migration scripts */
+interface LibraryVersions {
+  [libName: string]: {libVersion: string; appVersion: string}[];
+}
+
+/**
+ * Get the content of all migration files matching a glob
+ * @param glob
+ */
+const getMigrationFiles = async (glob: string | string[]): Promise<MigrationFileEntry[]> => {
+  return await Promise.all((await globby(glob, {fs}))
+    .map(async (fileName) => ({
+      fileName,
+      migrationObject: JSON.parse(await readFile(fileName, {encoding: 'utf8'}))
+    }))
+  );
+};
+
+/**
+ * Check if the lib version is valid
+ * @param libVersion
+ */
+const isValidLibVersion = (libVersion: Partial<LibraryVersions[string][number]>): libVersion is LibraryVersions[string][number] =>
+  !!libVersion.libVersion && !!libVersion.appVersion;
+
+/**
+ * Get the history of versions for each library in chronological order
+ * @param migrationFiles
+ */
+const getLibrariesVersions = (migrationFiles: MigrationFileEntry[]): LibraryVersions => {
+  const libraries: LibraryVersions = {};
+  for (const file of migrationFiles) {
+    if (file.migrationObject.libraries) {
+      for (const [lib, version] of Object.entries<string>(file.migrationObject.libraries)) {
+        if (!libraries[lib]) {
+          libraries[lib] = [];
+        }
+        const libVersion = {
+          libVersion: semver.coerce(version)?.raw,
+          appVersion: semver.coerce(file.migrationObject.version)?.raw
+        };
+
+        if (isValidLibVersion(libVersion) && !libraries[lib].some((e) => e.libVersion === libVersion.libVersion && e.appVersion === libVersion.appVersion)) {
+          libraries[lib].push(libVersion);
+        }
+      }
+    }
+  }
+  // Sort and dedupe versions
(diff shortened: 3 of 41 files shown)


vscaiceanu-1a (approved):

docs/cms-adapters/CMS_ADAPTERS.md
```diff
@@ -339,3 +339,58 @@ Make sure to expose them in the bundled application by adding them in the `files
 ```
 
 Also make sure to place them in a folder name `migration-scripts` in your packaged app or to set the `migrationScriptFolder` in your `cms.json`.
+
+### Case of dependencies on libraries
+
+If the libraries that you use are providing migration scripts, you need to aggregate them with your own to make the metadata-checks pass.
+
+You will need to specify in your migration script the list of libraries that you depend on and the current version of the library
```
pginoux-1A, line 347: the list of libraries that you depend on => those libraries (depends on + that have migration scripts)

packages/@o3r/extractors/builders.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "https://raw.githubusercontent.com/angular/angular-cli/master/packages/angular_devkit/architect/src/builders-schema.json",
+  "builders": {
```
kpanot, line 3: This builder should not be added to the project by the `ng-add` of the extractors package?

fpaul-1A: added

docs/cms-adapters/CMS_ADAPTERS.md
```diff
@@ -339,3 +339,58 @@ Make sure to expose them in the bundled application by adding them in the `files
 ```
 
 Also make sure to place them in a folder name `migration-scripts` in your packaged app or to set the `migrationScriptFolder` in your `cms.json`.
+
+### Case of dependencies on libraries
+
+If the libraries that you use are providing migration scripts, you need to aggregate them with your own to make the metadata-checks pass.
```
cpaulve-1A, line 345:
```suggestion
If the libraries that you use provide migration scripts, you need to aggregate them with your own to make the metadata-checks pass.
```

docs/cms-adapters/CMS_ADAPTERS.md
```diff
@@ -339,3 +339,58 @@ Make sure to expose them in the bundled application by adding them in the `files
...
+  "libraries": {
+    "@mylib/lib": "1.0.0"
+  },
+  "changes": [
+    // The changes specific to your project
+  ]
+}
+```
+
+Then you can automate the aggregation of migration scripts by adding the following builder in your `angular.json` file:
```
cpaulve-1A, line 363:
```suggestion
Then you can automate the aggregation of migration scripts by adding the `@o3r/extractors:aggregate-migration-scripts` builder in your `angular.json` file as follows:
```
Just to ensure that it is clear which builder does the work since you left one migration check builder in your example

vscaiceanu-1a (dismissed):

packages/@o3r/extractors/builders/aggregate-migration-scripts/index.ts
```diff
@@ -0,0 +1,169 @@
...
+  if (fs.existsSync(outputDirectory)) {
+    await rm(outputDirectory, {recursive: true});
+  }
+
+  for (const file of migrationFiles) {
+    const distFilePath = join(outputDirectory, basename(file.fileName));
+    if (!fs.existsSync(dirname(distFilePath))) {
+      await mkdir(dirname(distFilePath), {recursive: true});
+    }
+    await writeFile(distFilePath, JSON.stringify(file.migrationObject, null, 2) + '\n');
```
matthieu-crouzet, line 92: Shouldn't we use EOL ?

matthieu-crouzet (approved):

packages/@o3r/extractors/builders/aggregate-migration-scripts/schema.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/** Combine the local migration scripts of the current project with all the migration scripts of its dependencies. */
+export type AggregateMigrationScriptsSchema = JsonObject & {
+  /** Glob of the migration files to use. */
+  migrationDataPath: string | string[];
+
+  /** Path where the aggregated migration files should be written. */
+  outputDirectory: string;
+
+  /** Optional path where the libraries can be found (default to `require.resolve`) */
```
kpanot, line 11:
```suggestion
  /**
   * Optional path where the libraries can be found
   * @default module.paths
   */
```

packages/@o3r/test-helpers/src/utilities/verdaccio.ts
```diff
@@ -27,3 +29,37 @@ export function setupLocalRegistry() {
...
+      await promises.writeFile(npmrcLoggedTarget, `registry=${registry}/`);
+    }
+    execFileSync('npx', [
+      '--yes',
+      'npm-cli-login',
+      '-u', 'verdaccio',
+      '-p', 'verdaccio',
+      '-e', 'test@test.com',
+      '-r', registry,
+      '--config-path', npmrcLoggedTarget
```
cpaulve-1A, line 56: We could sanitize this parameter

packages/@o3r/extractors/builders/aggregate-migration-scripts/index.ts
```diff
@@ -0,0 +1,169 @@
...
+import { createBuilderWithMetricsIfInstalled, type MigrationFile } from '@o3r/extractors';
+import * as fs from 'node:fs';
+import { mkdir, readFile, rm, writeFile } from 'node:fs/promises';
+import { basename, dirname, join, posix, resolve } from 'node:path';
+import * as globby from 'globby';
+import * as semver from 'semver';
+import { AggregateMigrationScriptsSchema } from './schema';
+
+/** Maximum number of steps */
+const STEP_NUMBER = 3;
```
cpaulve-1A, line 11: Why 3? Is this a functional or a technical constraint?

packages/@o3r/test-helpers/src/utilities/verdaccio.ts
```diff
@@ -27,3 +29,29 @@
...
+    await promises.writeFile(npmrcLogged, `registry=${registry}/`);
+  }
+  execFileSync('npx', [
+    '--yes',
+    'npm-cli-login',
+    '-u', 'verdaccio',
+    '-p', 'verdaccio',
+    '-e', 'test@test.com',
+    '-r', registry,
+    '--config-path', npmrcLogged
```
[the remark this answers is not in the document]
cpaulve-1A: Maybe we could make sure that we stay in the same folder as the repository?

packages/@o3r/extractors/builders/aggregate-migration-scripts/index.ts
```diff
@@ -0,0 +1,169 @@
...
+      for (const [lib, version] of Object.entries<string>(file.migrationObject.libraries)) {
+        if (!libraries[lib]) {
+          libraries[lib] = [];
+        }
+        const libVersion = {
+          libVersion: semver.coerce(version)?.raw,
+          appVersion: semver.coerce(file.migrationObject.version)?.raw
+        };
+
+        if (isValidLibVersion(libVersion)) {
```
cpaulve-1A, line 61: Should we check for duplicates?

packages/@o3r/extractors/builders/aggregate-migration-scripts/index.ts
```diff
@@ -0,0 +1,169 @@
...
+  for (const file of migrationFiles) {
+    if (file.migrationObject.libraries) {
+      const appVersion = semver.coerce(file.migrationObject.version)?.raw;
+      for (const [lib, version] of Object.entries<string>(file.migrationObject.libraries)) {
+        const libIndex = librariesVersions[lib].findIndex((l) => l.appVersion === appVersion);
+        if (libIndex > 0) {
+          const newLibVersion = semver.coerce(version)?.raw;
+          const previousLibVersion = librariesVersions[lib][libIndex - 1].libVersion;
+          if (previousLibVersion !== newLibVersion) {
+            const libRange = new semver.Range(`>${previousLibVersion} <=${newLibVersion}`);
```
cpaulve-1A, line 112: Geniune question here: I am not sure I understand what we are doing:
`librariesVersion` contains the sorted list of all the version requested for each migration file.
For each migration file, we look for the previous version and look for the migration scripts in between, so why don't we just look for the migration script between the first requested version in `librariesVersion` and the latest one instead of this iteration ?

packages/@o3r/extractors/builders/aggregate-migration-scripts/index.ts
cpaulve-1A: Okay, just got to the part where it is used. Instead of having a step number, why don't we just list the steps and use the step length instead ?

packages/@o3r/rules-engine/schematics/index.it.spec.ts
```diff
@@ -31,7 +31,8 @@ describe('ng add rules-engine', () => {
     expect(diff.added).toContain('apps/test-app/cms.json');
     expect(diff.added).toContain('apps/test-app/placeholders.metadata.json');
     expect(diff.added).toContain('apps/test-app/tsconfig.cms.json');
-    expect(diff.added.length).toBe(12);
+    expect(diff.added).toContain('apps/test-app/migration-scripts/README.md');
+    expect(diff.added.length).toBe(13);
```
cpaulve-1A, line 35: What this number comes from? We have the 4 expected files, but what about the others? Could it be something we could compute ?
(eg: expectedFileList = ['app/test-app/cms.json', etc.]
expectedFileList.forEach((file) => expect(diff.added).toContain(file)))
expect(diff.added.length).toBe(? + expectedFileList.length) 
?

packages/@o3r/styling/schematics/index.it.spec.ts
```diff
@@ -29,7 +29,7 @@ describe('ng add styling', () => {
     await addImportToAppModule(applicationPath, 'TestComponentModule', 'src/components/test-component');
 
     const diff = getGitDiff(execAppOptions.cwd);
-    expect(diff.added.length).toBe(17);
+    expect(diff.added.length).toBe(18);
```
cpaulve-1A, line 32: Where does the 18 value comes from?

packages/@o3r/test-helpers/src/utilities/verdaccio.ts
```diff
@@ -27,3 +29,37 @@
...
+      await promises.writeFile(npmrcLoggedTarget, `registry=${registry}/`);
+    }
+    execFileSync('npx', [
+      '--yes',
+      'npm-cli-login',
+      '-u', 'verdaccio',
+      '-p', 'verdaccio',
+      '-e', 'test@test.com',
+      '-r', registry,
+      '--config-path', npmrcLoggedTarget
```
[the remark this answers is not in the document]
cpaulve-1A: Could we ensure that the file belongs to the same folder as the project?

packages/@o3r/test-helpers/src/utilities/verdaccio.ts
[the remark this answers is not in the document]
matthieu-crouzet: This path is meant to be used outside the repository and is reserved for a CI/local usage to publish on a local verdaccio. The risk to disclose secret is limited as of today, as in local, there is no risk to publish secrets on internet and on the CI/CD, we do not expose .

packages/@o3r/extractors/builders/aggregate-migration-scripts/index.ts
matthieu-crouzet: From what I see we are sorting on `appVersion` and here we are checking the `libVersion`

packages/@o3r/extractors/builders/aggregate-migration-scripts/index.ts
kpanot: Not sure why this comment was flagged as `resolved` when the change has not been applied.
I would prefer `\n` instead of `EOL` because we force `LF` even on Windows

kpanot (dismissed):

packages/@o3r/extractors/builders/aggregate-migration-scripts/index.ts
matthieu-crouzet: I tested with `EOL` and faced the same conclusion so I marked that comment as resolved without answering sorry

cpaulve-1A (approved):

PR #2007, opened 2024-07-30
feat(design): add type string support

kpanot:
## Proposed change

- feat: Add `string` type support
- fix: design token key name sanitizing


packages/@ama-sdk/showcase-sdk/package.json
```diff
@@ -117,6 +117,6 @@
   "engines": {
     "node": "^18.19.1 || ^20.11.1 || >=22.0.0",
     "npm": ">=4",
-    "yarn": "<5.0.0"
+    "yarn": "<2.0.0"
```
matthieu-crouzet, line 120: Why ?

.eslintrc.js
```diff
@@ -43,6 +43,7 @@ module.exports = {
       'rules': {
         '@o3r/json-dependency-versions-harmonize': ['error', {
           ignoredPackages: ['@o3r/build-helpers'],
+          ignoredDependencies: ['npm', 'yarn'],
```
matthieu-crouzet, line 46: do we have `npm` and `yarn` dependencies ?

kpanot: in `engines` properties, yes

packages/@ama-sdk/showcase-sdk/package.json
kpanot: A miss from the postinstall process.
I will rollback it

.eslintrc.js
kpanot: The story is more complicated:
- The `create` packages requires `yarn < 2.0.0`
- The workspace package.json explicitly forbid `npm` when some other packages are restriction on it.

The only compromise that can be found is:
- ignoring only `npm` at workspace level (instead of `npm` and `yarn`)
- override the rule regarding  the ignore`yarn` in `create` packages only
- remove the `yarn` and `npm` engines restriction in `showcase-sdk`

matthieu-crouzet (dismissed):

packages/@o3r/design/src/core/design-token/parsers/design-token.parser.ts
```diff
@@ -41,6 +42,9 @@ const getCssRawValue = (variableSet: DesignTokenVariableSet, {node, getType}: De
   } as typeof node;
 
   switch (checkNode.$type) {
+    case 'string': {
```
cpaulve-1A, line 45: You could add some unit-test to cover these and for us to understand what kind of sanitization we expect

matthieu-crouzet (approved):

fpaul-1A (approved):

cpaulve-1A (approved):

PR #2044, opened 2024-08-08
feat(design): add the capability to not render private token in metadata

kpanot:
## Proposed change

Add the capability to not render private token in metadata.

### DIFF
--- a/packages/@o3r/design/schemas/design-token.schema.json
+++ b/packages/@o3r/design/schemas/design-token.schema.json
@@ -59,25 +59,31 @@
       "type": "object",
       "properties": {
         "o3rTargetFile": {
+          "description": "Information regarding the path to file where the token requests to be generated",
           "type": "string"
         },
         "o3rPrivate": {
+          "description": "Determine if the token is flagged as private",
           "type": "boolean",
           "default": false
         },
         "o3rImportant": {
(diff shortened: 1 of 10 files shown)


packages/@o3r/design/schemas/design-token.schema.json
```diff
@@ -59,25 +59,31 @@
       "type": "object",
       "properties": {
         "o3rTargetFile": {
+          "description": "Information regarding the path to file where where the token requests to be generated",
```
fpaul-1A, line 62:
```suggestion
          "description": "Information regarding the path to file where the token requests to be generated",
```

packages/@o3r/design/schemas/design-token.schema.json
```diff
@@ -59,25 +59,31 @@
...
           "type": "string"
         },
         "o3rPrivate": {
+          "description": "Determine if the token is flagged as private",
           "type": "boolean",
           "default": false
         },
         "o3rImportant": {
+          "description": "Determine if the token should be flagged as important where generated",
```
fpaul-1A, line 71:
```suggestion
          "description": "Determine if the token should be flagged as important when generated",
```
?

vscaiceanu-1a (approved):

fpaul-1A (approved):

Issue #1693, opened 2024-04-22
[Feature]: Interface for ComponentIdentifier


mrednic-1A:
### Context

There are several places in the framework wherethe type 
`{ library: string; name: string }` is used explicitly.

### Proposal

Create an interface for this type (Maybe ComponentIdentifier) and share it across framework. 
Maybe _@o3r/core_ is a good candidate of module where to define the interface.

matthieu-crouzet:
I would propose `ItemIdentifier` as we have a method name `computeItemIdentifier` and we use this identifier for component and also configuration.

PR #2057, opened 2024-08-13
feat(#1693): create interface itemidentifier

matthieu-crouzet:
## Proposed change
Create interface ItemIdentifier

## Related issues

- :rocket: Feature #1693

### DIFF
--- a/packages/@o3r/core/src/core/interfaces/item-identifier.ts
+++ b/packages/@o3r/core/src/core/interfaces/item-identifier.ts
@@ -1,0 +1,13 @@
+/**
+ * Unique identifier of an item in the extracted metadata
+ */
+export interface ItemIdentifier {
+  /**
+   * Name of the library where the item is originally from
+   */
+  library: string;
+  /**
+   * Name of the item
+   */
+  name: string;
+}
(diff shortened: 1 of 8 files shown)


packages/@o3r/core/src/core/interfaces/item-identifier.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Identifier of an item
```
kpanot, line 2: I think the name and the description are too much generic.
The interface is actually used to identify components and configs in metadata (more generally in JSONs as it is also the case for Rules Engine's rules which are not metadata).

I would suggest to prefix the interface with `Otter` (or `o3r`) and give more details on the interface and fields' descriptions

packages/@o3r/core/src/core/interfaces/item-identifier.ts
```diff
@@ -0,0 +1,13 @@
...
+export interface ItemIdentifier {
+  /**
+   * Name of the library
+   */
+  library: string;
+  /**
+   * Name of the item
+   */
+  name: string;
+}
```
kpanot, line 13:
```suggestion
/**
 * Unique identifier of an item in the extracted metadata
 */
export interface ItemIdentifier {
  /**
   * Name of the library where the item is originally from
   */
  library: string;
  /**
   * Name of the item
   */
  name: string;
}
```

packages/@o3r/core/src/core/interfaces/item-identifier.ts
cpaulve-1A: I would not prefix all our interfaces with Otter as it already comes from an o3r package, however, if you want a prefix maybe `Metadata` would make sense?

kpanot (approved):

cpaulve-1A (approved):

PR #2059, opened 2024-08-14
feat(sdk): add the capability to clear dictionary field

### LINKED ISSUE #1802: [Bug]: Mandatory dictionary field remain mandatory when part of the request
### Package name

core, schematics

### Package version

*

### Reproduction steps

- Create a definition with mandatory field link to the dictionary (following example:)

```yaml
definitions:
  Amount:
    type: object
    description: Amount expressed as value and currency code.
    required:
      - value
      - currencyCode
    properties:
      value:
        type: integer
        format: int64
        minimum: 0
      rawValue:
        type: string
      currencyCode:
        type: string
        pattern: '[A-Z]{3}'
        x-dictionary-name: currency
        x-field-name: currency
        x-field-type: CurrencyDictionaryItem
```

- Use this definition in API response and in API request parameter

### Current result

First issue: The request function typing requires the parameter to be provided with the dictionary field (which can be complicated to provided by the application).
Second issue: The dictionary field is serialized and sent to the API (which can lead to call rejection if strongly checked by the API).

### Expected result

- The definition should have a dictionary mandatory in response but optional or removed in request (maybe via a type helper or a duplication of interface?)
- The serialization of the request should ignore the dictionary fields

### Additional comments

_No response_

kpanot:
## Proposed change

feat(sdk): add the capability to clear dictionary field

## Related issues

- :bug: Fixes #1802

### DIFF
--- a/docs/api-sdk/SDK_MODELS_HIERARCHY.md
+++ b/docs/api-sdk/SDK_MODELS_HIERARCHY.md
@@ -22,10 +22,17 @@ Example of a reviver:
 ```typescript
 export function reviveMyModel<T extends MyModel = MyModel>(data: any, dictionaries?: any): T  | undefined {
   if (!data) { return ; }
-  data.myField = reviverMyFiedType<MyFiedType>(data.myField, dictionaries);
+  data.myField = reviverMyFieldType<MyFieldType>(data.myField, dictionaries);
   return data as T;
 }
 ```
+
+> [!TIP]
+> The reviver can help remove dictionary fields from a revived object thanks to the `clearDictionaryFields` option:
+>
+> ```typescript
+> reviveMyModel<MyModel>(data, undefined, { clearDictionaryFields: true });
+> ```
 
 ## Core models (and base models extension)
 
--- a/packages/@ama-sdk/core/src/fwk/Reviver.ts
+++ b/packages/@ama-sdk/core/src/fwk/Reviver.ts
@@ -7,6 +7,21 @@ export interface ReviverOptions {
    * @default console
    */
   logger?: Logger;
+
+  /**
+   * Determine if the fields from dictionary should be removed from the given object.
+   * This options affects dictionary fields only if no dictionary is provided to the reviver function
+   * @default false
+   * @example Clear Dictionary fields from Data object
+   * ```typescript
+   * // Revived Data:
+   * const revivedData = reviveMyModel<MyModel>(data, myDictionary);
+   *
+   * // Remove dictionary from revived data :
+   * reviveMyModel(revivedData, undefined, { clearDictionaryFields: true });
+   * ```
+   */
+  clearDictionaryFields?: boolean;
 }
 
 /** Reviver type */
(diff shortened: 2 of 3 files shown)


matthieu-crouzet:
Shouldn't we re-generate the o3r showcase sdk with that ?

cpaulve-1A (approved):

docs/api-sdk/SDK_MODELS_HIERARCHY.md
```diff
@@ -20,13 +20,20 @@ The purpose of the revivers is to convert JSON parsed object to the model the re
...
-export function reviveMyModel<T extends MyModel = MyModel>(data: any, dictionaries?: any): T  | undefined {
+export function reviveMyModel<T extends MyModelNoDictionary = MyModel>(data: any, dictionaries?: any): T  | undefined {
   if (!data) { return ; }
   data.myField = reviverMyFiedType<MyFiedType>(data.myField, dictionaries);
   return data as T;
 }
 ```
 
+> [!TIP]
+> The reviver can help to remove the dictionary fields from a revived object when used without dictionary:
```
cpaulve-1A, line 31: So the one without the dictionnary is not the one generated by default?

kpanot: Sorry I am not sure to understand your question :S

docs/api-sdk/SDK_MODELS_HIERARCHY.md
```diff
@@ -20,13 +20,20 @@ The purpose of the revivers is to convert JSON parsed object to the model the re
...
+>
+> ```typescript
+> reviveMyModel<MyModelNoDictionary>(data);
```
mrednic-1A, line 34: The idea is to call the reviver by hand when you need a model without a revived dictionary field, when you call the API ?
Or, the reviver to remove the dictionary fields will be called any time before doing the api call?

matthieu-crouzet (dismissed):

docs/api-sdk/SDK_MODELS_HIERARCHY.md
```diff
@@ -22,11 +22,18 @@ Example of a reviver:
...
 export function reviveMyModel<T extends MyModel = MyModel>(data: any, dictionaries?: any): T  | undefined {
   if (!data) { return ; }
-  data.myField = reviverMyFiedType<MyFiedType>(data.myField, dictionaries);
+  data.myField = reviverMyFieldType<MyFieldType>(data.myField, dictionaries);
   return data as T;
 }
 ```
 
+> [!TIP]
+> The reviver can help to remove the dictionary fields from a revived object thanks to the option `clearDictionaryFields` :
```
vscaiceanu-1a, line 31:
```suggestion
> The reviver can help remove dictionary fields from a revived object thanks to the `clearDictionaryFields` option:
```

fpaul-1A (approved):

packages/@ama-sdk/core/src/fwk/Reviver.ts
```diff
@@ -7,6 +7,12 @@ export interface ReviverOptions {
    * @default console
    */
   logger?: Logger;
+
+  /**
+   * Determine if the fields from dictionary should be removed from the given object
```
fpaul-1A, line 12: it's not clear in the description that this option only applies when `dictionaries` is undefined

vscaiceanu-1a (approved):

pginoux-1A (approved):

PR #2062, opened 2024-08-14
fix: missing permissions for first pr comment workflow

matthieu-crouzet:
## Proposed change
Adding missing permissions for first pr comment workflow

[Failed workflow](https://github.com/AmadeusITGroup/otter/actions/runs/10388255377/job/28764058469)
[Needed permissions](https://docs.github.com/en/rest/issues/comments?apiVersion=2022-11-28#create-an-issue-comment)

### DIFF
--- a/.github/workflows/first-pr.yml
+++ b/.github/workflows/first-pr.yml
@@ -7,7 +7,7 @@ on:
 
 permissions:
   contents: read
-  issues: read
+  issues: write
   pull-requests: write
 
 jobs:


cpaulve-1A (approved):

kpanot (approved):

.github/workflows/first-pr.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 permissions:
   contents: read
-  issues: read
+  issues: write
```
kpanot, line 10: To follow the security guidelines, the `write` permissions should not be at job level instead?

matthieu-crouzet: As discussed this workflow contain and will contain only one job so it's equivalent

PR #2067, opened 2024-08-16
fix: clean up TODO comments

kpanot:
## Proposed change

Clean up todo


packages/@o3r/components/builders/component-extractor/index.ts
```diff
@@ -151,7 +151,6 @@ export default createBuilder(createBuilderWithMetricsIfInstalled<ComponentExtrac
    * @param libraries Libraries to watch
    */
   const watchFiles = (libraries: string[]): chokidar.FSWatcher => {
-    // TODO find a better way to watch files
```
matthieu-crouzet, line 154: Did we find a better way to watch those files ?

matthieu-crouzet (approved):

packages/@o3r/components/builders/component-extractor/index.ts
kpanot: Actually we are using chokibar  to watch the files, not sure about what we called a "better way" so I removed the comment.
If you find what be done in "better way" I am OK to re-put it with more explicit explanation

PR #2068, opened 2024-08-16
fix: clean skipLibCheck to limit to the necessary cases

kpanot:
## Proposed change

Clean `skipLibCheck` to limit to the necessary cases


matthieu-crouzet (dismissed):

packages/@o3r/analytics/tsconfig.build.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../../tsconfig.build",
   "compilerOptions": {
+    "skipLibCheck": true, // Due to Jasmine / Jest fixture compiling conflicting on types
```
sdo-1A, line 4:
```suggestion
    "skipLibCheck": true, // Due to Jasmine / Jest fixture compilation conflicting on types
```
I'm not sure if this is a good suggestion? if you agree please apply to other files

packages/@o3r/analytics/tsconfig.build.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../../tsconfig.build",
   "compilerOptions": {
+    "skipLibCheck": true, // Due to Jasmine / Jest fixture compilation conflicting on types
```
mrednic-1A, line 4: Could it be useful to have the list of libs causing this ? (@o3r/core, @o3r/schematics etc )

mrednic-1A (approved):

packages/@o3r/analytics/tsconfig.build.json
kpanot: It is actually all the modules with fixtures on Jasmine and Jest.
Where would you like to have this list? (And for which purpose? Not sure it is a argument enough strong to move out from Jasmine or Fixture mechanism :S)

mrednic-1A: I see. I was thinking to have the list of the packages which have caused the activation of the flag, here in the comment. But it's true that it can be found easily looking at the deps in package.json.

matthieu-crouzet (approved):

PR #2073, opened 2024-08-19
fix: css scrollable instructions

matthieu-crouzet:


apps/showcase/src/components/training/sdk-training-step/sdk-training-step-pres.component.html
```diff
@@ -1,15 +1,16 @@
-<div class="row mb-3 border-between ">
+<div class="d-flex h-100 gap-5">
   @if (showInstructions) {
-  <div class="col border-end overflow-y-auto" >
+  <div style="flex: 1;" class="instructions h-100 overflow-auto">
     <ng-content select=".step-instructions" />
   </div>
   }
-  <div class="col sticky-example" >
+  <div style="flex: 1;" class="d-flex h-100 overflow-hidden">
```
cpaulve-1A, line 7: why style?

PR #2079, opened 2024-08-21
feat(ama-sdk): create Mock Intercept angular plugin

kpanot:
## Proposed change

- Support for Angular ApiClient plugins
- Mock Intercept angular plugin

### DIFF
--- a/packages/@ama-sdk/core/package.json
+++ b/packages/@ama-sdk/core/package.json
@@ -36,6 +36,17 @@
       "node": "./dist/cjs/clients/*.js",
       "require": "./dist/cjs/clients/*.js"
     },
+    "./plugins/mock-intercept/angular": {
+      "module": "./dist/src/plugins/mock-intercept/mock-intercept.angular.js",
+      "esm2020": "./dist/src/plugins/mock-intercept/mock-intercept.angular.js",
+      "esm2015": "./dist/esm2015/plugins/mock-intercept/mock-intercept.angular.js",
+      "es2020": "./dist/cjs/plugins/mock-intercept/mock-intercept.angular.js",
+      "default": "./dist/cjs/plugins/mock-intercept/mock-intercept.angular.js",
+      "typings": "./dist/src/plugins/mock-intercept/mock-intercept.angular.d.ts",
+      "import": "./dist/src/plugins/mock-intercept/mock-intercept.angular.js",
+      "node": "./dist/cjs/plugins/mock-intercept/mock-intercept.angular.js",
+      "require": "./dist/cjs/plugins/mock-intercept/mock-intercept.angular.js"
+    },
     "./plugins/*": {
       "module": "./dist/src/plugins/*/index.js",
       "esm2020": "./dist/src/plugins/*/index.js",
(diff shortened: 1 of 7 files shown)


packages/@ama-sdk/core/src/plugins/mock-intercept/mock-intercept.angular.ts
```diff
@@ -0,0 +1,76 @@
...
+ * It will allow the user to delay the response or to handle the getResponse function provided with the mock (if present).
+ */
+export class MockInterceptAngular implements AngularPlugin {
+
+  constructor(protected options: MockInterceptFetchParameters) {}
+
+  public load(context: AngularPluginContext): PluginObservableRunner<HttpResponse<any>, AngularCall> {
+
+    if (!context.apiClient.options.requestPlugins.some((plugin) => plugin instanceof MockInterceptRequest)) {
+      throw new Error('MockInterceptFetch plugin should be used only with the MockInterceptRequest plugin');
```
matthieu-crouzet, line 20:
```suggestion
      throw new Error('MockInterceptAngular plugin should be used only with the MockInterceptRequest plugin');
```

packages/@ama-sdk/core/src/plugins/mock-intercept/mock-intercept.angular.ts
```diff
@@ -0,0 +1,76 @@
...
+                  return new HttpResponse<any>({
+                    ...responseCloned,
+                    body,
+                    url: responseCloned.url || undefined
+                  });
+                })
+              );
+
+            } catch {
+              (context.logger || console).error(`Failed to retrieve the latest mock for Operation ID ${operationId}, fallback to default mock`);
```
mrednic-1A, line 67: Is it a fallback to default mock or to default call (the real call)?

packages/@ama-sdk/core/package.json
```diff
@@ -36,6 +36,17 @@
       "node": "./dist/cjs/clients/*.js",
       "require": "./dist/cjs/clients/*.js"
     },
+    "./plugins/mock-intercept/angular": {
```
mrednic-1A, line 39: Is there a reason for declaring this separately, as it is handled by the declaration of `./plugins/*` just below ?

kpanot: Yes, we can not export the angular plugin in the `index.ts` (targeted by the declaration below) because we do not everybody using the mock interceptor to have a dependency on Angular :S

packages/@ama-sdk/core/src/plugins/mock-intercept/mock-intercept.angular.ts
kpanot: Copy from original fetch plugin but you're right, I will fix both of them

packages/@ama-sdk/core/package.json
mrednic-1A: ah, this is nice :)

mrednic-1A (approved):

packages/@ama-sdk/core/src/plugins/mock-intercept/mock-intercept.angular.ts
```diff
@@ -0,0 +1,76 @@
...
+      throw new Error('MockInterceptAngular plugin should be used only with the MockInterceptRequest plugin');
+    }
+
+    return {
+      transform: (call: AngularCall) => {
+        return from((
+          async () => {
+            await this.options.adapter.initialize();
+
+            let responsePromise = call;
```
fpaul-1A, line 29: looks weird to call it Promise when it's actually an Observable

fpaul-1A (approved):

PR #2080, opened 2024-08-21
fix: added generic to getConfig

guillaumerochelle:
## Proposed change

getConfig returns an observable of any.
To solve @typescript-eslint/no-unsafe-assignment eslint rule, getConfig can be made generic.

### DIFF
--- a/packages/@o3r/configuration/src/services/configuration/configuration.base.service.ts
+++ b/packages/@o3r/configuration/src/services/configuration/configuration.base.service.ts
@@ -94,8 +94,8 @@ export class ConfigurationBaseService {
    * @param defaultValue Default value of the configuration
    */
   public getComponentConfig<T extends Configuration>(id: string, defaultValue: T) {
-    return (source: Observable<Partial<T> | undefined>): Observable <T> => {
-      const componentConfigurationFromStore$ = this.getConfig(id) as Observable<T>;
+    return (source: Observable<Partial<T> | undefined>): Observable<T> => {
+      const componentConfigurationFromStore$ = this.getConfig<T>(id);
 
       return source.pipe(
         switchMap((overrideConfig) => componentConfigurationFromStore$.pipe(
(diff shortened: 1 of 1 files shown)


kpanot (approved):

packages/@o3r/configuration/src/services/configuration/configuration.base.service.ts
```diff
@@ -112,7 +112,7 @@ export class ConfigurationBaseService {
    * Get an observable of the configuration from store for a given component and merge it with the global config + the config overrides from the rules engine
    * @param id Id of the component
    */
-  public getConfig(id: string): Observable<any> {
+  public getConfig<T extends Configuration>(id: string): Observable<T> {
```
kpanot, line 115:
```suggestion
  public getConfig<T extends Configuration = Configuration>(id: string): Observable<T> {
```

To avoid breaking the application using `getConfig(...)` without type.

matthieu-crouzet (approved):

packages/@o3r/configuration/src/services/configuration/configuration.base.service.ts
guillaumerochelle: I can add it, but I think we already have the correct behavior without the default.
[Here is a typescript playground](https://www.typescriptlang.org/play/?ts=5.5.4#code/JYOwLgpgTgZghgYwgAgMIHsQ2Ac2QbwChkTkBtAawgE8AuZAZzClBwF16mWQcBuQgL6FCoSLEQoAgsggAPSCAAmDNJmx4ipZHE7NW-ISPDR4SZACEZ8iEpUYsuAsVIAjXdz6DhCTE2Q+HHAZ6ACUIHyhFAB4uVgAaVUCAPmQAXictAPVJek0tUh1kACI4IucSATjy-zVcc1zqrTdilzKtIQF+QgC-HAgwe3U05CiAFSsFZUT1JIAKYEV3VgBKenHUlLySKH6AVygQGsCGMgW2bTtavDgVUYNhAHoH5DBqAAcUdBhkAFtqQdw0jgSl+-yulmAl0C3V8YFBAJw0nSfQGV1mRSygKKy34PThfwRlmR-QR6MxOHM2K6Txe70+3z+ozpigR0khyCBIMZzMJyHZ5hhID83I+LKuSOQKIRUUkcwx4qpguF1CZot5xNRgSi5jl5MpOMez1eH2QX1BqogYsCbJUnMU5p54L5KgFeNpavF9AlUrR8utirdxsthPoRMlJN9eqpQA) which shows that config variables have the correct type inferred by TS.

cpaulve-1A (approved):

packages/@o3r/configuration/src/services/configuration/configuration.base.service.ts
kpanot: Fare enough it now the extended type is used as default, not need to explicit it :).

For info: the above method does not have the `= Configuration` because it is inferred from the second parameter.

PR #2083, opened 2024-08-23
fix(showcase): cache of dark theme generated output

kpanot:


apps/showcase/project.json
```diff
@@ -217,8 +217,7 @@
       "cache": true,
       "executor": "@o3r/design:generate-css",
       "outputs": [
-        "{options.defaultStyleFile}",
-        "{options.metadataOutput}"
+        "{options.defaultStyleFile}"
```
mrednic-1A, line 220: I think you could do the same for the horizon theme too.

mrednic-1A (approved):

matthieu-crouzet (approved):

PR #2094, opened 2024-08-26
fix: error message is empty on logged rule engine errors

cpaulve-1A:
## Proposed change

Catch returns an Error which can only be stringified with toString since JSON.stringify(new Error('Message') will always return {}

### DIFF
--- a/packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
+++ b/packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
@@ -267,7 +267,15 @@ export class RulesetExecutor {
               if (this.rulesEngine.debugMode) {
                 output.evaluation = handleRuleEvaluationDebug({ ...rule, inputFacts }, this.ruleset.name, output.actions, output.error, runtimeFactValues, factValues, oldFactValues);
               } else if (output.error) {
-                this.rulesEngine.logger?.error(output.error);
+                let errorMsg;
+                if (output.error instanceof Error) {
+                  errorMsg = output.error.toString();
+                } else if (typeof output.error === 'string') {
+                  errorMsg = output.error;
+                } else {
+                  errorMsg = JSON.stringify(output.error);
+                }
+                this.rulesEngine.logger?.error(errorMsg);
                 this.rulesEngine.logger?.warn(`Skipping rule ${rule.name}, and the associated ruleset`);
               }
               return output;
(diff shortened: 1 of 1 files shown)


mrednic-1A (dismissed):

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -267,7 +267,7 @@ export class RulesetExecutor {
               if (this.rulesEngine.debugMode) {
                 output.evaluation = handleRuleEvaluationDebug({ ...rule, inputFacts }, this.ruleset.name, output.actions, output.error, runtimeFactValues, factValues, oldFactValues);
               } else if (output.error) {
-                this.rulesEngine.logger?.error(output.error);
+                this.rulesEngine.logger?.error(output.error.toString());
```
matthieu-crouzet, line 270:
```suggestion
                this.rulesEngine.logger?.error(output.error instanceof Error ? output.error.toString() : typeof output.error === 'string' ? output.error : JSON.stringify(output.error));
```
as you can throw whatever you want 
when you catch you received an error of type `unknown`

matthieu-crouzet (approved):

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -267,7 +267,15 @@ export class RulesetExecutor {
               if (this.rulesEngine.debugMode) {
                 output.evaluation = handleRuleEvaluationDebug({ ...rule, inputFacts }, this.ruleset.name, output.actions, output.error, runtimeFactValues, factValues, oldFactValues);
               } else if (output.error) {
-                this.rulesEngine.logger?.error(output.error);
+                let errorMsg;
+                if (output.error instanceof Error) {
+                  errorMsg = output.error.toString();
+                } else if (typeof output.error === 'string') {
+                  errorMsg = output.error;
```
kpanot, line 274:
```suggestion
                if (output.error instanceof Error || typeof output.error === 'string') {
                  errorMsg = output.error.toString();
```

kpanot (approved):

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
fpaul-1A: maybe we could export this logic in a function like
```typescript
errorMsg = isJsonSerializable(output.error) ? JSON.stringify(output.error) : output.error.toString();
```

matthieu-crouzet (approved):

PR #2113, opened 2024-08-29
feat(sdk-core): migrate middleware auth plugin to ES6 crypto base

### LINKED ISSUE #2110: [Feature]: Use EcmaScript Crypto instead of Node implementation in SDK Plugins
### Context

The SDK plugin `MicroGatewayMiddlewareAuthenticationRequest` from the **@ama-sdk/core** package is depending on `node:crypto`.
To be able to use this plugin in both Node and Browser contexts, 2 different implementations is provided.

### Proposal

Removing the dependency to `node:crypto` and replace it by the [EcmaScript crypto](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) implementation

kpanot:
## Proposed change

feat(sdk-core): migrate middleware auth plugin to ES6 crypto base

## Related issues

- :rocket: Feature resolves #2110


cpaulve-1A (dismissed):

packages/@ama-sdk/core/src/plugins/mgw-mdw-auth/mgw-mdw-auth.request.ts
```diff
@@ -27,17 +24,14 @@ export async function sha256(value: string) {
  * @param value Value to hash
  * @param secretKey Secret cryptographic key
  */
-export function hmacSHA256(value: string, secretKey: string) {
```
vscaiceanu-1a, line 30: This function was exported. Do we risk breaking changes because of the `async`?

vscaiceanu-1a (dismissed):

packages/@ama-sdk/core/src/plugins/mgw-mdw-auth/mgw-mdw-auth.request.ts
kpanot: This function is exported to be tested, it does not make sense to be used by itself

fpaul-1A: maybe it should be exported in a different file then, to avoid having it in the barrel

kpanot: In term of breaking change it is even worse and will appear as function removal.
But ok, let's do it, soons like the best compromise :)

vscaiceanu-1a (approved):

fpaul-1A (approved):

Issue #2117, opened 2024-09-02
[Feature]: Turn SDK Mock API to ApiClient agnostic implementation


kpanot:
### Context

In the current implementation the generated `api-mock.ts` in the SDK is import the `ApiFetchClient`.
To be able to fully support the multi client strategy, the `api-mock.ts` file should be ApiClient agnostic.

### Proposal

- Change the implementation of the file to factory requiring the `ApiClient`
- Remove the Optional PeerDependency to `@ama-sdk/client-fetch` from the SDK shell.

PR #2118, opened 2024-09-02
feat(ama-sdk): move ApiFetchClient to dedicated package

kpanot:
## Proposed change

move ApiFetchClient to dedicated package


packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/package.json.template
```diff
@@ -116,6 +120,7 @@
     "@o3r/schematics": "<%= sdkCoreRange %>"
   },<% } %>
   "peerDependencies": {
+    "@ama-sdk/client-fetch": "<%= sdkCoreRange %>",
```
kpanot, line 123: Mandatory because of `api-mock.ts`.
Created #2117 (containing breaking change) to remove it.

packages/@ama-sdk/client-fetch/package.json
```diff
@@ -0,0 +1,128 @@
...
+    "@stylistic/eslint-plugin-ts": "~2.4.0",
+    "@swc/cli": "~0.4.0",
+    "@swc/core": "~1.7.0",
+    "@types/jest": "~29.5.2",
+    "@types/node": "^20.0.0",
+    "@types/uuid": "^9.0.0",
+    "@typescript-eslint/eslint-plugin": "^7.14.1",
+    "@typescript-eslint/parser": "^7.14.1",
+    "@typescript-eslint/utils": "^7.14.1",
+    "copyfiles": "^2.4.1",
```
matthieu-crouzet, line 101: why do we need `copyfiles` and `cpy-cli` ?

packages/@ama-sdk/client-fetch/schematics/ng-add/index.ts
```diff
@@ -0,0 +1,79 @@
...
+import * as path from 'node:path';
+import { NodeDependencyType } from '@schematics/angular/utility/dependencies';
+
+const devDependenciesToInstall: string[] = [
+
+];
+
+
+const reportMissingSchematicsDep = (logger: { error: (message: string) => any }) => (reason: any) => {
+  logger.error(`[ERROR]: Adding @ama-sdk/client-fetch has failed.
```
matthieu-crouzet, line 12: Is this package depends on `@o3r/core` ?

packages/@ama-sdk/client-fetch/src/fetch-plugin.ts
```diff
@@ -0,0 +1,46 @@
...
+  /** Abort controller to abort fetch call */
+  controller?: AbortController;
+}
+
+/**
+ * Interface of an async plugin starter
+ */
+export interface PluginAsyncStarter {
+  /** Determine if the action can start */
+  canStart?(): boolean | Promise<boolean>;
```
matthieu-crouzet, line 33: Question for my personal knowledge
is there a difference with 
```suggestion
  canStart?: () => boolean | Promise<boolean>;
```

packages/@ama-sdk/client-fetch/src/plugins/abort/abort.fetch.ts
```diff
@@ -0,0 +1,90 @@
+import type { FetchCall, FetchPlugin, FetchPluginContext, RequestOptions } from '@ama-sdk/core';
+
+interface AbortCallbackParameters {
+  /** URL targeted */
+  url: string;
+
+  /** Fetch call options */
+  options: RequestInit | RequestOptions;
+
+  // TODO Now supported for all the modern browsers - should become mandatory in @ama-sdk/core@11.0
```
matthieu-crouzet, line 10: Your PR will be merged in v11.3 or later 
shouldn't we set controller as mandatory for this new package ?

packages/@ama-sdk/client-fetch/testing/jest.config.ut.js
```diff
@@ -0,0 +1,21 @@
...
+const rootDir = path.join(__dirname, '..');
+
+/** @type {import('ts-jest/dist/types').JestConfigWithTsJest} */
+module.exports = {
+  ...getJestProjectConfig(rootDir, false),
+  displayName: require('../package.json').name,
+  rootDir,
+  fakeTimers: {
+    enableGlobally: true,
+    // TODO re-enable fake dates when issue fixed https://github.com/sinonjs/fake-timers/issues/437
```
matthieu-crouzet, line 12: Issue seems to be fixed

packages/@ama-sdk/core/src/clients/api-fetch-client.ts
```diff
@@ -16,13 +16,19 @@ import {BaseApiClientOptions} from '../fwk/core/base-api-constructor';
 import {CanceledCallError, EmptyResponseError, ResponseJSONParseError} from '../fwk/errors';
 import {ReviverType} from '../fwk/Reviver';
 
-/** @see BaseApiClientOptions */
+/**
+ * @see BaseApiClientOptions
+ * @deprecated Use the one exposed by {@link @ama-sdk/client-fetch}, will be removed in v13
```
fpaul-1A, line 21: shouldn't we remove it in v14 ?

packages/@ama-sdk/core/README.md
```diff
@@ -3,49 +3,49 @@
...
 
-This package contains all the [plugins](https://github.com/AmadeusITGroup/otter/tree/main/packages/%40ama-sdk/core/src/plugins), helpers and object definitions to dialog with an API following the `ama-sdk` architecture.
+This package contains a set of [plugins](https://github.com/AmadeusITGroup/otter/tree/main/packages/%40ama-sdk/core/src/plugins), helpers and object definitions to dialog with an API following the `ama-sdk` architecture.
 
 Please refer to the [ama-sdk-schematics](../schematics/README.md) package for getting started with an API based on `ama-sdk`.
 
 ## Available plugins
 
-- [abort](https://github.com/AmadeusITGroup/otter/tree/main/packages/%40ama-sdk/core/src/plugins/abort)
+- [abort](https://github.com/AmadeusITGroup/otter/tree/main/packages/%40ama-sdk/core/src/plugins/abort) *(deprecated)*
```
matthieu-crouzet, line 12: maybe we could give the link to the new plugin 
same for the others plugins

packages/@ama-sdk/core/src/clients/api-fetch-client.ts
matthieu-crouzet, line 21: Do we provide a ng-update or ng-add to handle that migration ?

kpanot: V12 or v13 but definitively not v14.
Not sure we want to keep dead code for 1 year and half

packages/@ama-sdk/client-fetch/schematics/ng-add/index.ts
kpanot: No but the run of the schematics does.
This is the generic message we display when `@o3r/schematics` is missing

packages/@ama-sdk/client-fetch/src/fetch-plugin.ts
kpanot: I copy pasted it from the core actually so I did not check the code.
I wonder if the diff is that this notation enforce the usage of class property.

packages/@ama-sdk/client-fetch/testing/jest.config.ut.js
kpanot: In version 12 but the current version used by jest is 10.
I will change the comment

packages/@ama-sdk/core/src/clients/api-fetch-client.ts
kpanot: I added a rule in the `ng-add` of `@ama-sdk/client-fetch` to do the migration.

fpaul-1A: we should probably discuss and update the documentation about breaking changes then https://github.com/AmadeusITGroup/otter/blob/main/CONTRIBUTING.md#rules-for-contributions

kpanot: Indeed the documentation is confusing:

> Note that only even major Otter versions allow breaking changes. This means we only allow breaking changes in n+2 major versions.

Here the second sentence guideline is respected but, even if it has be flagged as consequence of the first one, not the rule on "even major version".

Anyway, as discussed, we will need to adapt this part to make it more flexible

fpaul-1A (dismissed):

fpaul-1A (dismissed):

matthieu-crouzet (approved):

fpaul-1A (approved):

PR #2125, opened 2024-09-03
fix(#2124): compute the scope first while generating a new sdk

### LINKED ISSUE #2124: [Bug]: wrong scope computation when creating a sdk without scope in the name
### Package name

workspace

### Package version

11.1

### Reproduction steps

```shell
npm create @o3r repro-monorepo
cd repro-monorepo
npm run ng g sdk --  --name test
```

### Current result

the command `spec:regen` in the package.json of the sdk is 
```shell
npm run generate -- --generator-key -test && amasdk-clear-index

### Expected result

The command should be the following
```shell
npm run generate -- --generator-key test-sdk && amasdk-clear-index
```

### Additional comments

Currently in `@o3r/workspace:sdk` we consider `scope` as optional and `package` as mandatory for `@ama-sdk/schematics`, but it should be the opposite the default value of `package` is `sdk`

matthieu-crouzet:
## Proposed change
compute the scope first while generating a new sdk

## Related issues

- :bug: Fixes #2124

### DIFF
--- a/packages/@o3r/workspace/schematics/index.it.spec.ts
+++ b/packages/@o3r/workspace/schematics/index.it.spec.ts
@@ -20,7 +20,7 @@ describe('new otter workspace', () => {
   test('should add sdk to an existing workspace', () => {
     const { workspacePath, isInWorkspace, untouchedProjectsPaths } = o3rEnvironment.testEnvironment;
     const execAppOptions = {...getDefaultExecSyncOptions(), cwd: workspacePath};
-    packageManagerExec({script: 'ng', args: ['g', 'sdk', '@my-sdk/sdk']}, execAppOptions);
+    packageManagerExec({script: 'ng', args: ['g', 'sdk', 'my-sdk']}, execAppOptions);
 
     const diff = getGitDiff(execAppOptions.cwd);
     untouchedProjectsPaths.forEach(untouchedProject => {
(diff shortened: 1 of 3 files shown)


fpaul-1A (approved):

packages/@o3r/workspace/schematics/index.it.spec.ts
```diff
@@ -20,7 +20,7 @@ describe('new otter workspace', () => {
   test('should add sdk to an existing workspace', () => {
     const { workspacePath, isInWorkspace, untouchedProjectsPaths } = o3rEnvironment.testEnvironment;
     const execAppOptions = {...getDefaultExecSyncOptions(), cwd: workspacePath};
-    packageManagerExec({script: 'ng', args: ['g', 'sdk', '@my-sdk/sdk']}, execAppOptions);
+    packageManagerExec({script: 'ng', args: ['g', 'sdk', 'my-sdk']}, execAppOptions);
```
mrednic-1A, line 23: this change is not mandatory, right ?

matthieu-crouzet: This change is not mandatory indeed but here to avoid create a new integration test to check when the user provide a sdk name not following the asked format.

PR #2130, opened 2024-09-04
test: speed-up it tests

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/test-helpers/src/utilities/package-manager.ts
+++ b/packages/@o3r/test-helpers/src/utilities/package-manager.ts
@@ -19,6 +19,7 @@ type Command =
   | 'exec'
   | 'info'
   | 'install'
+  | 'ci'
   | 'publish'
   | 'run'
   | 'version'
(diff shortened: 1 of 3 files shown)


packages/@o3r/test-helpers/src/utilities/package-manager.ts
```diff
@@ -40,6 +42,7 @@ const PACKAGE_MANAGERS_CMD: {[packageManager in SupportedPackageManagers]: {[com
   },
   yarn: {
     add: ['yarn', 'add'],
+    ci: ['yarn', 'install', '--frozen-lockfile'],
```
kpanot, line 45: Stupid question: `--frozen-lockfile` is only for yarn 1, is it the one we are using here?

kpanot (dismissed):

packages/@o3r/test-helpers/src/utilities/package-manager.ts
fpaul-1A: should be `--immutable` I guess, my bad

kpanot (approved):

mrednic-1A (approved):

mrednic-1A:
Do we know how much do we gain, with approximation ?

fpaul-1A:
> Do we know how much do we gain, with approximation ?

not very much I'm afraid, probably less than 5%

kpanot (approved):

PR #2135, opened 2024-09-05
chore(github-actions): audit bundle reduction and target cascading deprecation

kpanot:


tools/github-actions/audit/package.json
```diff
@@ -30,7 +37,7 @@
     "@typescript-eslint/parser": "^7.14.1",
     "@typescript-eslint/utils": "^7.14.1",
     "@vercel/ncc": "~0.38.0",
-    "audit-types": "~0.6.0",
+    "audit-types": "0.6.2",
```
fpaul-1A, line 40: why do we need to pin the version ?

fpaul-1A (dismissed):

tools/github-actions/audit/package.json
kpanot: I don't, mistake was made by the IDE

fpaul-1A (approved):

sdo-1A (approved):

PR #2136, opened 2024-09-05
fix: wrong path calculation for sdk spec relative path in monorepo

### LINKED ISSUE #2123: [Bug]: Wrong path calculation while creating a sdk in a monorepo with a local spec file
### Package name

workspace

### Package version

11.1

### Reproduction steps

Create a file `repro-spec.yml` with any openapi spec inside
in the same folder rune the following commands
```shell
npm create @o3r repro-monorepo
cd repro-monorepo
npm run ng g sdk --  --name="@repro/sdk" --spec-path="../repro-spec.yml"
```

### Current result

ENOENT: no such file or directory, open 'libs/repro-spec.yml'

### Expected result

The generation should work

### Additional comments

It seems that the `specPath` is computed from the sdk folder instead of from where the command is run.

fpaul-1A:
## Related issues

- :bug: Fixes #2123 
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/workspace/schematics/index.it.spec.ts
+++ b/packages/@o3r/workspace/schematics/index.it.spec.ts
@@ -16,11 +16,15 @@ import { existsSync, promises as fs } from 'node:fs';
 import * as path from 'node:path';
 import type { PackageJson } from 'type-fest';
 
+const mocksFolder = path.join(__dirname, '..', 'testing', 'mocks');
+
 describe('new otter workspace', () => {
   test('should add sdk to an existing workspace', () => {
     const { workspacePath, isInWorkspace, untouchedProjectsPaths } = o3rEnvironment.testEnvironment;
     const execAppOptions = {...getDefaultExecSyncOptions(), cwd: workspacePath};
-    packageManagerExec({script: 'ng', args: ['g', 'sdk', 'my-sdk']}, execAppOptions);
+    expect(() =>
+      packageManagerExec({script: 'ng', args: ['g', 'sdk', 'my-sdk']}, execAppOptions)
+    ).not.toThrow();
 
     const diff = getGitDiff(execAppOptions.cwd);
     untouchedProjectsPaths.forEach(untouchedProject => {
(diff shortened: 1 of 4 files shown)


mrednic-1A (approved):

packages/@o3r/workspace/schematics/sdk/index.ts
```diff
@@ -48,6 +48,10 @@ function generateSdkFn(options: NgGenerateSdkSchema): Rule {
 
     const packageManager = getPackageManager({ workspaceConfig });
     const specExtension = options.specPackagePath ? path.extname(options.specPackagePath) : '.yaml';
+    // If spec path is relative to process.cwd, we need to make it relative to the project root
+    if (options.specPath?.startsWith('.')) {
```
kpanot, line 52: not covering all the cases, why not using `options.specPath && !path.isAbsolute(options.specPath)` ?

kpanot (approved):

packages/@o3r/workspace/schematics/index.it.spec.ts
```diff
@@ -31,8 +33,31 @@ describe('new otter workspace', () => {
...
+    const { workspacePath, isInWorkspace, untouchedProjectsPaths } = o3rEnvironment.testEnvironment;
+    const sdkPath = path.posix.join('libs', 'my-sdk-sdk');
+    const execAppOptions = {...getDefaultExecSyncOptions(), cwd: workspacePath};
+    await fs.copyFile(path.join(mocksFolder, 'easy-spec.yaml'), path.join(workspacePath, 'local-spec.yaml'));
+    packageManagerExec({script: 'ng', args: [
+      'g',
+      'sdk',
+      '@my-sdk/sdk',
+      '--spec-path', './local-spec.yaml'
+    ]}, execAppOptions);
```
matthieu-crouzet, line 46:
```suggestion
    expect(() => packageManagerExec({script: 'ng', args: [
      'g',
      'sdk',
      '@my-sdk/sdk',
      '--spec-path', './local-spec.yaml'
    ]}, execAppOptions)).not.toThrow();
```

packages/@o3r/workspace/schematics/index.it.spec.ts
```diff
@@ -48,6 +73,7 @@ describe('new otter workspace', () => {
     untouchedProjectsPaths.forEach(untouchedProject => {
       expect(diff.all.some(file => file.startsWith(path.posix.relative(workspacePath, untouchedProject)))).toBe(false);
     });
+    expect(diff.added).toContain(path.posix.join(sdkPath, 'open-api.yaml'));
```
matthieu-crouzet, line 76: We could also check that we have the model files generated no ?

packages/@o3r/workspace/schematics/sdk/index.ts
fpaul-1A: I didn't use `path.isAbsolute` because `options.specPath` can also be an URL, but indeed `startsWith('.')` is not enough

kpanot (approved):

matthieu-crouzet (approved):

PR #2160, opened 2024-09-12
fix(stylelint-plugin): remove noEmit from test tsconfig

kpanot:


fpaul-1A (approved):

mrednic-1A:
If tsc of tests is generating files, it could not conflict with the files generated by the actual build ?

kpanot:
> If tsc of tests is generating files, it could not conflict with the files generated by the actual build ?

Don't worry it does not work that way.
The Tsconfig.spec.json is used by Jest that does not generate any JS file :)

sdo-1A (approved):

mrednic-1A (approved):

mrednic-1A:
> > If tsc of tests is generating files, it could not conflict with the files generated by the actual build ?
> 
> Don't worry it does not work that way. The Tsconfig.spec.json is used by Jest that does not generate any JS file :)

Ok ok. thanks for the explanation :)

PR #2168, opened 2024-09-13
test: prevent race condition failure on it-tests

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/test-helpers/src/utilities/locker.ts
+++ b/packages/@o3r/test-helpers/src/utilities/locker.ts
@@ -100,8 +100,12 @@ export class Locker {
           resolve();
         }
       };
-      watcher = watch(this.options.lockFilePath, {persistent: false}, check);
-      watcher.on('error', check);
+      try {
+        watcher = watch(this.options.lockFilePath, {persistent: false}, check);
+        watcher.on('error', check);
+      } catch {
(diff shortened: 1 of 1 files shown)


kpanot (approved):

mrednic-1A (approved):

packages/@o3r/test-helpers/src/utilities/locker.ts
```diff
@@ -100,8 +100,12 @@ export class Locker {
           resolve();
         }
       };
-      watcher = watch(this.options.lockFilePath, {persistent: false}, check);
-      watcher.on('error', check);
+      try {
+        watcher = watch(this.options.lockFilePath, {persistent: false}, check);
+        watcher.on('error', check);
```
mrednic-1A, line 105: does it make sense to keep this line ? it is suppose to do what we do in the catch, or I am wrong :D

fpaul-1A: the reason I added a try/catch is for the rare case when the file has already been deleted in parallel when we start the watch (file not found exception)
the `on('error')` catch errors after the watch has been started
I would say we still need both

PR #2169, opened 2024-09-16
fix(telemetry): missing telemetry wrapper

matthieu-crouzet:
## Proposed change
Add the missing telemetry wrapper

### DIFF
--- a/packages/@o3r/design/builders/generate-css/index.ts
+++ b/packages/@o3r/design/builders/generate-css/index.ts
@@ -1,5 +1,6 @@
 import type { GenerateCssSchematicsSchema } from './schema';
 import { BuilderOutput, createBuilder } from '@angular-devkit/architect';
+import type { BuilderWrapper } from '@o3r/telemetry';
 import {
   getCssTokenDefinitionRenderer,
   getCssTokenValueRenderer,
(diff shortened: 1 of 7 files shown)


fpaul-1A (dismissed):

packages/@o3r/design/builders/generate-css/index.ts
```diff
@@ -1,5 +1,6 @@
 import type { GenerateCssSchematicsSchema } from './schema';
 import { BuilderOutput, createBuilder } from '@angular-devkit/architect';
+import { createBuilderWithMetricsIfInstalled } from '@o3r/extractors';
```
kpanot, line 3: Should be dynImport to remain optional
(same comment in the other files)

packages/@o3r/design/schematics/generate-css/index.ts
```diff
@@ -1,14 +1,14 @@
 import type { GenerateCssSchematicsSchema } from './schema';
 import type { Rule } from '@angular-devkit/schematics';
-import { globInTree } from '@o3r/schematics';
+import type { createSchematicWithMetricsIfInstalled, globInTree } from '@o3r/schematics';
```
fpaul-1A, line 3: `globInTree` seems to be more than just a type

fpaul-1A (approved):

kpanot (approved):

PR #2170, opened 2024-09-16
ci: check releases

fpaul-1A:
## Proposed change

Weekly job that tests that all currently supported versions are still passing the audits and it-tests
Check on IT-tests are currently disabled by default for the weekly job because of instability and discrepancy between yarn3 and yarn4 

Example of run: https://github.com/fpaul-1A/otter/actions/runs/12199946300

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/.github/workflows/check-release.yml
+++ b/.github/workflows/check-release.yml
@@ -1,0 +1,67 @@
+name: Check release
+
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        type: string
+        required: true
+        description: The branch, tag or SHA to checkout.
+      performAudit:
+        type: boolean
+        default: true
+        description: Run Audit
+      performITTests:
+        type: boolean
+        default: false
+        description: Run IT Tests
+  workflow_call:
+    inputs:
+      ref:
+        type: string
+        default: ''
+        description: The branch, tag or SHA to checkout.
+      performAudit:
+        type: boolean
+        default: true
+        description: Run Audit
+      performITTests:
+        type: boolean
+        default: false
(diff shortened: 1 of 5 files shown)


.github/workflows/check-release.yml
```diff
@@ -0,0 +1,46 @@
...
+        type: string
+        default: ''
+        description: The branch, tag or SHA to checkout.
+    secrets:
+      NX_CLOUD_ACCESS_TOKEN:
+        required: false
+        description: Token to use Nx Cloud token
+
+env:
+  YARN_ENABLE_HARDENED_MODE: 0
```
cpaulve-1A, line 22: why is it necessary?

fpaul-1A: it's not

.github/workflows/check-release.yml
```diff
@@ -0,0 +1,46 @@
...
+          install-jdk: 'true'
+      - run: yarn build:swagger-gen
+      - run: yarn build
+      - uses: ./tools/github-actions/upload-build-output
+        with:
+          artifactName: 'dist-${{ inputs.ref }}'
+
+  testRelease:
+    needs: [buildRelease]
+    uses: ./.github/workflows/it-tests.yml
```
matthieu-crouzet, line 43: as we have some instability on it-tests
should we add 1 or 2 `retry` on this step before it opens an issue ?

matthieu-crouzet: not the case anymore

kpanot (approved):

PR #2183, opened 2024-09-20
feat(sdk): exported Angular and Beacon client to dedicated package

### LINKED ISSUE #2128: [Feature]: Split SDK API Client to dedicated packages
### Context

In the current implementation, the package `@ama-sdk/core` contains the following items:
- Api helpers function (generating the final request parameters)
- Fetch Api Client
- Angular Api Client
- Beacon Api Client
- Reply and Request Plugins
- Fetch Plugins
- (Angular Plugins)
- Mock mechanism

To be able to expose all the items without forcing non mandatory dependencies, several entrypoints are exported from `@ama-sdk/core`. (example: Angular dependencies, several plugins, etc...).
The future integration of Tanstack Query will lead potentially to several plugins and Api Client requiring additional entrypoints.

### Proposal

Move the Client API to dedicated `@ama-sdk/client-*` packages.
This proposal comes with the following acceptance criteria:

- [ ] Move the plugins dedicated to the ApiClient to the new package
- [ ] Deprecate the usage of ApiClient and its plugins from `@ama-sdk/core`
- [ ] Provide schematics to support the migration

The change should affect the following ApiClients:
- [x] #2129
- [ ] BeaconApiClient
- [ ] AngularApiClient

kpanot:
## Proposed change

Split of Angular and Beacon API Client to their own dedicated package

## Related issues

- :rocket: Feature resolves #2128


packages/@ama-sdk/client-beacon/testing/jest.config.ut.js
```diff
@@ -0,0 +1,21 @@
...
+
+/** @type {import('ts-jest/dist/types').JestConfigWithTsJest} */
+module.exports = {
+  ...getJestProjectConfig(rootDir, false),
+  displayName: require('../package.json').name,
+  rootDir,
+  fakeTimers: {
+    enableGlobally: true,
+    // TODO re-enable fake dates when issue fixed https://github.com/sinonjs/fake-timers/issues/437 and v12 is used by Jest
+    doNotFake: ['Date']
```
fpaul-1A, line 13: not needed anymore

fpaul-1A: same for client-angular

fpaul-1A (approved):

kpanot:
> Could you document this change in the next major version migration guide?

Why?
This PR does not include breaking change and will be integrated in a minor version.
The migration guide should be updated on #2117 and on the v13 when removing the deprecated interfaces

vscaiceanu-1a:
> > Could you document this change in the next major version migration guide?
> 
> Why? This PR does not include breaking change and will be integrated in a minor version. The migration guide should be updated on #2117 and on the v13 (n+2) when removing the deprecated interfaces

In the v12 migration guide I would expect some words about this change: the list of deprecations mentioning that they will be removed in v13 and the fact that`ng add` of new packages will update the imports.

kpanot:
> > > Could you document this change in the next major version migration guide?
> > 
> > 
> > Why? This PR does not include breaking change and will be integrated in a minor version. The migration guide should be updated on #2117 and on the v13 (n+2) when removing the deprecated interfaces
> 
> In the v12 migration guide I would expect some words about this change: the list of deprecations mentioning that they will be removed in v13 and the fact that`ng add` of new packages will update the imports.
> 
> Users don't have to wait v13 until the interfaces will be removed, they can migrate to the new ones in v12. This is why I find this info useful in the v12 migration guide.

I disagree on this proposal for the following reasons:

- Not sure to understand why the deprecation should be listed in migration v12 when it is part of the v11.1
- I don't think that creating a "migration v12" file in the beginning of v11 is a good idea
- We discussed previously and already agreed to specify the deprecation in the minor version changelog

It you want to propose this, you can put it in the agenda of the next discussion session.
I don't think it should be a blocker of the merge of this PR, as it is not the first PR with deprecation and, even if the decision is taken to go with this proposal, the migration guide should be created with previous deprecations.

vscaiceanu-1a (approved):

PR #2200, opened 2024-09-26
feat(eslint): workflow to annotate the pull request when introducing new eslint issues

matthieu-crouzet:
## Proposed change

New workflow to annotate the pull request when introducing new ESLint issues

### DIFF
--- a/packages/@o3r/eslint-plugin/package.json
+++ b/packages/@o3r/eslint-plugin/package.json
@@ -31,6 +31,9 @@
     "type-fest": "^4.10.2"
   },
   "peerDependenciesMeta": {
+    "json-diff": {
+      "optional": true
+    },
     "jsonc-eslint-parser": {
       "optional": true
     },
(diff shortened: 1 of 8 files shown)


packages/@o3r/eslint-plugin/package.json
```diff
@@ -31,6 +31,9 @@
     "type-fest": "^4.10.2"
   },
   "peerDependenciesMeta": {
+    "json-diff": {
```
kpanot, line 34: not really optional in the current implementation: you should turn its import to dynImport or not export diff-import in public_apii

packages/@o3r/eslint-plugin/src/tools/diff-reports.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+export const mergeESLintReports = (pattern: Parameters<typeof sync>[0]) => sync(pattern).flatMap((filePath) => JSON.parse(readFileSync(filePath).toString()));
+
+/**
+ * Retrieve the list of introduced ESLint issues between two reports
+ * @param reportBase
+ * @param reportBranch
+ */
+export const getIntroducedESLintIssues = (reportBase: ESLint.LintResult[], reportBranch: ESLint.LintResult[]) => {
+  const reportDiff: [JsonDiffType, ESLint.LintResult][] = jsonDiff.diff(reportBase, reportBranch, { full: true });
```
cpaulve-1A, line 30: Are the report sorted?

Issue #689, opened 2023-08-28
Generate workspace with CI/CD capacity


kpanot:
### Context

The generator of Otter project provide a way to have an empty workspace in which the different Application/Library/SDK can be added.
Currently the Workspace is generated without Pipeline configuration.

### Proposal

The workspace should be generated with basic CI/CD.
This CICD generation schematics can be part of the workspace generator package/schematics or in an additional module.

This pipeline should contain the following items:
- Cascading pipeline
- Documentation regarding the branching strategy
- UT and Linter check pipeline
- Code Audit (including codeQL for Github)
- Label auto apply (with label configuration file)
- Package publication process

The following action should be added to the existing CI/CD by there dedicated package
- IT environment setup (Verdaccio) and process run

The schematics should support several environments (Github, Azure DevOps, Bitbucket, etc...) but in only the implementation of Github Action is required for this task.

sdo-1A:
Maybe add a parameter for the runner
Update the .npmrc (with environment variables for the publish that should be created by the user)

PR #2207, opened 2024-09-30
feat: o3r pipeline

vscaiceanu-1a:
## Related issues

- :rocket: Feature #689


packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/actions/setup/action.yml
```diff
@@ -0,0 +1,17 @@
+name: Setup and Install
+description: Sets up Node.js and installs dependencies
+
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/setup-node@0a44ba7841725637a19e28fa30b79a866c81b0a6 # v4.0.4
+      with:
+        node-version: 20
```
fpaul-1A, line 9: Should we setup the cache directly in `setup-node` action ?
```suggestion
    - uses: actions/setup-node@0a44ba7841725637a19e28fa30b79a866c81b0a6 # v4.0.4
      with:
        node-version: 20
        cache: <%= packageManager %>
```

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/actions/setup/action.yml
```diff
@@ -0,0 +1,17 @@
+name: Setup and Install
+description: Sets up Node.js and installs dependencies
+
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/setup-node@0a44ba7841725637a19e28fa30b79a866c81b0a6 # v4.0.4
```
matthieu-crouzet, line 7: Will it be update when we will use another version of this action ?

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
```diff
@@ -0,0 +1,54 @@
...
+
+  code-checks:
+    runs-on: <%= runner %>
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - uses: ./.github/actions/setup
+      - name: Unit tests
+        run: <%= packageManager %> run test
+      - name: Audit
+        run: <% if (packageManager === 'yarn') {%>yarn <%}%>npm audit
```
matthieu-crouzet, line 30: Why not using our own action `AmadeusITGroup/otter/tools/github-actions/audit` ?

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
```diff
@@ -0,0 +1,54 @@
...
+    runs-on: <%= runner %>
+    outputs:
+      nextVersionTag: ${{ steps.new-version.outputs.nextVersionTag }}
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - uses: ./.github/actions/setup
+      - name: New version
+        if: github.event_name != 'merge_group'
+        id: new-version
+        uses: AmadeusITGroup/otter/tools/github-actions/new-version@main
```
matthieu-crouzet, line 45: Should we recommend to use the `main` or a released version ?

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
```diff
@@ -0,0 +1,54 @@
...
+        run: <%= packageManager %> run lint:ci
+  version:
+    permissions:
+      contents: write
+    runs-on: <%= runner %>
+    outputs:
+      nextVersionTag: ${{ steps.new-version.outputs.nextVersionTag }}
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - uses: ./.github/actions/setup
```
kpanot, line 41: Not sure the setup is needed, only node is required and the checkout are required

packages/@o3r/pipeline/src/index.spec.ts
```diff
@@ -0,0 +1,5 @@
+// TODO be removed as soon as we have one test in this package
```
matthieu-crouzet, line 1: You could add a test for the `ng-add` schematics instead

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: New version
+        if: github.event_name != 'merge_group'
+        id: new-version
+        uses: AmadeusITGroup/otter/tools/github-actions/new-version@main
+        with:
+          defaultBranch: main
+          defaultBranchPrereleaseName: prerelease
+          releaseBranchRegExp: 'release\/(0|[1-9]\d*)\.(0|[1-9]\d*)(\.0-(?:next|prerelease|rc))?$'
```
kpanot, line 49: Could be defined as `const` at pipeline level

packages/@o3r/pipeline/tsconfig.build.json
```diff
@@ -0,0 +1,16 @@
...
+    "incremental": true,
+    "composite": true,
+    "module": "CommonJS",
+    "outDir": "./dist",
+    "rootDir": ".",
+    "tsBuildInfoFile": "build/.tsbuildinfo"
+  },
+  "include": [
+    "src/**/*.ts",
+    "cli/**/*.cts"
```
matthieu-crouzet, line 13: you don't have `cli`

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/actions/setup/action.yml
```diff
@@ -0,0 +1,17 @@
...
+      with:
+        node-version: 20
+<% if (packageManager === 'yarn') { %>
```
kpanot, line 10: Not required, the corepack enabling can/should be done in all cases

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
kpanot: We may want to configure Renovate to upgrade to upgrade the commit number with the latest from `release/*` branches

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
vscaiceanu-1a: Our action only supports yarn.

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
vscaiceanu-1a: Ok. I think we should also fix this in our own pipeline.

packages/@o3r/pipeline/package.json
```diff
@@ -0,0 +1,143 @@
+{
+  "name": "@o3r/pipeline",
+  "version": "0.0.0-placeholder",
+  "publishConfig": {
+    "access": "public"
+  },
+  "typings": "./dist/src/public_api.d.ts",
```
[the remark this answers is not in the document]
vscaiceanu-1a: It is not added at compile time because it's a `tsc`, not an `ng-packager` build

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
vscaiceanu-1a: do you mean as `env`?

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
matthieu-crouzet: In case of yarn we could use our action then and open an issue to support npm

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
kpanot: yes :)

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
vscaiceanu-1a: I agree that the audit action should support `npm` along with `yarn`. 
Now that I think about it, the audit step should not be here as it is supposed to be part of the security workflow.

packages/@o3r/pipeline/.gitignore
```diff
@@ -0,0 +1,2 @@
+/index.*
```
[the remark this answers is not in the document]
vscaiceanu-1a: nope

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
vscaiceanu-1a: oki :) done

kpanot: (The comment was for the 3 lines :S. I think they are constants that we want to highlight as they drive the branching model)

.renovaterc.json5
```diff
@@ -127,5 +127,11 @@
         "\"openApiSupportedVersion\": \"~(?<currentValue>[^\"]+)\""
       ]
     }
-  ]
+  ],
+  "github-actions": {
+    "fileMatch": [
+      "(^|/)(__dot__github|workflow-templates|\\.(?:github|gitea|forgejo)/(?:workflows|actions))/.+\\.ya?ml$",
```
kpanot, line 133: `__dot__` ? Would not be `[.]` or `\\.` instead ?

vscaiceanu-1a: here I am trying to override the default `fileMatch` [value](https://docs.renovatebot.com/modules/manager/github-actions/) so that it goes over my templates that are located in a `__dot__github` folder

packages/@o3r/pipeline/package.json
```diff
@@ -0,0 +1,120 @@
...
+  "exports": {
+    "./package.json": {
+      "default": "./package.json"
+    },
+    ".": {
+      "es2020": "./dist/src/public_api.js",
+      "default": "./dist/src/public_api.js",
+      "typings": "./dist/src/public_api.d.ts",
+      "node": "./dist/src/public_api.js",
+      "require": "./dist/src/public_api.js"
```
kpanot, line 36:
```suggestion
      "default": "./dist/src/public_api.js",
      "typings": "./dist/src/public_api.d.ts"
```

packages/@o3r/pipeline/package.json
```diff
@@ -0,0 +1,120 @@
...
+  },
+  "peerDependenciesMeta": {
+    "@o3r/core": {
+      "optional": true
+    },
+    "@o3r/schematics": {
+      "optional": true
+    }
+  },
+  "devDependencies": {
```
kpanot, line 57: Maybe there is a clean to do in the list of dev deps :S

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
```diff
@@ -0,0 +1,67 @@
...
+
+  version:
+    permissions:
+      contents: write
+    runs-on: <%= runner %>
+    outputs:
+      nextVersionTag: ${{ steps.new-version.outputs.nextVersionTag }}
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - uses: actions/setup-node@0a44ba7841725637a19e28fa30b79a866c81b0a6 # v4.0.4
```
kpanot, line 51: not sure this step is actually needed (sorry I said the opposite in a previous comment but I think it was a mistake)

vscaiceanu-1a: not needed indeed.

packages/@o3r/pipeline/testing/jest.config.ut.builders.js
```diff
@@ -0,0 +1,16 @@
...
+const rootDir = path.join(__dirname, '..');
+
+/** @type {import('ts-jest/dist/types').JestConfigWithTsJest} */
+module.exports = {
+  ...getJestProjectConfig(rootDir, false),
+  displayName: require('../package.json').name,
+  rootDir,
+  testPathIgnorePatterns: [
+    '<rootDir>/.*/templates/.*',
+    '<rootDir>/src/.*',
```
kpanot, line 12: why do you ignore this folder? (env if there is currently no test in it)

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/actions/setup/action.yml
```diff
@@ -0,0 +1,16 @@
...
+
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/setup-node@0a44ba7841725637a19e28fa30b79a866c81b0a6 # v4.0.4
+      with:
+        node-version: 20
+        cache: <%= packageManager %>
+    - name: Enable Corepack
+      shell: bash
```
kpanot, line 12: Both `shell: bash` should be removed.
They are enforcing `bash` to be installed for a command that does not require a specific shell.

packages/@o3r/pipeline/testing/jest.config.ut.builders.js
vscaiceanu-1a: this is the `ut.builders.js` file. No Builder tests should be under `/src` right?

packages/@ama-sdk/create/package.json
```diff
@@ -24,6 +24,7 @@
     "@angular-devkit/schematics-cli": "^18.0.5",
     "@angular/cli": "~18.2.0",
     "@o3r/schematics": "workspace:*",
+    "@o3r/telemetry": "workspace:^",
```
kpanot, line 27: Fix on #2253

packages/@o3r/create/package.json
```diff
@@ -19,6 +19,7 @@
   },
   "dependencies": {
     "@angular/cli": "~18.2.0",
+    "@o3r/telemetry": "workspace:^",
```
kpanot, line 22: Fix on https://github.com/AmadeusITGroup/otter/pull/2253

packages/@o3r/pipeline/README.md
```diff
@@ -0,0 +1,44 @@
...
+- `toolchain`: The DevOps toolchain to create. For now, only `GitHub` is supported.
+- `runner`: The CI runner. Default is `ubuntu-latest`.
+- `npmRegistry`: A private npm registry. By default, the public one will be used.
+
+### Usage
+
+Here is an example of how to use the `ng add` schematics with parameters:
+
+```shell
+ng add @o3r/pipeline --runner=windows-latest --npmRegistry=https://custom-registry.example.com
```
kpanot, line 38: I think you should add the code snippet of the "standard" usage (without the options) for copy/past purpose before writing this example

packages/@o3r/pipeline/package.json
```diff
@@ -0,0 +1,97 @@
+{
+  "name": "@o3r/pipeline",
+  "version": "0.0.0-placeholder",
+  "publishConfig": {
+    "access": "public"
+  },
+  "typings": "./dist/src/public_api.d.ts",
```
kpanot, line 7: forgot the `main` field

packages/@o3r/pipeline/schematics/ng-add/index.ts
```diff
@@ -0,0 +1,91 @@
...
+    const { createSchematicWithMetricsIfInstalled } = await import('@o3r/schematics').catch(() => {throw new Error(missingSchematicDependencyMessage); });
+    return createSchematicWithMetricsIfInstalled(ngAddFn)(options);
+  } catch (err) {
+    if (err instanceof Error && err.message === missingSchematicDependencyMessage) {
+      logger.warn(`[WARNING]: You are trying to add '@o3r/pipeline' on a project that does not have '@o3r/schematics' as a dependency.
+        Please run 'ng add @o3r/schematics' before for a more robust setup. Trying to guess the package manager.`);
+      return ngAddFn(options, true);
+    } else {
+      throw err;
+    }
```
kpanot, line 89: To make sure to not catch a legit error:

```suggestion
  return import('@o3r/schematics')
    .catch(() => {
      logger.warn(`You are trying to add '@o3r/pipeline' on a project that does not have '@o3r/schematics' as a dependency.
        Please run 'ng add @o3r/schematics' before for a more robust setup. Trying to guess the package manager.`);
    })
    .then((schematics) =>
      schematics ? schematics.createSchematicWithMetricsIfInstalled(ngAddFn)(options) : ngAddFn(options)
    );
```

packages/@o3r/pipeline/schematics/ng-add/index.ts
```diff
@@ -0,0 +1,91 @@
...
+      if (tree.exists('/yarn.lock')) {
+        packageManager = 'yarn';
+      }
+    } else {
+      const {
+        getPackageManager,
+        getWorkspaceConfig
+      } = await import('@o3r/schematics');
+      packageManager = getPackageManager({ workspaceConfig: getWorkspaceConfig(tree) });
+    }
```
kpanot, line 25:
```suggestion
    try {
      packageManager = await import('@o3r/schematics').getPackageManager({ workspaceConfig: getWorkspaceConfig(tree) });
    } catch {
      packageManager = tree.exists('/yarn.lock') ? 'yarn' : 'npm':
    }
```

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
```diff
@@ -0,0 +1,63 @@
...
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - name: New version
+        if: github.event_name != 'merge_group'
+        id: new-version
+        uses: AmadeusITGroup/otter/tools/github-actions/new-version@11
+        with:
+          defaultBranch: ${{ env.DEFAULT_BRANCH }}
+          defaultBranchPrereleaseName: ${{ env.DEFAULT_BRANCH_PRERELEASE_NAME }}
+          releaseBranchRegExp: ${{ env.RELEASE_BRANCH_REGEX }}
+      - name: Create release
```
kpanot, line 59: Maybe we should think about moving it out from the main (in action exposed by Otter?) to be easly replaced by another releasing tool

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
```diff
@@ -0,0 +1,63 @@
...
+      contents: write
+    runs-on: <%= runner %>
+    outputs:
+      nextVersionTag: ${{ steps.new-version.outputs.nextVersionTag }}
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - name: New version
+        if: github.event_name != 'merge_group'
+        id: new-version
+        uses: AmadeusITGroup/otter/tools/github-actions/new-version@11
```
kpanot, line 54: Should be git hash as other actions

packages/@o3r/pipeline/schematics/ng-add/index.ts
vscaiceanu-1a: all errors are caught and the ones that are not `missingSchematicDependencyMessage` are thrown again. I find it strange to  `import('@o3r/schematics')` a second time even if we know for sure that it isn't there.

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
vscaiceanu-1a: I find it a bit overkill to create an action just for a single line of shell, don't you think?

If you consider it is worth it, we can create an issue to expose it via an Otter GH Action that we will use here, in the template and in our libraries.

packages/@o3r/pipeline/README.md
vscaiceanu-1a: It is the case on line 20: `ng add @o3r/pipeline`

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
vscaiceanu-1a: https://github.com/AmadeusITGroup/otter/issues/2258 created

packages/@o3r/pipeline/schematics/ng-add/index.js
```diff
@@ -0,0 +1,13 @@
+/*
+
+This files is used to allow the usage of the builder within @o3r/framework mono-repository.
```
kpanot, line 3: from @sdo-1A :
```suggestion
This file is used to allow the use of the builder in the @o3r/framework mono-repository.
```

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/actions/setup/action.yml
```diff
@@ -0,0 +1,21 @@
...
+      run: corepack enable
+    - name: Install
+<% if (npmRegistry) { %>
```
kpanot, line 15: Maybe you should add it commented, or add documentation regarding it, if no `npmRegistry` are provided.
This env are not obvious for people adding registry after generation

vscaiceanu-1a: added some documentation regarding this

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
fpaul-1A: this one is a bit tricky since it is referring to the same repository as the generator (you cannot get the hash of the commit before committing the code :\)
I think we need to either:
- get the hash somehow at runtime during the generation and use it in the template
- or keep it like this and Renovate will convert it to an hash anyway when it will first run in the app

kpanot: Keeping `11` is not really satisfying neither as it will need to be manually changed every major and it will always really to latest 11 version even if a specific version of Otter is selected.
Without going as far as a hash, a good compromise can be to set the pin version via the `set-version` script/process at pre-publish time

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
vscaiceanu-1a: https://github.com/AmadeusITGroup/otter/issues/2292 created

.renovaterc.json5
```diff
@@ -127,5 +127,11 @@
...
+      "(^|/)action\\.ya?ml$"
```
fpaul-1A, line 134: the `fileMatch` rules are additive, no need to redeclare the default ones
```suggestion
      "/__dot__github/.*/.+\\.ya?ml$"
```
also, you'll need to remove `**/templates/**` from `ignorePaths` in this file to make it process the template files

.github/workflows/publish.yml
```diff
@@ -67,6 +67,9 @@ jobs:
         with:
           fetch-depth: 2
           ref: ${{ inputs.gitRef }}
+      - name: Update @o3r/pipeline package.json with git commit hash
+        run: |
+          jq '.o3rConfig.commitHash = "${{ github.sha }}"' packages/@o3r/pipeline/dist/package.json > tmp.json && mv tmp.json packages/@o3r/pipeline/dist/package.json
```
kpanot, line 72: Would be cleaner and more maintainable to be done in the `@o3r/pipeline` and potentially triggered by the nx task `publish`.
You can have access to the `GITHUB_SHA ` envvar as well from NX

.renovaterc.json5
```diff
@@ -7,8 +7,7 @@
   ],
   "timezone": "Europe/Paris",
   "ignorePaths": [
-    "**/node_modules/**",
-    "**/templates/**"
```
kpanot, line 11: Maybe cleaner to explicitly add your exception that allowing the whole `templates/` folders?

fpaul-1A: In most cases, the dependencies that we put in the package.json inside the templates are not hard-coded but evaluated at generation time (to match the deps defined in the package.json of the generating package)
However, I believe it wouldn't hurt to scan the dependencies even in `**/templates/**` folders, although files with `.template` suffix won't be scanned

kpanot: fare enough, anyway if we got issue with Renovate, we will rollback this change.

packages/@o3r/pipeline/scripts/write-git-hash.js
kpanot, line 1: Maybe you should move this file to `@o3r/build-helpers` as it can be used to any package with the same need.
The extension should be `.cjs` or `.mjs` to be explicit regarding the format

packages/@o3r/pipeline/scripts/write-git-hash.js
```diff
@@ -0,0 +1,25 @@
+const fs = require('fs');
+const path = require('path');
+
+const commitHash = process.env.GITHUB_SHA;
+
+if (!commitHash) {
+    console.error('GITHUB_SHA environment variable is not found.');
```
kpanot, line 7: should be `2 spaces` tab

packages/@o3r/pipeline/scripts/write-git-hash.js
```diff
@@ -0,0 +1,25 @@
...
+    process.exit(1);
+}
+
+const packageJsonPath = path.join(__dirname, '..', 'dist', 'package.json');
+
+try {
+    const data = fs.readFileSync(packageJsonPath, 'utf8');
```
kpanot, line 14:
```suggestion
  const data = fs.readFileSync(packageJsonPath, { encoding: 'utf8' });
```

tools/@o3r/build-helpers/scripts/write-git-hash.mjs
```diff
@@ -0,0 +1,33 @@
...
+ * The purpose of this script is add the git commit hash in package.json
+ */
+
+import minimist from 'minimist';
+import { readFileSync, writeFileSync } from 'node:fs';
+import { join, resolve } from 'node:path';
+
+const commitHash = process.env.GITHUB_SHA;
+if (!commitHash) {
+  console.error('GITHUB_SHA environment variable is not found.');
```
kpanot, line 11:
```suggestion
  console.error('GITHUB_SHA environment variable is not found. The git hash will not be added to the package.json file.');
```

kpanot (dismissed):

packages/@o3r/pipeline/schematics/ng-add/templates/github/__dot__github/workflows/main.yml
```diff
@@ -0,0 +1,66 @@
...
+
+  lint:
+    runs-on: <%= runner %>
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - uses: ./.github/actions/setup
+      - name: Lint
+        run: <%= packageManager %> run lint
+
+  version:
```
fpaul-1A, line 43: it seems to me that the goal of this job is to create a release, so the name `version` is not version fitting

vscaiceanu-1a: `release` seems more meaningful.

fpaul-1A (dismissed):

packages/@o3r/pipeline/README.md
```diff
@@ -0,0 +1,60 @@
...
+    COREPACK_INTEGRITY_KEYS: ""
+  shell: bash
+  run: npm ci
+```
+
+If you choose to run the schematic without specifying an `npmRegistry`, you may need to manually apply these changes afterwards.
+
+### GitHub workflow
+
+The generated pipeline ensures that your code is built, tested and linted on every push or pull request to the main and release branches.
```
kpanot, line 59:
```suggestion
The generated pipeline ensures that your code is built, tested and linted on every push or pull request to the main and release branches.\
```

fpaul-1A (approved):

kpanot (approved):

fpaul-1A (approved):

packages/@o3r/pipeline/schematics/ng-add/schema.ts
```diff
@@ -0,0 +1,15 @@
...
+export interface NgAddSchematicsSchema extends SchematicOptionObject {
+
+  /** The DevOps toolchain to create */
+  toolchain: Toolchains;
+
+  /** The CI runner */
+  runner: string;
+
+  /** A custom npm registry */
+  npmRegistry?: string | undefined;
```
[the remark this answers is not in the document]
vscaiceanu-1a: back in the day we had an issue: 
```typescript
interface I {
  a?: string;
}

// valid:
const v1: I = {};

// invalid:
const v2: I = { a: undefined };
```
but with explicit undefined:

```typescript
interface I {
  a?: string | undefined;
}

// valid:
const v1: I = {};
const v2: I = { a: undefined };
```

PR #2211, opened 2024-10-01
docs(renovate): enhance renovate documentation

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/docs/renovate/readme.md
+++ b/docs/renovate/readme.md
@@ -41,10 +41,9 @@ You will need the following config in your renovate instance to support the Otte
 ```javascript
 module.exports = {
   // ...
-  allowPostUpgradeCommandTemplating: true,
   allowedPostUpgradeCommands: [
     // Used to execute post-install scripts (like version-harmonize)
-    "^(npm|yarn) install$",
+    "^(npm|yarn) (ci|install)$",
 
     // Used to execute the migration scripts
     "^(npm|yarn) ng update[a-zA-Z0-9.= {}()#@'\/-]*$",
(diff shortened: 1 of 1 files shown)


docs/renovate/readme.md
```diff
@@ -60,3 +60,102 @@ module.exports = {
...
+      hostType: 'npm',
+      matchHost: '<url of your private registry>',
+      username: atob(process.env.PRIVATE_REGISTRY_READ_TOKEN)?.split(':')[0],
+      password: atob(process.env.PRIVATE_REGISTRY_READ_TOKEN)?.split(':')[1]
+    }
+  ],
+  allowPostUpgradeCommandTemplating: true,
+  allowedPostUpgradeCommands: [
+    // Used to execute post-install scripts (like version-harmonize)
+    "^(npm|yarn) install$",
```
kpanot, line 137: not sure about the `$`, you may not want to forbid options

kpanot (dismissed):

docs/renovate/readme.md
```diff
@@ -60,3 +60,102 @@ module.exports = {
   ]
 };
 ```
+
+### Complex case with private registries
+
+If some of your dependencies are coming from a private registry, you will have to add the token in the `.npmrc` or `.yarnrc.yml` to be able to fetch the new versions and trigger the install in postUpgradeCommands.
```
sdo-1A, line 66:
```suggestion
If some of your dependencies are coming from a private registry, you will have to add the token in the `.npmrc` or `.yarnrc.yml` to be able to fetch the new versions and trigger the install in `allowedPostUpgradeCommands`.
```

docs/renovate/readme.md
```diff
@@ -60,3 +60,102 @@ module.exports = {
...
+However, it is not recommended to put the tokens directly in the code of your repository, instead you should use secrets in your CI.
+
+A solution can be to set up the authentication at user level in the CI (`~/.npmrc`).
+This is only working when running `npx renovate` directly (instead of using the GitHub action which relies on docker images)
```
sdo-1A, line 70:
```suggestion
This is only working when running `npx renovate` directly (instead of using the GitHub action which relies on docker images).
```

docs/renovate/readme.md
```diff
@@ -60,3 +60,102 @@ module.exports = {
...
+```
+
+```text
+//.npmrc.ci
+@my-private-scope:registry=<url of your private registry>
+//<url of your private registry>:_auth="${PRIVATE_REGISTRY_READ_TOKEN}"
+```
+
+> [!NOTE]
+> The workflow of the bot don't have to be in the same repository as yours, you just need to make sure the token have the right rights
```
sdo-1A, line 161:
```suggestion
> The workflow of the bot doesn't have to be in the same repository as yours, you just need to make sure the tokens have the right rights.
```
(or "token has")

docs/renovate/readme.md
fpaul-1A: the goal is to limit the allowed commands to only what is really needed for our presets, and currently we do not add any options to the install

docs/renovate/readme.md
fpaul-1A: I am really talking about post-upgrade commands here, I've updated to clarify: `the post-upgrade commands from the presets`

sdo-1A (approved):

docs/renovate/readme.md
```diff
@@ -60,3 +60,102 @@ module.exports = {
...
+      hostType: 'npm',
+      matchHost: '<url of your private registry>',
+      username: atob(process.env.PRIVATE_REGISTRY_READ_TOKEN)?.split(':')[0],
+      password: atob(process.env.PRIVATE_REGISTRY_READ_TOKEN)?.split(':')[1]
+    }
+  ],
+  allowPostUpgradeCommandTemplating: true,
+  allowedPostUpgradeCommands: [
+    // Used to execute post-install scripts (like version-harmonize)
+    "^(npm|yarn) install$",
```
matthieu-crouzet, line 137: shouldn't it be `npm ci` ?

docs/renovate/readme.md
```diff
@@ -60,3 +60,102 @@ module.exports = {
...
+  platform: 'github',
+  repositories: [
+    // List of repositories to scan
+    // ...
+  ],
+  customEnvVariables: {
+    CI: 'true',
+    PRIVATE_REGISTRY_READ_TOKEN: process.env.PRIVATE_REGISTRY_READ_TOKEN
+  },
+  hostRules: [
```
cpaulve-1A, line 126: It could be interesting to mention that the hostRules is to allow the Renovate action while the env is for the postUpgradeTask (if I remember well :D)

docs/renovate/readme.md
```diff
@@ -60,3 +60,102 @@ module.exports = {
...
+      hostType: 'npm',
+      matchHost: '<url of your private registry>',
+      username: atob(process.env.PRIVATE_REGISTRY_READ_TOKEN)?.split(':')[0],
+      password: atob(process.env.PRIVATE_REGISTRY_READ_TOKEN)?.split(':')[1]
+    }
+  ],
+  allowPostUpgradeCommandTemplating: true,
+  allowedPostUpgradeCommands: [
+    // Used to execute post-install scripts (like version-harmonize)
+    "^(npm|yarn) (ci|install)$",
```
cpaulve-1A, line 137: Is there a need for an additional cat .npmrc.ci > ~/.npmrc in the post upgrade?

docs/renovate/readme.md
fpaul-1A: I am not even sure that the `hostRules` are still needed now that we set the authentication globally in `~/.npmrc`

docs/renovate/readme.md
fpaul-1A: no, the copy is done directly in the workflow (not in the postUpgradeCommands`)

docs/renovate/readme.md
fpaul-1A: after some tests, it turns out that this is very much needed :|
I've updated the doc accordingly

cpaulve-1A (approved):

PR #2212, opened 2024-10-02
fix(sdk-core): rollback api-mock.ts to use deprecate fetch client from core

kpanot:
## Proposed change

Rollback `api-mock.ts` to use deprecate fetch client from core.
This change will be part of #2117.

## Related issues

- 🔗  Related to: #2117

### DIFF
--- a/packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/resources/typescriptFetch/spec/api-mock.mustache
+++ b/packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/resources/typescriptFetch/spec/api-mock.mustache
@@ -1,6 +1,6 @@
 {{#apiInfo}}
 import { type ApiClient, isApiClient } from '@ama-sdk/core';
-import { ApiFetchClient, type BaseApiFetchClientConstructor } from '@ama-sdk/client-fetch';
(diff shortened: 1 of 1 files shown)


matthieu-crouzet (approved):

vscaiceanu-1a (approved):

packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/resources/typescriptFetch/spec/api-mock.mustache
```diff
@@ -1,6 +1,6 @@
 {{#apiInfo}}
 import { type ApiClient, isApiClient } from '@ama-sdk/core';
-import { ApiFetchClient, type BaseApiFetchClientConstructor } from '@ama-sdk/client-fetch';
+import { ApiFetchClient, type BaseApiFetchClientConstructor } from '@ama-sdk/core';
```
cpaulve-1A, line 3: Should we keep both lines separated?

kpanot: I would say yes, it is not leading to any issue and will reduce the diff change when migrating to `client-fetch` :)

cpaulve-1A (approved):

PR #2217, opened 2024-10-02
ci: no install needed for new version computation

vscaiceanu-1a:
## Proposed change

Only node and checkout are needed for computing the new version. No need to install.


matthieu-crouzet (approved):

kpanot (approved):

kpanot:
Cf. https://github.com/AmadeusITGroup/otter/pull/2207#discussion_r1783031205

PR #2224, opened 2024-10-02
fix: reduce number of log for rules engine action

cpaulve-1A:
a log is triggered for all the action blocks which does not scale if there are many rulesets

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/rules-engine/src/services/runner/rules-engine.runner.service.ts
+++ b/packages/@o3r/rules-engine/src/services/runner/rules-engine.runner.service.ts
@@ -74,9 +74,7 @@ export class RulesEngineRunnerService implements OnDestroy {
 
     this.subscription.add(
       this.events$.pipe(filter(() => this.enabled)).subscribe(async (events) => {
-        this.logger.debug('Start Rules Engine action executed');
         await this.executeActions(events);
-        this.logger.debug('All Rules Engine actions executed');
       })
     );
   }


packages/@o3r/rules-engine/src/services/runner/rules-engine.runner.service.ts
```diff
@@ -74,9 +74,7 @@ export class RulesEngineRunnerService implements OnDestroy {
 
     this.subscription.add(
       this.events$.pipe(filter(() => this.enabled)).subscribe(async (events) => {
-        this.logger.debug('Start Rules Engine action executed');
```
kpanot, line 77: I am globally not against this change as we already have this information provided in the rules engine debugger and providing debug log outside of the debugger is a bit meaningless.
But I am not sure to understand why it is a bugfix here. Reporting of debug log should not impact in anything the application (and I expect even to report much more debug log than what we do today). The display (or server report) of this level of log should be driven by the logger service client and should not require this change.

kpanot (approved):

fpaul-1A (approved):

packages/@o3r/rules-engine/src/services/runner/rules-engine.runner.service.ts
cpaulve-1A: I agree in principle, we will discuss it during the daily :)

PR #2239, opened 2024-10-07
fix: migration scripts not checking latest semver version with npm

fpaul-1A:
## Proposed change

**Context:** Migration scripts with npm implementation (also used for yarn1)

With the current implementation, when looking for a previous version, we will get the latest **published** version that match the range.
What we want is the most recent version according to semver.

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/extractors/src/core/comparator/package-managers-extractors/npm-file-extractor.helper.ts
+++ b/packages/@o3r/extractors/src/core/comparator/package-managers-extractors/npm-file-extractor.helper.ts
@@ -25,15 +25,27 @@ function sanitizeInput(input: string) {
  * @param packageDescriptor Package descriptor using the npm semver format (i.e. @o3r/demo@^1.2.3)
  * @param paths Paths of the files to extract
  */
-export function getFilesFromRegistry(packageDescriptor: string, paths: string[]): { [key: string]: string } {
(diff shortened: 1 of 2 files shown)


packages/@o3r/extractors/src/core/comparator/package-managers-extractors/npm-file-extractor.helper.ts
```diff
@@ -25,15 +25,26 @@ function sanitizeInput(input: string) {
  * @param packageDescriptor Package descriptor using the npm semver format (i.e. @o3r/demo@^1.2.3)
  * @param paths Paths of the files to extract
  */
-export function getFilesFromRegistry(packageDescriptor: string, paths: string[]): { [key: string]: string } {
+export async function getFilesFromRegistry(packageDescriptor: string, paths: string[]): Promise<Record<string, string>> {
+  const semver = await import('semver');
```
matthieu-crouzet, line 29: shouldn't it be in the `try` block ?

matthieu-crouzet (approved):

cpaulve-1A (dismissed):

matthieu-crouzet (dismissed):

kpanot (approved):

PR #2241, opened 2024-10-07
fix: use only exposed type to fill operators metadata

cpaulve-1A:
## Proposed change
Operators have exposed and supported type.
The first type is what to extract for the CMS and the second is to be used for type check in the code.
Left operand runtime type was overriding right operand metadata (idx = 2, hence > 1).
Now, only extract "ExposedTypes"

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.ts
+++ b/packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.ts
@@ -263,7 +263,6 @@ export class RulesEngineExtractor {
         if (!operatorDeclarations.length) {
           return;
         }
-
         operatorDeclarations.forEach((declaration) => {
           const operatorType = declaration.type.typeName.getText(source);
           const commentParsedDeclaration = this.commentParser.parseConfigDocFromNode(source, declaration);
@@ -291,9 +290,9 @@ export class RulesEngineExtractor {
               nbValues: 1
             };
           }
-
-          declaration.type.typeArguments?.forEach((argType, idx) => {
-            const operand = idx === 0 ? 'leftOperand' : 'rightOperand';
+          const operands: ('leftOperand' | 'rightOperand')[] = ['leftOperand', 'rightOperand'];
+          declaration.type.typeArguments?.slice(0, 2).forEach((argType, idx) => {
+            const operand = operands[idx];
             const operandObject = operator[operand]!;
 
             operandObject.nbValues = this.getTypeNbValue(argType);


fpaul-1A (approved):

packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.ts
```diff
@@ -293,21 +292,23 @@ export class RulesEngineExtractor {
           }
 
           declaration.type.typeArguments?.forEach((argType, idx) => {
```
matthieu-crouzet, line 294:
```suggestion
          declaration.type.typeArguments?.slice(0, 2).forEach((argType, idx) => {
```
if we are interested only in the 2 first elements

matthieu-crouzet (dismissed):

vscaiceanu-1a (approved):

packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.ts
```diff
@@ -293,21 +292,23 @@ export class RulesEngineExtractor {
...
-              delete operator[operand];
-            } else {
-              const typeData = this.extractComplexTypeData(typeNode, source, operandObject.nbValues);
-              operandObject.nbValues = typeData.nbValue;
-              operandObject.types = typeData.types;
-              if (operandObject.types.some((type) => type !== 'unknown' && allSupportedTypes.indexOf(type) === -1)) {
-                this.logger.warn(`Operator ${operator.id} has an unsupported type in operand ${operand}: "${operandObject.types.join(', ')}"`);
-                operandObject.types = ['unknown'];
+            if (idx <= 1) {
+              const operand = idx === 0 ? 'leftOperand' : 'rightOperand';
```
kpanot, line 296: (not mandatory but cooler :D :)

```suggestion
              const operands = ['leftOperand', 'rightOperand'];
              const operand = operands[idx];
```
(even better if the array in externalize)

packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.ts
```diff
@@ -291,9 +290,9 @@ export class RulesEngineExtractor {
               nbValues: 1
             };
           }
-
-          declaration.type.typeArguments?.forEach((argType, idx) => {
-            const operand = idx === 0 ? 'leftOperand' : 'rightOperand';
+          const operands: ('leftOperand' | 'rightOperand')[] = ['leftOperand', 'rightOperand'];
```
fpaul-1A, line 293:
```suggestion
          const operands = ['leftOperand', 'rightOperand'] as const;
```
?

fpaul-1A (approved):

vscaiceanu-1a (approved):

PR #2253, opened 2024-10-09
fix: missing telemetry dependencies in CLIs

kpanot:


matthieu-crouzet (approved):

packages/@ama-sdk/create/package.json
```diff
@@ -16,6 +16,14 @@
     "postbuild": "cpy package.json dist/",
     "prepare:publish": "prepare-publish ./dist"
   },
+  "peerDependencies": {
+    "@o3r/telemetry": "workspace:^"
```
fpaul-1A, line 20: Can this peer dependency ever be provided with npm create?

kpanot: If installed in global only

vscaiceanu-1a (approved):

PR #2271, opened 2024-10-10
fix: make typings more accurate by avoiding explict cast on consts

fpaul-1A:
## Proposed change

When creating an object that is intended to be read-only, we should let typescript infer the type.

Not good:
```typescript
interface NamedObjectMaybe {
  name?: string;
}
const UNMOVABLE_OBJECT: NamedObjectMaybe = {
  name: 'unicorn'
};

// UNMOVABLE_OBJECT.name is typed as string | undefined
```

Nice:
```typescript
interface NamedObjectMaybe {
  name?: string;
}
const UNMOVABLE_OBJECT = {
  name: 'unicorn'
} as const satisfies NamedObjectMaybe;

// UNMOVABLE_OBJECT.name is typed as 'unicorn' but we still validate that UNMOVABLE_OBJECT matches the interface
```
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/apps/github-cascading-app/src/cascading/interfaces.ts
+++ b/apps/github-cascading-app/src/cascading/interfaces.ts
@@ -64,7 +64,7 @@ export type CheckConclusion = 'cancelled' | 'neutral' | 'success' | 'failure' | 
 /**
  * Default configuration
  */
-export const DEFAULT_CONFIGURATION: Readonly<CascadingConfiguration> = {
+export const DEFAULT_CONFIGURATION = {
   ignoredPatterns: [] as string[],
   defaultBranch: '',
   cascadingBranchesPattern: '^releases?/\\d+\\.\\d+',
@@ -73,4 +73,4 @@ export const DEFAULT_CONFIGURATION: Readonly<CascadingConfiguration> = {
   labels: [] as string[],
   pullRequestTitle: '[cascading] from $origin to $target',
   branchNamePrefix: 'cascading'
-} as const;
+} as const satisfies Readonly<CascadingConfiguration>;
--- a/apps/showcase/src/app/app.module.ts
+++ b/apps/showcase/src/app/app.module.ts
@@ -33,14 +33,14 @@ import { AppRoutingModule } from './app-routing.module';
 import { AppComponent } from './app.component';
 
 
-const runtimeChecks: Partial<RuntimeChecks> = {
+const runtimeChecks = {
   strictActionImmutability: false,
   strictActionSerializability: false,
   strictActionTypeUniqueness: !isDevMode(),
   strictActionWithinNgZone: !isDevMode(),
   strictStateImmutability: !isDevMode(),
   strictStateSerializability: false
-};
+} as const satisfies Partial<RuntimeChecks>;
 
 registerLocaleData(localeEN, 'en-GB');
 registerLocaleData(localeFR, 'fr-FR');
--- a/apps/showcase/src/components/showcase/configuration/configuration-pres.config.ts
+++ b/apps/showcase/src/components/showcase/configuration/configuration-pres.config.ts
@@ -40,7 +40,7 @@ export interface ConfigurationPresConfig extends Configuration {
   shouldProposeRoundTrip: boolean;
 }
 
-export const CONFIGURATION_PRES_DEFAULT_CONFIG: ConfigurationPresConfig = {
+export const CONFIGURATION_PRES_DEFAULT_CONFIG = {
   inXDays: 7,
   destinations: [
     { cityName: 'London', available: true },
@@ -48,6 +48,6 @@ export const CONFIGURATION_PRES_DEFAULT_CONFIG: ConfigurationPresConfig = {
     { cityName: 'New-York', available: false }
   ],
   shouldProposeRoundTrip: false
-};
+} as const satisfies ConfigurationPresConfig;
 
 export const CONFIGURATION_PRES_CONFIG_ID = computeItemIdentifier('ConfigurationPresConfig', 'showcase');
--- a/apps/showcase/src/components/showcase/localization/localization-pres.translation.ts
+++ b/apps/showcase/src/components/showcase/localization/localization-pres.translation.ts
@@ -31,7 +31,7 @@ export interface LocalizationPresTranslation extends Translation {
   cityName: string;
 }
 
-export const translations: LocalizationPresTranslation = {
+export const translations = {
   welcome: 'o3r-localization-pres.welcome',
   welcomeWithCityName: 'o3r-localization-pres.welcomeWithCityName',
   question: 'o3r-localization-pres.question',
@@ -39,4 +39,4 @@ export const translations: LocalizationPresTranslation = {
   departureLabel: 'o3r-localization-pres.departureLabel',
   cityName: 'o3r-localization-pres.cityName',
   destinationPlaceholder: 'o3r-localization-pres.destinationPlaceholder'
-};
+} as const satisfies LocalizationPresTranslation;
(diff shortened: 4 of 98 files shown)


kpanot:
Do we really want systematically that?
It can be used in cases to avoid additional check or explicit `!` mark but

packages/@ama-sdk/client-beacon/src/api-beacon-client.ts
```diff
@@ -25,11 +25,11 @@ export interface BaseApiBeaconClientOptions extends BaseApiClientOptions {
 export interface BaseApiBeaconClientConstructor extends PartialExcept<Omit<BaseApiBeaconClientOptions, 'replyPlugins'>, 'basePath'> {
 }
 
-const DEFAULT_OPTIONS: Omit<BaseApiBeaconClientOptions, 'basePath'> = {
+const DEFAULT_OPTIONS = {
   replyPlugins: [] as never[],
   requestPlugins: [],
   enableTokenization: false
-};
+} as const satisfies Partial<BaseApiBeaconClientOptions>;
```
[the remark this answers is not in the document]
fpaul-1A: true, I was not sure if we added this type to try to match the type of the variable or the other way around
I'll keep `Omit<BaseApiBeaconClientOptions, 'basePath'>` then

apps/showcase/src/components/showcase/configuration/configuration-pres.config.ts
```diff
@@ -40,14 +40,14 @@ export interface ConfigurationPresConfig extends Configuration {
...
-export const CONFIGURATION_PRES_DEFAULT_CONFIG: ConfigurationPresConfig = {
+export const CONFIGURATION_PRES_DEFAULT_CONFIG = {
   inXDays: 7,
   destinations: [
     { cityName: 'London', available: true },
     { cityName: 'Paris', available: true },
     { cityName: 'New-York', available: false }
   ],
   shouldProposeRoundTrip: false
-};
```
matthieu-crouzet, line 51: We should change the generators accordingly and provide a migration script for that no ?

packages/@ama-sdk/client-fetch/src/api-fetch-client.ts
```diff
@@ -34,13 +34,13 @@ export interface BaseApiFetchClientOptions extends BaseApiClientOptions {
...
 
-const DEFAULT_OPTIONS: Omit<BaseApiFetchClientOptions, 'basePath'> = {
+const DEFAULT_OPTIONS = {
   replyPlugins: [new ReviverReply(), new ExceptionReply()],
   fetchPlugins: [],
   requestPlugins: [],
   enableTokenization: false,
   disableFallback: false
-};
+} as const satisfies Partial<BaseApiFetchClientOptions>;
```
matthieu-crouzet, line 43:
```suggestion
} as const satisfies Omit<BaseApiFetchClientOptions, 'basePath'>;
```

apps/showcase/src/app/app.module.ts
```diff
@@ -33,14 +33,14 @@ import { AppRoutingModule } from './app-routing.module';
...
-const runtimeChecks: Partial<RuntimeChecks> = {
+const runtimeChecks = {
   strictActionImmutability: false,
   strictActionSerializability: false,
   strictActionTypeUniqueness: !isDevMode(),
   strictActionWithinNgZone: !isDevMode(),
   strictStateImmutability: !isDevMode(),
   strictStateSerializability: false
-};
+} as const satisfies Partial<RuntimeChecks>;
```
matthieu-crouzet, line 43: I think this is part of the generated code so we should update it as well

packages/@o3r/rules-engine/src/devkit/rules-engine-devtools.service.ts
```diff
@@ -31,7 +31,7 @@ export class OtterRulesEngineDevtools {
     private readonly rulesEngineService: RulesEngineRunnerService,
     @Optional() @Inject(OTTER_RULES_ENGINE_DEVTOOLS_OPTIONS) options: RulesEngineDevtoolsServiceOptions) {
 
-    const eventsStackLimit = (options || OTTER_RULES_ENGINE_DEVTOOLS_DEFAULT_OPTIONS).rulesEngineStackLimit;
+    const eventsStackLimit = (options || OTTER_RULES_ENGINE_DEVTOOLS_DEFAULT_OPTIONS as RulesEngineDevtoolsServiceOptions).rulesEngineStackLimit;
```
matthieu-crouzet, line 34: Why do you need the cast here ?

matthieu-crouzet (commented):
Should we find a eslint rule that check that ?

packages/@o3r/rules-engine/src/devkit/rules-engine-devtools.service.ts
fpaul-1A: I'd like to discuss that we the team, here the compiler knows that `rulesEngineStackLimit` is never defined in `OTTER_RULES_ENGINE_DEVTOOLS_DEFAULT_OPTIONS`
But we want to make sure it's taken into account if we add it one day for some reason
That might be one reason why we wouldn't want more accurate typings

fpaul-1A:
> Should we find a eslint rule that check that ?

Not sure it's going to be easy, explicit cast is still relevant in a lot of cases
You would need to detect that the object is not modified anywhere in the code
To be discussed

packages/@o3r/configuration/src/devkit/configuration-devtools.token.ts
```diff
@@ -1,12 +1,12 @@
...
-export const OTTER_CONFIGURATION_DEVTOOLS_DEFAULT_OPTIONS: ConfigurationDevtoolsServiceOptions = {
+export const OTTER_CONFIGURATION_DEVTOOLS_DEFAULT_OPTIONS = {
   defaultLibraryName: '@o3r/components',
   defaultJsonFilename: 'partial-static-config.json',
   isActivatedOnBootstrap: false,
   isActivatedOnBootstrapWhenCMSContext: true
-};
+} as const satisfies ConfigurationDevtoolsServiceOptions;
 
 // eslint-disable-next-line max-len
```
vscaiceanu-1a, line 11: maybe to don't need this exception anymore?

apps/showcase/src/components/showcase/configuration/configuration-pres.config.ts
cpaulve-1A: Won't migration script be dangerous?

apps/showcase/src/components/showcase/localization/localization-pres.translation.ts
```diff
@@ -31,12 +31,12 @@ export interface LocalizationPresTranslation extends Translation {
...
+export const translations = {
   welcome: 'o3r-localization-pres.welcome',
   welcomeWithCityName: 'o3r-localization-pres.welcomeWithCityName',
   question: 'o3r-localization-pres.question',
   destinationLabel: 'o3r-localization-pres.destinationLabel',
   departureLabel: 'o3r-localization-pres.departureLabel',
   cityName: 'o3r-localization-pres.cityName',
   destinationPlaceholder: 'o3r-localization-pres.destinationPlaceholder'
-};
+} as const satisfies LocalizationPresTranslation;
```
kpanot, line 42: Same as previous comment, this may need to be changed in the generator as well

apps/github-cascading-app/src/cascading/interfaces.ts
```diff
@@ -64,7 +64,7 @@ export type CheckConclusion = 'cancelled' | 'neutral' | 'success' | 'failure' |
 /**
  * Default configuration
  */
-export const DEFAULT_CONFIGURATION: Readonly<CascadingConfiguration> = {
```
kpanot, line 67: As discussed, I think the exported const (function, fields, etc...) should keep the type of interface (potentially with `ReadOnly`) and not rely on `satisfies`.
For the following reasons:
- Avoid (future) breaking changes
- Keep the nature of the interface which is to offer a contract with the user

apps/showcase/src/components/showcase/configuration/configuration-pres.config.ts
kpanot: This change is a bit dangerous because it can impact the extractor code as well.
Let consider it in different PR is we want this change

PR #2273, opened 2024-10-10
feat(eslint): one import per line + import alphabetize + export all alphabetize

matthieu-crouzet:
## Proposed change
One import per line
Import alphabetize
Export all alphabetize


packages/@o3r/eslint-config/schematics/ng-add/vscode/index.ts
```diff
@@ -1,5 +1,9 @@
-import type { JsonObject } from '@angular-devkit/core';
-import { chain, type Rule } from '@angular-devkit/schematics';
+import type {
```
kpanot, line 1: 😢

packages/@o3r/eslint-config/schematics/ng-add/vscode/index.ts
```diff
@@ -13,6 +18,7 @@ export const updateVscode: Rule = async () => {
       const settings = tree.exists(vscodeSettingsPath) ? (tree.readJson(vscodeSettingsPath) || {}) as JsonObject : {};
       settings['eslint.useFlatConfig'] = true;
       settings['editor.defaultFormatter'] = 'dbaeumer.vscode-eslint';
+      settings['editor.foldingImportsByDefault'] = true;
```
matthieu-crouzet, line 21: Thanks to `unused-imports/no-unused-imports` we can collapse all imports by default :)

kpanot (approved):

packages/@o3r/eslint-config/src/rules/typescript/import.cjs
```diff
@@ -0,0 +1,25 @@
+/**
+ * @type {import('@typescript-eslint/utils').TSESLint.FlatConfig.ConfigArray}
+ */
+const config = [
+  {
+    name: '@o3r/overrides/import',
+    files: [
+      '**/*.{c,m,}{t,j}s'
+    ],
+    rules: {
```
mrednic-1A, line 10: We did not talked about it, but for imports I would add a trailing comma mandatory.
In this way if a new import is added at the end of the list only one line will appear on the diff.
Without trailing comma the diff will be on 2 lines.
https://eslint.style/rules/js/comma-dangle#options

eslint.shared.config.mjs
```diff
@@ -1,10 +1,9 @@
-import globals from 'globals';
+import o3rTemplate from '@o3r/eslint-config/template';
```
matthieu-crouzet, line 1: I don't understand why it does not go next to `import o3rConfig from '@o3r/eslint-config';` but if it is automatic now we don't really need to review the order, everyone will have the same

mrednic-1A (approved):

PR #2283, opened 2024-10-14
fix: metadata checks should look for cms-metadata in package.json for previous version

fpaul-1A:
## Proposed change

There can be a discrepancy between the path to cms metadata between 2 version of an application.
However, we only specify one path when running the metadata-checks.
We can use the path to the metadata specified in `package.json` under `cmsMetadata` to find the files even if the option passed to the script doesn't match.

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@o3r/extractors/src/core/comparator/metadata-comparison.helper.ts
+++ b/packages/@o3r/extractors/src/core/comparator/metadata-comparison.helper.ts
@@ -9,6 +9,7 @@ import { coerce, Range, satisfies } from 'semver';
 import type { PackageJson } from 'type-fest';
 import type { MetadataComparator, MigrationData, MigrationFile, MigrationMetadataCheckBuilderOptions } from './metadata-comparator.interface';
 import { getFilesFromRegistry, getLatestMigrationMetadataFile, getLocalMetadataFile, getVersionRangeFromLatestVersion } from './metadata-files.helper';
+import type { CmsMetadataData } from '../../interfaces';
 
 function checkMetadataFile<MetadataItem, MigrationMetadataItem, MetadataFile>(
   lastMetadataFile: MetadataFile,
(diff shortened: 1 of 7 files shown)


matthieu-crouzet (approved):

packages/@o3r/extractors/src/core/comparator/metadata-comparison.helper.ts
```diff
@@ -134,13 +135,23 @@ Detection of package manager runner will fallback on the one used to execute the
...
+  const previousPackageJson = JSON.parse(previousFile[packageJsonFileName]) as PackageJson & {cmsMetadata: CmsMetadataData};
+  context.logger.info(`Successfully fetched ${packageLocator} with version ${previousPackageJson.version}.`);
+
+  let metadata: MetadataFile;
+  const previousMetadataEntryFromPackageJson = previousPackageJson.cmsMetadata?.[options.packageJsonEntry];
+  if (previousMetadataEntryFromPackageJson) {
+    const previousFileFromPackageJson = await getFilesFromRegistry(packageLocator, [previousMetadataEntryFromPackageJson], packageManager, context.workspaceRoot);
+    metadata = JSON.parse(previousFileFromPackageJson[previousMetadataEntryFromPackageJson]) as MetadataFile;
+    context.logger.info(`Resolved metadata from "${previousPackageJson.cmsMetadata[options.packageJsonEntry]}".`);
+  } else {
```
mrednic-1A, line 147: Is it necessary to have the filed in package.json to have higher priority than the option passed in command line?. Usually we have the option in command line with the higher priority.

vscaiceanu-1a (approved):

mrednic-1A (approved):

PR #2287, opened 2024-10-15
fix(design): way to sort variable to refer existing variables

kpanot:
## Proposed change

fix(design): way to sort variable to refer existing variables

### DIFF
--- a/packages/@o3r/design/builders/generate-style/helpers/style-renderer-options.ts
+++ b/packages/@o3r/design/builders/generate-style/helpers/style-renderer-options.ts
@@ -10,6 +10,8 @@ import {
   getSassStyleContentUpdater,
   getSassTokenDefinitionRenderer,
   getSassTokenValueRenderer,
+  getTokenSorterByName,
+  getTokenSorterByRef,
   type SassTokenDefinitionRendererOptions,
   type SassTokenValueRendererOptions,
   type TokenKeyRenderer,
@@ -122,8 +124,23 @@ export const getStyleRendererOptions = (tokenVariableNameRenderer: TokenKeyRende
     }
   })(options.variableType || options.language);
 
+  /** Sorting strategy of variables based on selected language */
+  const tokenListTransforms = ((language) => {
+    switch (language) {
+      case 'scss':
+      case 'sass': {
+        return [getTokenSorterByName, getTokenSorterByRef];
+      }
+      case 'css':
+      default: {
+        return [getTokenSorterByName];
+      }
+    }
+  })(options.variableType || options.language);
+
   /** Option to be used by the style renderer */
   return {
+    tokenListTransforms,
     writeFile: writeFileWithLogger,
     styleContentUpdater,
     determineFileToUpdate,
--- a/packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.spec.ts
+++ b/packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.spec.ts
@@ -3,7 +3,7 @@ import { promises as fs } from 'node:fs';
 import { resolve } from 'node:path';
 import type { DesignTokenSpecification } from '../design-token-specification.interface';
 import type { DesignTokenVariableSet } from '../parsers';
-import { compareVariableByName, computeFileToUpdatePath, renderDesignTokens } from './design-token-style.renderer';
+import { computeFileToUpdatePath, getTokenSorterByName, getTokenSorterByRef, renderDesignTokens } from './design-token-style.renderer';
 
 describe('Design Token Renderer', () => {
   let exampleVariable!: DesignTokenSpecification;
@@ -78,7 +78,7 @@ describe('Design Token Renderer', () => {
       expect(writeFile).toHaveBeenCalledTimes(2);
     });
 
-    describe('the comparator', () => {
+    describe('the transformers sorting by name', () => {
       const firstVariable = '--example-color';
       const lastVariable = '--example-wrong-ref';
 
@@ -113,13 +113,47 @@ describe('Design Token Renderer', () => {
           readFile,
           existsFile,
           determineFileToUpdate,
-          variableSortComparator: (a, b) => -compareVariableByName(a, b)
+          tokenListTransforms: [(list) => (vars) => getTokenSorterByName(list)(vars).reverse()]
         });
 
         const contentToTest = result['styles.scss'];
 
         expect(contentToTest.indexOf(firstVariable)).toBeGreaterThan(contentToTest.indexOf(lastVariable));
       });
+
+      test('should execute the transform functions in given order', async () => {
+        const result: any = {};
+        const writeFile = jest.fn().mockImplementation((filename, content) => { result[filename] = content; });
+        const readFile = jest.fn().mockReturnValue('');
+        const existsFile = jest.fn().mockReturnValue(true);
+        const determineFileToUpdate = jest.fn().mockImplementation(computeFileToUpdatePath('.'));
+        const tokenListTransform = jest.fn().mockReturnValue([]);
+        await renderDesignTokens(designTokens, {
+          writeFile,
+          readFile,
+          existsFile,
+          determineFileToUpdate,
+          tokenListTransforms: [() => tokenListTransform, () => tokenListTransform]
+        });
+
+        expect(tokenListTransform).toHaveBeenCalledTimes(2 * 2); // twice on 2 files
+        expect(tokenListTransform).not.toHaveBeenNthCalledWith(1, []);
+        expect(tokenListTransform).toHaveBeenNthCalledWith(2, []);
+        expect(tokenListTransform).not.toHaveBeenNthCalledWith(3, []);
+        expect(tokenListTransform).toHaveBeenNthCalledWith(4, []);
+      });
+    });
+  });
+
+  describe('getTokenSorterByRef', () => {
+    it('should sort properly variables with multiple refs', () => {
+      const list = Array.from(designTokens.values());
+      const sortedTokens = getTokenSorterByRef(designTokens)(list);
+
+      expect(list.findIndex(({ tokenReferenceName }) => tokenReferenceName === 'example.post-ref'))
+        .toBeLessThan(list.findIndex(({ tokenReferenceName }) => tokenReferenceName === 'example.var1'));
+      expect(sortedTokens.findIndex(({ tokenReferenceName }) => tokenReferenceName === 'example.post-ref'))
+        .toBeGreaterThan(sortedTokens.findIndex(({ tokenReferenceName }) => tokenReferenceName === 'example.var1'));
     });
   });
 });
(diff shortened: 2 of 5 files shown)


packages/@o3r/design/builders/generate-style/helpers/style-renderer-options.ts
```diff
@@ -122,8 +124,23 @@ export const getStyleRendererOptions = (tokenVariableNameRenderer: TokenKeyRende
     }
   })(options.variableType || options.language);
 
+  /** Variable sorting strategy  based on selected language */
```
fpaul-1A, line 127:
```suggestion
  /** Sorting strategy of variables based on selected language */
```

kpanot: I will take your (mainly for the double space removal :D)

matthieu-crouzet (dismissed):

packages/@o3r/design/src/core/design-token/renderers/design-token.renderer.interface.ts
```diff
@@ -17,18 +17,29 @@ export type DesignContentFileUpdater = (variables: string[], file: string, style
...
    * Comparator to sort variable before rendering
-   * @default {@see compareVariableByName}
    * @param a first Design Token
    * @param b second Design Token
+   * @deprecated Use {@link tokenListTransform} instead. Will be removed in v13
    */
   variableSortComparator?: (a: DesignTokenVariableStructure, b: DesignTokenVariableStructure) => number;
 
+  /**
+   * List of tokens transform function. The transformation will be apply per file.
```
sdo-1A, line 38:
```suggestion
   * List of tokens transform functions. The transformation will be applied per file.
```

packages/@o3r/design/src/core/design-token/renderers/design-token.renderer.interface.ts
```diff
@@ -17,18 +17,29 @@ export type DesignContentFileUpdater = (variables: string[], file: string, style
...
+   * Note: the order of the outputted array will be kept when generating the code
```
sdo-1A, line 39:
```suggestion
   * Note: the order of the output array will be kept when generating the code
```

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.spec.ts
```diff
@@ -78,7 +78,7 @@ describe('Design Token Renderer', () => {
       expect(writeFile).toHaveBeenCalledTimes(2);
     });
 
-    describe('the comparator', () => {
+    describe('the transformers sorting by name', () => {
```
sdo-1A, line 81:
```suggestion
    describe('the transformers sorted by name', () => {
```
not sure about this suggestion?

cpaulve-1A: I would leave it as is, you read this methods as "It describes the transformers sorting by name", I am not sure "It describes the transformers sorted by name" is exactly what we mean here.

sdo-1A (approved):

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
```diff
@@ -22,9 +22,62 @@ export const computeFileToUpdatePath = (root = process.cwd(), defaultFile = 'sty
...
+export const getTokenSorterByName: TokenListTransform = (_variableSet) => (tokens) => tokens.sort((a, b) => a.getKey().localeCompare(b.getKey()));
+
+
+/**
+ * Compare Token variable by reference to write theme in reference order
+ * @param variableSet Complete set of the parsed Design Token
+ */
+export const getTokenSorterByRef: TokenListTransform = (variableSet) => {
+  /**
+   * Get the maximum number of reference levels for a Design Token, following its references
```
cpaulve-1A, line 42: Just to be sure I understand:
```
    "test": {
      "height": {
        "$extensions": {
          "o3rPrivate": true
        },
        "$value": 2.3,
        "$type": "number"
      },
      "width": {
        "$value": "{example.test.height}"
      },

      "border": {
        "$type": "border",
        "$value": {
          "color": "{example.color}",
          "style": "dashed",
          "width": "{example.test.width}"
        }
      }
```
border would return 2, width would return 1 and height 0?

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.spec.ts
```diff
@@ -113,13 +113,47 @@ describe('Design Token Renderer', () => {
...
+  });
+
+  describe('getTokenSorterByRef', () => {
+    it('should sort properly variables with multiple refs', () => {
+      const list = Array.from(designTokens.values());
+      const sortedTokens = getTokenSorterByRef(designTokens)(list);
+
+      expect(list.findIndex(({ tokenReferenceName }) => tokenReferenceName === 'example.post-ref'))
+        .toBeLessThan(list.findIndex(({ tokenReferenceName }) => tokenReferenceName === 'example.var1'));
+      expect(sortedTokens.findIndex(({ tokenReferenceName }) => tokenReferenceName === 'example.post-ref'))
```
cpaulve-1A, line 155: Where does post-ref comes from?

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
kpanot: exactly

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
```diff
@@ -22,9 +22,62 @@ export const computeFileToUpdatePath = (root = process.cwd(), defaultFile = 'sty
...
 
+/**
+ * Compare Token variable by name
+ * @param _variableSet Complete set of the parsed Design Token
+ */
+export const getTokenSorterByName: TokenListTransform = (_variableSet) => (tokens) => tokens.sort((a, b) => a.getKey().localeCompare(b.getKey()));
+
+
+/**
+ * Compare Token variable by reference to write theme in reference order
```
cpaulve-1A, line 37: What do you mean by this? Do you sort by crescent order on the depth of the reference? I am not sure I understand the method from the description

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
```diff
@@ -22,9 +22,62 @@ export const computeFileToUpdatePath = (root = process.cwd(), defaultFile = 'sty
  * Compare Token variable by name
  * @param a first token variable
  * @param b second token variable
+ * @deprecated use {@link getTokenSorterByName} instead. Will be removed in v13.
  */
 export const compareVariableByName = (a: DesignTokenVariableStructure, b: DesignTokenVariableStructure): number => a.getKey().localeCompare(b.getKey());
 
+/**
+ * Compare Token variable by name
```
cpaulve-1A, line 30:
```suggestion
 * Sort Token variable by name using the local alphabetical order
 ```
?

**Review by fpaul-1A — APPROVED** [review]

File: packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts [hunk 9]
```diff
@@ -22,9 +22,62 @@ export const computeFileToUpdatePath = (root = process.cwd(), defaultFile = 'sty
...
+      let hasChanged = false;
+      sortToken = sortToken.reduce((acc, token, idx) => {
+        const firstRef = sortToken.findIndex((t) => t.getReferences(variableSet).includes(token.tokenReferenceName));
+        if (firstRef > -1 && firstRef < idx) {
+          acc.splice(idx, 1);
+          acc.splice(firstRef, 0, token);
+          hasChanged = true;
+        }
+        return acc;
+      }, sortToken);
```
cpaulve-1A, line 71: Are you sure reduce is working with a copy of sortToken? It seems rather dangerous to modify what you are iterating on.

cpaulve-1A: image
In this case, it stops iterating after the 2 second element because idx > myTestArray.length
image

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.spec.ts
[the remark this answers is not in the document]
cpaulve-1A: yeah I have not get there yet :D

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
cpaulve-1A: especially with splice in the loop :)

kpanot: You actually have an issue in your example because you change the number of items in the original array (you could also give an example where you desynchronize the index of the loop by adding more item before the current index and get an issue as well).

In the current code there is:
- no array length change
- no change of number of items in left and right part between the index.

An example of very similar case is a standard function to revert an array:

```javascript
const testArray = ['a', 'b', 'c', 'd'];

testArray.reduce((acc, key, index) => {
    console.log(acc, key, index);
    acc.splice(index, 1);
    acc.unshift(key);
    return acc;
}, testArray);

console.log('result: ', testArray);
```

which output:
```
['a', 'b', 'c', 'd'] 'a' 0
['a', 'b', 'c', 'd'] 'b' 1
['b', 'a', 'c', 'd'] 'c' 2
['c', 'b', 'a', 'd'] 'd' 3
result: ['d', 'c', 'b', 'a']
```

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
kpanot: updated the description

matthieu-crouzet (approved):

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
cpaulve-1A: This is why I mentionned that in your particular case it works (see my remark "I know it should not be the case because you remove one element and add one before the idx, but it is not easy too read and it is risky for maintenance"), however I would not consider it a good practice as we could easily mess up the code while maintaining it.

fpaul-1A (approved):

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
kpanot: My bad I got only the first message on VsCode so I replied with the first one as context only.
In anycase, the code has changed, you can see if it fills better your expectations

cpaulve-1A (approved):

Issue #2292, opened 2024-10-16
[Feature] Audit GH Action to support npm


vscaiceanu-1a:
Today, the Audit GitHub action supports only `yarn`.
The goal is to support `npm` with the same reporting features as for `yarn. 

_Originally posted by @vscaiceanu-1a in https://github.com/AmadeusITGroup/otter/pull/2207#discussion_r1784176653_

PR #2297, opened 2024-10-17
feat(github-actions): support npm for audit action

fpaul-1A:
## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #2292


tools/github-actions/audit/action.yml
```diff
@@ -1,6 +1,10 @@
 name: 'Audit'
 description: 'Step to execute yarn audit and generate the report'
 inputs:
+  packageManager:
+    description: 'Package manager used to perform the audit'
+    required: false
+    default: yarn
```
matthieu-crouzet, line 7: don't we want to detect it by the presence of the `yarn.lock` file ?

tools/github-actions/audit/src/reports.ts
```diff
@@ -0,0 +1,198 @@
...
+ * report summary
+ * @param response
+ * @param severityThreshold
+ */
+export const computeNpmV2Report: OtterAuditReporter = (response: string, severityThreshold: Severity) => {
+  const reportJson = JSON.parse(response) as NPMAuditReportV2.Audit;
+  core.debug(response);
+  const nbVulnerabilities = severities.reduce((acc, severity) => acc + (reportJson.metadata.vulnerabilities[severity] ?? 0), 0);
+  let highestSeverityFound: Severity | undefined;
+  for (let index = severities.length - 1; index >= 0; index--) {
```
matthieu-crouzet, line 80: It seems that this loop is duplicated in the 4 helpers 
we may share the same code no ?

tools/github-actions/audit/src/reports.ts
```diff
@@ -0,0 +1,198 @@
...
+    const severity: Severity = severities[index];
+    if (reportJson.metadata.vulnerabilities[severity] > 0) {
+      highestSeverityFound = severity;
+      break;
+    }
+  }
+  return Object.values(reportJson.vulnerabilities)
+    .reduce<OtterAuditReport>((currentVulnerabilities, advisory) => {
```
matthieu-crouzet, line 88: we could also share the code inside the reduce if we provide a function to create the `vulnerability` pushed in `errors` and `warnings`

tools/github-actions/audit/src/reports.ts
```diff
@@ -0,0 +1,198 @@
...
+ * Data to build the action report summary
+ */
+export interface OtterAuditReport {
+  nbVulnerabilities: number;
+  errors: OtterAdvisory[];
+  warnings: OtterAdvisory[];
+  highestSeverityFound?: Severity;
+}
+
+export type OtterAuditReporter = (response: string, severityThreshold: Severity) => OtterAuditReport;
```
cpaulve-1A, line 37:
```suggestion
/** 
 * Method which formats the response of an audit api (yarn or npm) into a report that can be displayed in the 
 * Github action summary
 */
export type OtterAuditReporter = (response: string, severityThreshold: Severity) => OtterAuditReport;
```

tools/github-actions/audit/src/reports.ts
```diff
@@ -0,0 +1,198 @@
...
+/**
+ * Format the response from npm audit V1 in a common interface that will be used to build the
+ * report summary
+ * @param response
+ * @param severityThreshold
+ */
+export const computeNpmV1Report: OtterAuditReporter = (response: string, severityThreshold: Severity) => {
+  const reportJson = JSON.parse(response) as Audit;
+  core.debug(response);
+  const nbVulnerabilities = Object.values(reportJson.metadata.vulnerabilities as { [key: string]: number } || {}).reduce((acc, curr) => acc + curr, 0);
```
cpaulve-1A, line 48: NB: We could have also computed the highestSeverity in this reduce

tools/github-actions/audit/action.yml
```diff
@@ -10,11 +10,11 @@ inputs:
     required: false
     default: 'true'
   recursive:
-    description: 'Audit transitive dependencies as well'
+    description: 'Audit transitive dependencies as well (Yarn Berry only)'
     required: false
     default: 'true'
   environment:
-    description: 'Which environments to cover'
+    description: 'Which environments to cover ("production" or "development")'
```
vscaiceanu-1a, line 17: you could use `options` input in this case

tools/github-actions/audit/src/reports.ts
```diff
@@ -0,0 +1,179 @@
...
+  NPMAuditReportV2,
+  Severity,
+  YarnAudit,
+  YarnNpmAuditReport
+} from 'audit-types';
+type Audit = NPMAuditReportV1.Audit;
+type Advisory = NPMAuditReportV1.Advisory;
+
+/**
+ * Severities supported by yarn npm audit from the lowest to the highest criticality
```
vscaiceanu-1a, line 14: i guess it applies to all package managers, not only yarn?

tools/github-actions/audit/src/reports.ts
```diff
@@ -0,0 +1,179 @@
+import * as core from '@actions/core';
+import {
```
vscaiceanu-1a, line 2: can these be type imports?

tools/github-actions/audit/src/reports.ts
```diff
@@ -0,0 +1,179 @@
...
+};
+
+/**
+ * Format the response from yarn 2 and 3 npm audit in a common interface that will be used to build the
+ * report summary
+ * @param response
+ * @param severityThreshold
+ */
+export const computeYarn3Report: OtterAuditReporter = (response: string, severityThreshold: Severity) => {
+  core.info('Computing Report for Yarn 3');
```
vscaiceanu-1a, line 153: Could you retrieve the yarn version (2 or 3) from anywhere?
So that we don't confuse people using yarn 2.

tools/github-actions/audit/src/main.ts
```diff
@@ -1,137 +1,55 @@
...
-    const command = `yarn npm audit --environment ${environment} ${allWorkspaces ? '--all ' : ''}${recursive ? '--recursive ' : ''}--json`;
+    let auditReporter: OtterAuditReporter;
+    let auditCommand: string;
+    if (packageManager === 'yarn') {
+      const versionOutput = await getExecOutput('yarn --version', [], {cwd: process.env.GITHUB_WORKSPACE});
+      const version = Number.parseInt(versionOutput.stdout.split('.')[0], 10);
+      auditCommand = version <= 1 ?
+        `yarn audit ${environment === 'production' ? '--groups "dependencies peerDependencies" ' : ''}--json` :
+        `yarn npm audit --environment ${environment} ${allWorkspaces ? '--all ' : ''}${recursive ? '--recursive ' : ''}--json`;
+      auditReporter = version >= 4 ? computeYarn4Report : version >= 2 ? computeYarn3Report : computeYarn1Report;
```
vscaiceanu-1a, line 35: would it be more readable in a switch case?

tools/github-actions/audit/src/reports.ts
matthieu-crouzet: not sure as we import namespaces and not typings

tools/github-actions/audit/src/reports.ts
```diff
@@ -0,0 +1,179 @@
...
+ * report summary
+ * @param response
+ * @param severityThreshold
+ */
+export const computeNpmV2Report: OtterAuditReporter = (response: string, severityThreshold: Severity) => {
+  const reportJson = JSON.parse(response) as NPMAuditReportV2.Audit;
+  core.debug(response);
+  const { nbVulnerabilities, highestSeverityFound } = computeTotalsForSeverities((severity) => reportJson.metadata.vulnerabilities[severity]);
+  return Object.values(reportJson.vulnerabilities)
+    .reduce<OtterAuditReport>((currentVulnerabilities, advisory) =>
```
kpanot, line 99: I think the type is not needed (same in other similar places)
```suggestion
    .reduce((currentVulnerabilities, advisory) =>
```

tools/github-actions/audit/tsconfig.build.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
```
kpanot, line 2: May be better to extend the base one to facilitate upgrades and alignment

tools/github-actions/audit/tsconfig.eslint.json
```diff
@@ -1,6 +1,10 @@
 {
   "extends": "./tsconfig.json",
```
kpanot, line 2:
```suggestion
```
Not needed

tools/github-actions/audit/action.yml
fpaul-1A: I'm not sure we can set `type: choice` and `options` for inputs of a GitHub Action, it seems to be for workflow inputs
correct me if I'm wrong

tools/github-actions/audit/src/main.ts
fpaul-1A: I find this version easily readable

tools/github-actions/audit/action.yml
vscaiceanu-1a: Yup I think you're right

tools/github-actions/audit/src/reports.ts
fpaul-1A: I've updated to `Yarn 2 and 3`

tools/github-actions/audit/src/reports.ts
fpaul-1A: it seems to be needed, I've removed the cast on the acc though

tools/github-actions/audit/src/reports.ts
```diff
@@ -0,0 +1,183 @@
...
+  overview: string;
+  /** Name of the module affected by the vulnerability */
+  moduleName: string;
+}
+
+/**
+ * Data to build the action report summary
+ */
+export interface OtterAuditReport {
+  nbVulnerabilities: number;
```
kpanot, line 33: missing comments

tools/github-actions/audit/src/reports.ts
kpanot: This is because of the removal of the `as OtterAuditReport` in the initial value parameter, which is fine as well

tools/github-actions/audit/src/reports.ts
```diff
@@ -0,0 +1,188 @@
...
+    const severityTotal = severityAccessor(severity);
+    if (!highestSeverityFound && severityTotal > 0) {
+      highestSeverityFound = severity;
+    }
+    nbVulnerabilities += severityTotal;
+  }
+  return { highestSeverityFound, nbVulnerabilities };
+}
+
+function updateReportWithVulnerability(currentReport: OtterAuditReport, severityThreshold: Severity, vulnerability: OtterAdvisory): OtterAuditReport {
```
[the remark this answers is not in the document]
fpaul-1A: it is only needed for Yarn4 because it doesn't provide a summary
The loop to find the highest severity actually loops only over 5 elements, it's not the same loop as the one on the list of vulnerabilities
So no, I don't think there is any reason not to trust the summary and compute the highest severity on our own

matthieu-crouzet (approved):

cpaulve-1A (approved):

PR #2312, opened 2024-10-21
fix(design): create folder with files when does not exist

kpanot:
## Proposed change

fix(design): create folder with files when does not exist
fix(design): better sorting by name with grade

### DIFF
--- a/packages/@o3r/design/src/core/design-token/parsers/design-token.parser.ts
+++ b/packages/@o3r/design/src/core/design-token/parsers/design-token.parser.ts
@@ -1,4 +1,3 @@
-import { promises as fs } from 'node:fs';
 import type { DesignTokenVariableSet, DesignTokenVariableStructure, NodeReference, ParentReference } from './design-token-parser.interface';
 import type {
   DesignToken,
@@ -226,7 +225,7 @@ export const parseDesignToken = (specification: DesignTokenSpecification): Desig
 interface ParseDesignTokenFileOptions {
   /**
    * Custom function to read a file required by the token renderer
-   * @default {@see fs.promises.readFile}
+   * @default {@see import('node:fs/promises').readFile}
    * @param filePath Path to the file to read
    */
   readFile?: (filePath: string) => string | Promise<string>;
@@ -241,7 +240,7 @@ interface ParseDesignTokenFileOptions {
  * @param options
  */
 export const parseDesignTokenFile = async (specificationFilePath: string, options?: ParseDesignTokenFileOptions) => {
-  const readFile = options?.readFile || ((filePath: string) => fs.readFile(filePath, { encoding: 'utf8' }));
+  const readFile = options?.readFile || (async (filePath: string) => (await import('node:fs/promises')).readFile(filePath, { encoding: 'utf8' }));
   const context: DesignTokenContext = {
     basePath: dirname(specificationFilePath),
     ...options?.specificationContext
--- a/packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
+++ b/packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
@@ -1,9 +1,10 @@
 import type { DesignTokenVariableSet, DesignTokenVariableStructure } from '../parsers/design-token-parser.interface';
 import { getCssTokenDefinitionRenderer } from './css/design-token-definition.renderers';
 import { getCssStyleContentUpdater } from './css/design-token-updater.renderers';
-import { existsSync, promises as fs } from 'node:fs';
+import type { Logger } from '@o3r/core';
+import type { promises as fs } from 'node:fs';
 import { isAbsolute, resolve } from 'node:path';
-import type { DesignTokenRendererOptions, TokenListTransform } from './design-token.renderer.interface';
+import type { DesignTokenListTransform, DesignTokenRendererOptions } from './design-token.renderer.interface';
 
 /**
  * Retrieve the function that determines which file to update for a given token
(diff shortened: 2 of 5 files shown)


packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
```diff
@@ -28,16 +29,33 @@ export const compareVariableByName = (a: DesignTokenVariableStructure, b: Design
 
 /**
  * Sort Token variable by name using the local alphabetical order
- * @param _variableSet Complete set of the parsed Design Token
+ * @param _variables Complete set of the parsed Design Token
+ * @param options Parameters of the Design Token list transform function
  */
-export const getTokenSorterByName: TokenListTransform = (_variableSet) => (tokens) => tokens.sort((a, b) => a.getKey().localeCompare(b.getKey()));
+export const getTokenSorterByName: DesignTokenListTransform = (_variables, options) => {
+  const splitNameRegExp = /^(.*?)(\d+)$/;
```
matthieu-crouzet, line 36: Isn't `^(.*)(\d+)$` enough ?
Maybe I just don't understand the `?` here.

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
```diff
@@ -80,6 +98,31 @@ export const getTokenSorterByRef: TokenListTransform = (variableSet) => {
...
 
+/**
+ * Retrieve default file writer (based on Node `fs.promise.writeFile` interface)
+ * @param existsFile Function determining if the file exists
+ * @param logger Custom Logger
+ * @returns
+ */
+export const getDefaultFileWriter = (existsFile: (file: string) => boolean, logger?: Logger): typeof fs.writeFile => {
+  return async (file, ...args) => {
+    // eslint-disable-next-line @typescript-eslint/no-base-to-string
```
matthieu-crouzet, line 109: With the new ESLint config that we prepare we should add a comment to explain why we disable a rule following the format:
`// eslint-disable-next-line <rule-name> -- <explication>`

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
kpanot: `*?` is for non-greedy `*`, cf https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions/Quantifiers

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
```diff
@@ -80,6 +98,31 @@ export const getTokenSorterByRef: TokenListTransform = (variableSet) => {
...
+    const fileString = file.toString();
+    const { writeFile, mkdir } = await import('node:fs/promises');
+    if (!existsFile(fileString)) {
+      const { dirname } = await import('node:path');
+      try {
+        await mkdir(dirname(fileString), { recursive: true });
+      } catch {
+        // ignore folder creation failure
+      }
+    }
```
matthieu-crouzet, line 119: The file may not exist but the folder can 
why not doing 
```suggestion
      const { dirname } = await import('node:path');
      try {
        await mkdir(dirname(fileString), { recursive: true });
      } catch {
        // ignore folder creation failure
      }
```

packages/@o3r/design/src/core/design-token/renderers/design-token.renderer.interface.ts
```diff
@@ -17,10 +17,20 @@ export type DesignContentFileUpdater = (variables: string[], file: string, style
...
+   * Renderer the name of generated variable (without the prefix required by the target language)
+   */
+  tokenVariableNameRenderer?: TokenKeyRenderer;
+}
+
 /**
  * Function defining the way the variable should be sorted before being generated
  */
-export type TokenListTransform = (variableSet: DesignTokenVariableSet) => (tokens: DesignTokenVariableStructure[]) => DesignTokenVariableStructure[];
+export type DesignTokenListTransform = (variableSet: DesignTokenVariableSet, options?: DesignTTokenListTransformOptions) => (tokens: DesignTokenVariableStructure[]) => DesignTokenVariableStructure[];
```
matthieu-crouzet, line 33: Isn't it a breaking change ?

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
matthieu-crouzet: ooooh I learn something new so early today ! 😄

packages/@o3r/design/src/core/design-token/renderers/design-token.renderer.interface.ts
kpanot: Nop, the original `TokenListTransform` is still on **mian** branch

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
fpaul-1A: what about `/^([^\d]*)(\d+)$/` ?

kpanot: > what about `/^([^\d]*)(\d+)$/` ?
not matching the Token `var-1-black-100`

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
```diff
@@ -28,16 +29,33 @@ export const compareVariableByName = (a: DesignTokenVariableStructure, b: Design
 
 /**
  * Sort Token variable by name using the local alphabetical order
- * @param _variableSet Complete set of the parsed Design Token
+ * @param _variables Complete set of the parsed Design Token
+ * @param options Parameters of the Design Token list transform function
  */
-export const getTokenSorterByName: TokenListTransform = (_variableSet) => (tokens) => tokens.sort((a, b) => a.getKey().localeCompare(b.getKey()));
+export const getTokenSorterByName: DesignTokenListTransform = (_variables, options) => {
```
mrednic-1A, line 35: I am a bit confused by the unused parameter _variables. Couldn't it be removed ?

matthieu-crouzet (approved):

packages/@o3r/design/src/core/design-token/parsers/design-token.parser.ts
```diff
@@ -241,7 +240,7 @@ interface ParseDesignTokenFileOptions {
  * @param options
  */
 export const parseDesignTokenFile = async (specificationFilePath: string, options?: ParseDesignTokenFileOptions) => {
-  const readFile = options?.readFile || ((filePath: string) => fs.readFile(filePath, { encoding: 'utf8' }));
+  const readFile = options?.readFile || (async (filePath: string) => (await import('node:fs/promises')).readFile(filePath, { encoding: 'utf8' }));
```
cpaulve-1A, line 243: Why do we need to import it asynchronously?

kpanot: To be usable in non-node environment and remove the deps to it at file level.
Thanks to this change, we are able to replace these functions by some non-node based (and use the Design Token generation in a Chrome extension for example ;) )

fpaul-1A (approved):

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
kpanot: Can not be removed (because of the `options` next parameter that is used).
The only solution is to rename it `_` but it will be ugly in the jsdoc :S

mrednic-1A (approved):

PR #2267, opened 2024-10-10
feat: otter training - training sdk

sdo-1A:
## Proposed change

Creation of training SDK
This PR is mainly to check the package files, the specifications are likely to change when we update the content of the SDK training exercises.

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

PR #2314, opened 2024-10-21
fix: header of sdk readme file in schematics for clarity on versions

sdo-1A:
## Proposed change

Modify the header of the SDK readme file to clarify the version of the SDK spec and the version of OpenAPI used to generate the SDK.

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/readme.md
+++ b/packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/readme.md
@@ -1,6 +1,6 @@
 <% if (projectDescription) { %># <%=projectDescription%><% } %>
 
-### Based on OpenAPI spec 0.0.0
+### Based on API specification version 0.0.0 (using OpenAPI 0.0.0)
 
 The SDK contains 2 different parts:
 
--- a/packages/@ama-sdk/schematics/schematics/typescript/core/index.spec.ts
+++ b/packages/@ama-sdk/schematics/schematics/typescript/core/index.spec.ts
@@ -13,7 +13,7 @@ describe('Typescript Core Generator', () => {
   beforeEach(async () => {
     const runner = new SchematicTestRunner('@ama-sdk/schematics', collectionPath);
     const tree = Tree.empty();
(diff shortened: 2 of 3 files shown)


matthieu-crouzet (approved):

packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/readme.md
```diff
@@ -1,6 +1,6 @@
 <% if (projectDescription) { %># <%=projectDescription%><% } %>
 
-### Based on OpenAPI spec 0.0.0
+### Based on SDK spec version 0.0.0 (using OpenAPI 0.0.0)
```
kpanot, line 3: It is not really `SDK spec` but on `OpenAPI specification`.
The Specification version with not always aligned with the version of the SDK

sdo-1A: The goal was to clarify the versions being used as a follow up to this comment: https://github.com/AmadeusITGroup/otter/pull/2267#discussion_r1795601865 
Currently we get the impression that we are specifying the OpenAPI version which is not the case. 
Do you have a suggestion on how to make this heading more clear?

kpanot: Maybe you can write something like:

```javascript
`Based on specifification version ${specVersion[1] || '0.0.0'}

> The specification is following the standard ${generatorVersion[1]} ${generatorVersion[2]}`
```
(or with a better English :D)

What do you think?

sdo-1A: Good idea, it's more clear that way.
@cpaulve-1A what do you think? (since we implemented this fix together)

cpaulve-1A: We just need to ensure it will work with both on legacy and new message (careful, one is single line and the other one is multiline)

kpanot (approved):

PR #2315, opened 2024-10-21
feat(github-actions-release): get previous release for release note

### LINKED ISSUE #2256: Generate better release note for minor/major releases
### Context

In the current implementation, the release generator step is generating a release note based on the changes from the previous patch.
This is not efficient to generate the release note expected for new minor and major versions.

### Proposal

Calculate the proper version to provide to the `gh release` comment as `from` argument to generate the correct release note

kpanot:
## Proposed change

Generate Release note from the previous version in case of minor/major upgrade

## Related issues

- :rocket: Feature resolves #2256


tools/github-actions/release/action.yml
```diff
@@ -10,6 +10,10 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Previous release tag
+      if:  ${{!contains( inputs.version, '-' )) && endsWith( inputs.version, '0' )}}
+      shell: bash
+      run: echo "PREVIOUS_VERSION_TAG=$(gh release list --exclude-drafts --exclude-pre-releases --json tagName --limit 1 --jq '.[].tagName')" >> $GITHUB_ENV
```
fpaul-1A, line 16: if I understand correctly, this will extract the tag of the latest release published on GitHub (chronologically)
however I fail to understand how this can be the previous tag to use for the generation of the notes regardless of the `inputs.version`

kpanot: Yes, discovered after testing that it was chronological order (I don't know wh I was thinking it was semver ordered).
I am working on a fix

tools/github-actions/release/action.yml
```diff
@@ -7,9 +7,17 @@ inputs:
...
     required: true
+  previousVersionsLimit:
+    description: 'Limit of previous versions to consider for the release notes'
+    required: false
+    default: '100'
 runs:
   using: "composite"
   steps:
+    - name: Previous release tag
+      if:  ${{!contains( inputs.version, '-' )) && endsWith( inputs.version, '0' )}}
```
vscaiceanu-1a, line 18:
```suggestion
      if:  ${{!contains( inputs.version, '-' )) && endsWith( inputs.version, '.0' )}}
```

vscaiceanu-1a, line 18: How does this work for versions ending in `-prerelease.0`?

tools/github-actions/release/action.yml
```diff
@@ -7,9 +7,17 @@ inputs:
   target:
     description: 'The branch to target'
     required: true
+  previousVersionsLimit:
+    description: 'Limit of previous versions to consider for the release notes'
+    required: false
+    default: '100'
```
vscaiceanu-1a, line 13: I think we are missing some details in the documentation of this GH Action

tools/github-actions/release/action.yml
kpanot: the `!contains( inputs.version, '-' ) ` will exclude it

tools/github-actions/release/action.yml
kpanot: There is not details regarding the parameters in the documentation, there is a link referring this file to get the list.
Or did you mean the description is unclear? If so can you suggest one?

nx.json
```diff
@@ -363,7 +363,8 @@
         "{projectRoot}/action.yml",
         "{projectRoot}/LICENSE"
       ],
-      "outputs": ["{projectRoot}/packaged-action"]
+      "outputs": ["{projectRoot}/packaged-action"],
+      "dependsOn": ["build"]
```
fpaul-1A, line 367: is it really needed ?

kpanot: yes and I wonder why it was not the case before.
the `ncc` command build what is targeted by the `package.json`'s `main` field.
Currently, in our actions, this field is targeting the `tmp/index.js` which is the result of the `tsc` command triggered by the `build` (or `compile` and `github-action-compile`) task.

cpaulve-1A (dismissed):

tools/github-actions/release/action.yml
```diff
@@ -7,9 +7,17 @@ inputs:
...
 runs:
   using: "composite"
   steps:
+    - name: Previous release tag
+      if:  ${{(!contains( inputs.version, '-' )) && endsWith( inputs.version, '.0' )}}
+      shell: bash
+      run: echo "PREVIOUS_VERSION_TAG=$(gh release list --exclude-drafts --exclude-pre-releases --limit ${{ inputs.previousVersionsLimit }} --json tagName --template '[{{range .}}\"{{.tagName}}\",{{end}}""]{{\"\n\"}}' | node ./packaged-action/index.cjs ${{ inputs.version }})" >> $GITHUB_ENV
```
fpaul-1A, line 20:
```suggestion
      run: echo "PREVIOUS_VERSION_TAG=$(gh release list --exclude-drafts --exclude-pre-releases --limit ${{ inputs.previousVersionsLimit }} --json tagName --template '[{{range .}}\"{{.tagName}}\", \"{{end}}\"]{{\"\n\"}}' | node ./packaged-action/index.cjs ${{ inputs.version }})" >> $GITHUB_ENV
```

tools/github-actions/release/action.yml
vscaiceanu-1a: Since it is a particular behaviour for `.0` releases, I was thinking about adding some words in [the readme of the action](https://github.com/AmadeusITGroup/otter/blob/main/tools/github-actions/release/readme.md) explaining how the release notes is computed in case of `.0` releases.

tools/github-actions/release/action.yml
kpanot:
```suggestion
      run: echo "PREVIOUS_VERSION_TAG=$(gh release list --exclude-drafts --exclude-pre-releases --limit ${{ inputs.previousVersionsLimit }} --json tagName --template '[{{range .}}\"{{.tagName}}\",{{end}}\"\"]{{\"\n\"}}' | node ./packaged-action/index.cjs ${{ inputs.version }})" >> $GITHUB_ENV
```

Actually

vscaiceanu-1a (approved):

tools/github-actions/release/action.yml
```diff
@@ -7,9 +7,17 @@ inputs:
...
+    description: 'Limit of previous versions to consider for the release notes'
+    required: false
+    default: '100'
 runs:
   using: "composite"
   steps:
+    - name: Previous release tag
+      if:  ${{(!contains( inputs.version, '-' )) && endsWith( inputs.version, '.0' )}}
+      shell: bash
+      run: echo "PREVIOUS_VERSION_TAG=$(gh release list --exclude-drafts --exclude-pre-releases --limit ${{ inputs.previousVersionsLimit }} --json tagName --template '[{{range .}}"{{.tagName}}",{{end}}""]{{"\n"}}' | node ./packaged-action/index.cjs ${{ inputs.version }})" >> $GITHUB_ENV
```
fpaul-1A, line 20: are you sure the double quotes are kept in the output ?
I got this in local
image
image

kpanot: You forgot the `"` of the `echo` command, please try with the following command:

```bash
echo "PREVIOUS_VERSION_TAG=$(gh release list --exclude-drafts --exclude-pre-releases --limit 100 --json tagName --template '[{{range .}}"{{.tagName}}",{{end}}""]{{"\n"}}' | node ./tools/github-actions/release/packaged-action/index.cjs 11.2.0)"
```

The output (on Mac/Linux) is :

```
PREVIOUS_VERSION_TAG=v11.1.6
```

I would be interested to have an output on Windows, it may be different

fpaul-1A: my bad, I didn't run the `| node` part and got confused, but there is no problem if we pass the output to node before setting the variable

fpaul-1A (approved):

PR #2322, opened 2024-10-22
fix(eslint): manual fix of eslint errors on vscode-extension

matthieu-crouzet:


apps/vscode-extension/package.json
```diff
@@ -344,7 +344,9 @@
   },
   "dependencies": {
     "@o3r/design": "workspace:^",
-    "eslint": "^9.0.0"
+    "eslint": "^9.0.0",
+    "@o3r/eslint-plugin": "workspace:^",
```
fpaul-1A, line 348: there is no rule to put those in alphabetical order ?

matthieu-crouzet: Good catch it should not be in dependencies ! :)
Running yarn install should order them 
but in my case the linter added without my consent 😄

apps/vscode-extension/src/commands/generate/fixture.command.ts
```diff
@@ -25,7 +25,6 @@ const isHTMLTemplateDocument = (doc: vscode.TextDocument | undefined) => {
 /**
  * Generate component command
  * @param _context
- * @returns
```
vscaiceanu-1a, line 28: I am a bit surprised that this is a manual fix and not an automatic one.

apps/vscode-extension/src/commands/generate/store.command.ts
```diff
@@ -71,7 +70,7 @@ export function generateStoreGenerateCommand(_context: ExtensionContext, folder?
       modelIdPropName = 'id';
     }
 
-    const defaultOptions = await getSchematicDefaultOptions('@o3r/core:service');
+    const defaultOptions = await getSchematicDefaultOptions('@o3r/core:service') as Partial<{ path: string }>;
```
vscaiceanu-1a, line 73:
```suggestion
    const defaultOptions = await getSchematicDefaultOptions('@o3r/core:store') as Partial<{ path: string }>;
```

apps/vscode-extension/src/intellisense/configuration.ts
```diff
@@ -1,5 +1,12 @@
+import type {
+  O3rCategoriesTagsRuleOption,
+  O3rWidgetTagsRuleOption
+} from '@o3r/eslint-plugin';
+import type {
+  TSESLint,
+} from '@typescript-eslint/utils';
 import {
-  ESLint
+  ESLint,
```
fpaul-1A, line 9: trailing comma is on ?

packages/@o3r/eslint-plugin/src/public_api.ts
```diff
@@ -1 +1,6 @@
 export * from './index';
+export type * from './rules/typescript/matching-configuration-name/matching-configuration-name';
```
vscaiceanu-1a, line 2: where are these types used?

apps/vscode-extension/src/commands/generate/fixture.command.ts
matthieu-crouzet: because there is a rule that ask you to put a description if you put the `@returns`

packages/@o3r/eslint-plugin/src/public_api.ts
matthieu-crouzet: in the vscode extension

apps/vscode-extension/src/commands/generate/module.command.ts
```diff
@@ -13,18 +13,17 @@ import {
   stringifyOptions
 } from '../helpers';
 
+const getCurrentFolder = () => {
```
vscaiceanu-1a, line 16: You can use the method with the same name from `apps/vscode-extension/src/commands/helpers.ts`.
It has the exact same implementation.

apps/vscode-extension/src/intellisense/design-token.ts
```diff
@@ -18,20 +18,25 @@ type DesignTokenCache = {
...
-  const configFiles = (await Promise.all([angularFile, nxFile, projectFiles])).flat();
+  const configFilesArray = await Promise.all([angularFile, nxFile, projectFiles]);
+  const configFiles = configFilesArray.flat();
   if (configFiles.length === 0) {
     return;
   }
 
   const walkThroughProjectFiles = (node: any, mem: Set<string> = new Set()) => {
     if (typeof node === 'object' && node !== null && !Array.isArray(node)) {
+      /* eslint-disable @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-argument -- type is checked just above */
```
fpaul-1A, line 39: the type is not really checked, the problem is probably that the input has an `any` type

apps/vscode-extension/src/intellisense/design-token.ts
```diff
@@ -18,20 +18,25 @@ type DesignTokenCache = {
   tokenStructure: DesignTokenVariableSet;
 };
 
-const readFile = async (filePat: string) => (await vscode.workspace.fs.readFile(vscode.Uri.parse(filePat))).toString();
```
fpaul-1A, line 21: is this forced by a linter rule ?

matthieu-crouzet: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v55.0.0/docs/rules/no-await-expression-member.md

matthieu-crouzet: fallback change for this rule 
as we will discuss it later

fpaul-1A (approved):

fpaul-1A (approved):

kpanot (approved):

PR #2324, opened 2024-10-22
fix: do not dynamically import @o3r/rules-engine during rules-engine install

### LINKED ISSUE #2334: [Bug]: ng add @o3r/schematics command fails
### Package name

schematics

### Package version

11.2.*

### Reproduction steps

Create a new project without referring to @o3r/create, you can use for example the angular project create:
`npm create @angular test-app`
Try to run `npm run ng add @o3r/core` or `npm run ng add @o3r/schematics`

### Current result

379267778-3d01a5e2-1c00-41f5-a95e-c3ac0ffa564f
@o3r/schematics failes to access to its own code

### Expected result

ng add does not fail 

### Additional comments

Running twice the ng add @o3r/core command seems to be a workaround

cpaulve-1A:
## Proposed change
@o3r/schematics is not fully installed and not accessible during ng-add
schematics run. A relative import allows to access the library during
the installation process.

The use of composite in tsconfig.builders.json (current setup) allow to make reference
to the dist/src folder without risking the override of the src build.
## Related issues

- :bug: Fixes #2334

### DIFF
--- a/packages/@o3r/schematics/schematics/ng-add/index.ts
+++ b/packages/@o3r/schematics/schematics/ng-add/index.ts
@@ -1,5 +1,5 @@
 import type { Rule } from '@angular-devkit/schematics';
-import type { DependencyToAdd } from '@o3r/schematics';
+import { createSchematicWithMetricsIfInstalled, type DependencyToAdd, getExternalDependenciesVersionRange, setupDependencies } from '../../src/public_api';
 import { NodeDependencyType } from '@schematics/angular/utility/dependencies';
 import * as path from 'node:path';
 import type { NgAddSchematicsSchema } from './schema';
@@ -10,9 +10,8 @@ import type { NgAddSchematicsSchema } from './schema';
  */
 function ngAddFn(options: NgAddSchematicsSchema): Rule {
   const schematicsDependencies = ['@angular-devkit/architect', '@angular-devkit/schematics', '@angular-devkit/core', '@schematics/angular', 'globby'];
-  return async (_, context) => {
+  return (_, context) => {
     const packageJsonPath = path.resolve(__dirname, '..', '..', 'package.json');
-    const { getExternalDependenciesVersionRange, setupDependencies } = await import('@o3r/schematics');
     const dependencies = Object.entries(getExternalDependenciesVersionRange(schematicsDependencies, packageJsonPath, context.logger)).reduce((acc, [dep, range]) => {
       acc[dep] = {
         inManifest: [{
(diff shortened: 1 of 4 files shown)


tools/github-actions/new-version/packaged-action/index.js
cpaulve-1A, line 1: remove this file and the one after

packages/@o3r/schematics/schematics/ng-add/index.ts
```diff
@@ -12,7 +12,7 @@ function ngAddFn(options: NgAddSchematicsSchema): Rule {
   const schematicsDependencies = ['@angular-devkit/architect', '@angular-devkit/schematics', '@angular-devkit/core', '@schematics/angular', 'globby'];
   return async (_, context) => {
     const packageJsonPath = path.resolve(__dirname, '..', '..', 'package.json');
-    const { getExternalDependenciesVersionRange, setupDependencies } = await import('@o3r/schematics');
+    const { getExternalDependenciesVersionRange, setupDependencies } = await import('../../src/public_api');
```
kpanot, line 15: What issue is fixing this change?
Does it still need to be dynamic?

cpaulve-1A: Hi Kilian, this is a draft to test if it could fix the following error:
image

It is still WIP ;)

mrednic-1A: usually when we are in `schematics/builders` and we need to import from the package itself, we use the name of the package instead of the relative path. This is because the builds are done separately and the build of `schematics/builders` is usually after the build of `src` and it can override the output of `src` if the import is relative.
If we do this change we need to check that the output of the `src` build is not broken

cpaulve-1A: I will take it into account but even the ng add @o3r/schematics does not work today:
image

cpaulve-1A: @mrednic-1A I just tested - the schematics build is composite and won't modify the result of the compile task if it is already available 
If I run the compile, then modify the dist/src/index and run only the build-builder, I keep my modifications.
So I think it would be okay

matthieu-crouzet (dismissed):

fpaul-1A (approved):

mrednic-1A (approved):

packages/@o3r/schematics/schematics/ng-add/index.ts
kpanot: I think we can conclude it reply to the first question.
I think we are still missing the second:

> Does it still need to be dynamic?

kpanot (approved):

fpaul-1A (approved):

kpanot (approved):

PR #2326, opened 2024-10-23
fix(pipeline): missing dependencies

kpanot:
## Proposed change

fix(pipeline): missing dependencies
chore: update readme links


tools/github-actions/audit/packaged-action/LICENSE.txt
cpaulve-1A, line 1: why are these files impacted?

kpanot: because of the touch of `yarn.lock` which is part of the `namedInput: global` so trigger the `affected`

fpaul-1A (approved):

mrednic-1A (approved):

PR #2330, opened 2024-10-23
fix(eslint): manual fix of eslint errors on ama-sdk

matthieu-crouzet:
all clients packages are not handled here because not present since the rebase has not been done since a while

### DIFF
--- a/packages/@ama-sdk/core/src/clients/api-angular-client.ts
+++ b/packages/@ama-sdk/core/src/clients/api-angular-client.ts
@@ -165,13 +165,14 @@ export class ApiAngularClient implements ApiClient {
 
         const subscription = httpRequest.subscribe({
           next: (res) => data = res,
+          // eslint-disable-next-line @typescript-eslint/prefer-promise-reject-errors -- subscription forwards the error from the httpRequest to asyncResponse promise
           error: (err) => reject(err),
           complete: () => resolve(data)
         });
         metadataSignal?.throwIfAborted();
         metadataSignal?.addEventListener('abort', () => {
           subscription.unsubscribe();
-          reject(metadataSignal.reason);
+          reject(metadataSignal.reason instanceof Error ? metadataSignal.reason : new Error(metadataSignal.reason.toString()));
         });
       });
       response = await asyncResponse;
@@ -180,7 +181,7 @@ export class ApiAngularClient implements ApiClient {
       exception = new EmptyResponseError(e.message || 'Fail to Fetch', undefined, { apiName, operationId, url, origin });
     }
 
-    // eslint-disable-next-line no-console
+    // eslint-disable-next-line no-console -- `console.error` is supposed to be the default value if the `options` argument is not provided, can be removed in Otter v12.
     const reviver = getResponseReviver(revivers, response, operationId, { disableFallback: this.options.disableFallback, log: console.error });
     const replyPlugins = this.options.replyPlugins
       ? this.options.replyPlugins.map((plugin) => plugin.load<T>({
--- a/packages/@ama-sdk/core/src/clients/api-fetch-client.ts
+++ b/packages/@ama-sdk/core/src/clients/api-fetch-client.ts
@@ -162,12 +162,10 @@ export class ApiFetchClient implements ApiClient {
 
       const canStart = await Promise.all(loadedPlugins.map((plugin) => !plugin.canStart || plugin.canStart()));
       const isCanceledBy = canStart.indexOf(false);
-      if (isCanceledBy >= 0) {
+      asyncResponse = isCanceledBy >= 0
         // One of the fetch plugins cancelled the execution of the call
-        asyncResponse = Promise.reject(new CanceledCallError(`Is canceled by the plugin ${isCanceledBy}`, isCanceledBy, this.options.fetchPlugins[isCanceledBy], { apiName, operationId, url, origin }));
-      } else {
-        asyncResponse = fetch(url, options);
-      }
+        ? Promise.reject(new CanceledCallError(`Is canceled by the plugin ${isCanceledBy}`, isCanceledBy, this.options.fetchPlugins[isCanceledBy], { apiName, operationId, url, origin }))
+        : fetch(url, options);
 
       for (const plugin of loadedPlugins) {
         asyncResponse = plugin.transform(asyncResponse);
@@ -183,9 +181,10 @@ export class ApiFetchClient implements ApiClient {
     try {
       root = body ? JSON.parse(body) : undefined;
     } catch (e: any) {
-      exception = new ResponseJSONParseError(e.message || 'Fail to parse response body', response && response.status || 0, body, { apiName, operationId, url, origin });
+      exception = new ResponseJSONParseError(e.message || 'Fail to parse response body', (response && response.status) || 0, body, { apiName, operationId, url, origin });
     }
-    // eslint-disable-next-line no-console
+
+    // eslint-disable-next-line no-console -- `console.error` is supposed to be the default value if the `options` argument is not provided, can be removed in Otter v12.
     const reviver = getResponseReviver(revivers, response, operationId, { disableFallback: this.options.disableFallback, log: console.error });
     const replyPlugins = this.options.replyPlugins
       ? this.options.replyPlugins.map((plugin) => plugin.load<T>({
--- a/packages/@ama-sdk/core/src/fwk/Reviver.ts
+++ b/packages/@ama-sdk/core/src/fwk/Reviver.ts
@@ -33,6 +33,7 @@ export type ReviverType<T, V extends { [key: string]: any } = { [key: string]: a
 /**
  * Used in case of maps (dictionaries): All values of the map must be of the same type. reviveWithType will be called
  * for each of these elements.
+ * @param data
  * @param dictionaries
  * @param reviver
  * @param options Reviver options
@@ -44,8 +45,7 @@ export function reviveMap<T>(data: { [key: string]: any }, dictionaries: any = n
 
   const revived: { [key: string]: any } = {};
   for (const key in data) {
-    // eslint-disable-next-line no-prototype-builtins
-    if (data.hasOwnProperty(key)) {
+    if (Object.prototype.hasOwnProperty.call(data, key)) {
       revived[key] = reviver(data[key], dictionaries, options);
     }
   }
--- a/packages/@ama-sdk/core/src/fwk/api.helpers.ts
+++ b/packages/@ama-sdk/core/src/fwk/api.helpers.ts
@@ -57,27 +57,17 @@ export function filterUndefinedValues(object?: { [key: string]: string | undefin
  * Encodes this object to match application/x-www-urlencoded or multipart/form-data
  * @param data
  * @param type
- * @param data
- * @param type
  */
 export function processFormData(data: any, type: string): FormData | string {
   let encodedData: FormData | string;
 
-  /* eslint-disable guard-for-in */
   if (type === 'multipart/form-data') {
     const formData: FormData = new FormData();
-    for (const i in data) {
-      formData.append(i, data[i]);
-    }
+    Object.entries(data).forEach(([key, value]) => formData.append(key, value as any));
     encodedData = formData;
   } else {
-    const formData: string[] = [];
-    for (const i in data) {
-      formData.push(`${i}=${encodeURIComponent(data[i])}`);
-    }
-    encodedData = formData.join('&');
+    encodedData = Object.entries(data).map(([key, value]) => `${key}=${encodeURIComponent(value as any)}`).join('&');
   }
-  /* eslint-enable guard-for-in */
 
   return encodedData;
 }
@@ -121,17 +111,21 @@ export function tokenizeRequestOptions(tokenizedUrl: string, queryParameters: { 
  * Fallback to the lowest status code's reviver.
  * Does not try to match non-successful responses as error are handled separately
  * @param revivers
- * @param endpoint
  * @param response
- * @param options
- * @param options.disableFallback
- * @param options.log
+ * @param endpoint
+ * @param options `{ disableFallback: false, log: console.error }` by default
+ * @param options.disableFallback `false` by default
+ * @param options.log `() => {}` by default -- warning: default value will change to `console.error` in Otter v12.
  */
-export function getResponseReviver<T>(revivers: { [statusCode: number]: ReviverType<T> | undefined } | undefined | ReviverType<T>, response: Pick<Response, 'ok' | 'status'> | undefined,
-  // eslint-disable-next-line no-console
-  endpoint?: string, options: { disableFallback?: boolean; log?: (...args: any[]) => void } = { disableFallback: false, log: console.error }): ReviverType<T> | undefined {
+export function getResponseReviver<T>(
+  revivers: { [statusCode: number]: ReviverType<T> | undefined } | undefined | ReviverType<T>,
+  response: Pick<Response, 'ok' | 'status'> | undefined,
+  endpoint?: string,
+  options?: { disableFallback?: boolean; log?: (...args: any[]) => void }
+): ReviverType<T> | undefined {
+  // eslint-disable-next-line no-console -- set as default value
+  const { disableFallback = false, log: logMsg = options ? () => {} : console.error } = options ?? {};
   const logPrefix = `API status code error for ${endpoint || 'unknown'} endpoint`;
-  const logMsg = options.log || (() => {});
   if (!response || !response.ok) {
     return undefined;
   }
@@ -141,7 +135,7 @@ export function getResponseReviver<T>(revivers: { [statusCode: number]: ReviverT
   if (response.status && Object.keys(revivers).includes(`${response.status}`)) {
     return revivers[response.status];
   }
-  if (options?.disableFallback) {
+  if (disableFallback) {
     logMsg(`${logPrefix} - Missing ${response.status} from API specification - fallback is deactivated, no revive will run on this response`);
     return undefined;
   }
--- a/packages/@ama-sdk/core/src/fwk/core/api-client.ts
+++ b/packages/@ama-sdk/core/src/fwk/core/api-client.ts
@@ -70,7 +70,7 @@ export interface ApiClient {
   /**
    * Returns tokenized request options:
    * URL/query parameters for which sensitive parameters are replaced by tokens and the corresponding token-value associations
-   * @param tokenizedUrl URL for which parameters containing PII have been replaced by tokens
+   * @param url URL for which parameters containing PII have been replaced by tokens
    * @param queryParameters Original query parameters
    * @param piiParamTokens Tokens of the parameters containing PII
    * @param data Data to provide to the API call
@@ -94,11 +94,12 @@ export interface ApiClient {
  * @param client object to check
  */
 export function isApiClient(client: any): client is ApiClient {
-  return client
-    && !!client.options
-    && typeof client.extractQueryParams === 'function'
-    && typeof client.getRequestOptions === 'function'
-    && typeof client.prepareUrl === 'function'
-    && typeof client.processFormData === 'function'
-    && typeof client.processCall === 'function';
+  const apiClient: ApiClient | undefined = client;
+  return !!apiClient
+    && !!apiClient.options
+    && typeof apiClient.extractQueryParams === 'function'
+    && typeof apiClient.getRequestOptions === 'function'
+    && typeof apiClient.prepareUrl === 'function'
+    && typeof apiClient.processFormData === 'function'
+    && typeof apiClient.processCall === 'function';
 }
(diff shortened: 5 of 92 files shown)


packages/@ama-sdk/core/src/fwk/Reviver.ts
```diff
@@ -44,8 +45,7 @@ export function reviveMap<T>(data: { [key: string]: any }, dictionaries: any = n
 
   const revived: { [key: string]: any } = {};
   for (const key in data) {
-    // eslint-disable-next-line no-prototype-builtins
-    if (data.hasOwnProperty(key)) {
+    if (Object.prototype.hasOwnProperty.call(data, key)) {
```
fpaul-1A, line 48:
```suggestion
    if (key in data) {
```
is it equivalent ?

packages/@ama-sdk/core/src/fwk/date.spec.ts
```diff
@@ -94,6 +95,7 @@ describe('Date', () => {
 
   it('should support json conversions', () => {
     const originalDate = new utils.Date(new Date());
+    // eslint-disable-next-line unicorn/prefer-structured-clone -- we test the JSON conversion
```
fpaul-1A, line 98: maybe we should have a different eslint config for test files ?

packages/@ama-sdk/core/src/fwk/date.ts
```diff
@@ -1,46 +1,42 @@
...
  * @param digits minimal number of digits to display
  */
-export function pad(val: number, digits: number = 2): string {
+export function pad(val: number, digits = 2): string {
   const str = `${val}`;
   return '0'.repeat(Math.max(0, digits - str.length)) + str;
 }
 
+/**
+ * Removes timezone information from ISO8601 strings (Received from DAPI)
```
fpaul-1A, line 15:
```suggestion
 * Removes timezone information from ISO8601 strings
```

packages/@ama-sdk/core/src/fwk/date.ts
```diff
@@ -110,16 +100,10 @@ export namespace utils {
...
 
   export class DateTime extends CommonDate {
-
-    constructor()
-    constructor(value: number)
-    constructor(value: string)
-    constructor(date: _NativeDate)
-    constructor(dateTimeUtils: DateTime)
-    constructor(dateUtils: Date)
-    constructor(otherDate: any)
```
fpaul-1A, line 120: breaking ?

matthieu-crouzet: I still have a definition matching all of them 
I just regroup them

packages/@ama-sdk/core/src/fwk/date.spec.ts
matthieu-crouzet: Maybe but not for this use case
I think it's a particular case we want explicitly do JSON conversion
even in test we can use structuredClone instead of cloning with JSON conversion

packages/@ama-sdk/core/src/clients/api-angular-client.ts
```diff
@@ -165,13 +165,14 @@ export class ApiAngularClient implements ApiClient {
 
         const subscription = httpRequest.subscribe({
           next: (res) => data = res,
+          // eslint-disable-next-line @typescript-eslint/prefer-promise-reject-errors -- forwarding error from subscribe
```
cpaulve-1A, line 168:
```suggestion
          // eslint-disable-next-line @typescript-eslint/prefer-promise-reject-errors -- forwarding error to subscribe
```

packages/@ama-sdk/core/src/plugins/custom-info/custom-info.reply.ts
```diff
@@ -18,15 +18,18 @@ export class CustomInfoReply<G = { [key: string]: any }, V extends { [key: strin
    * Checks if the reply has a customInfo node
    * @param reply
    */
-  // eslint-disable-next-line no-use-before-define
-  public hasCustomInfo<T extends { [key: string]: any; customInfo?: K } | undefined = { [key: string]: any; customInfo: any },
-    K = { [key: string]: any }>(reply: T): reply is T & { customInfo: K & G } {
+  public hasCustomInfo<
+    // eslint-disable-next-line no-use-before-define -- order has no importance for typings
```
fpaul-1A, line 22: The order may have an importance when it comes to generic type default values
For example, if we want the default type to be `customInfo: T` instead of any, we cannot do that:
```
<
  T extends { [key: string]: any; customInfo?: K } | undefined = { [key: string]: any; customInfo: K},
  K = { [key: string]: any }
>
```
we could however do it the other way around
```
<
  K = { [key: string]: any },
  T extends { [key: string]: any; customInfo?: K } | undefined = { [key: string]: any; customInfo: K}
>
```
changing the order would be a breaking change tough
I'm ok to disable the rule here, just not sure about the justification

matthieu-crouzet: Could you propose a new sentence for the justification then please ?

packages/@ama-sdk/core/src/plugins/json-token/json-token.spec.ts
```diff
@@ -42,7 +42,11 @@ describe('Json Token', () => {
       const result = await runner.transform(options);
 
       if (typeof window !== 'undefined' && typeof window.sessionStorage !== 'undefined') {
+        /* eslint-disable jest/no-conditional-expect -- the same condition is verify above to setup the checked spy */
```
fpaul-1A, line 45:
```suggestion
        /* eslint-disable jest/no-conditional-expect -- the same condition is verified above to set up the checked spy */
```

packages/@ama-sdk/core/src/fwk/date.ts
fpaul-1A: you no longer have a definition that matches `constructor(otherDate: any)`

fpaul-1A: nevermind `constructor(...args: any[])` should match

packages/@ama-sdk/schematics/cli/clear-index.cts
```diff
@@ -14,36 +14,26 @@ import {
...
-
-  await Promise.all(
-    shouldRemoveModels
-      .map((model) => {
-        console.warn(`Removing ${model} model`);
-        return fs.unlink(resolve(baseDir, model, 'index.ts'));
-      })
-  );
+  const directories = models.filter((file) => statSync(resolve(baseDir, file)).isDirectory());
+  for (const model of directories) {
```
fpaul-1A, line 32: this is no longer parallelized

packages/@ama-sdk/schematics/cli/update-spec-from-npm.cts
```diff
@@ -44,8 +44,7 @@ const DEFAULT_SPEC_EXPORT_PATH_IN_NPM_MODULE = 'openapi';
 const supportedExtensions = [SPEC_JSON_EXTENSION, SPEC_YAML_EXTENSION, SPEC_YML_EXTENSION];
 
 if (help) {
-  // eslint-disable-next-line no-console
-  console.log(`This script can be used to update your local spec file from a given locally installed npm package.
+  logger.log(`This script can be used to update your local spec file from a given locally installed npm package.
```
fpaul-1A, line 47: This might be stupid but with this version the command `update-spec-from-npm --help --quiet` will log nothing

packages/@ama-sdk/schematics/schematics/helpers/tree-glob.ts
```diff
@@ -15,9 +15,9 @@ const walkThroughDir = (
   parents = ''
 ): string[] => {
   const folders = dir.subdirs.map((d) => `${parents}/${d}`);
-  const matchingFolders = folders.filter(filterFunction);
+  const matchingFolders = folders.filter((value, index, array) => filterFunction(value, index, array));
```
fpaul-1A, line 18: is that needed ?

packages/@ama-sdk/schematics/schematics/ng-update/index.ts
```diff
@@ -13,10 +11,12 @@ import {
   updateV10_3 as tsUpdateV10_3,
   updateV11_0 as tsUpdateV11_0
 } from './typescript';
+/* eslint-enable camelcase */
```
fpaul-1A, line 14: maybe we can keep it disabled for the whole file ?

packages/@ama-sdk/core/src/fwk/core/api-client.ts
```diff
@@ -94,11 +94,11 @@ export interface ApiClient {
...
-    && typeof client.prepareUrl === 'function'
-    && typeof client.processFormData === 'function'
-    && typeof client.processCall === 'function';
+  return !!client
+    && !!(client as ApiClient).options
+    && typeof (client as ApiClient).extractQueryParams === 'function'
+    && typeof (client as ApiClient).getRequestOptions === 'function'
+    && typeof (client as ApiClient).prepareUrl === 'function'
+    && typeof (client as ApiClient).processFormData === 'function'
+    && typeof (client as ApiClient).processCall === 'function';
```
kpanot, line 103:
```suggestion
  const apiClient: ApiClient | undefined = client;
  return !!apiClient
    && !!apiClient.options
    && typeof apiClient.extractQueryParams === 'function'
    && typeof apiClient.getRequestOptions === 'function'
    && typeof apiClient.prepareUrl === 'function'
    && typeof apiClient.processFormData === 'function'
    && typeof apiClient.processCall === 'function';
```

packages/@ama-sdk/core/src/plugins/timeout/timeout.fetch.ts
```diff
@@ -34,8 +34,8 @@ export type TimeoutStatus = 'timeoutStopped' | 'timeoutStarted';
  */
 function isImpervaCaptchaMessage(message: any): message is ImpervaCaptchaMessageData {
   return !!message && Object.prototype.hasOwnProperty.call(message, 'impervaChallenge')
```
kpanot, line 36:
```suggestion
  const impervaCaptchaMessage: ImpervaCaptchaMessageData | undefined = message;
  return !!impervaCaptchaMessage && Object.prototype.hasOwnProperty.call(impervaCaptchaMessage, 'impervaChallenge')
    && Object.prototype.hasOwnProperty.call(impervaCaptchaMessage.impervaChallenge, 'status')
    && Object.prototype.hasOwnProperty.call(impervaCaptchaMessage.impervaChallenge, 'type')
    && impervaCaptchaMessage.impervaChallenge.type === 'captcha';
```

packages/@ama-sdk/schematics/cli/clear-index.cts
```diff
@@ -14,36 +14,26 @@ import {
 import * as minimist from 'minimist';
 
 const argv = minimist(process.argv.slice(2));
-const { help } = argv;
+const { help, quiet } = argv;
 const baseDir = resolve(process.cwd(), 'src', 'models', 'base');
+const noop = () => undefined;
```
kpanot, line 19:
```suggestion
const noop = () => {};
```
not exactly the same type: `() => undefined` vs `() => void`

packages/@ama-sdk/schematics/cli/files-pack.cts
```diff
@@ -18,16 +18,18 @@ import type {
 
 const argv = minimist(process.argv.slice(2));
 const distFolder = argv.dist || 'dist';
-const baseDir = argv.cwd && path.resolve(process.cwd(), argv.cwd) || process.cwd();
-const { help, watch, noExports } = argv;
+const baseDir = (argv.cwd && path.resolve(process.cwd(), argv.cwd)) || process.cwd();
+const { help, watch, noExports, quiet } = argv;
+const noop = () => undefined;
```
kpanot, line 23:
```suggestion
const noop = () => {};
```
not exactly the same type: `() => undefined` vs `() => void`

packages/@ama-sdk/schematics/schematics/typescript/core/index.ts
```diff
@@ -225,7 +226,13 @@ function ngGenerateTypescriptSDKFn(options: NgGenerateTypescriptSDKCoreSchematic
     };
 
     const generateOperationFinder = async (): Promise<PathObject[]> => {
-      const specification: any = isJson ? tree.readJson(specDefaultPath) : (await import('js-yaml')).load(tree.readText(specDefaultPath));
+      let specification;
```
kpanot, line 229: we do not have a rule to avoid implicit `any`?

packages/@ama-sdk/swagger-builder/src/core/swagger-spec-merger.ts
```diff
@@ -130,7 +130,6 @@ export class SwaggerSpecMerger {
           break;
         }
 
-        case 'paths':
```
kpanot, line 133: I think we should, in contrary, have a rule to enforce expliciting all the possible cases of a switch case

packages/@o3r/eslint-config/src/rules/typescript/eslint.cjs
```diff
@@ -105,12 +105,6 @@ const config = [
       ],
       'no-unreachable': 'error',
       'no-unsafe-negation': 'error',
-      'no-use-before-define': [
```
kpanot, line 108: does it means that it is `error` without exception or `off`?

packages/@o3r/eslint-config/src/rules/typescript/unicorn.cjs
```diff
@@ -26,6 +26,7 @@ const config = [
       'unicorn/no-array-reduce': 'off',
       'unicorn/no-null': 'off',
       'unicorn/no-typeof-undefined': 'off',
+      'unicorn/prefer-code-point': 'off',
```
kpanot, line 29: why is it disabled?

packages/@ama-sdk/core/src/fwk/api.helpers.ts
```diff
@@ -121,17 +111,21 @@ export function tokenizeRequestOptions(tokenizedUrl: string, queryParameters: {
...
-  // eslint-disable-next-line no-console
-  endpoint?: string, options: { disableFallback?: boolean; log?: (...args: any[]) => void } = { disableFallback: false, log: console.error }): ReviverType<T> | undefined {
+export function getResponseReviver<T>(
+  revivers: { [statusCode: number]: ReviverType<T> | undefined } | undefined | ReviverType<T>,
+  response: Pick<Response, 'ok' | 'status'> | undefined,
+  endpoint?: string,
+  options?: { disableFallback?: boolean; log?: (...args: any[]) => void }
+): ReviverType<T> | undefined {
+  // eslint-disable-next-line no-console -- set as default value
+  const { disableFallback = false, log: logMsg = console.error } = options ?? {};
```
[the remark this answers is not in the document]
matthieu-crouzet: `{ log: () => {} } `

packages/@ama-sdk/schematics/cli/update-spec-from-npm.cts
matthieu-crouzet: I can fix it

packages/@ama-sdk/schematics/schematics/helpers/tree-glob.ts
matthieu-crouzet: yes because `<array>.filter` can take up to 4 arguments

packages/@ama-sdk/schematics/schematics/ng-update/index.ts
matthieu-crouzet: Yes, as the team prefer

packages/@ama-sdk/core/src/clients/api-fetch-client.ts
```diff
@@ -162,12 +162,10 @@ export class ApiFetchClient implements ApiClient {
 
       const canStart = await Promise.all(loadedPlugins.map((plugin) => !plugin.canStart || plugin.canStart()));
       const isCanceledBy = canStart.indexOf(false);
-      if (isCanceledBy >= 0) {
+      asyncResponse = isCanceledBy >= 0
```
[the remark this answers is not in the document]
matthieu-crouzet: [unicorn/prefer-ternary](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-ternary.md)

packages/@ama-sdk/core/src/clients/api-fetch-client.ts
```diff
@@ -183,10 +181,10 @@ export class ApiFetchClient implements ApiClient {
     try {
       root = body ? JSON.parse(body) : undefined;
     } catch (e: any) {
-      exception = new ResponseJSONParseError(e.message || 'Fail to parse response body', response && response.status || 0, body, { apiName, operationId, url, origin });
+      exception = new ResponseJSONParseError(e.message || 'Fail to parse response body', (response && response.status) || 0, body, { apiName, operationId, url, origin });
     }
-    // eslint-disable-next-line no-console
-    const reviver = getResponseReviver(revivers, response, operationId, { disableFallback: this.options.disableFallback, log: console.error });
+
+    const reviver = getResponseReviver(revivers, response, operationId, { disableFallback: this.options.disableFallback });
```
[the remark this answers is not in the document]
matthieu-crouzet: `console.error` is already the default value

packages/@ama-sdk/schematics/schematics/typescript/core/index.ts
matthieu-crouzet: As discussed earlier we set them as `warn` and they will be fixed later

packages/@ama-sdk/swagger-builder/src/core/swagger-spec-merger.ts
matthieu-crouzet: [unicorn/no-useless-switch-case](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v55.0.0/docs/rules/no-useless-switch-case.md)

packages/@o3r/eslint-config/src/rules/typescript/eslint.cjs
matthieu-crouzet: No it's disabled by default

packages/@o3r/eslint-config/src/rules/typescript/unicorn.cjs
fpaul-1A: I didn't like the fact that there was no auto-fix, it's an easy search and replace though

packages/@ama-sdk/swagger-builder/src/core/swagger-spec-merger.ts
cpaulve-1A: I can see value for both:
- having it explicitely written ensure that the developer did not forget a use case
- enforcing the default will make less boilerplate

matthieu-crouzet: But have we got a rule to enforce all use case ?

packages/@ama-sdk/core/src/fwk/api.helpers.ts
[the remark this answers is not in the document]
kpanot: Still a breaking change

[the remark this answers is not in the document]
matthieu-crouzet: fixed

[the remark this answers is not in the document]
matthieu-crouzet: I added a comment to say that the default value will change in v12

packages/@ama-sdk/swagger-builder/src/core/swagger-spec-merger.ts
matthieu-crouzet: we can use this one instead [@typescript-eslint/switch-exhaustiveness-check](https://typescript-eslint.io/rules/switch-exhaustiveness-check/)
with the following options
```json
{
    allowDefaultCaseForExhaustiveSwitch: false,
    requireDefaultForNonUnion: true,
}
```

packages/@ama-sdk/core/src/clients/api-fetch-client.ts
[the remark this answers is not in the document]
matthieu-crouzet: we keep it like that

packages/@o3r/eslint-config/src/rules/typescript/unicorn.cjs
matthieu-crouzet: After discussion with the team we keep this rule set to `off`

packages/@ama-sdk/core/src/fwk/Reviver.ts
matthieu-crouzet: As discussed it's not equivalent

fpaul-1A (approved):

kpanot (approved):

PR #2331, opened 2024-10-23
fix(eslint): manual fix of eslint errors on ama-terasu

matthieu-crouzet:


mrednic-1A (approved):

vscaiceanu-1a (approved):
LGTM

packages/@ama-terasu/cli/src/helpers/module.helper.ts
```diff
@@ -179,7 +184,8 @@ export const getLocalDependencies = async (): Promise<Record<string, string>> =>
  * @param options.localOnly Resolve module locally only
  * @returns list of modules to load
  */
-export const getCliModules = async (options: { localOnly: boolean } = { localOnly: false }): Promise<ModuleDiscovery[]> => {
+export const getCliModules = async (options?: { localOnly: boolean }): Promise<ModuleDiscovery[]> => {
```
kpanot, line 187: Why this change?
The caller is actually losing information regarding the default value :S.
So it should be kept as previously or `@default` should be specified (even if, in this case `undefined` is `falsy`, it is not the same as `false`)

matthieu-crouzet: The following rule was raising an error: [no-object-as-default-parameter](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v55.0.0/docs/rules/no-object-as-default-parameter.md)
I can specify the `@default` in tsdoc if you want

kpanot: There is no rule to enforce in in case of object used as function param?

fpaul-1A: I believe it already applies only for object
I would have liked that this triggered an error only when the object have multiple fields

matthieu-crouzet: As discussed with the team we keep this rule activated

PR #2339, opened 2024-10-24
fix: eslint configuration for windows

sdo-1A:
## Proposed change

Fix configuration of eslint to work for windows environment (absolute file paths and quotes in nx.json)

## Related issues

- :bug: Fixes #(issue)
- :rocket: Feature #(issue)

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -117,11 +117,11 @@
     "lint": {
       "options": {
         "cwd": "{projectRoot}",
-        "command": "yarn eslint '**/*.{m,c,}{j,t}s' '**/*.html' '**/*.json' --no-error-on-unmatched-pattern"
+        "command": "yarn eslint \"**/*.{m,c,}{j,t}s\" \"**/*.html\" \"**/*.json\" --no-error-on-unmatched-pattern"
       },
       "configurations": {
         "ci": {
-          "command": "yarn eslint '**/*.{m,c,}{j,t}s' '**/*.html' '**/*.json' --quiet --cache-location .cache/eslint --no-error-on-unmatched-pattern"
+          "command": "yarn eslint \"**/*.{m,c,}{j,t}s\" \"**/*.html\" \"**/*.json\" --quiet --cache-location .cache/eslint --no-error-on-unmatched-pattern"
         }
       },
       "executor": "nx:run-commands",
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,6 +1,6 @@
 import { sync } from 'globby';
 import { dirname, relative } from 'node:path';
-import { fileURLToPath } from 'node:url';
+import { fileURLToPath, pathToFileURL } from 'node:url';
 import shared from './eslint.shared.config.mjs';
 
 const __filename = fileURLToPath(import.meta.url);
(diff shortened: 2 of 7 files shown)


matthieu-crouzet (approved):

nx.json
```diff
@@ -117,11 +117,11 @@
     "lint": {
       "options": {
         "cwd": "{projectRoot}",
-        "command": "yarn eslint '**/*.{m,c,}{j,t}s' '**/*.html' '**/*.json' --no-error-on-unmatched-pattern"
+        "command": "yarn eslint \"**/*.{m,c,}{j,t}s\" \"**/*.html\" \"**/*.json\" --no-error-on-unmatched-pattern"
```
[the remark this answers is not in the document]
matthieu-crouzet: `'` were not working on Windows

previously we were using [@nx/eslint:lint](https://nx.dev/nx-api/eslint/executors/lint) where we had to set `lintFilePatterns` I replaced it with that to have the same command on each package (we can still configure it for each if needed)
or we can rediscussed it but it should not be part of this PR

tools/github-actions/audit/packaged-action/index.js
```diff
@@ -26189,9 +26189,9 @@ var __importStar = (this && this.__importStar) || function (mod) {
     return result;
 };
```
cpaulve-1A, line 26190: It seems strange only generated files have been modified in this PR and not the original main.ts?

matthieu-crouzet: She modified the nx.json, so it invalidated the cache, regenerated the code and linted it

fpaul-1A: why are we even linting the generated js code ?

cpaulve-1A: still, I don't think it is good that we have changes on something unrelated and that does not match the initial ts file

cpaulve-1A: Also, we should not lint generated files

nx.json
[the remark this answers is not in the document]
fpaul-1A: I think we had issues in the past were glob with double quotes like `"**/*"` were evaluated before the command, ending up with issues (line too long or something, don't remember)

fpaul-1A (approved):

nx.json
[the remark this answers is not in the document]
kpanot: I think indeed that Powershell evaluates `"**/*"` before executing the command which leads to issue, maybe the reason why `'` is not working need to be investigated :S.


> previously we were using [@nx/eslint:lint](https://nx.dev/nx-api/eslint/executors/lint) where we had to set `lintFilePatterns` I replaced it with that to have the same command on each package (we can still configure it for each if needed)
> or we can rediscussed it but it should not be part of this PR

I actually don't think it's a good solutions :S.
Having the patterns (and ignore patterns) in one (or several) dedicate files is much cleaner and configurable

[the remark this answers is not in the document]
matthieu-crouzet: We can change it later :)
it's not part of this PR :)

PR #2345, opened 2024-10-25
fix(eslint): manual fix of eslint errors on components

matthieu-crouzet:


packages/@o3r/components/schematics/index.it.spec.ts
```diff
@@ -3,23 +3,23 @@
  * @jest-environment @o3r/test-helpers/jest-environment
  * @jest-environment-o3r-app-folder test-app-components
  */
-const o3rEnvironment = globalThis.o3rEnvironment;
```
fpaul-1A, line 6: the point was to have this const linked to the "jsdoc" that we use to set up the jest environment

packages/@o3r/components/src/devkit/inspector/otter-inspector.helpers.spec.ts
```diff
@@ -27,11 +48,12 @@ class MockComponent {
   public translations: Translation = {
     mainKey: 'mainKey'
   };
+
   public analyticsEvents: AnalyticsEvents = {
     mainEvent: MockEvent
   };
 
-  // eslint-disable-next-line @typescript-eslint/naming-convention
+  // eslint-disable-next-line @typescript-eslint/naming-convention -- private field
```
fpaul-1A, line 56: private field that happens to be public ? :D

fpaul-1A (dismissed):

packages/@o3r/components/builders/component-extractor/helpers/component/component-class.extractor.ts
```diff
@@ -57,32 +63,28 @@ export class ComponentClassExtractor {
...
   /**
    * Get the component selector from the given decorator node.
-   *
    * @param decoratorNode The decorator node to get the component selector from
    */
   private getComponentSelector(decoratorNode: ts.Decorator) {
     if (this.isComponentDecorator(decoratorNode)) {
-      const matches = /selector:\s*['"](.*)['"]/.exec(decoratorNode.getText(this.source));
+      const matches = /selector:\s*["'](.*)["']/.exec(decoratorNode.getText(this.source));
       if (matches) {
```
kpanot, line 88: Why it has not be fixed by auto fix run earlier?

matthieu-crouzet: As I said in our chat I don't know why the auto fix didn't run on this package earlier :/

packages/@o3r/components/src/devkit/inspector/otter-inspector.helpers.spec.ts
matthieu-crouzet: We are in a mocked component

fpaul-1A: That doesn't change the fact that the justification for the rule is confusing (the field is clearly not private here)
Maybe instead you can say the the naming is imposed by the implementation in which it is a private field

packages/@o3r/components/builders/metadata-check/index.it.spec.ts
```diff
@@ -331,13 +354,15 @@ describe('check metadata migration', () => {
       packageManagerExec({ script: 'ng', args: ['run', `${appName}:check-metadata`] }, execAppOptionsWorkspace);
       throw new Error('should have thrown before');
     } catch (e: any) {
+      /* eslint-disable jest/no-conditional-expect -- always call as there is a throw always call in the try block */
```
fpaul-1A, line 357:
```suggestion
      /* eslint-disable jest/no-conditional-expect -- always called as there is a throw in the try block */
```

packages/@o3r/components/builders/metadata-check/index.it.spec.ts
```diff
@@ -331,13 +354,14 @@ describe('check metadata migration', () => {
       packageManagerExec({ script: 'ng', args: ['run', `${appName}:check-metadata`] }, execAppOptionsWorkspace);
       throw new Error('should have thrown before');
     } catch (e: any) {
-      expect(e.message).not.toBe('should have thrown before');
+      /* eslint-disable jest/no-conditional-expect -- always called as there is a throw in the try block */ expect(e.message).not.toBe('should have thrown before');
```
fpaul-1A, line 357:
```suggestion
      /* eslint-disable jest/no-conditional-expect -- always called as there is a throw in the try block */
      expect(e.message).not.toBe('should have thrown before');
```

fpaul-1A: same bellow

fpaul-1A (approved):

fpaul-1A (approved):

mrednic-1A (approved):

PR #2346, opened 2024-10-25
fix(eslint): manual fix of eslint errors on configuration

matthieu-crouzet:


cpaulve-1A (dismissed):

packages/@o3r/configuration/schematics/use-config-signal/index.ts
```diff
@@ -89,9 +89,9 @@ function ngUseConfigSignalFn(options: NgUseConfigSignalSchematicsSchema): Rule {
                   !member.name
                   || (
                     ts.isIdentifier(member.name)
-                    && member.name.escapedText !== 'config$'
-                    && member.name.escapedText !== 'config'
-                    && member.name.escapedText !== 'dynamicConfig$'
+                    && member.name.escapedText.toString() !== 'config$'
```
kpanot, line 92: A constant would be cleaner :)

packages/@o3r/configuration/src/devkit/configuration-devtools.console.spec.ts
```diff
@@ -40,7 +40,7 @@ describe('Configuration DevTools console', () => {
             configuration: {
               ids: ['@scope/package#componentTest'],
               entities: {
-                // eslint-disable-next-line @typescript-eslint/naming-convention
+                // eslint-disable-next-line @typescript-eslint/naming-convention -- id
```
kpanot, line 43: Creating a const for the name would avoir this exception (and it is used just a bit above)

packages/@o3r/configuration/src/devkit/configuration-devtools.message.spec.ts
```diff
@@ -56,7 +56,7 @@ describe('Configuration DevTools message', () => {
             configuration: {
               ids: ['@scope/package#componentTest'],
               entities: {
-                // eslint-disable-next-line @typescript-eslint/naming-convention
+                // eslint-disable-next-line @typescript-eslint/naming-convention -- id
                 '@scope/package#componentTest': {
```
kpanot, line 60: Same comme as before regarding the const

packages/@o3r/configuration/src/devkit/configuration-devtools.service.ts
```diff
@@ -60,11 +60,9 @@ export class OtterConfigurationDevtools {
...
           .filter((entity): entity is ConfigurationModel => !!entity)
           .map((entity) => {
-            const { id, ...configuration } = entity;
-            const parsedConfigurationName = parseConfigurationName(id);
-            const libraryName = parsedConfigurationName && parsedConfigurationName.library || this.options.defaultLibraryName;
-            const name = parsedConfigurationName && parsedConfigurationName.componentName || id;
-            return { name, config: configuration, library: libraryName } as CustomConfig;
+            const { id, ...config } = entity;
+            const { library = this.options.defaultLibraryName, componentName: name = id } = parseConfigurationName(id) || {};
+            return { name, config, library } as CustomConfig;
```
kpanot, line 65: Maybe `satisfies` would be cleaner

packages/@o3r/configuration/src/devkit/configuration-devtools.service.ts
```diff
@@ -60,11 +60,9 @@ export class OtterConfigurationDevtools {
...
           })
           .filter((conf): conf is CustomConfig => !!conf)
```
kpanot, line 67: Not needed line

fpaul-1A (approved):

packages/@o3r/configuration/src/devkit/configuration-devtools.console.spec.ts
matthieu-crouzet: Maybe a good example for the discussion for the different configuration between source and test :)

vscaiceanu-1a: Since these are few isolated cases, maybe it doesn't justify an extra config only for this.

matthieu-crouzet: it seems a bit overkill to create a constant just to avoid et eslint-disable in a test

fpaul-1A (approved):

packages/@o3r/configuration/src/devkit/configuration-devtools.console.spec.ts
kpanot: In that case the `const` can be justify as you want to align it with the `ids` above.
But I agree it is not always the case.

This can be an ajustement of the rule for our projects but it does not justify, for me, to have this exception for all the apps

packages/@o3r/configuration/schematics/use-config-signal/index.ts
```diff
@@ -89,9 +89,11 @@ function ngUseConfigSignalFn(options: NgUseConfigSignalSchematicsSchema): Rule {
                   !member.name
                   || (
                     ts.isIdentifier(member.name)
-                    && member.name.escapedText !== 'config$'
-                    && member.name.escapedText !== 'config'
-                    && member.name.escapedText !== 'dynamicConfig$'
+                    && ![
```
mrednic-1A, line 92: I am not sure that this is more performant than the `ifs`. anyway not important here

packages/@o3r/configuration/src/devkit/configuration-devtools.message.service.ts
```diff
@@ -78,8 +78,7 @@ export class ConfigurationDevtoolsMessageService implements OnDestroy, DevtoolsS
 
   /**
    * Function to handle the incoming messages from Otter Chrome DevTools extension
-   * @param event Event coming from the Otter Chrome DevTools extension
-   * @param message
+   * @param message message coming from the Otter Chrome DevTools extension
```
mrednic-1A, line 81: I think it's more readable if the description always starts with a capital letter

mrednic-1A (approved):

packages/@o3r/configuration/src/devkit/configuration-devtools.message.service.ts
matthieu-crouzet: We can discuss to add this rule [jsdoc/require-description-complete-sentence](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/.README/rules/require-description-complete-sentence.md)

matthieu-crouzet: As discussed if we go for this rule we will have a dedicated PR to activate it

PR #2347, opened 2024-10-25
fix(eslint): manual fix of eslint errors on core

matthieu-crouzet:


packages/@o3r/core/src/store/types.ts
```diff
@@ -65,27 +65,32 @@ export interface FailActionPayload<T> {
   error?: T;
 }
 
-// eslint-disable-next-line @typescript-eslint/naming-convention
-export type keep<T, K extends keyof T> = (Partial<T> & { [p in K]: T[p] });
+export type Keep<T, K extends keyof T> = (Partial<T> & { [p in K]: T[p] });
```
fpaul-1A, line 68: isn't it basically just a `Pick` ?

matthieu-crouzet: I don't want to refactor the code in this PR
I think Pick is equivalent to `type Pick = { [p in K]: T[p] }`
so here we keep all the fields but some becomes non mandatory but if you defined them you still have to match the initial typing

fpaul-1A: yes I read too fast
```suggestion
export type Keep<T, K extends keyof T> = Partial<T> & Pick<T,K>;
```

fpaul-1A (approved):

packages/@o3r/core/src/store/types.ts
```diff
@@ -65,27 +65,32 @@ export interface FailActionPayload<T> {
   error?: T;
 }
 
-// eslint-disable-next-line @typescript-eslint/naming-convention
-export type keep<T, K extends keyof T> = (Partial<T> & { [p in K]: T[p] });
+export type Keep<T, K extends keyof T> = Partial<T> & Pick<T, K>;
+
+/**
+ * @deprecated please use Keep instead, will be removed in v13.
```
kpanot, line 71:
```suggestion
 * @deprecated please use {@link Keep} instead, will be removed in v13.
```

kpanot (approved):

PR #2338, opened 2024-10-24
chore: rework on Nx Cache and Doc

kpanot:
## Proposed change

Rework on Nx Cache for clarification and simplification
Add documentation regarding Nx PAT new process

PR #2355, opened 2024-10-28
fix: missing optional dependencies

kpanot:
Extracted from #2338


apps/github-cascading-app/package.json
```diff
@@ -25,6 +25,7 @@
     "@actions/github": "^6.0.0",
     "@angular-devkit/build-angular": "~18.2.0",
     "@angular-devkit/core": "~18.2.0",
+    "@angular-devkit/schematics": "~18.2.0",
```
matthieu-crouzet, line 28: To reply to this comment https://github.com/AmadeusITGroup/otter/pull/2338#discussion_r1818524001
I'm not sure that we have usage of these packages in this GH app 
I don't see any `import` even `import type`

nx.json
matthieu-crouzet, line 1: I don't see changes in this file apart whitespace 
could we remove these changes from this PR ?

apps/github-cascading-app/package.json
kpanot: It was requested by yarn (not Nx):

image

matthieu-crouzet: I don't understand why it is requested as `@o3r/test-helpers` is only a devDep of `@o3r/eslint-plugin`

kpanot: It is used in `@o3r/test-helpers`'s `ng-add` so required as optional peerDeps.

The thing surprising is that there is no mandatory peerDep to `@angular-devkit/schematics` in the displayed packages. I wonder if `yarn` does not turn it mandatory when there is a recursive dependency loop of with the common optional peer dep :/

vscaiceanu-1a (approved):

apps/github-cascading-app/package.json
cpaulve-1A: image
I am not sure it is required? (Or am I missing something?)

cpaulve-1A: My bad, I didn't think you change the package.json dependencies when your reworked nx

kpanot: > the peerDep `@angular-devkit/schematics` is not optional in `@o3r/telemetry`, so this is probably why yarn requires it, even if `@o3r/telemetry` itself is not provided **EDIT**: I was looking at an old version of the code, the mandatory peer deps comes from `@o3r/schematics`, not `@o3r/telemetry`

Indeed I missed it, then it replies to @matthieu-crouzet original question.

> Yes, but `@o3r/telemetry` is a dep of `@o3r/schematics` a dep `@o3r/test-helpers`. As Matthieu mentionned, test-helper is only a dev dep of `@o3r/eslint-plugin`, why do we require it?

`@o3r/test-helpers` is not part of the dependencies of `github-cascading-app`.


> Still, looking at your other pr, at most these are optional peer dependency of @o3r/eslint-plugin. Why do we add it as a dependency for the github application when the optional peer dep is required only to run the ng-add?

(If I understand properly the question, sorry if it is not the case :S) `@o3r/eslint-plugin` is part of `github-cascading-app` dev dependencies because it is used by the linter. It is not relative to Nx rule neither Yarn dependency check.

PR #2356, opened 2024-10-28
fix(eslint): manual fix of eslint errors on packages @o3r/{l,m}*

matthieu-crouzet:


eslint.shared.config.mjs
```diff
@@ -37,6 +37,7 @@ export default [
         ...globals.browser,
         ...globals.node,
         ...globals.webextensions,
+        NodeJS: true,
```
kpanot, line 40: This is the case for `builders`, `cli`, `schematics`, `test` and `scripts`.
It may be very valuable to forbid `Node` in the other cases.
(same comment for `globals.node` and reverse comment for `globals.browser`)

packages/@o3r/logger/src/services/logger/logger.spec.ts
```diff
@@ -102,60 +121,81 @@ describe('Logger service', () => {
...
         client1 = {
           identify: jest.fn(),
           event: jest.fn(),
           getSessionURL: jest.fn().mockReturnValue('client.session.url1'),
-          stopRecording: jest.fn(),
-          resumeRecording: jest.fn(),
-          error: jest.fn(),
-          warn: jest.fn(),
-          log: jest.fn(),
+          stopRecording: stopRecordingMock1,
```
kpanot, line 183: Is it coming from a Rule?
This change is fine but a mockReset instead of a new instance can lead to issue if not correctly applied

eslint.shared.config.mjs
matthieu-crouzet: It's only for us, it's not the shared config

packages/@o3r/logger/src/services/logger/logger.spec.ts
matthieu-crouzet: Most of the time it's to fix [@typescript-eslint/unbound-method](https://typescript-eslint.io/rules/unbound-method/) 
I run the test in local for this package and they are still green :)

eslint.shared.config.mjs
kpanot: yes, my comment was for us :)

packages/@o3r/logger/src/services/logger/logger.spec.ts
kpanot: I knew the tests were fine.
The `jest.fn()` is not part of the exceptions handled by `eslint-plugin-jest` (as described in the linked doc)?

matthieu-crouzet: issue is not on `jest.fn()` but on `expect(client.log)` because we gave the reference to an unbound method

matthieu-crouzet: I got your point 
I was not aware this rule [jest/unbound-method](https://github.com/jest-community/eslint-plugin-jest/blob/HEAD/docs/rules/unbound-method.md)
This rule is not activated on our config 
we may check if we want to activate it or not

eslint.shared.config.mjs
matthieu-crouzet: I take the point 
I will do it in another PR

packages/@o3r/localization/src/tools/localization.service.ts
```diff
@@ -49,7 +48,7 @@ export class LocalizationService {
   /**
    * Internal subject that we use to track changes between keys only and translation mode
    */
-  // eslint-disable-next-line @typescript-eslint/naming-convention
+  // eslint-disable-next-line @typescript-eslint/naming-convention -- private field
   private readonly _showKeys$ = new BehaviorSubject(false);
```
vscaiceanu-1a, line 52:
```suggestion
  private readonly showKeys = new BehaviorSubject(false);
```

matthieu-crouzet: One of our recommendation is to have a suffix `$` when using observable or subject no ?

eslint.shared.config.mjs
matthieu-crouzet: https://github.com/AmadeusITGroup/otter/pull/2361

packages/@o3r/localization/schematics/ng-add/index.ts
```diff
@@ -72,6 +72,9 @@ function ngAddFn(options: NgAddSchematicsSchema): Rule {
       };
     });
     const registerDevtoolRule = await registerDevtools(options);
+    const schematicDefaultOptions = Object.fromEntries(['@o3r/core:component*'].map((name) => [name, {
```
fpaul-1A, line 75: are we really expecting more entries? or is it just a trick to make the linter happy?

matthieu-crouzet: For some schematics we have more entries 
So I did it to have the same format everywhere I saw it

packages/@o3r/localization/src/tools/localization.service.ts
vscaiceanu-1a: I don't think we recommend this for Subject, only for Observables.

packages/@o3r/localization/builders/helpers/localization.generator.ts
```diff
@@ -226,11 +226,12 @@ export class LocalizationExtractor {
...
   public getPatternsFromTsConfig() {
-    const tsconfigResult = ts.readConfigFile(this.tsconfigPath, ts.sys.readFile);
+    const tsconfigResult = ts.readConfigFile(this.tsconfigPath, (p) => ts.sys.readFile(p));
 
     if (tsconfigResult.error) {
-      // eslint-disable-next-line @typescript-eslint/no-base-to-string
-      const stringError = tsconfigResult.error.messageText.toString();
+      const stringError = typeof tsconfigResult.error.messageText === 'string'
+        ? tsconfigResult.error.messageText
+        : tsconfigResult.error.messageText.messageText;
```
fpaul-1A, line 234: is this really equivalent to `tsconfigResult.error.messageText.toString()`?

packages/@o3r/localization/schematics/localization-base/index.ts
```diff
@@ -463,11 +469,6 @@ export function updateLocalization(options: { projectName?: string | null | unde
   ]);
 }
 
-/**
- *
- * @param options
- * @param options.projectName
- */
 function updateI18nFn(options: { projectName?: string | undefined }): Rule {
```
sdo-1A, line 472: why did you remove the jsdoc ?

packages/@o3r/mobile/src/capacitor/helpers.spec.ts
```diff
@@ -9,9 +9,6 @@ import {
   isCapacitorContext
 } from './helpers';
 
-/**
- * @param platform the device platform: ios, android or web
- */
 function setCapacitorPlatform(platform: CapacitorPlatforms) {
```
sdo-1A, line 12: why did you remove the jsdoc ?

packages/@o3r/localization/builders/helpers/localization.generator.ts
matthieu-crouzet: if `tsconfigResult.error.messageText` was not of type string it will return `[object object]`, no ?
that's what the rule said

packages/@o3r/localization/schematics/localization-base/index.ts
matthieu-crouzet: It's a private function so jsdoc is not mandatory and this doc was not adding any values

packages/@o3r/localization/src/tools/localization.service.ts
kpanot: Indeed the private field name can be changed (or the rule adapted) to not disable `no-underscore-dangle` in several places

matthieu-crouzet: Fixed with the new config

packages/@o3r/logger/src/services/logger/logger.spec.ts
matthieu-crouzet: We will see later if we can/should use [jest/unbound-method](https://github.com/jest-community/eslint-plugin-jest/blob/HEAD/docs/rules/unbound-method.md)

packages/@o3r/logger/src/services/logger/logger.spec.ts
```diff
@@ -60,18 +60,37 @@ describe('Logger service', () => {
...
+      identifyMock.mockReset();
+      eventMock.mockReset();
+      getSessionURLMock.mockReset();
+      stopRecordingMock.mockReset();
+      resumeRecordingMock.mockReset();
+      errorMock.mockReset();
+      warnMock.mockReset();
+      logMock.mockReset();
+      createMetaReducerMock.mockReset();
+      getSessionURLMock.mockReturnValue('client.session.url');
```
fpaul-1A, line 83: that seems like a lot of boilerplate just to please the linter

packages/@o3r/localization/schematics/ng-update/v10-0/index.ts
```diff
@@ -36,4 +36,5 @@ function updateV10_0Fn(): Rule {
 /**
  * Update of Otter library V10.0
  */
+// eslint-disable-next-line @typescript-eslint/naming-convention -- contains the version number
```
mrednic-1A, line 39: is it not better to have it at the top of the file ?
We might still add migration functions.

packages/@o3r/logger/src/services/logger/logger.spec.ts
matthieu-crouzet: no need of changes with `jest/unbound-method`

packages/@o3r/localization/schematics/ng-update/v10-0/index.ts
matthieu-crouzet: That's something we discussed with the team and we said that it's better to disable only where it's needed
If you want we can re-discussed it to set a global config on the repo to disable this rule for all files inside an `schematics/ng-update` folder

fpaul-1A (approved):

mrednic-1A (approved):

PR #2361, opened 2024-10-28
fix(eslint): change config name for debugging + set only the globals needed for each project

matthieu-crouzet:


eslint.shared.config.mjs
```diff
@@ -25,35 +25,46 @@ export default [
...
   },
   {
-    name: '@o3r/globals',
+    name: '@o3r/framework/settings',
+    settings: {
+      'import/resolver': 'node'
+    }
+  },
+  {
+    name: '@o3r/framework/ecma-version',
```
kpanot, line 40: Can be merged with `@o3r/framework/settings`, they are both configs to align with `tsconfig.base.json`

kpanot (approved):

mrednic-1A (approved):

Issue #1518, opened 2024-03-19
[Feature]: Migrate Rules Engine debug panel to Design Factory


kpanot:
### Context

The Rules Engine debug panel is integrated within the Otter Chrome Extension and, since the Otter Chrome Extension has been migrated to the Design Factory, the different component with it should follow the same design pattern.

### Proposal

- [ ] Ensure the Rules Engine Panel follows the Design Factory pattern and recommendations
- [ ] Apply Design Factory color the the panel (thanks to CSS Variable based on the Design Token color set)

PR #2366, opened 2024-10-29
fix: automatic and manual fixes of eslint errors on rules engine

sdo-1A:
## Proposed change

Manual fix of ESLint errors for @o3r/rules-engine

## Related issues

*- No issue associated -*


packages/@o3r/rules-engine/schematics/index.it.spec.ts
```diff
@@ -3,8 +3,7 @@
  * @jest-environment @o3r/test-helpers/jest-environment
  * @jest-environment-o3r-app-folder test-app-rules-engine
  */
-const o3rEnvironment = globalThis.o3rEnvironment;
```
fpaul-1A, line 6: should be kept here

packages/@o3r/rules-engine/src/components/rules-engine/ruleset-history/ruleset-history-pres.template.html
```diff
@@ -7,6 +7,7 @@ <h4 class="mb-4">Ruleset Execution History</h4>
   </ng-template>
   <ul *ngIf="rulesetExecutions; else noRulesEngine" class="rulesets">
     <li *ngFor="let execution of rulesetExecutions" class="ruleset">
+      <!-- eslint-disable-next-line @angular-eslint/template/click-events-have-key-events, @angular-eslint/template/interactive-supports-focus -- need to refactor the div to accordion from DF -->
```
fpaul-1A, line 10: we can link the associated issue https://github.com/AmadeusITGroup/otter/issues/1518

packages/@o3r/rules-engine/src/engine/helpers/filter-ruleset-event.operator.spec.ts
```diff
@@ -134,68 +147,66 @@ describe('Filter rulesets event operator', () => {
...
-      done();
+  test('should consider only first ruleset', () => {
+    return new Promise((done) => {
+      rulesetsMapSubject$.pipe(
+        filterRulesetsEventStream(['ruleset1'])
+      ).subscribe((data) => {
+        expect(data.length).toBe(2);
+        done();
+      });
     });
```
fpaul-1A, line 164: can't we do something like this:
```suggestion
  test('should consider only first ruleset', async () => {
    const data = await firstValueFrom(rulesetsMapSubject$.pipe(
      filterRulesetsEventStream(['ruleset1'])
    );
    expect(data.length).toBe(2);
```
?

packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.ts
```diff
@@ -45,26 +61,27 @@ export class RulesEngineExtractor {
   private readonly commentParser = new ConfigDocParser();
 
   constructor(tsconfigPath: string, private readonly basePath: string, private readonly logger: LoggerApi) {
-    const {config} = ts.readConfigFile(tsconfigPath, ts.sys.readFile);
+    // eslint-disable-next-line @typescript-eslint/unbound-method -- parameter of `readConfigFile` is of type `(path: string) => string | undefined`
+    const { config } = ts.readConfigFile(tsconfigPath, ts.sys.readFile);
```
fpaul-1A, line 65: I assume the linter rule expects this instead
```suggestion
    const { config } = ts.readConfigFile(tsconfigPath, (p) => ts.sys.readFile(p));
```

packages/@o3r/rules-engine/src/engine/operator/operator.helpers.ts
```diff
@@ -37,20 +43,20 @@ export function executeOperator<L = unknown, R = unknown>(lhs: L, rhs: R, operat
  * @param operand value of one of the operands
  */
 export function numberValidator(operand: unknown): operand is number | string {
-  // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-  return operand !== '' && !Array.isArray(operand) && !isNaN(+`${operand}`);
+  // eslint-disable-next-line @typescript-eslint/restrict-template-expressions -- type is explicitly `unknown`
+  return operand !== '' && !Array.isArray(operand) && !Number.isNaN(+`${operand}`);
```
fpaul-1A, line 47:
```suggestion
  return operand !== '' && !Array.isArray(operand) && !Number.isNaN(Number(operand));
```
?

packages/@o3r/rules-engine/src/engine/operator/operators/basic.operators.ts
```diff
@@ -1,14 +1,21 @@
...
 
 /**
  * Check if a variable is equal to a specific value
  * @title is equal to
  */
 export const equals: Operator = {
   name: 'equals',
-  // eslint-disable-next-line
-  evaluator: (firstValue, secondValue) => firstValue == secondValue
+  evaluator: (firstValue, secondValue) => firstValue === secondValue
```
fpaul-1A, line 18: this no totally equivalent
`equals(4, '4')` was previously true, it is now false

packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.ts
```diff
@@ -45,26 +61,27 @@ export class RulesEngineExtractor {
...
    * Extract type definition of a type reference into schema file
-   *
    * @param type Name of the type to extract
    * @param sourceFile path to the code source file
    */
   private async extractTypeRef(type: string, sourceFile: string) {
-    const internalLibFiles = this.tsconfig.extraOptions?.otterSubModuleRefs?.length > 0 ?
-      await lastValueFrom(zip(...(this.tsconfig.extraOptions?.otterSubModuleRefs as string[]).map((value: string) => globby(value))).pipe(
+    const internalLibFiles = this.tsconfig.extraOptions?.otterSubModuleRefs?.length > 0
+      ? await lastValueFrom(zip(...(this.tsconfig.extraOptions?.otterSubModuleRefs as string[]).map((value: string) => globby(value))).pipe(
```
cpaulve-1A, line 76: It does not complain on the line length?

packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.ts
```diff
@@ -251,7 +263,7 @@ export class RulesEngineExtractor {
                     fact = {
                       ...fact,
                       type: 'string',
-                      // eslint-disable-next-line @typescript-eslint/no-base-to-string
+                      // eslint-disable-next-line @typescript-eslint/no-base-to-string -- convert non-null values to string
                       enum: schema.enum?.filter((value): value is JSONSchema7Type => !!value).map((v) => v!.toString())
```
cpaulve-1A, line 267: Does `JSONSchema7Type` has a valid toString method?

packages/@o3r/rules-engine/schematics/index.it.spec.ts
cpaulve-1A: Shy does it need to be dined before the import?

packages/@o3r/rules-engine/src/engine/helpers/filter-ruleset-event.operator.spec.ts
```diff
@@ -134,68 +147,66 @@ describe('Filter rulesets event operator', () => {
...
 
     rulesetsMapSubject$.pipe(
       filterRulesetsEventStream(['ruleset3'])
-    ).subscribe(data => {
+    ).subscribe((data) => {
       emittedActions = data;
     });
 
     await jest.advanceTimersByTimeAsync(500);
     expect(emittedActions).toBe(undefined);
```
kpanot, line 210: Not really relative to the fix but this can lead to issue because there is no unsubscription.
Maybe it can be fixed with:

```typescript
     const stop = new Subject<void>();
     const data = firstValueFrom(
       rulesetsMapSubject$.pipe(
         takeUntil(stop),
         filterRulesetsEventStream(['ruleset3'])
       ), {defaultValue: 'no emit'}
    );

    await jest.advanceTimersByTimeAsync(500);
    stop.next();
    await expect(data).resolves.toBe('no emit');
```

packages/@o3r/rules-engine/schematics/rules-engine-to-component/index.spec.ts
```diff
@@ -100,7 +106,7 @@ describe('Enable rules-engine on component', () => {
     await expect(runner.runSchematic('rules-engine-to-component', {
       projectName: 'test-project',
       path: 'test-folder/test-component-with-rules-engine.component.ts'
-    // eslint-disable-next-line max-len
+    // eslint-disable-next-line @stylistic/max-len -- keep error message on same line
     }, initialTree)).rejects.toThrow(new O3rCliError('Unable to add rules-engine: component "test-folder/test-component-with-rules-engine.component.ts" already has at least one of these properties: rulesEngineService.'));
```
cpaulve-1A, line 110: Same comment here about split on new keyword

packages/@o3r/rules-engine/src/components/rules-engine/rule-condition/rule-condition-pres.component.ts
```diff
@@ -35,18 +39,19 @@ export class RuleConditionPresComponent {
    */
   @Input()
   public set condition(condition: TopLevelCondition | undefined) {
+    // eslint-disable-next-line no-underscore-dangle -- false positive
```
cpaulve-1A, line 42: It is more that it is to identify a private object

packages/@o3r/rules-engine/src/engine/operator/operators/basic.operators.ts
```diff
@@ -1,14 +1,21 @@
...
+  UnaryOperator
+} from '../operator.interface';
 
 /**
  * Check if a variable is equal to a specific value
  * @title is equal to
  */
 export const equals: Operator = {
   name: 'equals',
-  // eslint-disable-next-line
```
kpanot, line 10: should not be removed but maybe disable `eqeqeq` only?

packages/@o3r/rules-engine/src/engine/operator/operators/date-based.operators.spec.ts
```diff
@@ -1,4 +1,6 @@
-import { executeOperator } from '..';
+import {
+  executeOperator
+} from '..';
```
kpanot, line 3: a more precise import would avoid issue :D

packages/@o3r/rules-engine/src/components/rules-engine/rule-condition/rule-condition-pres.component.ts
```diff
@@ -35,18 +39,19 @@ export class RuleConditionPresComponent {
...
     this._condition = condition;
     this.lhs = (condition as UnaryOperation | BinaryOperation)?.lhs ? this.getOperandName((condition as UnaryOperation | BinaryOperation).lhs) : 'undefined';
     this.rhs = (condition as BinaryOperation)?.rhs ? this.getOperandName((condition as BinaryOperation).rhs) : undefined;
   }
 
   public get condition(): TopLevelCondition | undefined {
+    // eslint-disable-next-line no-underscore-dangle -- false positive
```
cpaulve-1A, line 49: Same here, the explanation is less about false positive and more the usage of a private property. Isn't there a config to handle this use case? It is weird to allow something on private properties but prevent it when it is on usage (instead of definition)

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -112,8 +140,8 @@ export class RulesetExecutor {
       return undefined;
     } else if (isOperandFact(operand)) {
       const factValue = factsValue[operand.value];
-      // eslint-disable-next-line new-cap
-      return operand.path ? factValue && JSONPath({wrap: false, json: factValue, path: operand.path}) : factValue;
+      // eslint-disable-next-line new-cap -- convention for JSONPath
```
kpanot, line 143: can be renamed in `import` instead

packages/@o3r/rules-engine/schematics/index.it.spec.ts
fpaul-1A: to avoid the jsdoc being associated with the import
The const itself doesn't need to be first but the comment needs to be on the first line of the file for Jest to use it

packages/@o3r/rules-engine/src/engine/structure.ts
```diff
@@ -37,17 +40,20 @@ export interface BinaryOperation {
...
 
 /** Nested Condition */
-// eslint-disable-next-line no-use-before-define
+
 export type NestedCondition = UnaryOperation | BinaryOperation | TopLevelCondition;
 /** All Condition */
-// eslint-disable-next-line id-blacklist
-export type AllConditions = {all: NestedCondition[]; any?: never; not?: never};
+
+// eslint-disable-next-line id-denylist -- any is a condition type
```
kpanot, line 47:
```suggestion
// eslint-disable-next-line id-denylist -- `any` is a conditional keyword enforced by rule interface
```
?

packages/@o3r/rules-engine/src/engine/structure.ts
```diff
@@ -72,7 +78,7 @@ export interface Rule {
   /** Name of the rule*/
   name: string;
   /** rootElement of the rule, that contains either a block, either an action list */
-  // eslint-disable-next-line no-use-before-define
+
```
kpanot, line 81: can be removed

packages/@o3r/rules-engine/src/engine/structure.ts
```diff
@@ -112,7 +118,7 @@ export interface RuleBlock extends RuleElement {
 
 /** All supported blocks (supporting nested structure) */
 export type AllBlock =
-// eslint-disable-next-line no-use-before-define
+
```
kpanot, line 121: can be removed

packages/@o3r/rules-engine/src/fact/fact.spec.ts
```diff
@@ -40,6 +47,7 @@ describe('Rules engine fact', () => {
     subjectFact.next('test4');
     await jest.runAllTimersAsync();
 
+    // eslint-disable-next-line @typescript-eslint/unbound-method -- method is not called here, we use reference for the mock engine
```
kpanot, line 50: this is part of `jest/unbound-method` rule

packages/@o3r/rules-engine/src/components/rules-engine/rule-key-value/rule-key-value-pres.template.html
```diff
@@ -3,12 +3,22 @@
...
+    (click)="shouldLimitCharactersForOldValue = !shouldLimitCharactersForOldValue"
+    (keyup.enter)="shouldLimitCharactersForOldValue = !shouldLimitCharactersForOldValue"
+    tabindex="0"
+  >
+    {{isOldValuePrimitiveType ? oldValue : (oldValue | json)}}
+  </pre>
   <button (click)="copyToClipBoard(oldValue)" *ngIf="isClipBoardFeatureAvailableForOldValue" title="Copy to clipboard">📋</button>
   →
 </ng-container>
 <pre class="input-value"
```
cpaulve-1A, line 15: shouldn't it be a button in that case? We can have a button without actual styling and get it similar visually

packages/@o3r/rules-engine/src/engine/operator/operators/basic.operators.ts
cpaulve-1A: Especially when the value might come from a JSON file (if one of the value is a const defined on a CMS)

cpaulve-1A: And this is true for the whole file :)

packages/@o3r/rules-engine/src/fact/fact.spec.ts
```diff
@@ -31,6 +37,7 @@ describe('Rules engine fact', () => {
     factsService.register();
     await jest.runAllTimersAsync();
 
+    // eslint-disable-next-line @typescript-eslint/unbound-method -- method is not called here, we use reference for the mock engine
     expect(mockEngine.upsertFacts).toHaveBeenCalledTimes(1);
```
cpaulve-1A, line 41: Is it still failing after the import of the jest-unbound-method?

packages/@o3r/rules-engine/src/services/runner/rules-engine.runner.service.spec.ts
```diff
@@ -1,39 +1,99 @@
...
+} from '../../stores/index';
+import {
+  RulesEngineRunnerService
+} from './rules-engine.runner.service';
 
 describe('Rules engine service', () => {
-
   let service: RulesEngineRunnerService;
-  // eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents
+
```
cpaulve-1A, line 96: line could be removed here :)

packages/@o3r/rules-engine/schematics/ng-add/index.ts
```diff
@@ -92,18 +105,13 @@ function ngAddFn(options: NgAddSchematicsSchema): Rule {
           }]
         };
       });
+    const defaultSchematicOptions = Object.fromEntries([
+      '@o3r/core:component',
+      '@o3r/core:component-container'
+    ].map((name) => [name, { useRulesEngine: undefined }]));
```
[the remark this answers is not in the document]
fpaul-1A: `Object.fromEntries` convert the array to an object

packages/@o3r/rules-engine/src/engine/helpers/filter-ruleset-event.operator.spec.ts
fpaul-1A: `firstValueFrom` should be enough to close the subscription, I don't think you need the `takeUntil`

packages/@o3r/rules-engine/schematics/index.it.spec.ts
matthieu-crouzet: either we can disable the rule import/first for all our it.spec.ts (not in the recommended config) 
either we can set 
```
/* eslint-disable import/first -- for it test we want to have `o3rEnvironment` linked to the jsdoc of the jest environment setup */
```
before import

packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.ts
matthieu-crouzet: 141 and max value is 200 so it seems okay

packages/@o3r/rules-engine/src/components/rules-engine/rule-condition/rule-condition-pres.component.ts
matthieu-crouzet: after the rebase this issue should not be raised anymore

packages/@o3r/rules-engine/builders/rules-engine-extractor/index.ts
```diff
@@ -56,46 +76,43 @@ export default createBuilder(createBuilderWithMetricsIfInstalled<RulesEngineExtr
...
   const newFactList = (await Promise.all(
-    (await Promise.all(options.factFilePatterns.map((pattern) => globby(pattern, {cwd: context.currentDirectory}))))
+    (await Promise.all(options.factFilePatterns.map((pattern) => globby(pattern, { cwd: context.currentDirectory }))))
       .flat()
       .map((file, idx, arr) => {
         context.reportProgress(idx, arr.length, `Parsing fact from ${file}`);
         return extractor.extractFacts(resolve(context.currentDirectory, file), schemaFolder, `./${SCHEMA_FOLDER}`);
       })
-  )).reduce<MetadataFact[]>((acc, factList) => [...acc, ...factList], []);
+  )).flat();
```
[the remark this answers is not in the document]
matthieu-crouzet: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-array-flat.md

packages/@o3r/rules-engine/src/engine/debug/helpers.spec.ts
```diff
@@ -1,13 +1,23 @@
...
+  handleRuleEvaluationDebug,
+  retrieveRulesetTriggers
+} from './helpers';
 
 const currentEvaluationsOutput = [
   {
     'actions': [],
     'evaluation': {
-      'timestamp': 1639037495870,
+      'timestamp': 1_639_037_495_870,
```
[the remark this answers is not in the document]
matthieu-crouzet: agree but I think it comes with autofix for numbers ... 
maybe there is a configuration for that

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
matthieu-crouzet: Should not be a problem anymore after the rebase

packages/@o3r/rules-engine/src/fact/fact.spec.ts
matthieu-crouzet: Will not be raised anymore after rebase

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
matthieu-crouzet: It would be easier to identify without renaming 
I prefer the justification
because someone could remove the rename at some point 
or if we change the eslint config the rename will be useless

packages/@o3r/rules-engine/src/fact/fact.spec.ts
matthieu-crouzet: will be fixed with the rebase

packages/@o3r/rules-engine/src/engine/helpers/filter-ruleset-event.operator.spec.ts
kpanot: It will lead to a promise never resolve, I don't know if the subscription is close when there is no ref to this promise anymore.
Maybe it is also simpler to check the `non-fulfilled` status by explicitly completing the observable

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
kpanot: Indeed, but this is also one of the purposes of the renaming: giving a name that respect your conventions.
So, if it would have been a new dev, we would have renamed so it would not shocked me to do the same here.
That being said, I do not have a strong opinion on it

packages/@o3r/rules-engine/src/components/rules-engine/rule-key-value/rule-key-value-pres.template.html
matthieu-crouzet: It should not be a button it is only a click to change the css class to have the ellipsis or not 
I don't want the screen reader to be aware of that because the innerContent will be the same and does not explain the action that a button will made 
for me in this case we want only the click and the keyboard navigation but not a button

packages/@o3r/rules-engine/builders/rules-engine-extractor/helpers/rules-engine.extractor.ts
sdo-1A: we may need to rewrite this line in another PR, but this is not the goal of this PR

packages/@o3r/rules-engine/schematics/ng-add/index.ts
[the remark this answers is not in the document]
matthieu-crouzet: https://github.com/AmadeusITGroup/otter/pull/2384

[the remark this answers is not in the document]
sdo-1A: https://github.com/AmadeusITGroup/otter/pull/2384

packages/@o3r/rules-engine/schematics/index.it.spec.ts
sdo-1A: https://github.com/AmadeusITGroup/otter/pull/2384

packages/@o3r/rules-engine/src/services/runner/rules-engine.runner.service.spec.ts
```diff
@@ -94,22 +153,23 @@ describe('Rules engine service', () => {
...
-      done();
+  it('should handle linked components and validity range properly', () => {
+    return new Promise((done) => {
+      service.events$.pipe(take(1)).subscribe((actions) => {
+        expect(actions.length).toBe(1);
+        expect(actions[0].actionType).toBe('UPDATE_LOCALISATION');
+        done();
+      });
+      store.dispatch(setRulesetsEntities({ entities: jsonOneRulesetValidOneRuleNoCond.ruleSets }));
     });
```
fpaul-1A, line 170:
```suggestion
  it('should handle linked components and validity range properly', async () => {
    const eventsPromise = firstValueFrom(service.events$);
    store.dispatch(setRulesetsEntities({ entities: jsonOneRulesetValidOneRuleNoCond.ruleSets }));
    const actions = await eventsPromise ;
    expect(actions.length).toBe(1);
    expect(actions[0].actionType).toBe('UPDATE_LOCALISATION');
```

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
mrednic-1A: I don;t like to have extra code to satisfy the linter. I would keep the disable part.

packages/@o3r/rules-engine/src/engine/helpers/filter-ruleset-event.operator.spec.ts
fpaul-1A: my bad, I didn't realize that we were checking that the observable doesn't emit

packages/@o3r/rules-engine/src/services/runner/rules-engine.runner.service.spec.ts
matthieu-crouzet: We deactivated the linter for this test to not refactor the test in this PR

kpanot (approved):

fpaul-1A (approved):

PR #2370, opened 2024-10-30
fix(eslint): manual fix of eslint errors on package schematics

matthieu-crouzet:


packages/@o3r/schematics/src/utility/collection.spec.ts
```diff
@@ -43,18 +42,18 @@ const angularJsonNoGeneric: WorkspaceSchema = {
...
-describe('Get schematics options', () => {
-  const createFakeContext = (collection: string, name: string): any => ({
-    schematic: {
-      description: {
-        collection: {
-          name: collection
-        },
-        name
-      }
+const createFakeContext = (collection: string, name: string): any => ({
```
vscaiceanu-1a, line 45: wasn't this rule disabled for tests? (I am ok with the change, just curious)

packages/@o3r/schematics/src/utility/package-version.ts
```diff
@@ -28,14 +28,13 @@ export function getPackageVersion(packageJsonPath: string) {
...
 export const findClosestPackageJson = (currentPath: string, tree?: Tree): string | undefined => {
-  const dir = (tree ? posix.dirname : dirname)(currentPath);
+  const dir = (tree ? posix.dirname.bind(posix) : dirname)(currentPath);
   if (dir === currentPath || currentPath === '/') {
     return undefined;
   }
 
-  const packageJsonPath = (tree ? posix.join : join)(dir, 'package.json');
-  return (tree ? tree.exists : existsSync)(packageJsonPath) ? packageJsonPath : findClosestPackageJson(dir, tree);
+  const packageJsonPath = (tree ? posix.join.bind(posix) : join)(dir, 'package.json');
```
vscaiceanu-1a, line 38: from which rule is this coming?

fpaul-1A (approved):

packages/@o3r/schematics/src/utility/collection.spec.ts
matthieu-crouzet: We said that we will deactivate this rule after the migration, for now we try to fix a maximum when possible and meaningful

packages/@o3r/schematics/src/utility/package-version.ts
matthieu-crouzet: @typescript-eslint/unbound-method

vscaiceanu-1a (approved):

PR #2376, opened 2024-10-30
fix(pipeline): action version and update npmRegistry

vscaiceanu-1a:
## Proposed change

The version for Otter GitHub actions is incorrect.

Also, if a registry is already present in `.npmrc` but a value is passed to the ng add schematics, it should be updated.

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@o3r/pipeline/schematics/ng-add/index.ts
+++ b/packages/@o3r/pipeline/schematics/ng-add/index.ts
@@ -13,10 +13,10 @@ function ngAddFn(options: NgAddSchematicsSchema): Rule {
 
   return async (tree, context) => {
     const packageJsonPath = path.resolve(__dirname, '..', '..', 'package.json');
-    const ownPackageJson = JSON.parse(fs.readFileSync(packageJsonPath, { encoding: 'utf8' })) as PackageJson & { o3rConfig?: { commitHash?: string } };
-    const commitHash = ownPackageJson.o3rConfig?.commitHash;
+    const ownPackageJson = JSON.parse(fs.readFileSync(packageJsonPath, { encoding: 'utf8' })) as PackageJson & { config?: { o3r?: { commitHash?: string } } };
+    const commitHash = ownPackageJson.config?.o3r?.commitHash;
     const ownVersion = ownPackageJson.version;
-    const actionVersionString = commitHash ? `${commitHash} # v${ownVersion}` : ownVersion;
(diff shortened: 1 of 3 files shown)


packages/@o3r/pipeline/schematics/ng-add/index.ts
```diff
@@ -57,7 +57,12 @@ function ngAddFn(options: NgAddSchematicsSchema): Rule {
           tree.create(npmrcPath, npmRegistry);
         } else {
           const npmrcContent = tree.readText(npmrcPath);
-          tree.overwrite(npmrcPath, npmrcContent.concat(`\n${npmRegistry}`));
+          const registryPattern = /^registry=.*$/m;
+          const newRegistryLine = `registry=${npmRegistry}`;
+          const newContent = registryPattern.test(npmrcContent)
```
kpanot, line 62: Why do you want to enforce your registry if already a default one is specified?

vscaiceanu-1a: Only when a custom one is passed in the schematic's options. 
Today when we generate an npm otter monorepo it comes with `registry=http://registry.npmjs.org/`

kpanot (approved):

packages/@o3r/pipeline/schematics/ng-add/index.ts
```diff
@@ -16,7 +16,7 @@ function ngAddFn(options: NgAddSchematicsSchema): Rule {
     const ownPackageJson = JSON.parse(fs.readFileSync(packageJsonPath, { encoding: 'utf8' })) as PackageJson & { o3rConfig?: { commitHash?: string } };
     const commitHash = ownPackageJson.o3rConfig?.commitHash;
```
matthieu-crouzet, line 17: Sorry I didn't see it in your previous PR but `o3rConfig` is not aligned with what we decided for telemetry. We set `o3r.config.telemetry`, shouldn't we have `o3r.config.commitHash` to be aligned ?

vscaiceanu-1a: well spotted. I will update in this PR

vscaiceanu-1a: it's actually `config.o3r`, not `o3r.config`. PR updated

fpaul-1A (approved):

kpanot (approved):

PR #2377, opened 2024-10-30
fix(design): remove Node requirement in core files

kpanot:
## Proposed change

Remove Node requirement in core files of the Design Tool.
The purpose is to be able to run the design token generation in a non-Node environment

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
+++ b/packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
@@ -3,20 +3,66 @@ import { getCssTokenDefinitionRenderer } from './css/design-token-definition.ren
 import { getCssStyleContentUpdater } from './css/design-token-updater.renderers';
 import type { Logger } from '@o3r/core';
 import type { promises as fs } from 'node:fs';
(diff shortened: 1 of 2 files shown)


mrednic-1A (approved):

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
```diff
@@ -3,20 +3,66 @@ import { getCssTokenDefinitionRenderer } from './css/design-token-definition.ren
...
-  }
+export const computeFileToUpdatePath = (root = '.', defaultFile = 'styles.scss') => (token: DesignTokenVariableStructure) => {
+  return token.extensions.o3rTargetFile
+    ? getFilePath(token.extensions.o3rTargetFile, token.context?.basePath || root)
+    : defaultFile;
+};
 
-  return defaultFile;
+/**
+ * Retrieve the function that determines which file to update for a given token
```
fpaul-1A, line 45: should we document that this function only works in a Node environment?

fpaul-1A (approved):

PR #2378, opened 2024-10-31
fix(localization): path generated in the angular.json in case of monorepo

kpanot:
Path generated in the angular.json in case of monorepo.

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@o3r/localization/schematics/localization-base/index.ts
+++ b/packages/@o3r/localization/schematics/localization-base/index.ts
@@ -90,6 +90,8 @@ export function updateLocalization(options: { projectName?: string | null | unde
       workspaceProject.architect = {};
     }
 
+    const projectBasePath = workspaceProject.root.replace(/[/\\]$/, '');
+
     workspaceProject.architect['generate-translations'] ||= {
       builder: '@o3r/localization:localization',
       options: {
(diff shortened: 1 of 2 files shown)


packages/@o3r/localization/schematics/localization-base/index.ts
```diff
@@ -98,12 +100,12 @@ export function updateLocalization(options: { projectName?: string | null | unde
         locales: [
           'en-GB'
         ],
-        assets: [`${mainAssetsFolder}/locales`],
-        outputPath: `${devResourcesFolder}/localizations`
+        assets: [`${projectBasePath}/${mainAssetsFolder}/locales`],
```
matthieu-crouzet, line 103: shouldn't we add a test somewhere to be sure we will not remove that fix at some point ? or understand if we should supports both cases

matthieu-crouzet (approved):

packages/@o3r/localization/schematics/localization-base/index.ts
kpanot: We do not have e2e for an application outside of the mono-repo.
I can add a UT

mrednic-1A (approved):

PR #2379, opened 2024-10-31
feat(design): add sorting way based on regexps

kpanot:
## Proposed change

Add a way to sort variables based on provided RegExps

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@o3r/design/builders/generate-style/helpers/style-renderer-options.ts
+++ b/packages/@o3r/design/builders/generate-style/helpers/style-renderer-options.ts
@@ -2,6 +2,7 @@ import type { BuilderContext } from '@angular-devkit/architect';
 import {
   type CssTokenDefinitionRendererOptions,
   type CssTokenValueRendererOptions,
+  type DesignTokenListTransform,
   type DesignTokenRendererOptions,
   type DesignTokenVariableStructure,
   getCssStyleContentUpdater,
@@ -12,6 +13,7 @@ import {
   getSassTokenValueRenderer,
   getTokenSorterByName,
   getTokenSorterByRef,
+  getTokenSorterFromRegExpList,
   type SassTokenDefinitionRendererOptions,
   type SassTokenValueRendererOptions,
   type TokenKeyRenderer,
@@ -19,6 +21,7 @@ import {
 } from '../../../src/public_api';
 import type { GenerateStyleSchematicsSchema } from '../schema';
 import { resolve } from 'node:path';
+import { readFileSync } from 'node:fs';
 
 export const getStyleRendererOptions = (tokenVariableNameRenderer: TokenKeyRenderer | undefined , options: GenerateStyleSchematicsSchema, context: BuilderContext): DesignTokenRendererOptions => {
 
@@ -113,14 +116,35 @@ export const getStyleRendererOptions = (tokenVariableNameRenderer: TokenKeyRende
 
   /** Sorting strategy of variables based on selected language */
   const tokenListTransforms = ((language) => {
+    const customSorter: DesignTokenListTransform[] = [];
+    if (options.sortOrderPatternsFilePath) {
+      try {
+        const regExps = (JSON.parse(readFileSync(resolve(context.workspaceRoot, options.sortOrderPatternsFilePath), {encoding: 'utf8'})) as string[])
+          .map((item) => new RegExp(item.replace(/^\/(.*)\/$/, '$1')));
+        customSorter.push(getTokenSorterFromRegExpList(regExps));
+      } catch (err: any) {
+        if (err?.code === 'ENOENT') {
+          context.logger.warn(`The specified RegExp file ${options.sortOrderPatternsFilePath} is not found in ${context.workspaceRoot}`);
+        } else {
+          context.logger.warn(`Error during the parsing of ${options.sortOrderPatternsFilePath}.`);
+          if (err instanceof Error) {
+            context.logger.warn(err.message);
+            context.logger.debug(err.stack || 'no stack');
+          } else {
+            context.logger.debug(JSON.stringify(err, null, 2));
+          }
+        }
+        context.logger.warn(`The ordered list will be ignored.`);
+      }
+    }
     switch (language) {
       case 'scss':
       case 'sass': {
-        return [getTokenSorterByName, getTokenSorterByRef];
+        return [getTokenSorterByName, ...customSorter, getTokenSorterByRef];
       }
       case 'css':
       default: {
-        return [getTokenSorterByName];
+        return [getTokenSorterByName, ...customSorter];
       }
     }
   })(options.variableType || options.language);
--- a/packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.spec.ts
+++ b/packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.spec.ts
@@ -3,7 +3,13 @@ import { promises as fs } from 'node:fs';
 import { resolve } from 'node:path';
 import type { DesignTokenGroup, DesignTokenSpecification } from '../design-token-specification.interface';
 import type { DesignTokenVariableSet } from '../parsers';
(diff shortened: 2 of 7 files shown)


packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.spec.ts
```diff
@@ -157,6 +163,45 @@ describe('Design Token Renderer', () => {
...
+    it('should sort properly based on regExps', () => {
+      const regExps = [
+        /override$/,
+        /shadow/
+      ];
+      const list = Array.from(designTokens.values());
+      const sortedTokens = getTokenSorterFromRegExpList(regExps)(designTokens)(list);
+
+      expect(list.findIndex(({ tokenReferenceName }) => tokenReferenceName === 'example.var-expect-override'))
+        .toBeGreaterThan(list.findIndex(({ tokenReferenceName }) => tokenReferenceName === 'example.var1'));
```
matthieu-crouzet, line 176: We are doing several time `list.findIndex(({ tokenReferenceName }) => tokenReferenceName === 'example.var1')` for example
we can grouped them in a variable to improve a bit the performances no ?

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
```diff
@@ -98,6 +98,31 @@ export const getTokenSorterByRef: DesignTokenListTransform = (variableSet) => {
...
 
+/**
+ * Reorganize the Tokens based on a ordered list of RegExps
+ * @param regExps Ordered list of regular expressions defining the order of the Tokens.
+ */
+export const getTokenSorterFromRegExpList: (regExps: RegExp[]) => DesignTokenListTransform = (regExps) => (_variableSet, options) => {
+
+  return (tokens) =>
+    tokens
+      .map((token) => ({ index: regExps.findIndex((r) => r.test(token.getKey(options?.tokenVariableNameRenderer))), token }))
```
matthieu-crouzet, line 109: We should validate that the regexp does not have a `g` flag
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test#using_test_on_a_regex_with_the_global_flag
no ?

kpanot: I think it is up to the user to care about it.
We can add a word regarding it in the doc, but I don't think you should forbid it

matthieu-crouzet: I agree to add a message in the doc 
but I will at least raised a warning because the behavior is totally uncontrol in this case (I would prefer that we throw an error, I don't see use case where it could have a sense)
If in the code we have `token.getKey(options?.tokenVariableNameRenderer).match(r)`
the user will not have the problem and it's not their role to check how we coded it.

packages/@o3r/design/builders/generate-style/helpers/style-renderer-options.ts
```diff
@@ -113,14 +116,25 @@ export const getStyleRendererOptions = (tokenVariableNameRenderer: TokenKeyRende
...
+      } catch {
+        context.logger.warn(`The specified RegExp file ${options.sortOrderPatternsFilePath} is not reachable or not correctly formatted.`);
+        context.logger.warn(`The order list will be ignored`);
+      }
+    }
     switch (language) {
       case 'scss':
       case 'sass': {
-        return [getTokenSorterByName, getTokenSorterByRef];
+        return [getTokenSorterByName, ...customSorter, getTokenSorterByRef];
```
mrednic-1A, line 133: for my understanding.
This is sorting the variables by name and then applies the custom filter and then the ref one, the ref one having the highest priority ?

mrednic-1A (dismissed):

packages/@o3r/design/builders/generate-style/helpers/style-renderer-options.ts
kpanot: This is correct

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
```diff
@@ -98,6 +99,38 @@ export const getTokenSorterByRef: DesignTokenListTransform = (variableSet) => {
...
+ * Each regexp is applied only to last part of the Token name (before key rendering).
+ * @param regExps Ordered list of regular expressions defining the order of the Tokens.
+ * @param applyRendererName Determine if the regexps are apply to the rendered Token key. If `false`, it will be applied to the Token key's name (last part of the Token name).
+ */
+export const getTokenSorterFromRegExpList: (regExps: RegExp[], applyRendererName?: boolean) => DesignTokenListTransform = (regExps, applyRendererName = false) => (_variableSet, options) => {
+
+  const applyRegExp = (token: DesignTokenVariableStructure, regExp: RegExp) => (applyRendererName
+    ? token.getKey(options?.tokenVariableNameRenderer)
+    : token.tokenReferenceName.split(TOKEN_KEY_SEPARATOR).at(-1)!
+  ).match(regExp);
```
matthieu-crouzet, line 113: Maybe add a comment so we don't move back to `Regexp.test(string)` as recommended by our new ESLint recommendation

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.spec.ts
fpaul-1A, line 1: it could be good to add a test with a incorrectly formatted RegExp that triggers an exception

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.ts
```diff
@@ -98,6 +99,39 @@ export const getTokenSorterByRef: DesignTokenListTransform = (variableSet) => {
   };
 };
 
+/**
+ * Reorganize the Tokens based on a ordered list of regexps.
+ * Each regexp is applied only to last part of the Token name (before key rendering).
+ * @param regExps Ordered list of regular expressions defining the order of the Tokens.
+ * @param applyRendererName Determine if the regexps are apply to the rendered Token key. If `false`, it will be applied to the Token key's name (last part of the Token name).
```
fpaul-1A, line 106:
```suggestion
 * Reorganize the Tokens based on an ordered list of regexps.
 * Each regexp is applied only to the last part of the Token name (before key rendering).
 * @param regExps Ordered list of regular expressions defining the order of the Tokens.
 * @param applyRendererName Determine if the regexps are applied to the rendered Token key. If `false`, it will be applied to the Token key's name (last part of the Token name).
```

packages/@o3r/design/builders/generate-style/helpers/style-renderer-options.ts
```diff
@@ -113,14 +116,25 @@ export const getStyleRendererOptions = (tokenVariableNameRenderer: TokenKeyRende
 
   /** Sorting strategy of variables based on selected language */
   const tokenListTransforms = ((language) => {
+    const customSorter: DesignTokenListTransform[] = [];
+    if (options.sortOrderPatternsFilePath) {
+      try {
+        const regExps = (JSON.parse(readFileSync(resolve(context.workspaceRoot, options.sortOrderPatternsFilePath), {encoding: 'utf8'})) as string[])
+          .map((item) => new RegExp(item.replace(/^\/(.*)\/$/, '$1')));
+        customSorter.push(getTokenSorterFromRegExpList(regExps));
+      } catch {
```
fpaul-1A, line 125: maybe the exception message could be helpful to determine the cause of the error

packages/@o3r/design/src/core/design-token/renderers/design-token-style.renderer.spec.ts
kpanot: I suppose you are talking about the builder part, not this one as the regexp are strongly typed.
Should be part of #2385

fpaul-1A: yes the builder part where we do a `new RegExp`

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #2389, opened 2024-10-31
feat(2388): check relevance of migration files items on metadata checks

### LINKED ISSUE #2388: [Feature]: Check relevance of migration files items on metadata checks
### Context

When describing the breaking changes on migration-files, we are checking that the metadata removal are properly documented in the migration file.
However, we are not checking that the removal documented in the migration scripts really correspond to a metadata removal.

### Proposal

We should add a feature that checks that all the migration items documented in the migration files are used during the metadata checks.
This feature can be disabled with an option.

fpaul-1A:
## Proposed change

Add an optional check on migration files to detect unused migration items.

## Related issues

* :rocket: Feature resolves #2388

### DIFF
--- a/packages/@o3r/extractors/src/core/comparator/metadata-comparator.interface.ts
+++ b/packages/@o3r/extractors/src/core/comparator/metadata-comparator.interface.ts
@@ -26,12 +26,17 @@ export interface MetadataComparator<MetadataItem, MigrationMetadataItem, Metadat
   isSame?: (item1: MetadataItem, item2: MetadataItem) => boolean;
 
   /**
+   * Returns true is the contentType is supported by the comparator
+   * @param contentType Content type of the migration item
+   */
+  isRelevantContentType: (contentType: string) => boolean;
+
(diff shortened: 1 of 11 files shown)


packages/@o3r/extractors/src/core/comparator/metadata-comparator.interface.ts
```diff
@@ -25,13 +25,19 @@ export interface MetadataComparator<MetadataItem, MigrationMetadataItem, Metadat
...
+   * Returns true is the contentType is supported by the comparator
+   * @param contentType Content type of the migration item
+   */
+  isRelevantContentType: (contentType: string) => boolean;
+
   /**
    * Returns true if a migration item matches a metadata item.
    * @param metadataItem Metadata item
    * @param migrationItem Migration item
    * @param metadataType Type of the metadata
```
mrednic-1A, line 38: could be removed from param list

kpanot (approved):

PR #2393, opened 2024-11-04
fix(eslint): fix the build

matthieu-crouzet:


apps/chrome-devtools/tsconfig.build.components.json
```diff
@@ -1,8 +1,13 @@
 {
   "extends": "./tsconfig.build",
   "compilerOptions": {
-    "composite": true
+    "composite": false
```
fpaul-1A, line 4: why is it not working?

matthieu-crouzet: This tsconfig is used to compile with esbuild and should not be a composite 
I don't understand what has changed and raised an issue
but setting composite to false (which is the expected value) solve the issue

fpaul-1A (approved):

vscaiceanu-1a (approved):

PR #2397, opened 2024-11-04
fix(2396): metadata checks should run on prerelease tags

### LINKED ISSUE #2396: [Bug]: Metadata checks don't take into account prerelease tags
### Package name

extractors

### Package version

11

### Reproduction steps

Run the metadata checks on an application that have prerelease tags:
tags: `1.0.0`, `1.1.0-rc`
Run the metadata checks on the tag `2.0.0`

### Current result

The check is done on tag `1.0.0`

### Expected result

The check should be done on tag `1.1.0-rc`

### Additional comments

This is link the function that we use the get the list of tags `npm view` that doesn't return the prerelease tags when using a range

fpaul-1A:
## Proposed change

- Use `npm view <package> versions` instead of `npm view <package@range> version` to include the prerelease tags
- Extends the `NpmSemverResolver` used for yarn to include prerelease tags

## Related issues

* :bug: Fix resolves #2396

### DIFF
--- a/packages/@o3r/extractors/src/core/comparator/package-managers-extractors/custom-npm-semver-resolver.ts
+++ b/packages/@o3r/extractors/src/core/comparator/package-managers-extractors/custom-npm-semver-resolver.ts
@@ -1,0 +1,56 @@
+import { npmHttpUtils, NpmSemverFetcher, NpmSemverResolver } from '@yarnpkg/plugin-npm';
+import { Descriptor, miscUtils, Package, ResolveOptions, structUtils } from '@yarnpkg/core';
+import { Range, SemVer, valid } from 'semver';
+
+/**
+ * Unexposed constant from yarn https://github.com/yarnpkg/berry/blob/master/packages/plugin-npm/sources/constants.ts
+ */
+const PROTOCOL = `npm:`;
+
+/**
+ * Extends NpmSeverResolver to support prerelease tags
+ * Check original code from https://github.com/yarnpkg/berry/blob/master/packages/plugin-npm/sources/NpmSemverResolver.ts
+ */
+export class CustomNpmSemverResolver extends NpmSemverResolver {
+  /** @inheritDoc */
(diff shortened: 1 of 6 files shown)


kpanot (dismissed):

packages/@o3r/extractors/src/core/comparator/package-managers-extractors/custom-npm-semver-resolver.ts
```diff
@@ -0,0 +1,55 @@
...
+    });
+
+    // If there are versions that aren't deprecated, use them
+    const finalCandidates = noDeprecatedCandidates.length > 0
+      ? noDeprecatedCandidates
+      : candidates;
+
+    finalCandidates.sort((a, b) => {
+      return -a.compare(b);
+    });
```
kpanot, line 41:
```suggestion
    finalCandidates.sort((a, b) => -a.compare(b));
```

packages/@o3r/extractors/src/core/comparator/package-managers-extractors/npm-file-extractor.helper.ts
```diff
@@ -31,18 +31,22 @@ export async function getFilesFromRegistry(packageDescriptor: string, paths: str
   const tempDirPath = join(tmpdir(), tempDirName);
   let extractedFiles: { [key: string]: string } = {};
   mkdirSync(tempDirPath);
+  const [,packageName,packageRange] = sanitizeInput(packageDescriptor).match(/^(.*?)(?:\b@([^@]+))?$/) || [];
```
kpanot, line 34: not sure the `@` is forbidden in `prerelease` tag:

```suggestion
  const [,packageName,packageRange] = sanitizeInput(packageDescriptor).match(/^(.*?)(?:\b@(.+))?$/) || [];
```

kpanot (approved):

kpanot (approved):

matthieu-crouzet (approved):

matthieu-crouzet (approved):

PR #2401, opened 2024-11-04
fix: linked components should take validity range into account

### LINKED ISSUE #2400: [Bug]: Rulesets with linked components do not take validity range into account
### Package name

rules-engine

### Package version

10.4.0

### Reproduction steps

In the `rulesets.json` used in the showcase app, set the `validityRange` with values that should disable the rulesets (i.e. a `from` date in the future).

### Current result

The ruleset is still executed.

### Expected result

It should not be executed.

### Additional comments

I believe it should be treated with high priority.

vscaiceanu-1a:
## Related issues

* :bug: Fix resolves #2400

### DIFF
--- a/packages/@o3r/rules-engine/src/stores/rulesets/rulesets.selectors.spec.ts
+++ b/packages/@o3r/rules-engine/src/stores/rulesets/rulesets.selectors.spec.ts
@@ -100,6 +100,31 @@ describe('RuleSets Selector tests', () => {
     }
   };
 
+  const r9: RulesetsModel = {
+    id: 'r9',
+    name: 'r9_name',
+    rules: [],
+    validityRange: {
+      from: afterTomorrow.toISOString()
+    },
+    linkedComponent: {
+      library: '@mylibrary',
+      name: 'thisComponentWillBeIgnored'
+    },
+    linkedComponents: {
+      or: [
+        {
+          library: '@mylibrary',
+          name: 'mycomponent'
+        },
+        {
+          library: '@mylibrary',
+          name: 'mycomponent2'
+        }
+      ]
+    }
+  };
+
   it('should return the ruleset in range', () => {
     const state: RulesetsState = {
       ids: [r1.id, r2.id],
@@ -110,8 +135,9 @@ describe('RuleSets Selector tests', () => {
       requestIds: []
     };
     const allRuleSetsArray = selectors.selectAllRulesets.projector(state);
+    const rulesetsInRange = selectors.selectRuleSetsInRange.projector(allRuleSetsArray);
 
-    expect(selectors.selectActiveRuleSets.projector(allRuleSetsArray)).toEqual(['r1']);
+    expect(selectors.selectActiveRuleSets.projector(rulesetsInRange)).toEqual(['r1']);
   });
 
   it('should return the rulest with no validity or on demand', () => {
@@ -124,8 +150,9 @@ describe('RuleSets Selector tests', () => {
       requestIds: []
     };
     const allRuleSetsArray = selectors.selectAllRulesets.projector(state);
+    const rulesetsInRange = selectors.selectRuleSetsInRange.projector(allRuleSetsArray);
 
-    expect(selectors.selectActiveRuleSets.projector(allRuleSetsArray)).toEqual(['r3']);
+    expect(selectors.selectActiveRuleSets.projector(rulesetsInRange)).toEqual(['r3']);
   });
 
   it('should exclude the one not in validity range', () => {
@@ -138,8 +165,9 @@ describe('RuleSets Selector tests', () => {
       requestIds: []
     };
     const allRuleSetsArray = selectors.selectAllRulesets.projector(state);
+    const rulesetsInRange = selectors.selectRuleSetsInRange.projector(allRuleSetsArray);
 
-    expect(selectors.selectActiveRuleSets.projector(allRuleSetsArray)).toEqual(['r3']);
+    expect(selectors.selectActiveRuleSets.projector(rulesetsInRange)).toEqual(['r3']);
   });
 
   it('should not find valid rulesets', () => {
@@ -152,8 +180,9 @@ describe('RuleSets Selector tests', () => {
       requestIds: []
     };
     const allRuleSetsArray = selectors.selectAllRulesets.projector(state);
+    const rulesetsInRange = selectors.selectRuleSetsInRange.projector(allRuleSetsArray);
 
-    expect(selectors.selectActiveRuleSets.projector(allRuleSetsArray)).toEqual([]);
+    expect(selectors.selectActiveRuleSets.projector(rulesetsInRange)).toEqual([]);
   });
 
   it('should find a ruleset valid starting from a date in the past', () => {
@@ -167,8 +196,9 @@ describe('RuleSets Selector tests', () => {
       requestIds: []
     };
     const allRuleSetsArray = selectors.selectAllRulesets.projector(state);
+    const rulesetsInRange = selectors.selectRuleSetsInRange.projector(allRuleSetsArray);
 
-    expect(selectors.selectActiveRuleSets.projector(allRuleSetsArray)).toEqual(['r5']);
+    expect(selectors.selectActiveRuleSets.projector(rulesetsInRange)).toEqual(['r5']);
   });
 
   it('should find a ruleset valid until a date in the future', () => {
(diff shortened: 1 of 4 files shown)


packages/@o3r/rules-engine/src/stores/rulesets/rulesets.selectors.ts
```diff
@@ -93,6 +101,9 @@ export const selectRuleSetLinkComponents = createSelector(
   (ruleSets) =>
     ruleSets
       .reduce((acc: Record<string, string[]>, ruleSet: Ruleset) => {
+        if (!checkValidityRange(ruleSet)) {
```
kpanot, line 104: Not sure it is the correct way to fix it.
Why not taking `selectActiveRuleSets` instead of `selectAllRulesets `?  It sounds like make more sense :S

(same below)

packages/@o3r/rules-engine/src/stores/rulesets/rulesets.selectors.ts
```diff
@@ -43,32 +71,12 @@ export const selectActiveRuleSets = createSelector(
...
-
-    return to && to.getTime() >= time;
+    return checkValidityRange(ruleSet);
   }).map((ruleSet: Ruleset) => ruleSet.id));
 
-/*
+/**
  * Assign rulesetId to a component
- * @deprecated; It will be rmeoved in v12 with the selector using it
+ * @deprecated It will be rmeoved in v12 with the selector using it
```
matthieu-crouzet, line 79:
```suggestion
 * @deprecated It will be removed in v12 with the selector using it
```

packages/@o3r/rules-engine/src/stores/rulesets/rulesets.selectors.ts
matthieu-crouzet: It seems that linkedComponents are filtered out 
https://github.com/AmadeusITGroup/otter/pull/2401/files#diff-927d077c8a2d283c3eafef045849507ef3adebbd7f609c7f4bace09b1dbdf99bR70

packages/@o3r/rules-engine/src/stores/rulesets/rulesets.selectors.ts
```diff
@@ -30,6 +30,34 @@ export const selectRulesetsStorePendingStatus = createSelector(selectRulesetsSta
...
+
+  if (from) {
+    return from.getTime() <= time;
+  }
+  return to && to.getTime() >= time;
+}
+
 /**
  * Returns only the rulesets ids which are not onDemand and in the validity range
  * OnDemand takes precedence over validity range.
```
matthieu-crouzet, line 63: It seems that the current behavior was planned 
if after clarification we want to change it we should update the documentation accordingly

packages/@o3r/rules-engine/src/stores/rulesets/rulesets.selectors.ts
kpanot: Indeed (the name was less clear than the description) but the comment remains valid:
It is always cleaner to explicitly chain the selecters than re-applying the same function everywhere.

```typescript
const selectRuleSetsToApply = createSelector(
  selectAllRulesets,
  (ruleSets) => ruleSets.filter((ruleSet: Ruleset) => {
  const validity = ruleSet.validityRange;
  if (!validity || !validity.from && !validity.to) {
    return true;
  }

  const from = validity.from && new Date(validity.from);
  const to = validity.to && new Date(validity.to);

  if (to && !isValidDate(to) || from && !isValidDate(from)) {
    return false;
  }
  const time = Date.now();
  if (to && from) {
    return from.getTime() <= time && to.getTime() >= time;
  }

  if (from) {
    return from.getTime() <= time;
  }
  return to && to.getTime() >= time;
};

export const selectActiveRuleSets = createSelector(
  selectRuleSetsToApply,
  (ruleSets) => ruleSets.filter((ruleSet: Ruleset) => (!(ruleSet.linkedComponents?.or?.length || ruleSet.linkedComponent)));

export const selectRuleSetLinkComponents = createSelector(
  selectRuleSetsToApply,
  (ruleSets) => // ....

```

packages/@o3r/rules-engine/src/stores/rulesets/rulesets.selectors.ts
vscaiceanu-1a: JS doc updated. The `docs/` part has to be done in another PR on main because it does not exist on 10.4 yet

packages/@o3r/rules-engine/src/stores/rulesets/rulesets.selectors.ts
```diff
@@ -62,13 +55,25 @@ export const selectActiveRuleSets = createSelector(
...
+ * Returns the rulesets ids which are not onDemand and in the validity range
+ */
+export const selectActiveRuleSets = createSelector(
+  selectRuleSetsInRange,
+  (ruleSets) => ruleSets.filter((ruleSet: Ruleset) => {
+    if (ruleSet.linkedComponents?.or?.length || ruleSet.linkedComponent) {
+      return false;
+    }
+    return true;
   }).map((ruleSet: Ruleset) => ruleSet.id));
```
kpanot, line 72:
```suggestion
  selectRuleSetsInRange,
  (ruleSets) => ruleSets
    .filter((ruleSet: Ruleset) => (!(ruleSet.linkedComponents?.or?.length || ruleSet.linkedComponent)))
    .map((ruleSet: Ruleset) => ruleSet.id));
```

packages/@o3r/rules-engine/src/stores/rulesets/rulesets.selectors.ts
```diff
@@ -31,18 +31,11 @@ export const selectRulesetsStorePendingStatus = createSelector(selectRulesetsSta
 const isValidDate = (d: any) => !isNaN(d) && d instanceof Date;
 
 /**
- * Returns only the rulesets ids which are not onDemand and in the validity range
- * OnDemand takes precedence over validity range.
- * Only if the ruleset is NOT onDemand (this means it is taken into consideration for the runs), the validity is checked
+ * Returns the rulesets which are in the validity range, if provided
  */
-export const selectActiveRuleSets = createSelector(
+export const selectRuleSetsInRange = createSelector(
```
kpanot, line 36: you way not want to expose it

packages/@o3r/rules-engine/src/stores/rulesets/rulesets.selectors.spec.ts
```diff
@@ -138,8 +163,9 @@ describe('RuleSets Selector tests', () => {
       requestIds: []
     };
     const allRuleSetsArray = selectors.selectAllRulesets.projector(state);
+    const rulesetsInRange = selectors.selectRuleSetsInRange.projector(allRuleSetsArray);
 
-    expect(selectors.selectActiveRuleSets.projector(allRuleSetsArray)).toEqual(['r3']);
+    expect(selectors.selectActiveRuleSets.projector(rulesetsInRange)).toEqual(['r3']);
```
kpanot, line 168: Does this change mandatory? In the end  `allRuleSetsArray` is reached behind `selectRuleSetsInRange`

vscaiceanu-1a: it is mandatory because the input selector of `selectActiveRuleSets`. It is no longer `selectAllRulesets` but `selectRuleSetsInRange`

kpanot (approved):

packages/@o3r/rules-engine/src/stores/rulesets/rulesets.selectors.ts
```diff
@@ -62,13 +55,22 @@ export const selectActiveRuleSets = createSelector(
...
+export const selectActiveRuleSets = createSelector(
+  selectRuleSetsInRange,
+  (ruleSets) => ruleSets
+    .filter((ruleSet: Ruleset) => (!(ruleSet.linkedComponents?.or?.length || ruleSet.linkedComponent)))
+    .map((ruleSet: Ruleset) => ruleSet.id));
+
+/**
  * Assign rulesetId to a component
- * @deprecated; It will be rmeoved in v12 with the selector using it
+ * @deprecated It will be removed in v12 with the selector using it
```
fpaul-1A, line 73: this function is not exported, why can't we just remove it?

packages/@o3r/rules-engine/src/stores/rulesets/rulesets.selectors.ts
```diff
@@ -62,13 +55,22 @@ export const selectActiveRuleSets = createSelector(
...
+);
 
-/*
+/**
+ * Returns the rulesets ids which are not onDemand and in the validity range
+ */
+export const selectActiveRuleSets = createSelector(
+  selectRuleSetsInRange,
+  (ruleSets) => ruleSets
+    .filter((ruleSet: Ruleset) => (!(ruleSet.linkedComponents?.or?.length || ruleSet.linkedComponent)))
```
fpaul-1A, line 68:
```suggestion
    .filter((ruleSet: Ruleset) => !(ruleSet.linkedComponents?.or?.length || ruleSet.linkedComponent))
```

fpaul-1A (approved):

packages/@o3r/rules-engine/src/stores/rulesets/rulesets.selectors.ts
vscaiceanu-1a: because it is used in `selectRuleSetLinkComponents` which is exported

kpanot (approved):

PR #2410, opened 2024-11-06
chore: setup Nx Cloud Id for PAT support

kpanot:
## Proposed change

Setup Nx Cloud Id for PAT support
See https://nx.dev/ci/recipes/security/personal-access-tokens for details

## Related issues

*- No issue associated -*

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,11 +21,14 @@ on:
 
 env:
   YARN_ENABLE_HARDENED_MODE: 0
+  NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
 
 jobs:
   # Check the integrity of yarn lock
   yarn_lock_check:
     runs-on: ubuntu-latest
+    env:
+      NX_SKIP_NX_CACHE: true
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - uses: ./tools/github-actions/setup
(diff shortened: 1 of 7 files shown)


nx.json
```diff
@@ -515,5 +515,14 @@
...
   },
   "useDaemonProcess": false,
-  "cacheDirectory": ".cache/nx"
+  "cacheDirectory": ".cache/nx",
+  "nxCloudId": "63e6951e043da20dd8321aec",
+  "tasksRunnerOptions": {
+    "default": {
+      "runner": "nx-cloud",
+      "options": {
+        "nxCloudId": "63e6951e043da20dd8321aec"
```
fpaul-1A, line 524: according to the docs, we should add `nx-cloud` devDependency for it to work on Nx <= 19.6 (https://github.com/nrwl/nx/issues/28414 :'( )

kpanot: Not sure to understand where in the doc it flags `nx-cloud` as mandatory devDep (and it is working locally today without it).
`nx-cloud` is convenient to register global on your machine but it is not mandatory in the dev in the repo (as the examples in the docs are `npx nx-cloud ...` based

fpaul-1A: image

kpanot: Indeed :S
It's working without it today but let's add it then

.github/workflows/main.yml
```diff
@@ -21,11 +21,14 @@ on:
...
 env:
   YARN_ENABLE_HARDENED_MODE: 0
+  NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
 
 jobs:
   # Check the integrity of yarn lock
   yarn_lock_check:
     runs-on: ubuntu-latest
+    env:
+      NX_SKIP_NX_CACHE: true
```
fpaul-1A, line 31: Not related to your change but don't we want the Nx cache for this step?

fpaul-1A (approved):

.github/workflows/main.yml
kpanot: It is requires by `nx-cloud` runner when run in CI.
cf: https://github.com/AmadeusITGroup/otter/actions/runs/11702511447/job/32590842412

cpaulve-1A (approved):

PR #2414, opened 2024-11-06
feat(training): use markdown for step instructions

fpaul-1A:
## Proposed change

Instead of writing the trainings instructions in HTML, we can use markdown.
With this, we can easily enable the code highlighting.

Example:
https://fpaul-1a.github.io/otter/#/sdk-training#2
Generated from https://github.com/fpaul-1A/otter/blob/main/apps/showcase/src/assets/trainings/sdk/steps/typescript-sdk/instructions.md

image

## Related issues

*- No issue associated -*


apps/showcase/src/app/app.module.ts
```diff
@@ -110,7 +111,8 @@ export function registerCustomComponents(): Map<string, any> {
     StylingDevtoolsModule,
     LocalizationDevtoolsModule,
     ConfigurationDevtoolsModule,
-    MonacoEditorModule.forRoot()
+    MonacoEditorModule.forRoot(),
+    MarkdownModule.forRoot()
```
mrednic-1A, line 115: can't it be provided in the training route only, to have it lazy loaded ?

fpaul-1A: actually with standalone component I can just remove this import

sdo-1A (approved):

apps/showcase/src/helpers/marked-alert-extension.ts
```diff
@@ -0,0 +1,89 @@
...
+      }
+    },
+    extensions: [
+      {
+        name: 'alert',
+        level: 'block',
+        renderer({ meta, tokens = [] }) {
+          let tmpl = `<div dfInsertAlertIcon class="${meta.className} ${meta.className}-${meta.variant}">\n`;
+          tmpl += `<p class="${meta.titleClassName}">`;
+          tmpl += meta.icon;
```
matthieu-crouzet, line 78: maybe we could have the code like that to avoid duplication of the `<span>`
```suggestion
          tmpl += `<span class="df-insert-icon ${meta.icon}"></span>`;
```
```ts
const iconVariantMapping: Record<string, string> = {
  note: 'fa-info-circle',
  tip: 'fa-lightbulb',
  important: 'fa-info-circle',
  warning: 'fa-exclamation-triangle',
  caution: 'fa-minus-circle'
};

File: apps/showcase/src/app/app.module.ts [hunk 3]
```diff
@@ -27,10 +27,12 @@ import { ConsoleLogger, Logger, LOGGER_CLIENT_TOKEN, LoggerService } from '@o3r/
 import { OTTER_RULES_ENGINE_DEVTOOLS_OPTIONS, RulesEngineRunnerModule } from '@o3r/rules-engine';
 import { OTTER_STYLING_DEVTOOLS_OPTIONS, StylingDevtoolsModule } from '@o3r/styling';
 import { HIGHLIGHT_OPTIONS } from 'ngx-highlightjs';
+import { CLIPBOARD_OPTIONS, provideMarkdown } from 'ngx-markdown';
 import { MonacoEditorModule } from 'ngx-monaco-editor-v2';
-import { DatePickerHebrewInputPresComponent, ScrollBackTopPresComponent, SidenavPresComponent } from '../components/utilities';
+import { ClipboardButtonPresComponent, DatePickerHebrewInputPresComponent, ScrollBackTopPresComponent, SidenavPresComponent } from '../components/utilities';
 import { AppRoutingModule } from './app-routing.module';
 import { AppComponent } from './app.component';
+import { markedAlert } from '../helpers/marked-alert-extension';
```
sdo-1A, line 35: order of imports?

matthieu-crouzet (approved):

apps/showcase/src/app/app.module.ts
sdo-1A: nevermind, it will automatically be fixed

PR #2419, opened 2024-11-07
chore: exclude build.gradle.kts from OxyFlax codeowners setup for intellij-extension folder

OxyFlax:
## Proposed change

Removing my ownership of the build.gradle.kts file to avoid blocking the merges for simple version change

## Related issues

Currently without this change I'm being requested to review PRs for just renovate version bumps

*- No issue associated -*

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -14,3 +14,4 @@
 
 ## Default reviewers for Intellij extension
 /apps/intellij-extension/ @OxyFlax
+/apps/intellij-extension/build.gradle.kts @AmadeusITGroup/otter_admins


matthieu-crouzet (approved):

.github/CODEOWNERS
```diff
@@ -14,3 +14,4 @@
 
 ## Default reviewers for Intellij extension
 /apps/intellij-extension/ @OxyFlax
+/apps/intellij-extension/build.gradle.kts
```
fpaul-1A, line 17: I'm not sure how the codeowners format works, shouldn't it be:
```suggestion
/apps/intellij-extension/build.gradle.kts @AmadeusITGroup/otter_admins
```
?
From what I seem to understand from the examples of the doc, lefting it empty will override the default

OxyFlax: Good catch!

kpanot (approved):

matthieu-crouzet (approved):

PR #2429, opened 2024-11-08
fix(telemetry): detach fetch process

kpanot:
## Proposed change

Detach report sending process from Node main process to avoid long fetch process to lock the main thread

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@o3r/telemetry/src/sender/index.ts
+++ b/packages/@o3r/telemetry/src/sender/index.ts
@@ -55,19 +55,32 @@ export type SendDataFn = (data: MetricData, logger?: { error: (msg: string) => v
 /**
  * Send metric to a Amadeus Log Server
  * @param data Metrics to report
- * @param _logger Optional logger to provide to the function
- * @param _logger.error
+ * @param logger Optional logger to provide to the function
  */
-export const sendData: SendDataFn = async (data: MetricData, _logger?: { error: (msg: string) => void }) => {
-  const message = JSON.stringify(data);
-  const body = JSON.stringify({
-    messages: [{
-      applicationName: 'OTTER',
-      message
-    }]
-  });
-  await fetch('https://uat.digital-logging.saas.amadeus.com/postUILogs', {
-    method: 'POST',
-    body
-  });
+export const sendData: SendDataFn = (data, logger) => {
+  let body!: string;
+  try {
+    const message = JSON.stringify(data);
+    body = JSON.stringify({
+      messages: [{
+        applicationName: 'OTTER',
+        message
+      }]
+    });
+  } catch (e: any) {
+    return Promise.reject(e);
+  }
+
+  setTimeout(() => {
+    void fetch('https://uat.digital-logging.saas.amadeus.com/postUILogs', {
+      method: 'POST',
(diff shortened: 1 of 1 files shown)


packages/@o3r/telemetry/src/sender/index.ts
```diff
@@ -55,19 +55,32 @@ export type SendDataFn = (data: MetricData, logger?: { error: (msg: string) => v
...
+  try {
+    const message = JSON.stringify(data);
+    body = JSON.stringify({
+      messages: [{
+        applicationName: 'OTTER',
+        message
+      }]
+    });
+  } catch (e: any) {
+    return Promise.reject(e);
```
matthieu-crouzet, line 71: Linter will complain that e is maybe not an error 
If you can disable it please 
https://eslint.org/docs/latest/rules/prefer-promise-reject-errors

matthieu-crouzet: maybe we can prefer doing 
```suggestion
   const err = (e instanceof Error ? e : new Error(e));
   logger?.error(err.stack || err.toString());
```
and do not throw at all 
and remove all `sendData(...).catch` in the different handler as we don't want the sendData to throw

kpanot: Not really, here we clearly want to reject.
In case of `async function` a throw would have done the job (and we would have done it) but we are in a simple function and we have consumer calling the `.reject(...)` to the return of this function.

For my point of view, if the linter complain about that, we should or ignore this line or change the rule complaining to it.

matthieu-crouzet: I don't get the point of why we want to reject ?
no one should be looking for the success or failure of that function as we don't care if it succeed or not, no ?

matthieu-crouzet: That's my point we should remove that code in every handler and only set it once in this method and do not reject

kpanot: Not needed on this bug fix.
I don't want to do additional unnecessary change on a bug fix to please a future linter rule.

matthieu-crouzet (approved):

packages/@o3r/telemetry/src/sender/index.ts
```diff
@@ -55,19 +55,32 @@ export type SendDataFn = (data: MetricData, logger?: { error: (msg: string) => v
...
+
+  setTimeout(() => {
+    void fetch('https://uat.digital-logging.saas.amadeus.com/postUILogs', {
+      method: 'POST',
+      body
+    }).catch((e) => {
+      const err = (e instanceof Error ? e : new Error(e));
+      logger?.error(err.stack || err.toString());
+    });
+  }, 1).unref();
```
fpaul-1A, line 82:
```suggestion
  }, 0).unref();
```
should work too, right?

kpanot: According to https://nodejs.org/api/timers.html#settimeoutcallback-delay-args `0` will be defaulted to `1`

PR #2473, opened 2024-11-15
feat(cascading): option to bypass non-latest minor branch

kpanot:
## Proposed change

Option to bypass non-latest minor branch on cascading bot

## Related issues

*- No issue associated -*

### DIFF
--- a/.github/.cascadingrc.json
+++ b/.github/.cascadingrc.json
@@ -1,6 +1,5 @@
 {
   "$schema": "../apps/github-cascading-app/schemas/config.schema.json",
-  "bypassReviewers": true,
   "labels": ["cascading"],
   "ignoredPatterns": [
     "-next$"
--- a/apps/github-cascading-app/schemas/config.schema.json
+++ b/apps/github-cascading-app/schemas/config.schema.json
@@ -31,6 +31,11 @@
       "description": "Pattern determining if the branch is part of the cascading strategy",
       "default": "^releases?/\\d+\\.\\d+"
     },
+    "onlyCascadeOnHighestMinors": {
+      "type": "boolean",
+      "description": "Determine if the branches for which a higher minor version exists should be skipped during the cascading",
+      "default": false
+    },
     "versionCapturePattern": {
       "type": "string",
       "description": "Pattern containing a capture to extract the version of a cascading branch",
--- a/apps/github-cascading-app/src/cascading/cascading.spec.ts
+++ b/apps/github-cascading-app/src/cascading/cascading.spec.ts
@@ -207,6 +207,91 @@ describe('Cascading Application', () => {
       expect(logger.info).toHaveBeenCalledWith('The branch test-cascading/1.0 is the last branch of the cascading. The process will stop.');
     });
 
+    describe('on onlyCascadeOnHighestMinors options', () => {
+      it('should ignore branches not to the latest minor when true', async () => {
+        customization.loadConfiguration = customization.loadConfiguration.mockResolvedValue({
+          ...DEFAULT_CONFIGURATION,
+          cascadingBranchesPattern: 'test-cascading/.*',
+          onlyCascadeOnHighestMinors: true,
+          ignoredPatterns: []
+        });
+        customization.getBranches = customization.getBranches.mockResolvedValue([
+          'test-cascading/1.0',
+          'test-cascading/1.1',
+          'test-cascading/1.2',
+          'test-cascading/2.0'
+        ]);
+        customization.isBranchAhead = customization.isBranchAhead.mockResolvedValue(true);
+        customization.createBranch = customization.createBranch.mockResolvedValue();
+        customization.getPullRequests = customization.getPullRequests.mockResolvedValue([]);
+        customization.createPullRequest = customization.createPullRequest.mockResolvedValue({
+          id: 1,
+          originBranchName: '',
+          isOpen: true,
+          mergeable: true,
+          body: render(mockBasicTemplate, { isConflicting: false, targetBranch: 'main', currentBranch: 'release/0.1', bypassReviewers: true }, { async: false })
+        });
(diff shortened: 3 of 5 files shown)


apps/github-cascading-app/schemas/config.schema.json
```diff
@@ -31,6 +31,11 @@
       "description": "Pattern determining if the branch is part of the cascading strategy",
       "default": "^releases?/\\d+\\.\\d+"
     },
+    "excludeBranchesWithHigherMinor": {
```
fpaul-1A, line 34: I find this name very confusing, it seems that it's actually doing the exact opposite `onlyCascadeOnHighestMinors`

apps/github-cascading-app/schemas/config.schema.json
```diff
@@ -31,6 +31,11 @@
       "description": "Pattern determining if the branch is part of the cascading strategy",
       "default": "^releases?/\\d+\\.\\d+"
     },
+    "onlyCascadeOnHighestMinors": {
+      "type": "boolean",
+      "description": "Determine if the branches, for which an higher minor version exists, should not not be targeted by then cascading pull requests",
```
fpaul-1A, line 36:
```suggestion
      "description": "Determine if the branches for which a higher minor version exists should be skipped during the cascading",
```

apps/github-cascading-app/src/cascading/cascading.ts
```diff
@@ -15,6 +15,9 @@ export const CASCADING_BRANCH_PREFIX = 'cascading';
 /** Time (in ms) to wait before re-checking the mergeable status of a PR */
 export const RETRY_MERGEAGLE_STATUS_CHECK_TIMING = 3000;
 
+/** Object representing a branch with the version determine from its name */
```
fpaul-1A, line 18:
```suggestion
/** Object representing a branch with the version determined from its name */
```

fpaul-1A (dismissed):

apps/github-cascading-app/src/cascading/cascading.spec.ts
```diff
@@ -198,6 +198,93 @@ describe('Cascading Application', () => {
...
+        ]);
+        customization.isBranchAhead = customization.isBranchAhead.mockResolvedValue(true);
+        customization.createBranch = customization.createBranch.mockResolvedValue();
+        customization.getPullRequests = customization.getPullRequests.mockResolvedValue([]);
+        customization.createPullRequest = customization.createPullRequest.mockResolvedValue({
+          id: 1,
+          originBranchName: '',
+          isOpen: true,
+          mergeable: true,
+          body: render(mockBasicTemplate, { isConflicting: false, targetBranch: 'main', currentBranch: 'release/0.1', bypassReviewers: true }, { async: false })
```
cpaulve-1A, line 224: release/0.1 is not part of the list of supported branches line 212

mrednic-1A (dismissed):

apps/github-cascading-app/src/cascading/cascading.spec.ts
kpanot: not a problem it is unused mock to unlock the process (it is the object to generate the PR template, so not the purpose of this test :))

cpaulve-1A (approved):

apps/github-cascading-app/src/cascading/interfaces.ts
```diff
@@ -34,7 +36,10 @@ export interface PullRequestContext {
   currentBranch: string;
   /** Cascading Pull Request Target Branch */
   targetBranch: string;
-  /** Determine if the reviewers are bypassed */
+  /**
+   * Determine if the reviewers are bypassed
+   * Note: This option is not supported on Github anymore due to Github Api change.
```
matthieu-crouzet, line 41: Why are not we deprecating it ?

.github/.cascadingrc.json
```diff
@@ -2,6 +2,7 @@
   "$schema": "../apps/github-cascading-app/schemas/config.schema.json",
   "bypassReviewers": true,
   "labels": ["cascading"],
+  "onlyCascadeOnHighestMinors": true,
```
cpaulve-1A, line 5: We should probably not activate it for us

.github/.cascadingrc.json
```diff
@@ -2,6 +2,7 @@
   "$schema": "../apps/github-cascading-app/schemas/config.schema.json",
   "bypassReviewers": true,
```
vscaiceanu-1a, line 3: this can be removed

vscaiceanu-1a (commented):
Is there a case where we want to skip non-latest minors?
Wouldn't this create incoherence and confusion among the different branches / code levels?

apps/github-cascading-app/src/cascading/interfaces.ts
kpanot: I did not deprecate this option because I think that as the cascading is already today platform agnostic (event if it is implemented only for Github) and this option still make sense for supporting platform, it would be weird to remove it in 1 year to follow the unique implementation to potentially seeing it reappear after a while because of a new implementation.
I did make more sense for me to keep it and warn the Github users instead, but I am OK to change if needed.

PS: To implement another platform support, there is no need to extract code, it's the bot that should switch the implementation according to the received message. So today it is already ready to support multi platform as it is

kpanot:
> Is there a case where we want to skip non-latest minors? Wouldn't this create incoherence and confusion among the different branches / code levels?

Not sure to understand why.
As for the policy, we considered the latest minor branches as supported/maintained branches. Which mean that all the other branches are considered as archived (if they are not destroyed).
When we cascade, we want to do it only on the considered branched automatically, why would it create "incoherence" to skip unmaintained branches?

fpaul-1A (approved):

mrednic-1A (approved):

PR #2475, opened 2024-11-15
fix: make typings more accurate by avoiding explict cast on consts

kpanot:
## Proposed change

When creating an object that is intended to be read-only, we should let typescript infer the type.

Not good:
```typescript
interface NamedObjectMaybe {
  name?: string;
}
const UNMOVABLE_OBJECT: NamedObjectMaybe = {
  name: 'unicorn'
};

// UNMOVABLE_OBJECT.name is typed as string | undefined
```

Nice:
```typescript
interface NamedObjectMaybe {
  name?: string;
}
const UNMOVABLE_OBJECT = {
  name: 'unicorn'
} as const satisfies NamedObjectMaybe;

// UNMOVABLE_OBJECT.name is typed as 'unicorn' but we still validate that UNMOVABLE_OBJECT matches the interface
```
## Related issues

* :octocat: Pull Request #2271


fpaul-1A (approved):

sdo-1A (approved):

mrednic-1A (dismissed):

matthieu-crouzet:
As discussed wait for https://github.com/AmadeusITGroup/otter/pull/2192 before merging this PR

fpaul-1A (dismissed):

apps/chrome-devtools/src/services/state.service.ts
```diff
@@ -207,11 +207,11 @@ export class StateService {
         ...(override ? undefined : state.stylingVariables),
         ...changes.stylingVariables
       }).filter((entry): entry is [string, string] => entry[1] !== null));
-      const stateOverrides: StateOverride = {
+      const stateOverrides = {
         configurations: Object.keys(configurationOverrides).length > 0 ? configurationOverrides : undefined,
         localizations: Object.keys(localizationOverrides || {}).length > 0 ? localizationOverrides : undefined,
         stylingVariables: Object.keys(stylingOverrides).length > 0 ? stylingOverrides : undefined
-      };
+      } satisfies StateOverride;
```
matthieu-crouzet, line 214: Why not `as const satifies StateOverride` for this one ?

apps/showcase/src/components/showcase/configuration/configuration-pres.config.ts
```diff
@@ -45,14 +45,14 @@ export interface ConfigurationPresConfig extends Configuration {
   shouldProposeRoundTrip: boolean;
 }
 
-export const CONFIGURATION_PRES_DEFAULT_CONFIG: ConfigurationPresConfig = {
+export const CONFIGURATION_PRES_DEFAULT_CONFIG: Readonly<ConfigurationPresConfig> = {
```
matthieu-crouzet, line 48: Shouldn't we adapt the generator as well ? (maybe a ng update)

apps/showcase/src/components/showcase/localization/localization-pres.translation.ts
```diff
@@ -33,12 +33,12 @@ export interface LocalizationPresTranslation extends Translation {
   cityName: string;
 }
 
-export const translations: LocalizationPresTranslation = {
+export const translations: Readonly<LocalizationPresTranslation> = {
```
matthieu-crouzet, line 36: same remark as for configuration

packages/@o3r/analytics/src/services/event-track/event-track.service.ts
```diff
@@ -44,12 +44,12 @@ import {
 } from './event-track.configuration';
 
 /** The initial value of the performance measurements */
-export const performanceMarksInitialState: PerfEventPayload = {
+export const performanceMarksInitialState: Readonly<PerfEventPayload> = {
```
matthieu-crouzet, line 47: Why do we put `Readonly` on some but not on all of the `as const` ?

apps/showcase/src/components/showcase/configuration/configuration-pres.config.ts
matthieu-crouzet: Could be done in another PR but at least we should open an issue to not forget it

packages/@o3r/analytics/src/services/event-track/event-track.service.ts
kpanot: Not sure I got the question :S.
`readonly<...>` is mandatory when the constant is declared `as const` but, as it is exposed, should also have an explicit type following an interface.
When there are not exported, it does not need to have the type defined but only to `satisfy` it. So no need `Readonly<...>`.

apps/showcase/src/components/showcase/configuration/configuration-pres.config.ts
kpanot: Issue opened: #2530

apps/showcase/src/components/showcase/localization/localization-pres.translation.ts
kpanot: Issue opened: #2530

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #2486, opened 2024-11-19
feat: move schematic to package dependencies to simplify ng-add

kpanot:
## Proposed change

move @o3r/schematic package to packages dependencies to simplify ng-add and supporting it outside of Otter project.

## Related issues

*- No issue associated -*


kpanot:
Temporary closed due to the large amount of dependencies bring by it

PR #2491, opened 2024-11-20
fix: add missing devDep to schematics

kpanot:
## Proposed change

fix: add missing devDep to schematics

## Related issues

*- No issue associated -*


cpaulve-1A (dismissed):

packages/@o3r/schematics/src/rule-factories/ng-add/ng-add.helpers.ts
```diff
@@ -3,10 +3,10 @@ import { NodePackageInstallTask } from '@angular-devkit/schematics/tasks';
 import { NodeDependencyType } from '@schematics/angular/utility/dependencies';
 import { mkdirSync, readFileSync, writeFileSync } from 'node:fs';
 import * as path from 'node:path';
-import { lastValueFrom } from 'rxjs';
 import type { PackageJson } from 'type-fest';
 import type { NgAddPackageOptions } from '../../tasks/index';
 import { getPackageManager } from '../../utility/index';
+import { lastValueFrom } from 'rxjs';
```
fpaul-1A, line 9: Why this change?

packages/@o3r/stylelint-plugin/schematics/ng-add/index.ts
```diff
@@ -32,7 +33,7 @@ function ngAddFn(options: NgAddSchematicsSchema): Rule {
       setupDependencies
     } = await import('@o3r/schematics');
     // eslint-disable-next-line @typescript-eslint/naming-convention
-    const { NodeDependencyType } = await import('@schematics/angular/utility/dependencies');
+    const { NodeDependencyType } = await import('@schematics/angular/utility/dependencies').catch(() => ({ NodeDependencyType: { Dev: 'devDependencies' as NodeDependencyType.Dev } }));
```
fpaul-1A, line 36: We should probably avoid shadowing the type `NodeDependencyType` with the variable of the same name

kpanot: Actually only the string value is used on this one but changing the interface would be a breaking change :S

fpaul-1A: I'm just talking about renaming the const, how would that be a breaking change?

kpanot: Oooh sorry, I just understand your point.
Actually there are the same from the same import except that on is a type in sync and the impl in dynImport.
renaming one or the other would be almost more confusing, don't you think?
(The linter does not complain)

vscaiceanu-1a (approved):

packages/@o3r/application/schematics/ng-add/index.ts
```diff
@@ -44,15 +44,20 @@ function ngAddFn(options: NgAddSchematicsSchema): Rule {
...
       );
 
-      if (isImported(sourceFile, 'prefersReducedMotion', '@o3r/application')) {
-        context.logger.info('[LOG]: prefersReducedMotion from @o3r/application is already imported.');
+      if (!sourceFile) {
+        context.logger.warn('No Typescript executor detected, the ng-add process will be skipped.');
+        return tree;
+      }
+
+      if (isImported?.(sourceFile, 'prefersReducedMotion', '@o3r/application')) {
```
matthieu-crouzet, line 59: Shouldn't we log a warning if `isImported` is `undefined`

packages/@o3r/stylelint-plugin/schematics/ng-add/index.ts
matthieu-crouzet: The new one maybe ! :)

packages/@o3r/schematics/src/rule-factories/ng-add/ng-add.helpers.ts
```diff
@@ -121,7 +120,10 @@ export function ngAddPackages(packages: string[], options?: Omit<NgAddPackageOpt
         hideOutput: false,
         quiet: false
       } as any));
-      await lastValueFrom(context.engine.executePostTasks());
+
+      await new Promise<void>((resolve, reject) =>
```
fpaul-1A, line 124: why is that needed?
by the look of it, the only difference is that we hide the exception on error

fpaul-1A (approved):

packages/@o3r/new-version/package.json
```diff
@@ -45,7 +45,9 @@
     "@nx/jest": "~19.5.0",
     "@o3r/build-helpers": "workspace:^",
     "@o3r/eslint-plugin": "workspace:^",
+    "@o3r/schematics": "workspace:^",
     "@o3r/test-helpers": "workspace:^",
+    "@schematics/angular": "~18.2.0",
```
matthieu-crouzet, line 50: Why should we add this one ?

fpaul-1A (approved):

packages/@o3r/new-version/package.json
kpanot: It's a mandatory peer dependency of `@o3r/schematics`

matthieu-crouzet (approved):

PR #2496, opened 2024-11-21
chore: clean Nx config

kpanot:
## Related issues

*- No issue associated -*

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -25,16 +25,13 @@
       "project",
       "{workspaceRoot}/yarn.lock",
       "{workspaceRoot}/tsconfig.base.json",
+      "{workspaceRoot}/tsconfig.build.json",
       "{workspaceRoot}/nx.json",
       "{workspaceRoot}/.nxignore",
       "{workspaceRoot}/tools/@o3r/build-helpers/scripts/*.mjs"
     ],
-    "base-build": [
-      "base",
-      "{workspaceRoot}/tsconfig.build.json"
-    ],
     "source": [
-      "base-build",
+      "base",
       "{projectRoot}/tsconfig.build.json",
       "{projectRoot}/tsconfig.build.*.json",
       "{projectRoot}/src/**/*",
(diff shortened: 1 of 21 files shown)


packages/@o3r/telemetry/project.json
```diff
@@ -26,12 +26,8 @@
       "options": {
         "main": "packages/@o3r/telemetry/src/public_api.ts",
         "tsConfig": "packages/@o3r/telemetry/tsconfig.build.json",
-        "outputPath": "packages/@o3r/telemetry/dist",
-        "clean": false
```
fpaul-1A, line 30: are you sure there is no problem removing that?
there are several tasks writing on `dist` and I remember we had issues in the past where some files were missing on `dist`

tools/@o3r/build-helpers/project.json
```diff
@@ -3,11 +3,5 @@
...
   "projectType": "library",
   "sourceRoot": "packages/@o3r/build-helpers/scripts",
-  "tags": [],
-  "targets": {
-    "build": {
-      "executor": "nx:noop",
-      "dependsOn": []
-    }
-  }
+  "tags": []
```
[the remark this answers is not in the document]
kpanot: Sorry I forgot to flag the PR on `Draft`.
Yes, the issue is still present (not on Nx Cache but build directly).
The issue is that when a package does not have `build` task, it cannot be excluded via `!` in `explicitDependencies`

packages/@o3r/telemetry/project.json
kpanot: Sorry I forgot to turn the PR as draft from the beginning.
Yes, it is an issue for the cache re-hydration, the problem is when the `compile` task is re-run but not the `builder` ones.
I got issue with wrong cache-miss and through it was because of it but after investigation it was mainly due to `schemas`. So I rolled back this change and created a dedicated task to schemas copies (when not part of `build` directly)

nx.json
```diff
@@ -397,7 +415,7 @@
       "cache": true
     },
     "prepare": {
-      "cache": true,
```
matthieu-crouzet, line 400: why is it not cached anymore ?

fpaul-1A (approved):

packages/@o3r/design/package.json
```diff
@@ -32,7 +32,7 @@
       "default": "./schemas/*.json"
     },
     ".": {
-      "es2020": "./dist/src/public_api.js",
+      "module": "./dist/src/public_api.js",
       "default": "./dist/src/public_api.js",
```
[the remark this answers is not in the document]
kpanot: Maybe it should be fixed in all packages then :S.
Here I renamed it because I got a warning saying that `es2020` is not valid

PR #2499, opened 2024-11-22
feat(training): loading progress indicator

fpaul-1A:
## Proposed change

{5A8C1177-3106-49F4-9B4F-93F27B019926}

## Related issues

*- No issue associated -*

### DIFF
--- a/apps/showcase/src/services/webcontainer/webcontainer-runner.ts
+++ b/apps/showcase/src/services/webcontainer/webcontainer-runner.ts
@@ -1,16 +1,22 @@
-import { DestroyRef, inject, Injectable } from '@angular/core';
+import { DestroyRef, inject, Injectable, signal } from '@angular/core';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { type FileSystemTree, type IFSWatcher, WebContainer, type WebContainerProcess } from '@webcontainer/api';
 import { Terminal } from '@xterm/xterm';
 import {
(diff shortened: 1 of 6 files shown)


kpanot (approved):

apps/showcase/src/services/webcontainer/webcontainer-runner.ts
```diff
@@ -147,10 +174,35 @@ export class WebContainerRunner {
     const process = await instance.spawn(command, args, {cwd: cwd});
     this.commandOutput.process.next(process);
     const exitCode = await process.exit;
-    if (exitCode !== 0) {
+    if (exitCode === 0) {
```
matthieu-crouzet, line 177: Shouldn't we prefer a switch case, in case we have to handle others exitCode or it should never happen ?

matthieu-crouzet (approved):

apps/showcase/src/services/webcontainer/webcontainer-runner.ts
fpaul-1A: I think it's a matter of preference, I don't see any good reason to choose one or the other.
I could argue that with `if` we can more easily include ranges `if (existCode >= 100 && exitCode < 200)`
you can also do it with `switch(true) { case (existCode >= 100 && exitCode < 200): ... }` but at the end it looks more like an `if` than a `switch/case`

sdo-1A (approved):

PR #2502, opened 2024-11-22
fix(eslint): fix typescript import resolution

matthieu-crouzet:


apps/vscode-extension/src/intellisense/configuration.ts
```diff
@@ -8,12 +14,6 @@ import type {
 import {
   ESLint,
 } from 'eslint';
```
fpaul-1A, line 16: the imports don't seem to be in alphabetical order

matthieu-crouzet: https://github.com/AmadeusITGroup/otter/pull/2502/files#diff-4bc2155348e1323d934d8ba058c7cfdef2426d74133eb1281fcc81b1f66e76b9R28
this flag the import as `core`
import are not only in alphabetical order but also following ["builtin", "external", "internal", "parent", "sibling", "index", "object", "type"]
https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md

kpanot (approved):

PR #2508, opened 2024-11-25
fix(eslint): fix remaining linter issues

matthieu-crouzet:


packages/@o3r-training/showcase-sdk/eslint.local.config.mjs
```diff
@@ -13,14 +13,10 @@ export default [
   {
     name: '@o3r-training/showcase-sdk/ignores',
     ignores: [
-      'src/api',
-      'src/models/base',
-      'src/spec/api-mocks.ts',
-      'src/spec/operation-adapter.ts',
-      'dist',
```
fpaul-1A, line 20: I assume this is already ignored?

fpaul-1A (approved):

packages/@o3r-training/showcase-sdk/eslint.local.config.mjs
matthieu-crouzet: yes already ignored in the shared config at the root of the monorepo

fpaul-1A (approved):

cpaulve-1A (approved):

PR #2510, opened 2024-11-25
fix: clean unreferenced cached values of translations and configs

### LINKED ISSUE #2509: [Bug]: shareReplay keeps cached value in memory
### Package name

configuration, localization

### Package version

10.2

### Reproduction steps

Use the translation pipe or function often in the app

### Current result

Translation results are kept in memory even though they have no one that is observing them

### Expected result

Memory should be cleaned up :)

### Additional comments

We had been running into some issues in regards to memory consumption and in the traces I have found that lots of objects were related to translations. For about 6 months I had overridden the `translate` function from `LocalizationService` with the addition of the `refCount: true` to `shareReplay`. As a result we observed a drop of memory usage (when using our app in a long flow), specifically for the OperatorSubscriber (that is used internally by the `shareReplay` to keep the value "alive").

(Value on the left with `x22170` and `x77596` are the amount of objects retained and the rightmost value indicated the retained size in bytes)
Before override:
Screenshot 2024-11-25 at 09 14 13

After override:
Screenshot 2024-11-25 at 09 14 37

Functionally speaking we have not observed any bugs in these months, so for the translate change I'm very confident to apply that change. For the configuration I've seen that some more `shareReplay(1)` exist, their footprint is minimal since we call them once per component, therefore they are almost negligible. However I think they should be cleaned up as well when a component is no longer observing these values. There are some more occurences of this type of caching in the code, but I didn't check if they could benefit us.

ntrotner:
## Proposed change
If translation or config is no longer subscribed to the value should be cleaned up

## Related issues

* :bug: Fix #2509

### DIFF
--- a/packages/@o3r/configuration/src/core/configuration.ts
+++ b/packages/@o3r/configuration/src/core/configuration.ts
@@ -58,6 +58,6 @@ export function getConfiguration<T extends Record<string, unknown>>(defaultValue
         return deepFill(defaultValue, configOverride);
       }),
       distinctUntilChanged((prev, current) => JSON.stringify(prev) === JSON.stringify(current)),
-      shareReplay(1)
+      shareReplay({ refCount: true, bufferSize: 1 })
     );
 }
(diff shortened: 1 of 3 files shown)


fpaul-1A:
you may want to change the target branch if you want the fix on `release/11.4` or `release/11.5.0-rc`
if you target `main`, the fix will be available on `11.6`

ntrotner:
@fpaul-1A thx for the info, I've rebased it to 11.4

fpaul-1A (approved):

kpanot (approved):

PR #2516, opened 2024-11-26
fix(rules-engine): allow custom action in the rulesets schema

matthieu-crouzet:
## Proposed change

Today if we create a custom action handler, we can't write a `rulesets.json` validating the schema exposed.


packages/@o3r/rules-engine/schemas/rulesets.schema.json
```diff
@@ -166,11 +166,31 @@
...
+      "description": "Content of a custom action",
+      "properties": {
+        "actionType": {
+          "type": "string",
+          "description": "Action Type",
+          "pattern": "^(?!(?:SET_FACT|UPDATE_PLACEHOLDER|UPDATE_CONFIG|UPDATE_ASSET|UPDATE_LOCALISATION)$).*$"
+        },
+        "value": {}
+      },
+      "additionalItems": false,
```
deathpat, line 188: I think this deserves to be true

packages/@o3r/rules-engine/schemas/rulesets.schema.json
```diff
@@ -166,11 +166,31 @@
...
       ]
     },
+    "RulesEngineCustomActionBlock": {
+      "type": "object",
+      "description": "Content of a custom action",
+      "properties": {
+        "actionType": {
+          "type": "string",
+          "description": "Action Type",
+          "pattern": "^(?!(?:SET_FACT|UPDATE_PLACEHOLDER|UPDATE_CONFIG|UPDATE_ASSET|UPDATE_LOCALISATION)$).*$"
```
kpanot, line 184: Unfortunately you cannot do it that way without hardcoding value defined in other packages.
There is no way to do a reference instead?

packages/@o3r/rules-engine/schemas/rulesets.schema.json
```diff
@@ -166,11 +166,31 @@
...
+        },
+        "value": {}
+      },
+      "additionalItems": true,
+      "required": [
+        "actionType",
+        "value"
```
kpanot, line 191: Not sure it should be required

matthieu-crouzet: I think it has to be specified but with any value 
image
image

fpaul-1A (approved):

kpanot (approved):

PR #2518, opened 2024-11-26
deprecate(rules-engine): actionHandlers will become protected in v13

matthieu-crouzet:
## Proposed change
It's not a good practice to expose a `Set`.
I added a method `registerActionHandlers` to replace the usage of `actionHandlers.add(...)`.
It will be future proof if we want to do something else when an action handler is registered.

### DIFF
--- a/packages/@o3r/rules-engine/schematics/ng-update/v11.6/use-register-action-handlers.spec.ts
+++ b/packages/@o3r/rules-engine/schematics/ng-update/v11.6/use-register-action-handlers.spec.ts
@@ -1,0 +1,100 @@
+import { callRule, type SchematicContext, Tree } from '@angular-devkit/schematics';
+import { lastValueFrom } from 'rxjs';
+import { useRegisterActionHandlers } from './use-register-action-handlers';
+
+let initialTree: Tree;
+let tree: Tree;
+const untouchedFile = 'untouchedFile.ts';
+const fileWithActionHandlerAdd = 'fileWithActionHandlerAdd.ts';
+const fileWithActionHandlerDelete = 'fileWithActionHandlerDelete.ts';
+const multiplePresenceFile = 'multiplePresence.ts';
+
+describe('useRegisterActionHandlers', () => {
+  beforeEach(() => {
+    initialTree = Tree.empty();
+  });
+
+  it('should not touch files without references', async () => {
+    initialTree.create(untouchedFile, 'export {};');
+    tree = await lastValueFrom(callRule(useRegisterActionHandlers, initialTree, {} as SchematicContext));
+    expect(tree.readText(untouchedFile)).toBe(initialTree.readText(untouchedFile));
+  });
+
+  it('should change file with a reference to actionHandlers.add', async () => {
+    initialTree.create(fileWithActionHandlerAdd, `
+      import {inject, runInInjectionContext} from '@angular/core';
+      import {RulesEngineRunnerService} from '@o3r/rules-engine';
+      import {appConfig} from './app/app.config';
+      import {AppComponent} from './app/app.component';
+      import {PopupActionHandler} from './services/popup-action-handler';
+
+      bootstrapApplication(AppComponent, appConfig)
+        .then((m) => {
+          runInInjectionContext(m.injector, () => {
+            inject(RulesEngineRunnerService);
+            ruleEngine.actionHandlers.add(inject(PopupActionHandler));
+          });
+        })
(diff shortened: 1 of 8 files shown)


packages/@o3r/rules-engine/src/services/runner/rules-engine.runner.service.ts
```diff
@@ -119,29 +122,37 @@ export class RulesEngineRunnerService implements OnDestroy {
...
    */
   public upsertRulesets(ruleSets: Ruleset[]) {
     this.store.dispatch(setRulesetsEntities({entities: ruleSets}));
   }
 
+  /**
+   * Add action handlers in the rules engine
+   * @param actionHandlers
+   */
+  public registerActionHandlers(...actionHandlers: RulesEngineActionHandler[]) {
```
kpanot, line 152: To be iso you should propose an `unregister`

kpanot (approved):

packages/@o3r/rules-engine/schematics/ng-update/v11.6/use-register-action-handlers.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  it('should not touch files without references', async () => {
+    expect(initialTree.readText(untouchedFile)).not.toContain('actionHandlers.add');
+    expect(initialTree.readText(untouchedFile)).not.toContain('actionHandlers.delete');
+    tree = await lastValueFrom(callRule(useRegisterActionHandlers, initialTree, {} as SchematicContext));
+    expect(tree.readText(untouchedFile)).toBe(initialTree.readText(untouchedFile));
+  });
+
+  it('should change file with a reference to actionHandlers.add', async () => {
+    expect(initialTree.readText(fileWithActionHandlerAdd)).toContain('actionHandlers.add');
```
fpaul-1A, line 81: Isn't it weird to do an `expect` on the initial tree? We are just testing the code of the `beforeEach`?

packages/@o3r/rules-engine/schematics/ng-update/v11.6/use-register-action-handlers.ts
```diff
@@ -0,0 +1,20 @@
+import type { Rule, Tree } from '@angular-devkit/schematics';
+import { findFilesInTree } from '@o3r/schematics';
+
+/**
+ * Replace `actionHandlers.add` or `actionHandlers.delete` by `registerActionHandlers` or `unregisterActionHandlers` in file
+ * @param tree
+ */
+export const useRegisterActionHandlers: Rule = (tree: Tree) => {
+  const files = findFilesInTree(tree.root, (file) => /actionHandlers\.(add|delete)/.test(tree.readText(file)));
```
fpaul-1A, line 9: Shouldn't we be more restrictive on the regex to avoid impacting unrelated code?
```suggestion
  const files = findFilesInTree(tree.root, (file) => /\.actionHandlers\.(add|delete)/.test(tree.readText(file)));
```

fpaul-1A (approved):

packages/@o3r/rules-engine/schematics/ng-update/v11.6/use-register-action-handlers.spec.ts
matthieu-crouzet: I added it to ensure that it is well present in the initial tree 
because I did the `find & replace` after writing the test, so indeed  `actionHandlers.add` was never present

fpaul-1A (approved):

packages/@o3r/rules-engine/schematics/ng-update/v11.6/use-register-action-handlers.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+          runInInjectionContext(m.injector, () => {
+            inject(RulesEngineRunnerService);
+            ruleEngine.actionHandlers.add(inject(PopupActionHandler));
+          });
+        })
+        // eslint-disable-next-line no-console
+        .catch(err => console.error(err));
+    `);
+    tree = await lastValueFrom(callRule(useRegisterActionHandlers, initialTree, {} as SchematicContext));
+    expect(tree.readText(fileWithActionHandlerAdd)).not.toContain('actionHandlers.add');
```
fpaul-1A, line 42: you don't test that it contains `registerActionHandlers` ?

kpanot (approved):

PR #2521, opened 2024-11-27
feat: use takeUntilDestroyed instead of subscriptions to unsubscribe

### LINKED ISSUE #1874: [Feature]: Use takeUntilDestroyed instead of subscription array
### Context

Since angular v17 we can use `takeUntilDestroyed` from `@angular/core/rxjs-interop`

### Proposal

We should replace all subscription array with the unsubscription in the `ngOnDestroy` by `takeUntilDestroy`.
We should modify our schematic that generate subscription array

matthieu-crouzet:
## Proposed change

Use `takeUntilDestroyed` instead of subscriptions to unsubscribe.

## Related issues

* :rocket: Feature resolves #1874

### DIFF
--- a/docs/analytics/TRACK_EVENTS.md
+++ b/docs/analytics/TRACK_EVENTS.md
@@ -197,9 +197,12 @@ You can mark the time the loading is rendered.
 ```typescript
 // app component
 ...
-  constructor(private router: Router, public trackEventsService: EventTrackService) {}
+  private readonly destroyRef = inject(DestroyRef);
+  private readonly router = inject(Router);
+  private readonly trackEventsService = inject(EventTrackService);
+
   ngOnInit() {
-    this.subscriptions.push(this.router.events.subscribe((event) => this.setLoadingIndicator(event)));
+    this.router.events.pipe(takeUntilDestroyed(this.destroyRef)).subscribe((event) => this.setLoadingIndicator(event));
     ...
   }
   setLoadingIndicator(event: Event) {
@@ -239,27 +242,34 @@ For example, on the availability page, mark _data ready_ when the calendar and o
 ```typescript
 // upsell page component
 ...
-export class UpsellComponent implements OnInit, OnDestroy, Configurable<UpsellConfig> {
+export class UpsellComponent implements OnInit, Configurable<UpsellConfig> {
   ...
-  constructor(public trackEventsService: EventTrackService, private store: Store<AirOffersStore & AirSearchCriteriaStore & CartStore & AirCalendarStore>) {
-    ...
-  }
+  private readonly destroyRef = inject(DestroyRef);
+  private readonly store = inject(Store<
+    AirOffersStore
+    & AirSearchCriteriaStore
+    & CartStore
+    & AirCalendarStore
+  >);
+
+  public readonly trackEventsService = inject(EventTrackService);
 
   ngOnInit() {
     const airCalendarReady$ = this.store.pipe(
         select(selectAirCalendarState),
+        takeUntilDestroyed(this.destroyRef),
         filter((state) => state.isPending === false && state.isFailure === false)
       );
     const airOffersReady$ = this.store.pipe(
       select(selectAirOffersIds),
+      takeUntilDestroyed(this.destroyRef),
       filter((ids) => !!ids.length)
     );
-    this.subscriptions.push(
-      combineLatest(airCalendarReady$, airOffersReady$)
-        .pipe(take(1))
-        .subscribe(([_airCalendar, _airOffersIds]) => {
-          this.trackEventsService.markDataReady();  /// ----> mark data ready when both calendar and offres data are in the store
-        }));
+    void firstValueFrom(
+      combineLatest([airCalendarReady$, airOffersReady$])
+    ).then(([_airCalendar, _airOffersIds]) => {
+      this.trackEventsService.markDataReady();  // ----> mark data ready when both calendar and offers data are in the store
+    });
   }
   ...
 }
--- a/packages/@o3r/components/src/tools/placeholder/placeholder.component.ts
+++ b/packages/@o3r/components/src/tools/placeholder/placeholder.component.ts
@@ -3,12 +3,17 @@ import {
   ChangeDetectionStrategy,
   ChangeDetectorRef,
   Component,
+  DestroyRef,
+  inject,
   Input,
   OnDestroy,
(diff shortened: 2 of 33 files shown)


docs/analytics/TRACK_EVENTS.md
```diff
@@ -197,9 +197,12 @@ You can mark the time the loading is rendered.
...
 // app component
 ...
-  constructor(private router: Router, public trackEventsService: EventTrackService) {}
+  private readonly destroyRef = inject(DestroyRef);
+  private readonly router = inject(Router);
+  private readonly trackEventsService = inject(EventTrackService);
+
   ngOnInit() {
-    this.subscriptions.push(this.router.events.subscribe((event) => this.setLoadingIndicator(event)));
+    this.router.events.pipe(takeUntilDestroyed(this.destroyRef)).subscribe((event) => this.setLoadingIndicator(event));
```
fpaul-1A, line 205: isn't `this.destroyRef` implicit ?

packages/@o3r/components/src/tools/component-replacement/c11n.directive.ts
```diff
@@ -95,8 +95,6 @@ export class C11nDirective<
     if (changes.component && changes.component.currentValue) {
 
       if (this.componentRef) {
-        this.componentSubscriptions.forEach((s) => s.unsubscribe());
```
fpaul-1A, line 98: this one might still be needed to cleanup the subscriptions when changing the `componentRef`

packages/@o3r/components/src/tools/placeholder/placeholder.component.ts
```diff
@@ -63,43 +65,42 @@ export class PlaceholderComponent implements OnInit, OnDestroy, AfterViewChecked
...
-              isPending: placeholders?.isPending
-            })),
-            distinctUntilChanged((previous, current) =>
-              previous.id === current.id &&
-              previous.isPending === current.isPending &&
-              previous.orderedTemplates?.map(placeholder => placeholder.renderedTemplate).join('') ===
-              current.orderedTemplates?.map(placeholder => placeholder.renderedTemplate).join('')
-            )
+    this.id$.pipe(
+      filter((id): id is string => !!id),
```
fpaul-1A, line 69: missing `takeUntilDestroyed(this.destroyRef)` ?

packages/@o3r/components/src/tools/component-replacement/c11n.directive.ts
matthieu-crouzet: fallback to ensure no regression on this one

fpaul-1A (approved):

packages/@o3r/rules-engine/src/services/runner/rules-engine.runner.service.ts
```diff
@@ -73,17 +71,13 @@ export class RulesEngineRunnerService implements OnDestroy {
...
-      this.events$.pipe(filter(() => this.enabled)).subscribe(async (events) => {
-        await this.executeActions(events);
-      })
-    );
+    this.store.pipe(
+      select(selectAllRulesets),
+      takeUntilDestroyed()
+    ).subscribe((rulesets: Ruleset[]) => this.engine.upsertRulesets(rulesets));
+    this.events$.pipe(takeUntilDestroyed(), filter(() => this.enabled)).subscribe(async (events) => {
+      await this.executeActions(events);
```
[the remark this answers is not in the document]
matthieu-crouzet: It will have to be 
```suggestion
    this.events$.pipe(takeUntilDestroyed(), filter(() => this.enabled)).subscribe((events) => {
      void this.executeActions(events);
```
But I don't get what is simpler 
It's a part of the code I didn't touch

docs/analytics/TRACK_EVENTS.md
```diff
@@ -239,27 +242,29 @@ For example, on the availability page, mark _data ready_ when the calendar and o
...
-  }
+  private readonly destroyRef = inject(DestroyRef);
+
+  public readonly trackEventsService = inject(EventTrackService);
+  private readonly store = inject(Store<AirOffersStore & AirSearchCriteriaStore & CartStore & AirCalendarStore>);
 
   ngOnInit() {
     const airCalendarReady$ = this.store.pipe(
         select(selectAirCalendarState),
+        takeUntilDestroyed(this.destroyRef),
```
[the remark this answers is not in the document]
matthieu-crouzet: Here you will change the behavior of the example
I would keep it like that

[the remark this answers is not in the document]
kpanot: No, it is adding `takeUntilDestroyed(this.destroyRef)` that changes the behavior.
As explained, currently the `markDataReady()` can be triggered after the service destruction (which seems to make more sense), it is not the case with the proposed change.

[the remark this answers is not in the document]
matthieu-crouzet: The subscription was saved in an array and unsubscribe on service destruction

apps/chrome-devtools/src/app-devtools/theming-panel/color.pipe.ts
```diff
@@ -32,7 +32,7 @@ export class HexColorPipe implements PipeTransform {
   name: 'contrast',
   standalone: true
 })
-export class ConstrastPipe implements PipeTransform {
+export class ContrastPipe implements PipeTransform {
```
fpaul-1A, line 35: this is unrelated to the PR and could be done separately

fpaul-1A (approved):

packages/@o3r/components/src/tools/placeholder/placeholder.component.ts
```diff
@@ -84,41 +88,40 @@ export class PlaceholderComponent implements OnInit, OnDestroy, AfterViewChecked
...
-              isPending: placeholders?.isPending
-            })),
-            distinctUntilChanged((previous, current) =>
-              previous.id === current.id
-              && previous.isPending === current.isPending
-              && previous.orderedTemplates?.map((placeholder) => placeholder.renderedTemplate).join('')
-              === current.orderedTemplates?.map((placeholder) => placeholder.renderedTemplate).join('')
-            )
+    this.id$.pipe(
+      filter((id): id is string => !!id),
```
mrednic-1A, line 92: I think here we need a 'takeUntilDestroy' too

PR #2523, opened 2024-11-27
fix(training): monaco editor tooltips not fully visible

fpaul-1A:
## Proposed change

image

Before:
image

## Related issues

*- No issue associated -*

### DIFF
--- a/apps/showcase/src/components/training/code-editor-view/code-editor-view.component.scss
+++ b/apps/showcase/src/components/training/code-editor-view/code-editor-view.component.scss
@@ -26,3 +26,7 @@ ngx-monaco-editor {
 .editor-view {
   background: #1d1d1d;
 }
+
+.monaco-overflow-widgets {
+  z-index: var(--o3r-header-zindex);
+}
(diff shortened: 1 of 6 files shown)


apps/showcase/src/components/training/code-editor-view/code-editor-view.component.scss
```diff
@@ -26,3 +26,7 @@ ngx-monaco-editor {
 .editor-view {
   background: #1d1d1d;
 }
+
+.monaco-overflow-widgets {
+  z-index: 1022;
```
kpanot, line 31: magic number? :D

kpanot (approved):

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.scss
fpaul-1A: Yeah... I could add a comment on that, I need it to be greater than the header which is `1021`

sdo-1A (dismissed):

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.scss
fpaul-1A: updated with css var

matthieu-crouzet (approved):

PR #2542, opened 2024-12-02
feat(config): generate default config with readonly and as const

### LINKED ISSUE #2530: [Feature]: Generate Localization and Configuration object with `as const` contrains
### Context

The purpose is to properly expose configuration and localization with `as const` annotation to avoid outside mutation.

### Proposal

- [ ] Update generator templates for Localization and Configuration
- [ ] Update/check metadata extractor for supporting the new variable definition format
- [ ] *(Bonus: propose an `ng-update` script to help the migration)*

matthieu-crouzet:
## Related issues
 :bug: Fix #2530

### DIFF
--- a/packages/@o3r/components/builders/component-extractor/helpers/component/component-config.extractor.ts
+++ b/packages/@o3r/components/builders/component-extractor/helpers/component/component-config.extractor.ts
@@ -365,10 +365,36 @@ export class ComponentConfigExtractor {
           if (ts.isVariableDeclaration(declarationNode)) {
             let isConfigImplementation = false;
             declarationNode.forEachChild((vNode) => {
-              if (ts.isTypeReferenceNode(vNode) && configurationInformationWrapper.configurationInformation!.name === vNode.getText(this.source)) {
+              if (
+                ts.isTypeReferenceNode(vNode)
+                && ts.isIdentifier(vNode.typeName)
+                && (
+                  vNode.typeName.escapedText.toString() === configurationInformationWrapper.configurationInformation!.name
+                  || (
+                    vNode.typeName.escapedText.toString() === 'Readonly'
+                    && vNode.typeArguments?.[0]
+                    && ts.isTypeReferenceNode(vNode.typeArguments?.[0])
+                    && ts.isIdentifier(vNode.typeArguments?.[0].typeName)
+                    && vNode.typeArguments[0].typeName.escapedText.toString() === configurationInformationWrapper.configurationInformation!.name
+                  )
+                )
+              ) {
                 isConfigImplementation = true;
-              } else if (isConfigImplementation && ts.isObjectLiteralExpression(vNode)) {
-                vNode.forEachChild((propertyNode) => {
+              } else if (
+                isConfigImplementation
+                && (
+                  ts.isObjectLiteralExpression(vNode)
+                  || (
+                    ts.isAsExpression(vNode)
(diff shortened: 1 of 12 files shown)


packages/@o3r/localization/schematics/ng-update/v11-6/index.ts
```diff
@@ -0,0 +1,28 @@
+import type {
+  Rule,
+  Tree,
+} from '@angular-devkit/schematics';
+import {
+  createSchematicWithMetricsIfInstalled,
+  findFilesInTree,
+} from '@o3r/schematics';
+
+const regexp = /translations\s*:\s*(?:Readonly<)?([A-Z][\w]*)(?:>)?\s*=\s*({[^;]+})\s*(as\s*const)?\s*;/g;
```
kpanot, line 10: To handle different linter rule guidelines it can be an easy fix to turn the `;` non-mandatory :
```suggestion
const regexp = /translations\s*:\s*(?:Readonly<)?([A-Z][\w]*)(?:>)?\s*=\s*({[^;]+})\s*(as\s*const)?\s*;?/g;
```

kpanot (approved):

packages/@o3r/localization/schematics/ng-update/v11-6/index.ts
matthieu-crouzet: It's more difficult to detect the end of the object then
`;` is used before in the regexp for that `({[^;]+})`

kpanot: Then I propose:
```
/translations\s*:\s*(?:Readonly<)?([A-Z][\w]*)(?:>)?\s*=\s*({(?:[^:]+:\s*(?:".*"|'.*'|\s*,\s*))*\s*})\s*(as\s*const)?\s*;?/g
```

packages/@o3r/components/builders/component-extractor/helpers/component/component-config.extractor.ts
```diff
@@ -365,10 +365,36 @@ export class ComponentConfigExtractor {
           if (ts.isVariableDeclaration(declarationNode)) {
             let isConfigImplementation = false;
             declarationNode.forEachChild((vNode) => {
-              if (ts.isTypeReferenceNode(vNode) && configurationInformationWrapper.configurationInformation!.name === vNode.getText(this.source)) {
+              if (
```
mrednic-1A, line 368: Could not manage to understand from the changes if this is a backward compatible update. So, I am asking it ? :)

matthieu-crouzet: Yes it supposed to be

packages/@o3r/localization/schematics/ng-update/v11-6/index.ts
matthieu-crouzet: Does not work for complex object like the one in this example 
https://regex101.com/r/2HTPCa/1

mrednic-1A (approved):

packages/@o3r/localization/schematics/ng-update/v11-6/index.ts
kpanot: Indeed but it is working for localisation, which is what you want to parse, don't you?
(And it does support `;` within the translation key even if I am not sure we do support it farer in the extractor 😅)

PR #2544, opened 2024-12-02
feat(training): file navigation inside code editor

fpaul-1A:
## Proposed change

Enable `Go to Definition` feature (Ctrl+click), auto complete and avoid errors on imports.

image

## Related issues

*- No issue associated -*

### DIFF
--- a/apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
+++ b/apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
@@ -1,21 +1,22 @@
 import {
   AsyncPipe,
-  JsonPipe,
 } from '@angular/common';
 import {
   ChangeDetectionStrategy,
   Component,
+  computed,
+  effect,
   ElementRef,
   inject,
-  Input,
-  OnChanges,
+  input,
   OnDestroy,
-  SimpleChanges,
+  untracked,
   ViewChild,
   ViewEncapsulation,
 } from '@angular/core';
 import {
   takeUntilDestroyed,
+  toSignal,
 } from '@angular/core/rxjs-interop';
 import {
   FormBuilder,
@@ -33,6 +34,7 @@ import {
 import {
   AngularSplitModule,
 } from 'angular-split';
+import type * as Monaco from 'monaco-editor';
 import {
   MonacoEditorModule,
 } from 'ngx-monaco-editor-v2';
@@ -46,6 +48,7 @@ import {
   debounceTime,
   distinctUntilChanged,
   filter,
+  firstValueFrom,
   from,
   map,
   Observable,
@@ -53,17 +56,25 @@ import {
   share,
   skip,
   startWith,
+  Subject,
   switchMap,
 } from 'rxjs';
 import {
   checkIfPathInMonacoTree,
 } from '../../../helpers/monaco-tree.helper';
 import {
+  flattenTree,
   WebContainerService,
 } from '../../../services';
 import {
   CodeEditorControlComponent,
 } from '../code-editor-control';
+
+declare global {
+  interface Window {
+    monaco: typeof Monaco;
+  }
+}
 
 /** ngx-monaco-editor options language - determined based on file extension */
 const editorOptionsLanguage: Record<string, string> = {
@@ -96,7 +107,6 @@ export interface TrainingProject {
     AsyncPipe,
     CodeEditorControlComponent,
     FormsModule,
-    JsonPipe,
     MonacoEditorModule,
     NgxMonacoTreeComponent,
     ReactiveFormsModule,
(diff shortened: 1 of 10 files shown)


apps/showcase/src/services/webcontainer/webcontainer.service.ts
```diff
@@ -126,4 +126,32 @@ export class WebContainerService {
...
+        const indexFiles = await Promise.all(files.filter((entry) => entry.isFile() && entry.name.endsWith('.d.ts')).map(async (indexFile) => ({
+          filePath: `file:///${path}/${dirEntry.name}/${indexFile.name}`,
+          content: await instance.fs.readFile(`${basePath}/${dirEntry.name}/${indexFile.name}`, 'utf8')
+        })));
+        return [
+          ...indexFiles,
+          ...(maxDepth > 1 ? await this.getDeclarationTypes(project, maxDepth - 1, `${path}/${dirEntry.name}`) : [])
+        ];
+      }
+    }))).flat().filter((entry) => !!entry);
```
matthieu-crouzet, line 155:
```suggestion
    return (await Promise.all(dependencies.map(async (dirEntry) => {
      if (dirEntry.isDirectory() && !dependenciesBlackList.test(dirEntry.name) && (dependenciesWhiteList.test(dirEntry.name) || dependenciesWhiteList.test(path))) {
        const files = await instance.fs.readdir(`${basePath}/${dirEntry.name}`, { encoding: 'utf8', withFileTypes: true });
        const indexFiles = await Promise.all(files.filter((entry) => entry.isFile() && entry.name.endsWith('.d.ts')).map(async (indexFile) => ({
          filePath: `file:///${path}/${dirEntry.name}/${indexFile.name}`,
          content: await instance.fs.readFile(`${basePath}/${dirEntry.name}/${indexFile.name}`, 'utf8')
        })));
        return [
          ...indexFiles,
          ...(maxDepth > 1 ? await this.getDeclarationTypes(project, maxDepth - 1, `${path}/${dirEntry.name}`) : [])
        ];
      }
      return [];
    }))).flat();
```

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
```diff
@@ -167,23 +177,71 @@ export class CodeEditorViewComponent implements OnDestroy, OnChanges {
...
+      value,
+      language: editorOptionsLanguage[fileExtension || ''] || '',
+      uri: `file:///${fileName}`
+    };
+  });
+
   constructor() {
+    effect(async () => {
+      const project = this.project();
+      await untracked(async () => {
```
kpanot, line 236: is the `untracked` necessary here?
I suppose it is used for the `this.loadNewProject()` but it also use only `this.project()` an already tracked signal.

apps/showcase/src/services/webcontainer/webcontainer.service.ts
```diff
@@ -126,4 +126,32 @@ export class WebContainerService {
...
   }
+
+  /**
+   * Retrieve all typescript declaration files from web-container
+   * @param project
+   * @param maxDepth
+   * @param path
+   */
+  public async getDeclarationTypes(project: string, maxDepth = 20, path = 'node_modules'): Promise<{ filePath: string; content: string }[]> {
+    // TODO read that from project devDependencies?
```
kpanot, line 137: devDependencies can be removed according to the builder (as it is with ng-packagr)

apps/showcase/src/services/webcontainer/webcontainer.service.ts
```diff
@@ -126,4 +126,32 @@ export class WebContainerService {
...
+    const dependenciesWhiteList = /@ama-sdk|@angular|@o3r|rxjs|@types/;
```
kpanot, line 138: why not using array instead of regexp?

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
fpaul-1A: it's not necessary, but I'm only interested in tracking `this.project()` so I've put it there to make sure we don't track anything else by mistake

apps/showcase/src/services/webcontainer/webcontainer.service.ts
fpaul-1A: Here, I'm talking about the `devDependencies` of the tutorial app inside the web-container

apps/showcase/src/services/webcontainer/webcontainer.service.ts
fpaul-1A: I though it might be more convenient if we want to use patterns on the files like `@angular.*` or something

matthieu-crouzet (approved):

apps/showcase/src/services/webcontainer/webcontainer.service.ts
kpanot: I think an array + startsWith would be easier to maintain.
(btw, if you keep regex you may want to add `^` because if it does not contains `@` it may match unwanted deps)

fpaul-1A: I've added the `^` and `$`
anyway, this may change after we implement the TODO about dev deps

matthieu-crouzet (approved):

cpaulve-1A (approved):

PR #2545, opened 2024-12-02
fix: mock webcontainer api and xterm for unit test

cpaulve-1A:
## Related issues

*- No issue associated -*


matthieu-crouzet (approved):

fpaul-1A (approved):

apps/showcase/src/app/sdk-training/sdk-training.spec.ts
```diff
@@ -9,6 +9,15 @@ import {
 describe('SdkTrainingComponent', () => {
   let component: SdkTrainingComponent;
   let fixture: ComponentFixture<SdkTrainingComponent>;
+  global.fetch = jest.fn(() => Promise.resolve({
```
kpanot, line 12: Maybe cleaner to be set in the `beforeEach` to renew the mock before each test.
Or, if the single mock is done on purpose, maybe moving it outside of the `describe` as it is affecting globally.

(same comment for the other tests)

apps/showcase/testing/setup-jest.ts
```diff
@@ -1 +1,30 @@
 import 'jest-preset-angular/setup-jest';
+
+jest.mock('@webcontainer/api',
```
kpanot, line 3: Maybe more readable to move the both mocks on external `testing/mocks/webcontainer-api.ts` and `testing/mocks/xterm-xtern.ts` that you can import in this file?

fpaul-1A (approved):

matthieu-crouzet (approved):

apps/showcase/testing/setup-jest.ts
kpanot: I was thinking:

```typescript
// in webcontainer-api.mock.ts

class WebContainerApiMock {
  public static boot = jest.fn(() => Promise.resolve({
    on: jest.fn(() => jest.fn())
  }));
}
jest.mock('@webcontainer/api',
  () => ({
    WebContainer: WebContainerApiMock
  }), {
    virtual: true
  }
);
```

and in this file just:
```typescript
import './mocks/webcontainer-api.mock';
```

kpanot (approved):

PR #2548, opened 2024-12-03
fix: do not publish training package; not needed

mrednic-1A:
## Proposed change
Remove the '@o3r-training/showcase-sdk' package from the published packages. It is not needed to be installed during the training and has no value outside the training.

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@o3r-training/showcase-sdk/package.json
+++ b/packages/@o3r-training/showcase-sdk/package.json
@@ -13,9 +13,7 @@
   "types": "index.d.ts",
   "sideEffects": false,
   "deprecated": "This package is intended for testing purposes only.",
-  "publishConfig": {
-    "access": "public"
-  },
+  "private": true,
   "exports": {
     "./package.json": {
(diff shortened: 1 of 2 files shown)


matthieu-crouzet (dismissed):

packages/@o3r-training/showcase-sdk/package.json
```diff
@@ -13,9 +13,6 @@
   "types": "index.d.ts",
   "sideEffects": false,
   "deprecated": "This package is intended for testing purposes only.",
-  "publishConfig": {
```
fpaul-1A, line 16: should we add `"private": true` to make sure it's never targeted for publish

mrednic-1A: If it's for that, then yes.

matthieu-crouzet (approved):

fpaul-1A (approved):

PR #2576, opened 2024-12-10
chore: increase number of nx parallel task for build

kpanot:
## Proposed change

increase number of nx parallel task for build

## Related issues

*- No issue associated -*

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,6 +39,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: [yarn_lock_check]
     env:
+      NX_PARALLEL: 3
       NX_SKIP_NX_CACHE: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/heads/release') }}
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -39,6 +39,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: [yarn_lock_check]
     env:
+      NX_PARALLEL: 3
```
fpaul-1A, line 42: Can't we set that directly in GitHub settings?

kpanot: We can but it will be apply to all nx command, not sure it is what we want.
I saw a time degradation for `build` task but not for `Unit test` which seems to be faster

fpaul-1A: do we really want to have a different value depending on the job? I think we should have the same value everywhere

matthieu-crouzet: If we decide to not have the same value everywhere I would expect a comment to explain why do we not use the default one

PR #2578, opened 2024-12-11
fix: vscode watcher issue with ts upgrade

kpanot:
## Proposed change

fix: vscode watcher issue with ts upgrade
chore: align dependencies

## Related issues

*- No issue associated -*

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -56,14 +56,9 @@
   "files.associations": {
     "*.json5": "jsonc"
   },
-  "files.exclude": {
-    "**/.DS_Store": true,
-    "**/.git": true,
-    "**/Thumbs.db": true
-  },
   "files.readonlyExclude": {
-    "packages/@o3r-training/*sdk/src/api/*/index.ts": true,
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -94,17 +96,6 @@
     "@o3r/core:*"
   ],
   "otter.extract.styling.prefix": "o3r",
-  "search.exclude": {
```
matthieu-crouzet, line 97: Why do you remove this field ?

kpanot: If not specified, it inherits of `files.exclude` (which include git ignored files).
The added value of specifying it is for `.yarn/release/yarn.cjs` and `.yarn/sdks` but we will lose the benefit of relying on `files.exclude`

matthieu-crouzet (dismissed):

.vscode/settings.json
kpanot: edit: After testing, with the option `search.useIgnoreFiles` explicitly set, the `search.exclude` property extends the ignored files.
So I put it back with a minimized value

matthieu-crouzet (approved):

PR #2579, opened 2024-12-11
feat(training): persistence of code modifications

matthieu-crouzet:
## Proposed change
Add persistence in `localStorage` of code modifications in a training step.
When the user refresh or come back on the training on which code modifications were made, we prompt a dialog to ask if he wants to keep the changes or discard them.
image

With this feature we think also to add a `Reset` or `Discard changes` button, it will be done in another PR, unless someone think it's a must inside this PR.

### DIFF
--- a/apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
+++ b/apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
@@ -6,12 +6,10 @@ import {
   Component,
   computed,
   effect,
-  ElementRef,
   inject,
   input,
   OnDestroy,
   untracked,
-  ViewChild,
   ViewEncapsulation,
 } from '@angular/core';
 import {
@@ -25,6 +23,9 @@ import {
   FormsModule,
   ReactiveFormsModule,
 } from '@angular/forms';
+import {
+  DfModalService,
+} from '@design-factory/design-factory';
 import {
   LoggerService,
 } from '@o3r/logger';
@@ -44,6 +45,7 @@ import {
 } from 'ngx-monaco-tree';
 import {
   BehaviorSubject,
+  combineLatest,
   combineLatestWith,
   debounceTime,
   distinctUntilChanged,
@@ -51,9 +53,10 @@ import {
   firstValueFrom,
   from,
   map,
+  merge,
   Observable,
   of,
-  share,
+  shareReplay,
   skip,
   startWith,
   Subject,
@@ -69,6 +72,9 @@ import {
 import {
   CodeEditorControlComponent,
 } from '../code-editor-control';
+import {
+  SaveCodeDialogComponent,
+} from '../save-code-dialog';
 
 declare global {
   interface Window {
@@ -134,9 +140,6 @@ export class CodeEditorViewComponent implements OnDestroy {
    */
   private readonly loggerService = inject(LoggerService);
 
-  @ViewChild('monacoOverflowWidgets')
-  private readonly monacoOverflowWidgets!: ElementRef;
-
   /**
    * Allow to edit the code in the monaco editor
    */
@@ -163,7 +166,7 @@ export class CodeEditorViewComponent implements OnDestroy {
         : of([])
     ),
     filter((tree) => tree.length > 0),
-    share()
+    shareReplay({ bufferSize: 1, refCount: true })
   );
 
   /**
@@ -205,17 +208,24 @@ export class CodeEditorViewComponent implements OnDestroy {
       readOnly: (this.editorMode() === 'readonly'),
       automaticLayout: true,
       scrollBeyondLastLine: false,
-      overflowWidgetsDomNode: this.monacoOverflowWidgets.nativeElement,
+      fixedOverflowWidgets: true,
       model: this.model()
     }))
   );
 
-  private readonly fileContentLoaded$ = this.form.controls.file.valueChanges.pipe(
+  private readonly modalService = inject(DfModalService);
+  private readonly forceReload = new Subject<void>();
+  private readonly forceSave = new Subject<void>();
+
+  private readonly fileContentLoaded$ = combineLatest([
+    this.form.controls.file.valueChanges,
+    this.forceReload.pipe(startWith(undefined))
+  ]).pipe(
     takeUntilDestroyed(),
     combineLatestWith(this.cwdTree$),
-    filter(([path, monacoTree]) => !!path && checkIfPathInMonacoTree(monacoTree, path.split('/'))),
-    switchMap(([path]) => from(this.webContainerService.readFile(`${this.project().cwd}/${path}`).catch(() => ''))),
-    share()
+    filter(([[path], monacoTree]) => !!path && checkIfPathInMonacoTree(monacoTree, path.split('/'))),
+    switchMap(([[path]]) => from(this.webContainerService.readFile(`${this.project().cwd}/${path}`).catch(() => ''))),
+    shareReplay({ refCount: true, bufferSize: 1 })
   );
 
   private readonly fileContent = toSignal(this.fileContentLoaded$);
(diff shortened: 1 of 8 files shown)


apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
```diff
@@ -238,10 +254,14 @@ export class CodeEditorViewComponent implements OnDestroy {
...
+        this.cwd$.next(project.cwd);
       });
     });
+    this.forceReload.subscribe(async () => {
+      await this.cleanAllModelsFromMonaco();
+      await this.loadAllProjectFilesToMonaco();
+    });
     this.form.controls.code.valueChanges.pipe(
       distinctUntilChanged(),
       skip(1),
```
matthieu-crouzet, line 267: `debounceTime` is set to 300, I think we should increase a bit this number to avoid to much write on the webcontainer (that will trigger app (re)build) and (now with this PR) write on the `localStorage`.
I would propose at least 1sec without typing before we store the changes, what do you think ?

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
```diff
@@ -382,10 +405,33 @@ export class CodeEditorViewComponent implements OnDestroy {
...
   }
 
+  public ngOnInit() {
+    const { cwd } = this.project();
+    const savedState = localStorage.getItem(cwd);
+    if (savedState) {
+      const modal = this.modalService.open(SaveCodeDialogComponent, { backdrop: 'static' });
+      void modal.result.then((positiveReply) => {
+        if (positiveReply) {
+          this.cwdTree$.pipe(take(1)).subscribe(async () => {
```
kpanot, line 415:
```suggestion
          await firstValueFrom(this.cwdTree$);
```
?

kpanot (dismissed):

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
matthieu-crouzet: `ngOnInit` is not async, I could do it with a `.then` but it's the same no ?

kpanot: It's inside the `then((positiveReply) => ...)` already, just need to be`then(async (positiveReply) => ...)`.
But not problem to not do the change :D

matthieu-crouzet: indeed I didn't remind that it was already in a `then`
I will do the change when I will rebase

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
fpaul-1A: I'm ok to auto-save after 1s of inactivity, maybe we can also trigger it on `Ctrl+S`

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
```diff
@@ -382,10 +411,33 @@ export class CodeEditorViewComponent implements OnDestroy {
...
+    const savedState = localStorage.getItem(cwd);
+    if (savedState) {
+      const modal = this.modalService.open(SaveCodeDialogComponent, { backdrop: 'static' });
+      void modal.result.then(async (positiveReply) => {
+        if (positiveReply) {
+          await firstValueFrom(this.cwdTree$);
+          const state = JSON.parse(savedState);
+          await Promise.all(Object.entries<string>(state).map(async ([path, text]) => {
+            await this.webContainerService.writeFile(`${cwd}/${path}`, text);
+          }));
```
fpaul-1A, line 425:
```suggestion
          await Promise.all(Object.entries<string>(state).map(([path, text]) =>
            this.webContainerService.writeFile(`${cwd}/${path}`, text);
          ));
```

apps/showcase/src/components/training/save-code-dialog/save-code-dialog.component.ts
```diff
@@ -0,0 +1,30 @@
...
+import {
+  NgbActiveModal,
+} from '@ng-bootstrap/ng-bootstrap';
+
+@Component({
+  selector: 'code-editor-terminal',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  standalone: true,
+  imports: [],
+  template: `
```
fpaul-1A, line 15: it might be better to create an html file to have the linting, autocomplete, ...

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
```diff
@@ -238,26 +255,43 @@ export class CodeEditorViewComponent implements OnDestroy {
...
       filter((text): text is string => !!text),
       takeUntilDestroyed()
     ).subscribe((text: string) => {
       if (!this.project) {
         this.loggerService.error('No project found');
         return;
       }
+      if (text !== this.fileContent()) {
+        const { cwd } = this.project();
+        localStorage.setItem(cwd, JSON.stringify({
```
cpaulve-1A, line 283: Can we try updating and saving a change on the yarn.lock and verify it will not be truncated?

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
```diff
@@ -382,10 +411,32 @@ export class CodeEditorViewComponent implements OnDestroy {
     }
   }
 
+  public ngOnInit() {
+    const { cwd } = this.project();
+    const savedState = localStorage.getItem(cwd);
+    if (savedState) {
+      const modal = this.modalService.open(SaveCodeDialogComponent, { backdrop: 'static' });
+      void modal.result.then(async (positiveReply) => {
```
cpaulve-1A, line 419: Can we try to ask the question once we have the cwdTree to avoid blocking the user too long?

apps/showcase/src/components/training/save-code-dialog/save-code-dialog.template.html
```diff
@@ -0,0 +1,10 @@
+<div class="modal-header">
+  <h2 class="modal-title">Code modifications detected</h2>
```
cpaulve-1A, line 2:
```suggestion
  <h2 class="modal-title">Training in progress detected</h2>
```

apps/showcase/src/components/training/save-code-dialog/save-code-dialog.template.html
```diff
@@ -0,0 +1,10 @@
...
+</div>
+<div class="modal-body">
+  <p>Do you want to keep your previous changes or discard them?</p>
+</div>
+<div class="modal-footer">
+  <button type="button" class="btn btn-outline-danger me-5" (click)="activeModal.close(false)">Discard</button>
+  <button type="button" class="btn btn-primary" ngbAutofocus (click)="activeModal.close(true)">Keep</button>
```
cpaulve-1A, line 9:
```suggestion
  <button type="button" class="btn btn-primary" ngbAutofocus (click)="activeModal.close(true)">Resume</button>
```

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
fpaul-1A: actually, should we block the user at all? we just need to block the editor view, no need to prevent the navigation or any other action?
I was thinking about putting the prompt on the right side to block the editor view but keep the rest of the showcase available

apps/showcase/src/components/training/save-code-dialog/save-code-dialog.template.html
sdo-1A: I personally prefer the original sentence `Code modifications detected` ...

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
cpaulve-1A: Why not, let's see what it could look like

matthieu-crouzet: I updated the screenshot

cpaulve-1A (dismissed):

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
matthieu-crouzet: In our use case we have a package-lock.json I don't know if there is a difference for the length of the file but it seems ok I didn't notice any latency

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
```diff
@@ -316,10 +339,48 @@ export class CodeEditorViewComponent implements OnDestroy {
...
+    const savedState = localStorage.getItem(cwd);
+    if (!savedState) {
+      return;
+    }
+    const state = JSON.parse(savedState);
+    const hasDiscrepancies = (await Promise.all(
+      Object.entries(state).map(async ([path, text]) =>
+        text !== (await this.webContainerService.readFile(`${this.project().cwd}/${path}`).catch(() => undefined))
+      ))
+    ).some((hasDiscrepancy) => !!hasDiscrepancy);
```
fpaul-1A, line 357: We don't have to wait the response for all the files to dhow the dialog, as soon as we spot one discrepancy
Maybe something like:
```suggestion
    const hasDiscrepancies = (await Promise.any(
      Object.entries(state).map(([path, text]) => new Promise((resolve, reject) =>
        text !== (await this.webContainerService.readFile(`${this.project().cwd}/${path}`).catch(() => undefined)) ? resolve(true) : reject()
      ))
    ).catch(() => false);
```

fpaul-1A (approved):

cpaulve-1A (approved):

PR #2582, opened 2024-12-12
fix: remove pattern for ref property in localization.metadata.schema.json

guillaumerochelle:
There is no point validating the ref if we don't validate the key.


matthieu-crouzet (approved):

kpanot (approved):

amadeus-pmeneguzzi:
I approve this change !

PR #2590, opened 2024-12-13
fix: base code navigation on webcontainer instead of project initial value

cpaulve-1A:
## Proposed change
Project only contains the description of the exercise and does not reflect the trainee changes.
The file list should reflect the webcontainer state

Fixes also e2e tests and add a prefix for training directories to avoid name conflicts

## Related issues

*- No issue associated -*

### DIFF
--- a/apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
+++ b/apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
@@ -167,6 +167,11 @@ export class CodeEditorViewComponent implements OnDestroy {
   );
 
   /**
+   * Signal that reflects the current value of the monaco tree in the project directory
+   */
+  public cwdTreeSignal = toSignal(this.cwdTree$, { initialValue: [] });
+
+  /**
    * Form with the selected file and its content which can be edited in the Monaco Editor
    */
   public form: FormGroup<{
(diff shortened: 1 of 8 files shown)


apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
```diff
@@ -110,7 +111,8 @@ export interface TrainingProject {
     MonacoEditorModule,
     NgxMonacoTreeComponent,
     ReactiveFormsModule,
-    AngularSplitModule
+    AngularSplitModule,
+    JsonPipe
```
fpaul-1A, line 115: this doesn't seem to be used

apps/showcase/src/services/webcontainer/webcontainer.service.ts
```diff
@@ -36,7 +36,7 @@ export class WebContainerService {
 
   public monacoTree$ = this.monacoTree.pipe(
     distinctUntilChanged((a, b) => JSON.stringify(a) === JSON.stringify(b)),
-    share()
+    shareReplay(1)
```
fpaul-1A, line 39:
```suggestion
    shareReplay({ bufferSize: 1, refCount: true })
```

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
cpaulve-1A: Leftovers from some testing :D

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #2600, opened 2024-12-16
chore(eslint-plugin): rely on base template parser

kpanot:
## Proposed change

chore(eslint-plugin): rely on base template parser instead of angular-eslint one

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@o3r/eslint-plugin/src/rules/template/no-inner-html/no-inner-html.spec.ts
+++ b/packages/@o3r/eslint-plugin/src/rules/template/no-inner-html/no-inner-html.spec.ts
@@ -1,22 +1,26 @@
 import {
+  meta,
+  parseForESLint,
+} from '@angular-eslint/template-parser';
+import {
   RuleTester,
 } from '@angular-eslint/test-utils';
-import {
-  templateParser,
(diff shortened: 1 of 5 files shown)


packages/@o3r/eslint-plugin/src/rules/template/no-inner-html/no-inner-html.spec.ts
```diff
@@ -1,22 +1,26 @@
...
-  templateParser,
-} from 'angular-eslint';
 import noInnerHTMLRule, {
   name,
 } from './no-inner-html';
 
 const ruleTester = new RuleTester({
   languageOptions: {
-    parser: templateParser
+    parser: {
```
matthieu-crouzet, line 14: It seems to be exactly the same, no ?
https://github.com/angular-eslint/angular-eslint/blob/main/packages/angular-eslint/src/index.ts#L13
image

matthieu-crouzet: Like `typescript-eslint`, `angular-eslint` built a package where you should have everything accessible instead of needed to know in which package is what

PR #2608, opened 2024-12-18
fix(core): add support for "null" type

### LINKED ISSUE #2601: [Bug]: deepFill not handling null values correctly
### Package name

store-sync

### Package version

10.4

### Reproduction steps

- Register storage sync
- Merge reducer is called
- Objects to merge contain null values

### Current result

App gets stuck and store crashes

### Expected result

No crash :)

### Additional comments

In our specific case we have an object `base`:
```json
{
    selection: { ... }
}
```

and `source`
```json
{
    selection: null
}
```

When deepFill is called at some point the merge happens for the `selection` key, where an `object` and `null` are merged. Because of `source = null` the check is not sufficient to catch that case: https://github.com/AmadeusITGroup/otter/blob/56eae2bfcfce2548537d42cbae3f0d9fc2b95730/packages/%40o3r/core/src/utils/deep-fill/deep-fill.ts#L60

This result in an error where our app is no longer responding to routing requests:
Screenshot 2024-12-11 at 14 07 19 copy

kpanot:
## Proposed change

Add the support of the `null` as deepFill object field value

## Related issues

* :bug: Fix resolves #2601

### DIFF
--- a/packages/@o3r/core/src/utils/deep-fill/deep-fill.spec.ts
+++ b/packages/@o3r/core/src/utils/deep-fill/deep-fill.spec.ts
@@ -12,6 +12,20 @@ describe('Deep fill function', () => {
     expect(deepFill(base, source)).toEqual(base);
   });
 
+  it('should support "null" value in source', () => {
+    const base = { selection: { field: 'test-field' }, a: 'string' };
+    const source = { selection: null } as any;
+
+    expect(deepFill(base, source)).toEqual({ selection: null, a: 'string' });
+  });
+
+  it('should support "null" value in base', () => {
+    const base = { selection: null, a: 'string' } as any;
+    const source = { selection: { field: 'test-field' } };
+
+    expect(deepFill(base, source)).toEqual({ selection: { field: 'test-field' }, a: 'string' });
+  });
+
   it('should keep properties from base not present in the source', () => {
     const base = Object.freeze({a: 1, b: '2', c: true});
     const source = Object.freeze({c: false, a: 3});
--- a/packages/@o3r/core/src/utils/deep-fill/deep-fill.ts
+++ b/packages/@o3r/core/src/utils/deep-fill/deep-fill.ts
@@ -66,9 +66,11 @@ export function deepFill<T extends { [x: string]: any }>(base: T, source?: { [x:
   }
   const newObj = {...base};
   for (const key in base) {
-    if (key in source && typeof base[key] === typeof source[key]) {
+    if (source[key] === null) {
+      newObj[key] = immutablePrimitive(null, additionalMappers);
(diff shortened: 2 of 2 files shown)


packages/@o3r/core/src/utils/deep-fill/deep-fill.ts
```diff
@@ -43,7 +43,7 @@ export function immutablePrimitive(obj: any, additionalMappers?: PrimitiveRevive
     return new Date(obj);
   } else if (obj instanceof Object) {
     // eslint-disable-next-line no-use-before-define
-    return deepFill(obj, obj, additionalMappers);
+    return obj === null ? null : deepFill(obj, obj, additionalMappers);
```
fpaul-1A, line 46: can `obj` be null if it's `instanceof Object` ?

packages/@o3r/core/src/utils/deep-fill/deep-fill.spec.ts
```diff
@@ -12,6 +12,20 @@ describe('Deep fill function', () => {
     expect(deepFill(base, source)).toEqual(base);
   });
 
+  it('should support "null" value in source', () => {
+    const base = { selection: { field: 'test-field' }, a: 'string' };
+    const source = { selection: null } as any;
+
+    expect(deepFill(base, source)).toEqual(base);
```
fpaul-1A, line 19: it this really what we expect from a deepFill?
it would make sense to me that `deepFill({a: 1}, {a: null})` would result in `{a: null}`
`deepFill({a: 1}, {a: undefined})` results in `{a: 1}`
`deepFill({a: null}, {a: 1})` results in `{a: 1}`

packages/@o3r/core/src/utils/deep-fill/deep-fill.ts
matthieu-crouzet: image

packages/@o3r/core/src/utils/deep-fill/deep-fill.ts
```diff
@@ -66,9 +66,11 @@ export function deepFill<T extends { [x: string]: any }>(base: T, source?: { [x:
   }
   const newObj = {...base};
   for (const key in base) {
-    if (key in source && typeof base[key] === typeof source[key]) {
+    if (key in source && typeof base[key] === typeof source[key] && source[key] !== null) {
```
fpaul-1A, line 69: you don't need to change this condition if you move the `else if ` block up
```suggestion
    if (source[key] === null) {
      newObj[key] = immutablePrimitive(null, additionalMappers);
    } else if (key in source && typeof base[key] === typeof source[key]) {
```

fpaul-1A (approved):

cpaulve-1A (approved):

PR #2635, opened 2024-12-19
fix(#2628): generate eol LF config

### LINKED ISSUE #2628: [Feature]: @o3r/create should generate end-of-line LF config
### Context

When generating a project using `@o3r/create`, do we want the generator to automatically add the LF end-of-line config?

### Proposal

To handle the LF config, we could:
1. Add the following line in `.editorconfig` file: `end_of_line = lf`.
2. Generate a `.gitattributes` file containing at least this line: `* text eol=lf`.

matthieu-crouzet:
## Proposed change

Generate eol LF config

## Related issues

* :bug: Fix resolves #2628

### DIFF
--- a/packages/@o3r/create/src/index.ts
+++ b/packages/@o3r/create/src/index.ts
@@ -5,6 +5,7 @@ import {
   type SpawnSyncOptionsWithBufferEncoding,
 } from 'node:child_process';
 import {
+  existsSync,
   readFileSync,
   writeFileSync,
 } from 'node:fs';
@@ -264,6 +265,21 @@ const prepareWorkspace = (relativeDirectory = '.', projectPackageManager = 'npm'
       ));
     }
   }
+
+  if (argv.minimal) {
+    const editorconfigPath = resolve(cwd, '.editorconfig');
+    const editorconfig = existsSync(editorconfigPath) ? readFileSync(editorconfigPath, { encoding: 'utf8' }) : '';
+    const newEditorconfig = /\[[*]\]/.test(editorconfig)
+      ? editorconfig.replace(/(\[[*]\])/, '$1\nend_of_line = lf')
+      : editorconfig.concat('[*]\nend_of_line = lf');
+    writeFileSync(editorconfigPath, newEditorconfig);
+  }
+  const gitAttributesPath = resolve(cwd, '.gitattributes');
+  writeFileSync(
+    gitAttributesPath,
+    (existsSync(gitAttributesPath) ? readFileSync(gitAttributesPath, { encoding: 'utf8' }) : '')
+      .concat('* text eol=lf')
+  );
 
   exitProcessIfErrorInSpawnSync(INSTALL_PROCESS_ERROR_CODE, spawnSync(runner, ['install'], spawnSyncOpts));
 };
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -95,7 +95,9 @@
   "search.exclude": {
     "**/*.code-search": true,
     "**/.yarn/**": true,
-    "**/github-actions/*/packaged-action/**": true
+    "**/github-actions/*/packaged-action/**": true,
+    "**/.yarn": true,
+    "**/.pnp.*": true
   },
   "typescript.enablePromptUseWorkspaceTsdk": true,
   "typescript.tsdk": ".yarn/sdks/typescript/lib"
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -20,7 +20,7 @@ packageExtensions:
       "@swc/types": "*"
   "@typescript-eslint/rule-tester@*":
     dependencies:
-      "@typescript-eslint/parser": ~8.17.0
+      "@typescript-eslint/parser": ~8.18.0
   "@angular-eslint/eslint-plugin-template@*":
     dependencies:
       "@typescript-eslint/types": "^8.0.0"
--- a/packages/@ama-sdk/schematics/package.json
+++ b/packages/@ama-sdk/schematics/package.json
@@ -144,7 +144,7 @@
     "@swc/helpers": "~0.5.0",
     "@commitlint/cli": "^19.0.0",
     "@commitlint/config-conventional": "^19.0.0",
-    "@typescript-eslint/eslint-plugin": "~8.17.0",
+    "@typescript-eslint/eslint-plugin": "~8.18.0",
     "jest-junit": "~16.0.0",
     "lint-staged": "^15.0.0",
     "minimist": "^1.2.6",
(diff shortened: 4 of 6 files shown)


packages/@o3r/create/src/index.ts
```diff
@@ -265,6 +266,21 @@ const prepareWorkspace = (relativeDirectory = '.', projectPackageManager = 'npm'
     }
   }
 
+  if (argv.minimal) {
+    const editorconfigPath = resolve(cwd, '.editorconfig');
```
kpanot, line 270: Why is it implemented in this file instead of be part of `core`?
To remember, this file should contain exclusively what it mandatory to run Ng generation and the add of Otter core.
If it is not the case of these file updates you may need to move them in core.

I also think that in any case, as the generators require to be have the `lf` constrain, the core schematics should do this update (as well if it is mandatory here).

packages/@o3r/create/src/index.ts
```diff
@@ -265,6 +266,21 @@ const prepareWorkspace = (relativeDirectory = '.', projectPackageManager = 'npm'
     }
   }
 
+  if (argv.minimal) {
```
kpanot, line 269: Not sure to understand this option.
If it is needed, you should also provide documentation regarding it in readme.
(and potentially tests :D)

packages/@o3r/create/src/index.ts
matthieu-crouzet: For me it's part of the file for the monorepo configuration I would expect to have it defined directly when I create the monorepo 
I don't expect that it changes if someone run `ng add @o3r/core` on an exisiting monorepo

packages/@o3r/create/src/index.ts
matthieu-crouzet: It is already documented [here]( https://github.com/AmadeusITGroup/otter/blob/main/packages/%40o3r/create/README.md#others-available-options) whne you go on this link https://v17.angular.io/cli/new#options

packages/@o3r/create/src/index.ts
matthieu-crouzet: What do you think about that @AmadeusITGroup/otter_contributors  ?

packages/@o3r/create/src/index.ts
matthieu-crouzet: In their case they are not generating the `.editorconfig` if they have the option `minimal` set to true

packages/@o3r/create/src/index.ts
fpaul-1A: I also think this should only be generated at `create` time, not on `ng add`

cpaulve-1A: Same as Florian and Matthieu. I would not mess with an existing project. 
It makes sense to keep it aligned with the current project and in this case the file is created during the create process.

kpanot: But this is not at all the purpose of the create script.
If it is a mandatory for our generators then as soon as we add Otter to a project we may need this change.
If it is not, but part of the Otter guidelines, then it should be bring by a `ng add` (even if it is done only once).

Please remember that the create script is not mandatory to make the other installation work, should remain as simple as possible and should not be the one applying guidelines (or setup default value)

packages/@o3r/create/src/index.ts
kpanot: > It is already documented [here](https://github.com/AmadeusITGroup/otter/blob/main/packages/%40o3r/create/README.md?rgh-link-date=2024-12-20T06%3A39%3A53Z#others-available-options), you can go on this link [v17.angular.io/cli/new#options](https://v17.angular.io/cli/new#options)

This is not really a fare answer :D.
The option itself is not documented but, as part of Angular option, is supported.
I think you can add a comment (with a link potentially) in the code regarding it.

packages/@o3r/create/src/index.ts
matthieu-crouzet: If we decide to set it up in a `ng add`, I would recommend to set it inside `@o3r/workspace` and not `@o3r/core`

packages/@o3r/workspace/schematics/ng-add/project-setup.ts
```diff
@@ -40,6 +40,16 @@ import type {
   NgAddSchematicsSchema,
 } from './schema';
 
+const updateEditorConfig: Rule = (tree) => {
+  const editorconfigPath = '.editorconfig';
+  const editorconfig = tree.exists(editorconfigPath) ? tree.readText(editorconfigPath) : '';
+  const newEditorconfig = /\[[*]\]/.test(editorconfig)
```
kpanot, line 46: Maybe you should not append `end_of_line` if already defined in the file

```typescript
if (editorConfig.includes('end_of_line')) {
  return tree;
}
```

kpanot (dismissed):

packages/@o3r/create/src/index.ts
```diff
@@ -265,6 +266,16 @@ const prepareWorkspace = (relativeDirectory = '.', projectPackageManager = 'npm'
...
   }
 
+  const gitAttributesPath = resolve(cwd, '.gitattributes');
+  if (existsSync(gitAttributesPath)) {
+    const content = readFileSync(gitAttributesPath, { encoding: 'utf8' });
+    if (!content.includes('* text eol=')) {
+      writeFileSync(gitAttributesPath, '* text eol=lf\n'.concat(content));
+    }
+  } else {
+    writeFileSync(gitAttributesPath, '* text eol=lf');
```
fpaul-1A, line 276: this will also apply to binary files (images, fonts, ...) if my understanding is correct, there is no problems with that?

kpanot: Indeed, if they are not flagged as binary explicitly with something like:

```ini
# Binary files
*.png binary
*.jpg binary
*.gif binary
*.ico binary
```

matthieu-crouzet: As we are in a create and that Angular for now is not generating it.
I created the file without checking its presence or content.
the content generated is 
```
* text eol=lf

# Binary files
*.png binary
*.jpg binary
*.gif binary
*.jar binary
*.ico binary
```

fpaul-1A (dismissed):

packages/@o3r/create/src/index.ts
```diff
@@ -265,6 +265,20 @@ const prepareWorkspace = (relativeDirectory = '.', projectPackageManager = 'npm'
...
+    resolve(cwd, '.gitattributes'),
+    [
+      '* text eol=lf',
+      '',
+      '# Binary files',
+      '*.png binary',
+      '*.jpg binary',
+      '*.gif binary',
+      '*.jar binary',
+      '*.ico binary'
```
kpanot, line 278:
```suggestion
      '*.ico binary',
      ''
```

kpanot (approved):

fpaul-1A (approved):

PR #2638, opened 2024-12-19
fix: test and visual testing update for otter training

cpaulve-1A:
## Related issues

*- No issue associated -*


apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
```diff
@@ -195,7 +195,9 @@ export class CodeEditorViewComponent implements OnDestroy {
    */
   private readonly monacoPromise = firstValueFrom(this.newMonacoEditorCreated.pipe(
     map(() => window.monaco)
-  ));
+  )).catch(
+    (err) => console.error(err)
```
kpanot, line 199: Should not use the logger instead?

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
```diff
@@ -195,7 +195,9 @@ export class CodeEditorViewComponent implements OnDestroy {
    */
   private readonly monacoPromise = firstValueFrom(this.newMonacoEditorCreated.pipe(
```
kpanot, line 196: To have a better typing for `monacoPromise`, maybe you can caught error in the observable and use the second parameter of `firstValueFrom` to define `{ defaultValue: undefined }`

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
matthieu-crouzet: I don't think we set a logger for now on the Otter showcase 
If we don't have one set up, I would expect to do it in another PR

apps/showcase/src/components/training/save-code-dialog/save-code-dialog.component.ts
```diff
@@ -12,6 +12,7 @@ import {
   changeDetection: ChangeDetectionStrategy.OnPush,
   standalone: true,
   imports: [],
+  providers: [NgbActiveModal],
```
matthieu-crouzet, line 15: Why do you have to add this ?
I don't see it on the [bootstrap exemple](https://ng-bootstrap.github.io/#/components/modal/examples#component)

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
matthieu-crouzet: Currently the typing is `Promise<void | typeof Monaco>`, the `void` comes from the `.catch` so adding `{ defaultValue: undefined }` will change the type to `Promise<void | typeof Monaco | undefined>`
if we want to have `Promise<typeof Monaco | undefined>`, we should either have 
```typescript
(err) => {
      console.error(err);
      return undefined;
}
```
or 
```
.catch<undefined>(...)
```

I checked the implementation of the option `defaultValue` it is used only if the subject `newMonacoEditorCreated` complete before providing a value.
[implementation of `firstValueFrom`](https://github.com/ReactiveX/rxjs/blob/f9f07f14ff15cc5f81e4b34005286efb4b5cabc1/packages/rxjs/src/internal/firstValueFrom.ts#L56)

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
fpaul-1A: shouldn't we let it `throw`? Lot of things will be broken if we don't have the global `monaco` instance

apps/showcase/src/components/training/save-code-dialog/save-code-dialog.component.ts
cpaulve-1A: I choose to put it here instead of inside the unit test to keep the SaveCodeDialog as a standalone, but maybe I should add it in the unit tests themselves.

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
cpaulve-1A: I will go for the second as the issue comes from the `firstValueFrom`.

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
cpaulve-1A: It will throw if the monaco editor is never instantiated and the newMonacoEditorCreated subject is completed.
This is the case in the unit tests. 
I could catch the monacoPromise in the tests themselves but that would require making it public and I prefer adding a catch than transforming a private property into a public one for testing purposes.

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
cpaulve-1A: Actually I think defaultValue might better fit my use case here.

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.ts
fpaul-1A: we should be able to test monaco-editor in the unit-tests? only the webcontainer part should be mocked?

apps/showcase/src/components/training/save-code-dialog/save-code-dialog.component.ts
matthieu-crouzet: ok, I don't have opinion if it should be done one UT or in the component itself

fpaul-1A: I had a similar problem with `markdown` and I had to put it on the tests otherwise the component would not receive the instance provided globally (the global config for markdown was not applied)

cpaulve-1A: I will go for the same solution in that case

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #2648, opened 2024-12-23
fix(workspace): the tsconfig links to generate libraries

### LINKED ISSUE #2665: [Bug]: tsconfig.base.json and tsconfig.build.json are not properly generated
### Package name

core

### Package version

11.5.3

### Reproduction steps

Create a project with @o3r/create, then generate an app and a lib:
```sh
npm create @o3r project
ng g application app
ng g library lib
```

### Current result

- File `tsconfig.base.json` is not generated.
- File `tsconfig.build.json` is not generated.
- The app and lib tsconfig files extend `../../tsconfig.json` root file.
- In the lib `tsconfig.spec.json` file, there is a reference to path `./tsconfig.build.composite.json`, but this file was not generated.
- The path mapping in `tsconfig.json` file is not properly configured: the generated path mapping is `"lib": ["libs/lib/dist", "libs/lib/src/public-api"]`, which does not allow for a proper IDE navigation. A proper config could also ease local lib development (today you have to run `ng build lib --watch` to have your lib changes live reloaded).

### Expected result

The generated configuration should be similar to the one of o3r monorepo, with `tsconfig.base.json` and `tsconfig.build.json` files generated, and the app and lib properly referencing these files. Path mappings should also be properly configured.

### Additional comments

_No response_

kpanot:
## Proposed change

fix(workspace): the tsconfig links to generate libraries

It fixes the following points:
- generate the tsconfig.build.json (mandatory for lib standalone publishing)
- remove the `dist/` path from the `tsconfig.base.json` which is leading to wrong path mapping when the library was locally built
- register properly the lib in both tsconfig files
- update the tsconfig.lib.prod.json to extend the correct tsconfig

## Related issues

* :bug: Fix resolves #2665

### DIFF
--- a/packages/@o3r/workspace/schematics/index.it.spec.ts
+++ b/packages/@o3r/workspace/schematics/index.it.spec.ts
@@ -90,7 +90,7 @@ describe('new otter workspace', () => {
     expect(() => packageManagerRunOnProject('@my-sdk/sdk', isInWorkspace, { script: 'spec:upgrade' }, execAppOptions)).not.toThrow();
   });
 
-  test('should add a library to an existing workspace', () => {
+  test('should add a library to an existing workspace', async () => {
     const { workspacePath } = o3rEnvironment.testEnvironment;
     const execAppOptions = { ...getDefaultExecSyncOptions(), cwd: workspacePath };
     const libName = 'test-library';
@@ -115,6 +115,10 @@ describe('new otter workspace', () => {
     expect(existsSync(path.join(workspacePath, 'project'))).toBe(false);
     generatedLibFiles.forEach((file) => expect(existsSync(path.join(inLibraryPath, file))).toBe(true));
     expect(() => packageManagerRunOnProject(libName, true, { script: 'build' }, execAppOptions)).not.toThrow();
+
+    // check tsconfig.lib.prod.json override
+    const tsconfigLibProd = JSON.parse(await fs.readFile(path.join(inLibraryPath, 'tsconfig.lib.prod.json'), { encoding: 'utf8' }));
+    expect(!!tsconfigLibProd.extends && existsSync(path.resolve(inLibraryPath, tsconfigLibProd.extends))).toBe(true);
   });
 
   test('should generate a monorepo setup', async () => {
--- a/packages/@o3r/workspace/schematics/library/index.spec.ts
+++ b/packages/@o3r/workspace/schematics/library/index.spec.ts
@@ -20,24 +20,14 @@ const collectionPath = path.join(__dirname, '..', '..', 'collection.json');
 
 describe('New module generator', () => {
   let initialTree: Tree;
-
   beforeEach(() => {
     initialTree = Tree.empty();
-    initialTree.create('/tsconfig.base.json', JSON.stringify({
-      compilerOptions: {
-        paths: {}
-      }
-    }));
-    initialTree.create('/tsconfig.build.json', JSON.stringify({
-      compilerOptions: {
-        paths: {}
-      }
-    }));
+    initialTree.create('angular.json', '{"version": 1, "projects": {} }');
+    initialTree.create('package.json', '{ "version": "0.0.0-test" }');
   });
 
   it('should generate the minimum mandatory files', async () => {
-    initialTree.create('angular.json', '{"version": 1, "projects": {} }');
-    initialTree.create('package.json', '{ "version": "0.0.0-test" }');
+    initialTree.create('/tsconfig.base.json', JSON.stringify({}));
     initialTree.create('/packages-test/my-new-module/package.json', '{ "version": "0.0.0-test" }');
     initialTree.create('/packages-test/my-new-module/ng-package.json', '{  }');
     const runner = new SchematicTestRunner('schematics', collectionPath);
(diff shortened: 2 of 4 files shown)


packages/@o3r/workspace/schematics/library/templates/ng/tsconfig.lib.prod.json.template
```diff
@@ -0,0 +1,4 @@
+/* For IDE usage only */
```
vscaiceanu-1a, line 1: can you update the integration tests to check for the new files?

fpaul-1A, line 1: Unless I'm missing something, this is actually used the the prod build of the lib, not by the IDE

packages/@o3r/workspace/schematics/library/rules/rules.ng.ts
```diff
@@ -85,26 +85,48 @@ export function ngGenerateModule(options: NgGenerateModuleSchema & { targetPath:
...
+      configFile.compilerOptions.paths[options.packageJsonName] = [
+        path.posix.join(relativeTargetPath, 'src', 'public-api')
+      ];
+      tree.overwrite(tsconfigPath, JSON.stringify(configFile, null, 2));
     });
+
+    if (tsconfigBuild && tree.exists(tsconfigBuild)) {
+      const configFile = tree.readJson(tsconfigBuild) as TsConfigJson;
+      configFile.compilerOptions!.paths![options.packageJsonName].unshift(path.posix.join(relativeTargetPath, 'dist'));
+      path.posix.join(relativeTargetPath, 'dist');
```
fpaul-1A, line 127: is this line doing anything?

packages/@o3r/workspace/schematics/library/index.spec.ts
```diff
@@ -55,12 +45,34 @@ describe('New module generator', () => {
 
     expect(spy).toHaveBeenNthCalledWith(1, '@schematics/angular', 'library', expect.anything());
     expect(tree.exists('/packages-test/my-new-module/tsconfig.json')).toBe(true);
+    expect(tree.exists('/packages-test/my-new-module/tsconfig.lib.prod.json')).toBe(true);
```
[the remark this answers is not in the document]
kpanot: Yes we are, thx to the `MergeStrategy.Overwrite` but to calm your doubt, I specified the test.

fpaul-1A (dismissed):

packages/@o3r/workspace/schematics/index.it.spec.ts
```diff
@@ -115,6 +115,10 @@ describe('new otter workspace', () => {
     expect(existsSync(path.join(workspacePath, 'project'))).toBe(false);
     generatedLibFiles.forEach((file) => expect(existsSync(path.join(inLibraryPath, file))).toBe(true));
     expect(() => packageManagerRunOnProject(libName, true, { script: 'build' }, execAppOptions)).not.toThrow();
+
+    // check tsconfig.lib.prod.json override
+    const tsconfigLibProd = JSON.parse(await fs.readFile(path.join(inLibraryPath, 'tsconfig.lib.prod.json'), { encoding: 'utf8' }));
```
vscaiceanu-1a, line 120: why not `readFileSync`?

kpanot: Just to use something already imported.
It is not changing anything, does it? :D

packages/@o3r/workspace/schematics/library/rules/rules.ng.ts
```diff
@@ -85,26 +85,47 @@ export function ngGenerateModule(options: NgGenerateModuleSchema & { targetPath:
...
-            path.posix.join(relativeTargetPath, 'dist'),
-            path.posix.join(relativeTargetPath, 'src', 'public-api')
-          ];
-          tree.overwrite(tsconfigPath, JSON.stringify(configFile, null, 2));
-        } else {
-          context.logger.warn(`${tsconfigPath} does not contain path mapping, the edition will be skipped`);
+    let tsconfigBuild = findConfigFileRelativePath(tree, ['tsconfig.build.json'], '/');
+
+    if (!tsconfigBase) {
+      context.logger.error('no tsconfig found in the workspace to register the library');
```
matthieu-crouzet, line 91:
```suggestion
      context.logger.error('No TSConfig found in the workspace to register the library.');
```

packages/@o3r/workspace/schematics/library/rules/rules.ng.ts
```diff
@@ -85,26 +85,47 @@ export function ngGenerateModule(options: NgGenerateModuleSchema & { targetPath:
...
+        angularCompilerOptions: {
+          compilationMode: 'partial'
         }
-      } else {
-        context.logger.warn(`No TsConfig file found at ${tsconfigPath}`);
-      }
+      };
+      const tsconfigBuildPath = tsconfigBuild || 'tsconfig.build.json';
+      tree.create(tsconfigBuildPath, JSON.stringify(content, null, 2));
+      tsconfigBuild ||= tsconfigBuildPath;
```
matthieu-crouzet, line 108:
```suggestion
      tsconfigBuild ||= 'tsconfig.build.json';
      tree.create(tsconfigBuild, JSON.stringify(content, null, 2));
```

matthieu-crouzet (dismissed):

packages/@o3r/workspace/schematics/index.it.spec.ts
vscaiceanu-1a: I was thinking it would be better to align with the rest (like `existsSync`) but sure, it's the same thing in the end.

vscaiceanu-1a (dismissed):

packages/@o3r/workspace/schematics/index.it.spec.ts
kpanot: The rest of the file is using `fs.readFile` (example line 130), in other tests we do use both.
Not sure there is something align to (except `fs.readFile` if you thinking aligning at file level)

fpaul-1A (approved):

PR #2685, opened 2025-01-13
feat(chrome-devtools): add a panel to see the current value of all available facts

### LINKED ISSUE #2476: [Feature]: Exposed facts in the o3r debug panel
### Context

The debug panel currently provides access to the ruleset history. Additionally, it would be highly beneficial to display the available facts along with their values.

There are primarily two scenarios where this feature would be useful:

1. Business analysts could easily review what facts are available and determine if they can be utilized in rulesets.
2. For developers, when introducing a new fact, it would simplify the process of verifying that the correct value is assigned without the need to create any rulesets.

### Proposal

Made available in the UI (o3r debug panel) the facts information.

fpaul-1A:
## Proposed change

image
image

## Related issues

*- No issue associated -*

* :rocket: Feature resolves #2476


packages/@o3r/rules-engine/src/components/rules-engine/facts-snapshot/facts-snapshot.component.ts
```diff
@@ -0,0 +1,61 @@
...
+    FormsModule,
+    ReactiveFormsModule
+  ],
+  encapsulation: ViewEncapsulation.None
+})
+export class FactsSnapshotComponent {
+  /**
+   * Full list of available facts with their current value
+   */
+  public facts = input<{ factName: string; value: Facts }[]>([]);
```
matthieu-crouzet, line 37: Shouldn't it be required ?
```suggestion
  public facts = input.required<{ factName: string; value: Facts }[]>();
```

packages/@o3r/rules-engine/src/components/rules-engine/facts-snapshot/facts-snapshot.style.scss
```diff
@@ -0,0 +1,15 @@
+@import '../styling/shared-mixin';
```
matthieu-crouzet, line 1: Could we use `@use` instead 
and maybe we could have a dedicated file for the variables, as here you don't use any mixins

packages/@o3r/rules-engine/src/components/rules-engine/facts-snapshot/facts-snapshot.template.html
```diff
@@ -0,0 +1,26 @@
+<section>
+  @if (facts().length < 1) {
+    <div class="alert alert-danger m-2" role="alert">
```
matthieu-crouzet, line 3: Maybe it should not be `danger`

apps/chrome-devtools/src/app-devtools/app.component.ts
```diff
@@ -112,6 +112,7 @@ export class AppComponent {
   });
 
   public rulesetExecutions$ = this.rulesetHistoryService.rulesetExecutions$;
+  public facts$ = this.rulesetHistoryService.facts$;
```
kpanot, line 115: Is there a reason to not be `readonly` as the others?

packages/@o3r/rules-engine/src/components/rules-engine/facts-snapshot/facts-snapshot.component.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (search) {
+      const matchString = new RegExp(search.replace(/[\s#$()*+,.?[\\\]^{|}-]/g, '\\$&'), 'i');
+      return this.facts().filter(({ factName, value }) =>
+        matchString.test(factName)
+        || (typeof value === 'object'
+          ? matchString.test(JSON.stringify(value))
+          : matchString.test(String(value)))
+      );
+    } else {
+      return this.facts();
```
kpanot, line 58: As used in the `if` and `else` it can be set in constant on top of the function to simplify the reading computed function

packages/@o3r/rules-engine/src/components/rules-engine/facts-snapshot/facts-snapshot.component.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * Search terms
+   */
+  public search = signal('');
+
+  /**
+   * Filtered list of facts using search terms
+   */
+  public filteredFacts = computed(() => {
```
kpanot, line 47: these fields can be flagged as `readonly`

kpanot (approved):

matthieu-crouzet (approved):

mrednic-1A (approved):

packages/@o3r/rules-engine/src/components/rules-engine/rule-actions/rule-actions-pres.template.html
```diff
@@ -1,72 +1,89 @@
...
-          </div>
+@if (actions?.length === 0 && runtimeOutputs.length === 0) {
+  <div class="ruleset-panel-category-body empty">
+    No action
+  </div>
+} @else {
+  <ul class="ruleset-panel-category-body">
+    @for (action of actions; track $index) {
+      <li>
+        @if (isActionBlock(action)) {
```
cpaulve-1A, line 10: Is it possible to use a predicate in a pipe and avoid call to functions in the template ?

fpaul-1A: it doesn't seem feasible to do type predicate inference using Angular pipes (as far as I can tell)
the performance impact is very low here as the function is just accessing a property of the object and the scope of the change detection is limited

cpaulve-1A (approved):

PR #2699, opened 2025-01-16
feat(pipeline): set pipeline version to node LTS version

### LINKED ISSUE #2697: [Feature]: @o3r/pipeline should use node lts
### Context

Github pipelines should by default use nodejs lts version (so currently 22).

Currently, calling `ng add @o3r/pipeline` sets up a pipeline with node 20.

### Proposal

Update this line:
https://github.com/AmadeusITGroup/otter/blob/main/packages/%40o3r/pipeline/schematics/ng-add/templates/github/__dot__github/actions/setup/action.yml#L9
and wherever else the node version would be set.

kpanot:
## Proposed change

feat(pipeline): set pipeline version to node LTS version

## Related issues

* :rocket: Feature resolves #2697

### DIFF
--- a/.renovaterc.json5
+++ b/.renovaterc.json5
@@ -3,6 +3,7 @@
   "extends": [
     "github>AmadeusITGroup/otter//tools/renovate/base",
     "github>AmadeusITGroup/otter//tools/renovate/group/design-factory",
+    "github>AmadeusITGroup/otter//tools/renovate/managers/github-node-version.json5",
     ":labels(dependencies)"
   ],
   "timezone": "Europe/Paris",
@@ -131,7 +132,7 @@
   ],
   "github-actions": {
     "fileMatch": [
-      "/__dot__github/.*/.+\\.ya?ml$"
+      "/(?:__dot__|__empty__\\.)github/.*/.+\\.ya?ml$"
     ]
   }
 }
--- a/docs/renovate/readme.md
+++ b/docs/renovate/readme.md
@@ -47,13 +47,13 @@ module.exports = {
 
     // Used to execute the migration scripts
     "^(npm|yarn) ng update[a-zA-Z0-9.= {}()#@'\/-]*$",
-    
+
     // Used to regenerate the SDK
     "^(npm|yarn) run spec:upgrade$",
 
     // Used to execute the migration scripts of ama-sdk
     "^(npm|yarn) exec schematics @ama-sdk\/schematics:migrate[a-zA-Z0-9.= {}()#'\/-]*$",
-    
+
     // Used for yarn upgrade when using PnP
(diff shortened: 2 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -65,7 +65,7 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
         with:
-          node-version: 20
+          node-version: lts/*
```
fpaul-1A, line 68: shouldn't we let Renovate handle these?

kpanot: I was thinking about it as well but I think it would have been complicated to work on the pipelines templates (as they are not valid yml).
I can see 3 choices:
- align our pipeline and templates to `lts` version
- disalign the Otter repo pipeline (based on renovate upgrade) and the templates which should be manually upgrade
- investigate on a way to make Renovate work on both Otter repo pipeline and templates to upgrade and align the version

I pick the first one which seems to best compromise for me but we can discuss if people disagree about it

fpaul-1A: I would personally prefer to use Renovate to avoid having surprises on Node update
I don't think it matters that much that the `yml` is invalid, Renovate mostly uses regex to parse the file (https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/github-actions/extract.ts)
Also, we've already included the templates in Renovate https://github.com/AmadeusITGroup/otter/pull/2207/files#diff-167211c795925f325dd081ae988bb88f96c1f5780374c8df77c469a4b7a87ff6L11

kpanot: You convinced, i created the Renovate rule

cpaulve-1A (dismissed):

fpaul-1A (approved):

tools/renovate/base.json
```diff
@@ -8,7 +8,8 @@
     "group:recommended",
     "group:test",
     "group:linters",
-    "helpers:pinGitHubActionDigestsToSemver"
+    "helpers:pinGitHubActionDigestsToSemver",
+    "github>AmadeusITGroup/otter//tools/renovate/managers/github-node-version.json5"
```
fpaul-1A, line 12: are we confident enough to add this in the `base` preset?
shouldn't we test it in our config first?

kpanot: - [ ] register the `github-node-version.json5` rule to the `base` rule

fpaul-1A (approved):

matthieu-crouzet (approved):

vscaiceanu-1a (approved):

PR #2717, opened 2025-01-22
fix(workspace): update nx scope in generated executors

kpanot:
## Proposed change

fix(workspace): update nx scope in generated executors

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@o3r/workspace/schematics/library/templates/nx/project.json.template
+++ b/packages/@o3r/workspace/schematics/library/templates/nx/project.json.template
@@ -34,7 +34,7 @@
       }
     },
     "lint": {
-      "executor": "@nrwl/linter:eslint",
+      "executor": "@nx/eslint:lint",
       "configurations": {
         "ci": {
           "quiet": true,
@@ -49,7 +49,7 @@
       }
     },
     "test": {
-      "executor": "@nrwl/jest:jest",
+      "executor": "@nx/jest:jest",
       "options": {
         "jestConfig": "<%= projectRoot %>/jest.config.js",
         "silent": true


packages/@o3r/workspace/schematics/library/templates/nx/project.json.template
```diff
@@ -34,7 +34,7 @@
       }
     },
     "lint": {
-      "executor": "@nrwl/linter:eslint",
+      "executor": "@nx/linter:eslint",
```
fpaul-1A, line 37: shouldn't it be `@nx/eslint:lint` ?

packages/@o3r/workspace/schematics/library/templates/nx/project.json.template
```diff
@@ -34,7 +34,7 @@
       }
     },
     "lint": {
-      "executor": "@nrwl/linter:eslint",
```
matthieu-crouzet, line 37: we don't use this executor as it does not support mjs (at least when we did the migration to ESLint v9 and flat config)

fpaul-1A (approved):

packages/@o3r/workspace/schematics/library/templates/nx/project.json.template
kpanot: In the origin the PR were made to fix the `@nrwl/jest:jest` which is still used.
I fixed the `@nrwl/linter:eslint` because it was also invalid.
I create the issue #2735 to handle the migration to `eslint` cli execution.

matthieu-crouzet (approved):

PR #2726, opened 2025-01-23
fix: dev install for @o3r/pipeline

vscaiceanu-1a:
## Proposed change

The `pipeline` package should be added as a dev dependency. 

We should reflect if this is the case for other packages as well.

### DIFF
--- a/packages/@o3r/schematics/src/utility/wrapper.ts
+++ b/packages/@o3r/schematics/src/utility/wrapper.ts
@@ -91,7 +91,7 @@ export const createSchematicWithMetricsIfInstalled: SchematicWrapper = (schemati
         message: `
 Would you like to share anonymous data about the usage of Otter builders and schematics with the Otter Team at Amadeus ?
 It will help us to improve our tools.
-For more details and instructions on how to change these settings, see https://github.com/AmadeusITGroup/otter/blob/main/docs/telemetry/PRIVACY_NOTICE.md.
+For more details and instructions on how to change these settings, see https://github.com/AmadeusITGroup/otter/blob/main/docs/telemetry/PRIVACY_NOTICE.md
         `,
         default: false
       } as const satisfies Question;
--- a/packages/@o3r/extractors/src/core/wrapper.ts
+++ b/packages/@o3r/extractors/src/core/wrapper.ts
@@ -58,7 +58,7 @@ export const createBuilderWithMetricsIfInstalled: BuilderWrapper = (builderFn) =
         message: `
 Would you like to share anonymous data about the usage of Otter builders and schematics with the Otter Team at Amadeus ?
 It will help us to improve our tools.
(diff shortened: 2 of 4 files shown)


packages/@o3r/pipeline/README.md
```diff
@@ -17,7 +17,7 @@ This module contains tooling around DevOps toolchains.
 ## How to install
 
 ```shell
-ng add @o3r/pipeline
+ng add -D @o3r/pipeline
```
fpaul-1A, line 20: is that a real option? can't find it in the docs

matthieu-crouzet: I don't think it's an option 
But if I remember well we can set a property in the package.json so Angular save it as a devDep and not a dep

matthieu-crouzet: https://angular.dev/tools/cli/schematics-for-libraries#define-dependency-type

vscaiceanu-1a: yes, `-D` doesn't work. I've updated `package.json` and it's looking good.

sdo-1A (dismissed):

tools/github-actions/audit/packaged-action/index.js
matthieu-crouzet, line 1: You should not have changes on github-actions

tools/github-actions/new-version/packaged-action/LICENSE.txt
```diff
@@ -94,33 +94,6 @@ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
 IN THE SOFTWARE.
 
 @o3r/new-version
-Copyright Amadeus SAS
```
kpanot, line 97: should not be removed

sdo-1A (approved):

packages/@o3r/schematics/src/utility/wrapper.ts
```diff
@@ -91,7 +91,7 @@ export const createSchematicWithMetricsIfInstalled: SchematicWrapper = (schemati
         message: `
 Would you like to share anonymous data about the usage of Otter builders and schematics with the Otter Team at Amadeus ?
 It will help us to improve our tools.
-For more details and instructions on how to change these settings, see https://github.com/AmadeusITGroup/otter/blob/main/docs/telemetry/PRIVACY_NOTICE.md.
+For more details and instructions on how to change these settings, see https://github.com/AmadeusITGroup/otter/blob/main/docs/telemetry/PRIVACY_NOTICE.md
```
kpanot, line 94: Not sure to understand this change but ok :D

kpanot (approved):

packages/@o3r/schematics/src/utility/wrapper.ts
vscaiceanu-1a: Well it's quite funny. The `.` makes the link in the logs invalid.

fpaul-1A (approved):

PR #2727, opened 2025-01-23
fix: build builder and schematics for o3r lib

### LINKED ISSUE #2714: [Bug]: TS issues in generated file libs/library/schematics/ng-add/index.ts
### Package name

core

### Package version

11.5.3

### Reproduction steps

Create a project with o3r generator, then generate a library:
```shell
npm create @o3r project
cd project
ng g library lib
```

### Current result

There are TS issues in the generated `libs/library/schematics/ng-add/index.ts` file:

1. `fs.readFileSync` is used, but `fs` is not imported. We should add `import * as fs from 'node:fs';` in the imports section.

2. Replace `const dependenciesToInstall = []` by `const dependenciesToInstall: string[] = []` to avoid lint issues. Same for `dependenciesToNgAdd`.

3. In `ngAddFn` function, the line `return async (tree, context) => {` raises a lint issue because `context` is not used. We could replace it by `_context`.

4. Lines 36 to 38 have compilation issues. 
To fix them, replace:
```ts
    const { getProjectNewDependenciesTypes, getPackageInstallConfig, applyEsLintFix, install } = await import('@o3r/schematics');
    // current package version
    const version = JSON.stringify(fs.readFileSync(packageJsonPath)).version;
    const dependencies = [...dependenciesToInstall, ...dependenciesToNgAdd].reduce((acc, dep) => {
```
By:
```ts
    const { getProjectNewDependenciesTypes, getPackageInstallConfig, applyEsLintFix, setupDependencies, getWorkspaceConfig } = await import('@o3r/schematics');
    const version = (JSON.parse(fs.readFileSync(packageJsonPath, { encoding: 'utf8' })) as PackageJson).version;
    const workspaceProject = options.projectName ? getWorkspaceConfig(tree)?.projects[options.projectName] : undefined;
    const dependencies = [...dependenciesToInstall, ...dependenciesToNgAdd].reduce((acc, dep) => {
```

### Expected result

All the above issues should be fixed in the template file that is used to generate the `ng-add/index.ts` file.

### Additional comments

_No response_

### LINKED ISSUE #2715: [Bug]: missing cpy script in file libs/library/package.json
### Package name

core

### Package version

11.5.3

### Reproduction steps

Create a project with o3r generator, then generate a library:
```shell
npm create @o3r project
cd project
ng g library lib
```

### Current result

In the generated `libs/library/package.json` file, there are occurrences to a `cpy` script, for example:
```json
"prepare:build:builders": "npm run cpy 'collection.json' dist && npm run cpy 'schematics/**/*.json' dist/schematics",
```

However, this `cpy` script is not generated, so running this command will fail with:
```shell
npm error Missing script: "cpy"
```

### Expected result

If we want to generate scripts containing `npm run cpy`, then we should also generate the `cpy` script.

### Additional comments

_No response_

vscaiceanu-1a:
## Related issues

*- No issue associated -*

* :bug: Fix resolves #2714
* :bug: Fix resolves #2715

### DIFF
--- a/packages/@o3r/core/schematics/rule-factories/otter-environment/index.ts
+++ b/packages/@o3r/core/schematics/rule-factories/otter-environment/index.ts
@@ -18,15 +18,23 @@ import {
   TYPES_DEFAULT_FOLDER,
 } from '@o3r/schematics';
 import generateEnvironments from '@schematics/angular/environments/index';
+import type {
+  TsConfigJson,
+} from 'type-fest';
 import * as ts from 'typescript';
 
 const editTsConfigJson = (tree: Tree) => {
-  if (tree.exists('/tsconfig.json')) {
-    const tsConfig = ts.parseConfigFileTextToJson('/tsconfig.json', tree.readText('/tsconfig.json')).config;
-    if (tsConfig.compilerOptions?.noPropertyAccessFromIndexSignature) {
-      delete tsConfig.compilerOptions.noPropertyAccessFromIndexSignature;
+  const tsConfigPath = '/tsconfig.json';
+  if (tree.exists(tsConfigPath)) {
+    const tsConfig = ts.parseConfigFileTextToJson(tsConfigPath, tree.readText(tsConfigPath)).config as TsConfigJson;
+    if (tsConfig.compilerOptions) {
+      if (tsConfig.compilerOptions.noPropertyAccessFromIndexSignature) {
+        delete tsConfig.compilerOptions.noPropertyAccessFromIndexSignature;
+      }
+      tsConfig.compilerOptions.moduleResolution = 'node';
+      tsConfig.compilerOptions.declaration = true;
     }
-    tree.overwrite('/tsconfig.json', JSON.stringify(tsConfig, null, 2));
+    tree.overwrite(tsConfigPath, JSON.stringify(tsConfig, null, 2));
   }
   return tree;
 };
--- a/packages/@o3r/core/schematics/ng-add-create/index.ts
+++ b/packages/@o3r/core/schematics/ng-add-create/index.ts
@@ -16,7 +16,7 @@ import {
 import {
   createSchematicWithMetricsIfInstalled,
   findConfigFileRelativePath,
-  getPackageManagerRunner,
+  getPackageManagerExecutor,
 } from '@o3r/schematics';
 import type {
   PackageJson,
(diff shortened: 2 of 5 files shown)


packages/@o3r/core/schematics/rule-factories/otter-environment/index.ts
```diff
@@ -18,15 +18,23 @@ import {
...
+  TsConfigJson,
+} from 'type-fest';
 import * as ts from 'typescript';
 
 const editTsConfigJson = (tree: Tree) => {
-  if (tree.exists('/tsconfig.json')) {
-    const tsConfig = ts.parseConfigFileTextToJson('/tsconfig.json', tree.readText('/tsconfig.json')).config;
-    if (tsConfig.compilerOptions?.noPropertyAccessFromIndexSignature) {
-      delete tsConfig.compilerOptions.noPropertyAccessFromIndexSignature;
+  const tsConfigPath = '/tsconfig.json';
```
kpanot, line 27: should not be instead something like:

```suggestion
  const tsConfigPath =['/tsconfig.lib.jdon', '/tsconfig.app.json', '/tsconfig.json'].filter((file) => tree.exists(file));
```

and run on all the file found?

vscaiceanu-1a: this is only about the root `tsconfig` file so there won't be app/lib ones

vscaiceanu-1a: waiting for https://github.com/AmadeusITGroup/otter/pull/2648 to be merged and to adapt this PR

packages/@o3r/core/schematics/rule-factories/otter-environment/index.ts
```diff
@@ -18,15 +18,23 @@ import {
...
+  if (tree.exists(tsConfigPath)) {
+    const tsConfig = ts.parseConfigFileTextToJson(tsConfigPath, tree.readText(tsConfigPath)).config as TsConfigJson;
+    if (tsConfig.compilerOptions) {
+      if (tsConfig.compilerOptions.noPropertyAccessFromIndexSignature) {
```
cpaulve-1A, line 31: Why do we need to update their tsconfig?

vscaiceanu-1a: Because this is where the `compilerOptions` that are incompatible with our project are generated.
To avoid any confusion it is best to update them at the root rather than to add them in the `tsconfig.build.json` which is expanding the root one.

In this particular case, we are removing `noPropertyAccessFromIndexSignature` because of some lint issues.

mrednic-1A (approved):

matthieu-crouzet (approved):

PR #2736, opened 2025-01-27
chore: upgrade nx major

fpaul-1A:
## Related issues

*- No issue associated -*


packages/@o3r/telemetry/project.json
```diff
@@ -5,6 +5,7 @@
   "sourceRoot": "packages/@o3r/telemetry/src",
   "prefix": "o3r",
   "implicitDependencies": [
+    "!core",
```
fpaul-1A, line 8: I had to add this to avoid circular dep

packages/@o3r/schematics/project.json
```diff
@@ -4,6 +4,9 @@
   "projectType": "library",
   "sourceRoot": "packages/@o3r/schematics/src",
   "prefix": "o3r",
+  "implicitDependencies": [
+    "!core"
```
fpaul-1A, line 8: I had to add this to avoid circular dep

packages/@o3r/eslint-plugin/src/rules/json/json-dependency-versions-harmonize/json-dependency-versions-harmonize.ts
```diff
@@ -123,7 +123,7 @@ export default createRule<[VersionsHarmonizeOptions, ...any], 'versionUpdate' |
               .filter((depGroup): depGroup is AST.JSONObjectExpression => depGroup.type === 'JSONObjectExpression')
               .forEach((depGroup) => {
                 const report = (name: string, resolvedName: string, dep: AST.JSONProperty, range: string | undefined, bestRange: string | undefined) => {
-                  if (bestRange && bestRange !== range) {
+                  if (bestRange && bestRange !== range && !range?.startsWith('workspace:')) {
```
fpaul-1A, line 126: version harmonize was throwing an error during migration

nx.json
```diff
@@ -615,5 +615,6 @@
         "nxCloudId": "63e6951e043da20dd8321aec"
       }
     }
-  }
+  },
+  "useLegacyCache": true
```
matthieu-crouzet, line 619: Why don't we use the new cache system ?

fpaul-1A: it seems to fail on the CI without it
that, or just the fact that I've changed the `nx.json` forced a refresh of the cache
I think we should keep it disabled for now before further investigation

packages/@o3r/schematics/project.json
matthieu-crouzet: Maybe because of `require.resolve('@o3r/core/package.json');`
(same for telemetry)

tsconfig.base.json
```diff
@@ -110,21 +110,21 @@
       "@o3r/components": [
         "packages/@o3r/components/src/public_api"
       ],
-      "@o3r/components/schemas/*": [
-        "packages/@o3r/components/schemas/*"
-      ],
       "@o3r/components/*": [
         "packages/@o3r/components/src/*/public_api"
       ],
+      "@o3r/components/schemas/*": [
```
matthieu-crouzet, line 116: shouldn't we have first the most precise path to not have the global one matching first ?

packages/@o3r/schematics/project.json
fpaul-1A: yes it's the `require.resolve` which was not caught before

tsconfig.base.json
fpaul-1A: it seems that the order actually matters :\

nx.json
kpanot: Can you create and link an issue for it (to not forget to investigate)?

kpanot (approved):

kpanot:
The `20.4` is out since today, do you think it is worth to update to it directly?

fpaul-1A:
> The `20.4` is out since today, do you think it is worth to update to it directly?

The `20.3.3` was not working for me, I can give the `20.4` a try
We can already upgrade to `20.2` in the meantime

matthieu-crouzet (approved):

PR #2737, opened 2025-01-27
fix(github-actions): undeterministic github actions packaging

fpaul-1A:
## Related issues

*- No issue associated -*


matthieu-crouzet (dismissed):

cpaulve-1A (dismissed):

vscaiceanu-1a (dismissed):

matthieu-crouzet (approved):

nx.json
```diff
@@ -267,7 +267,18 @@
         "{projectRoot}/dist/package.json",
         "{projectRoot}/dist/readme.json",
         "{projectRoot}/dist/LICENSE"
-      ]
+      ],
+      "configurations": {
+        "default": {
+          "dependsOn": []
+        },
```
kpanot, line 274: is the `default` configuration required or the `"dependsOn": []` can be apply to the task directly and be overridden by `force-rebuild` config?

kpanot (approved):

nx.json
fpaul-1A: It turns out that we cannot use `dependsOn` in a `configuration`, I must have imagined things

package.json
```diff
@@ -16,7 +16,7 @@
     "build:tools": "yarn nx run-many --target=build --projects=eslint-plugin,workspace",
     "build:lint": "yarn nx run-many --target=build --projects=eslint-plugin",
     "build:swagger-gen": "yarn nx run-many --target=build-swagger",
-    "publish": "yarn nx run-many --target=publish --exclude=tag:private --nx-bail",
+    "publish": "yarn nx run-many --target=prepare-publish --exclude-task-dependencies && yarn nx run-many --target=publish --exclude=tag:private --nx-bail --exclude-task-dependencies",
```
kpanot, line 19: It seems that we run the publish task with `excludeTaskDependencied` in npmjs and verdaccio publishes.
Is there cases where publish task would need it's deps then?

fpaul-1A: probably not

cpaulve-1A (approved):

kpanot (approved):

PR #2754, opened 2025-01-29
fix(schematics): fix skip install task

kpanot:
## Proposed change

fix(schematics): fix skip install task
fix(workspace): enforce install when pre-commit requested

## Related issues

*- No issue associated -*


packages/@o3r/schematics/src/rule-factories/ng-add/dependencies.ts
```diff
@@ -170,7 +170,7 @@ export const setupDependencies = (options: SetupDependenciesOptions): Rule => {
     const ngAddToRun = new Set(Object.keys(options.dependencies)
       .filter((dep) => options.ngAddToRun?.some((pattern) => typeof pattern === 'string' ? pattern === dep : pattern.test(dep))));
     const requiringInstallList = new Set(Object.entries(options.dependencies).filter(([, { requireInstall }]) => requireInstall).map(([dep]) => dep));
```
fpaul-1A, line 172: This doesn't seem to be used anymore, we can probably change it to something like:
```suggestion
    const isInstallRequired = Object.values(options.dependencies).some(({ requireInstall }) => requireInstall);
```

packages/@o3r/workspace/schematics/ng-add/project-setup.ts
```diff
@@ -125,7 +125,7 @@ export const prepareProject = (options: NgAddSchematicsSchema): Rule => {
       filterPackageJsonScripts,
       setupDependencies({
         dependencies,
-        skipInstall: options.skipInstall,
+        skipInstall: options.skipInstall && options.skipPreCommitChecks,
```
fpaul-1A, line 128: it looks weird to have something being installed when we explicitly set `--skip-install`

kpanot: I did not find any other way that removing the default value of `skipInstall` (as `husky` have to be installed to setup pre-commit hooks).
If you have other better proposal I am ok to change it

packages/@o3r/workspace/schematics/ng-add/schema.json
```diff
@@ -11,8 +11,7 @@
     },
     "skipInstall": {
       "type": "boolean",
-      "description": "Skip the install process",
-      "default": true
+      "description": "Skip the install process"
```
mrednic-1A, line 14: you could add that if not defined, the skipPrecommitChecks will take over.

vscaiceanu-1a:
are there any new tests / assertions that we can do in the integration tests to validate this fix?

packages/@o3r/workspace/schematics/ng-add/project-setup.ts
```diff
@@ -125,10 +126,10 @@ export const prepareProject = (options: NgAddSchematicsSchema): Rule => {
       filterPackageJsonScripts,
       setupDependencies({
         dependencies,
-        skipInstall: options.skipInstall,
+        skipInstall,
         ngAddToRun: internalPackagesToInstallWithNgAdd,
         scheduleTaskCallback: (taskIds) => {
-          if (!options.skipPreCommitChecks) {
+          if (!skipInstall && !options.skipPreCommitChecks) {
```
cpaulve-1A, line 132: There could be a warning in the case the user requested to skip the install but not the precommit checks

kpanot:
> are there any new tests / assertions that we can do in the integration tests to validate this fix?

Regarding the nature of the fix (default value of the generator option leading to an installation) I think the IT should be the tests covering it and the scenario is the same a previously.
Unfortunately this change it impacting only terminal with stdin (which is why we cannot see the issue in the current ITs)

vscaiceanu-1a (approved):

PR #2773, opened 2025-01-31
fix: routing module needs partial BuildTimeProperties

vscaiceanu-1a:
## Proposed change

Routing module needs partial BuildTimeProperties


packages/@o3r/routing/src/tools/app-server-routing.module.ts
```diff
@@ -38,7 +38,7 @@ export class AppServerRoutingModule {
    * Injects the APP_BASE_HREF with a custom factory
    * @param config The application environment configuration
    */
-  public static forRoot(config: BuildTimeProperties): ModuleWithProviders<AppServerRoutingModule> {
+  public static forRoot(config: Partial<BuildTimeProperties>): ModuleWithProviders<AppServerRoutingModule> {
```
matthieu-crouzet, line 41: If we provide a Partial then I think, we need to prefill with the default values
`{ provide: ENVIRONMENT_CONFIG_TOKEN, useValue: { ...DEFAULT_BUILD_PROPERTIES, ...config } }`

vscaiceanu-1a: yes, it's best to avoid any breaking change for this small fix.
I spread with the default and keep the factory as before, without Partial

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #2787, opened 2025-02-03
chore: align typescript-eslint deps

vscaiceanu-1a:
## Proposed change

Align some deps

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -16,7 +16,7 @@ packageExtensions:
       "@swc/types": "*"
   "@typescript-eslint/rule-tester@*":
     dependencies:
-      "@typescript-eslint/parser": ~8.21.0
+      "@typescript-eslint/parser": ~8.22.0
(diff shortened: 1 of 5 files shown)


fpaul-1A (approved):

kpanot (approved):

.yarnrc.yml
```diff
@@ -16,7 +16,7 @@ packageExtensions:
       "@swc/types": "*"
   "@typescript-eslint/rule-tester@*":
     dependencies:
-      "@typescript-eslint/parser": ~8.21.0
+      "@typescript-eslint/parser": ~8.22.0
```
cpaulve-1A, line 19: Do we still need to enforce it?

kpanot: It's still only in its devDependencies so I would say yes if we are using typings from `@typescript-eslint/parser`.

kpanot (approved):

PR #2789, opened 2025-02-03
fix: avoid daylight saving switch issue on utils.Date

cpaulve-1A:
## Context 

Italy changed their daylight saving logic several time in the past (may prior 1980, march following then).

As a result, the offset computation for lightsaving switch prior to 1980 are incorrect:
image
Here, the timezone offset should be -60 prior to the timechange (2am).
A valid offset would be the following (working example with France daylight saving switch in 1992):
image

This results in invalid utils.Date which are serialized to the day before.

## Proposed change

Setting utils.Date time to 12:00 avoids this error (12:00 being correctly computed as after the time change).
It also reduces the impact of an incorrect offset.

## Related issues

*- No issue associated -*


matthieu-crouzet (dismissed):

packages/@ama-sdk/core/src/fwk/date.ts
```diff
@@ -75,9 +75,9 @@ export namespace utils {
     constructor(year: number, month: number, date?: number)
     constructor(...args: any[]) {
       if (args && typeof args[0] === 'string' && args[0].lastIndexOf('T') < 0) {
-        args[0] = `${args[0]}T00:00:00Z`;
+        args[0] = `${args[0]}T12:00:00Z`;
```
kpanot, line 78: Maybe you can add a comment or a test case it UT to explain the issue (for the future us)?

kpanot (approved):

packages/@ama-sdk/core/src/fwk/date.ts
cpaulve-1A: I will add a comment because I think we have limitation on jest to support timezone change in the test

cpaulve-1A: https://stackoverflow.com/questions/56261381/how-do-i-set-a-timezone-in-my-jest-config

fpaul-1A (approved):

fpaul-1A (dismissed):

kpanot (dismissed):

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #2792, opened 2025-02-04
Fix: add @o3r/eslint-config via @o3r/workspace

### LINKED ISSUE #2482: [Bug]: several peerDeps are flagged as optional in @o3r/eslint-plugin but usage of it are exposed at the root
### Package name

eslint-plugin

### Package version

11.4.1

### Reproduction steps

Use `@o3r/eslint-plugin` without having `yaml-eslint-parser` installed.

### Current result

Issue while running linter because `yaml-eslint-parser` is not installed.

### Expected result

To not introduce breaking changes, in v11 we can remove the optional flags.
For v12 we can expose sub-entries like `@o3r/eslint-plugin/yaml`

### Additional comments

Same issue for `jsonc-eslint-parser` and `@angular-eslint/template-parser`

### LINKED ISSUE #2766: [Bug]: adding eslint-config-otter when adding @o3r/workspace fails on a project with eslint
### Package name

workspace

### Package version

11.5.0

### Reproduction steps

```shell
ng new --no-create-application my-workspace
cd my-workspace
npm install -D eslint
ng add @o3r/workspace # make sure angular and otter version are aligned to avoid peer dep errors

You already have ESLint installed. Would you like to add Otter config rules for ESLint?
Otherwise, you can add them later via this command: ng add @o3r/eslint-config-otter Yes
```

### Current result

Collection "@o3r/eslint-config-otter" cannot be resolved.

### Expected result

Adds workspace along with Otter eslint correctly. 

### Additional comments

_No response_

vscaiceanu-1a:
## Related issues

* :bug: Fix resolves #2766

### DIFF
--- a/packages/@o3r/eslint-config-otter/schematics/ng-add/index.ts
+++ b/packages/@o3r/eslint-config-otter/schematics/ng-add/index.ts
@@ -1,3 +1,4 @@
+import * as fs from 'node:fs';
 import * as path from 'node:path';
 import {
   chain,
@@ -5,6 +6,9 @@ import {
   SchematicContext,
   Tree,
 } from '@angular-devkit/schematics';
+import type {
+  PackageJson,
+} from 'type-fest';
 import {
   updateLinterConfigs,
 } from './linter';
@@ -50,6 +54,7 @@ function ngAddFn(options: NgAddSchematicsSchema): Rule {
       getO3rPeerDeps,
       getProjectNewDependenciesTypes,
       removePackages,
+      registerPackageCollectionSchematics,
       getPackageInstallConfig
     } = await import('@o3r/schematics');
     const depsInfo = getO3rPeerDeps(path.resolve(__dirname, '..', '..', 'package.json'), true, /^@(?:o3r|ama-sdk|eslint-)/);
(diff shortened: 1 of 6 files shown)


.husky/pre-commit
```diff
@@ -1,3 +0,0 @@
-yarn lint-staged
```
kpanot, line 1: Faster than `-n` :D ?

packages/@o3r/eslint-config-otter/schematics/ng-add/index.ts
```diff
@@ -78,6 +84,7 @@ function ngAddFn(options: NgAddSchematicsSchema): Rule {
       });
 
     return () => chain([
+      registerPackageCollectionSchematics(packageJson),
```
kpanot, line 87: Not sure to understand the goal of calling this function in the 2 files.
The purpose of this function is to register the schematics exposed by the package to ng cli to not have to specify the package name to use them.
This 2 packages have only `ng-add` and `ng-update`, not sure there is schematics to register :S

.husky/pre-commit
vscaiceanu-1a: in case of a cherry pick, yes :)

vscaiceanu-1a:
better fix in https://github.com/AmadeusITGroup/otter/pull/2795

Issue #2784, opened 2025-02-03
[Feature]: Add rule id in rules-engine logs


fpaul-1A:
### Context

When errors are logged today we get a message like
`Error: Invalid left operand: undefined`
This is not very helpful for debugging

### Proposal

We should log the rule.id with the error messages

Narendra-Kamath:
@fpaul-1A - raised a PR https://github.com/AmadeusITGroup/otter/pull/2793 - workflows need approval to run. Kindly review. Thank you!

PR #2793, opened 2025-02-04
feat(rules-engine): added rule ID for logger error message

Narendra-Kamath:
## Proposed change

Added rule ID for the output error message in rules engine while evaluating a rule. This helps to log the error message with an associated rule ID, instead of just the error text.

## Related issues

https://github.com/AmadeusITGroup/otter/issues/2784

### DIFF
--- a/packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
+++ b/packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
@@ -300,7 +300,7 @@ export class RulesetExecutor {
               if (this.rulesEngine.debugMode) {
                 output.evaluation = handleRuleEvaluationDebug(rule, this.ruleset.name, output.actions, output.error, runtimeFactValues, factValues, oldFactValues, inputFacts);
               } else if (output.error) {
-                this.rulesEngine.logger?.error(output.error);
+                this.rulesEngine.logger?.error(`Error while evaluating rule ID: ${rule.id}`, output.error);
                 this.rulesEngine.logger?.warn(`Skipping rule ${rule.name}, and the associated ruleset`);
               }
               return output;


packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -292,9 +292,9 @@ export class RulesetExecutor {
                   acc[id] = factValues![index];
                   return acc;
                 }, {}), runtimeFactValues);
-              } catch (error) {
+              } catch (error: any) {
```
kpanot, line 295: per default `error` is already `any` if I am not mistaken

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
```diff
@@ -292,9 +292,9 @@ export class RulesetExecutor {
...
                 output.actions = undefined;
-                output.error = error;
+                output.error = `Error while evaluating rule ID: ${rule.id}, ${error}`;
```
kpanot, line 297: Here you rely on the serialization to string of the error, I don't think it is a good idea.
It may be simpler to modify the `logger.error` to add the information relative to the rule ID

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
Narendra-Kamath: this was added to avoid lint error. The line will be reverted.

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
Narendra-Kamath: Added to logger.error instead of manipulating the output error object

kpanot (approved):

Narendra-Kamath:
@kpanot - I do not have permission to add reviewers, it requires 2 reviewers with write permission to merge the change. Also, no permission to add labels. Please do the needful! Thank you!

packages/@o3r/rules-engine/src/engine/ruleset-executor.ts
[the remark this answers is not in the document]
Narendra-Kamath: Raised a new PR with base branch from release/12.0.0-rc - https://github.com/AmadeusITGroup/otter/pull/2798 

Current PR can be closed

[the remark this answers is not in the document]
Narendra-Kamath: Closing this PR since a new PR has been raised to release/12.0.0-rc 

new PR reference - https://github.com/AmadeusITGroup/otter/pull/2798

PR #2804, opened 2025-02-05
fix: starting file does not open on training container

cpaulve-1A:
## Proposed change
Invalid model as options seems to impact the load of Monaco Editor

As monaco editor does not need the definition of model as an input and options,
simplify the options

## Related issues

*- No issue associated -*


matthieu-crouzet (approved):

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.spec.ts
```diff
@@ -12,7 +15,10 @@ describe('CodeEditorViewComponent', () => {
 
   beforeEach(async () => {
     await TestBed.configureTestingModule({
-      imports: [CodeEditorViewComponent]
+      imports: [CodeEditorViewComponent],
+      providers: [
+        { provide: NGX_MONACO_EDITOR_CONFIG, useValue: { baseUrl: `` } }
```
fpaul-1A, line 20: I don't understand why this is needed, `NGX_MONACO_EDITOR_CONFIG` is provided by `MonacoEditorModule` which is imported by `CodeEditorViewComponent`

fpaul-1A (approved):

apps/showcase/src/components/training/code-editor-view/code-editor-view.component.spec.ts
kpanot, line 20: strange the linter did not complain for ` `` ` instead of `''`

cpaulve-1A: I will recheck

cpaulve-1A: I think it is only provided by the forRoot of the MonacoEditorModule, which is not imported in the test 😉

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #2817, opened 2025-02-07
chore: upgrade angular 19.1

fpaul-1A:
## Related issues

*- No issue associated -*


packages/@ama-sdk/core/src/utils/crypto.ts
```diff
@@ -11,7 +11,7 @@ declare global {
  * Converts an ArrayBuffer to a string.
  * @param buf ArrayBuffer to convert
  */
-export function ab2str(buf: ArrayBuffer) {
+export function ab2str(buf: ArrayBufferLike) {
```
vscaiceanu-1a, line 14: why did you have to do this change?

fpaul-1A: I had the following error 
```
error TS2345: Argument of type 'ArrayBufferLike' is not assignable to parameter of type 'ArrayBuffer'.
```
caused by typescript upgrade to 5.7 https://devblogs.microsoft.com/typescript/announcing-typescript-5-7/#typedarrays-are-now-generic-over-arraybufferlike

packages/@o3r/workspace/package.json
```diff
@@ -22,7 +22,7 @@
   "main": "./dist/src/public_api.js",
   "types": "./dist/src/public_api.d.ts",
   "peerDependencies": {
-    "@angular-devkit/architect": "^0.1900.0",
+    "@angular-devkit/architect": "~0.1901.0",
```
sdo-1A, line 25:
```suggestion
    "@angular-devkit/architect": "^0.1900.0",
```
peer dependencies should not change right?

packages/@ama-sdk/core/src/utils/crypto.ts
vscaiceanu-1a: It rings a bell.

sdo-1A (approved):

mrednic-1A (approved):

PR #2829, opened 2025-02-10
fix: use createSchematicWithOptionsFromWorkspace on every schematics

### LINKED ISSUE #2822: [Bug]: The schematic wrapper `createSchematicWithOptionsFromWorkspace` should be used around each schematic
### Package name

global

### Package version

11.6

### Reproduction steps

- Specify a default option like this, in the `angular.json`:
```json5
{
  ...
  "schematics": {
       '*:*': {
          customOption: 'value'
        }
   }
}
```
- Run a schematic that does not have the wrapper `createSchematicWithOptionsFromWorkspace`

### Current result

The schematic will be run without the option `customOption`.

### Expected result

The schematic should be run with the option `customOption`.

### Additional comments

Each schematic should be wrapped by the wrapper `createSchematicWithOptionsFromWorkspace`, created with https://github.com/AmadeusITGroup/otter/pull/2769 to fix https://github.com/AmadeusITGroup/otter/issues/2758 .

matthieu-crouzet:
## Related issues

* :bug: Fix resolves #2822


packages/@ama-sdk/client-angular/schematics/ng-add/index.ts
```diff
@@ -87,6 +87,9 @@ function ngAddFn(options: NgAddSchematicsSchema): Rule {
  * @param options
  */
 export const ngAdd = (options: NgAddSchematicsSchema): Rule => async (_, { logger }) => {
-  const { createSchematicWithMetricsIfInstalled } = await import('@o3r/schematics').catch(reportMissingSchematicsDep(logger));
-  return createSchematicWithMetricsIfInstalled(ngAddFn)(options);
+  const {
+    createSchematicWithMetricsIfInstalled,
+    createSchematicWithOptionsFromWorkspace
+  } = await import('@o3r/schematics').catch(reportMissingSchematicsDep(logger));
+  return createSchematicWithOptionsFromWorkspace(createSchematicWithMetricsIfInstalled(ngAddFn))(options);
```
mrednic-1A, line 94: we are calling the 2 createSchematic everywhere. 
can't we have a generic wrapper 'createSchematic' which will call these 2 ? 
and if we'll have in the future another one (which is less probably) we could just add it there

packages/@o3r/pipeline/schematics/ng-add/index.ts
```diff
@@ -59,6 +59,7 @@ function ngAddFn(options: NgAddSchematicsSchema): Rule {
     const baseTemplateSource = apply(url(`./templates/${options.toolchain}`), [
       template({
         ...options,
+        npmRegistry: options.npmRegistry || '',
```
fpaul-1A, line 62: is that relevant to the PR ?

fpaul-1A (approved):

packages/@ama-sdk/client-angular/schematics/ng-add/index.ts
```diff
@@ -87,6 +87,8 @@ function ngAddFn(options: NgAddSchematicsSchema): Rule {
  * @param options
  */
 export const ngAdd = (options: NgAddSchematicsSchema): Rule => async (_, { logger }) => {
-  const { createSchematicWithMetricsIfInstalled } = await import('@o3r/schematics').catch(reportMissingSchematicsDep(logger));
-  return createSchematicWithMetricsIfInstalled(ngAddFn)(options);
+  const {
+    createOtterSchematic
+  } = await import('@o3r/schematics').catch(reportMissingSchematicsDep(logger));
+  return createOtterSchematic(ngAddFn)(options);
```
kpanot, line 93: Not a decision taken part of this PR by should we really stop the `ng-add` if `@o3r/schematics'` is not provided?
It seems a bit drastic for telemetry and additional options, maybe a warning message and defaulting to `ngAddFn(options)` would be clearer

kpanot (approved):

packages/@ama-sdk/client-angular/schematics/ng-add/index.ts
matthieu-crouzet: You can open an issue so we can discuss that with the team 😄

PR #2830, opened 2025-02-10
fix: adjust timezone based on new date timezone

cpaulve-1A:
## Proposed change

Changing timezone from 00 to 12 introduced a breaking change and does not  work for DateTime issue

Recompute timezone based on the difference between the prior timezone and the new date one

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@ama-sdk/core/src/fwk/date.spec.ts
+++ b/packages/@ama-sdk/core/src/fwk/date.spec.ts
@@ -47,6 +47,14 @@ describe('DateTime', () => {
     expect(dateUtils).toEqual(dateCompare);
   });
 
+  it('should support daylight saving time switch', () => {
+    expect((new Date('1972-05-28T00:00:00')).getTimezoneOffset()).toBe(-120);
+    expect((new Date('1972-05-27T23:59:59')).getTimezoneOffset()).toBe(-60);
+    const dateUtils = new utils.DateTime('1972-05-28T00:00:00');
+    // At midnight in italy, for this date, time should be 1 am
+    expect(dateUtils.toJSON()).toEqual('1972-05-28T01:00:00.000');
+  });
+
   it('should ignore timezone of datetime', () => {
     const date1 = '2015-03-25T12:00:00-02:00';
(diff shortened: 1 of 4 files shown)


packages/@ama-sdk/core/src/fwk/date.spec.ts
```diff
@@ -104,7 +104,7 @@ describe('Date', () => {
   it('should be converted to a Js DateTime', () => {
     const date1 = '1988-06-07';
 
-    const dateCompare = (new utils.DateTime('1988-06-07T12:00:00Z'));
```
kpanot, line 107: Not sure to understand the breaking change introduced by the previous change but if it is easy to test, maybe it can be valuable to add a UT regarding it

cpaulve-1A: I cannot switch the TZ on time and I need a specific use case so I needed to force the timezone to Italy for the whole test suite (which is not ideal but it is not possible to activate it on a test basis...)

packages/@ama-sdk/core/src/fwk/date.ts
```diff
@@ -77,9 +85,9 @@ export namespace utils {
       if (args && typeof args[0] === 'string' && args[0].lastIndexOf('T') < 0) {
         // Set time to 12 to limit scenarios where the native date api is not able to correctly
```
fpaul-1A, line 86: the comment is no longer relevant

fpaul-1A (dismissed):

fpaul-1A (dismissed):

packages/@ama-sdk/core/src/fwk/date.spec.ts
kpanot: Actually I just meant to secure the minutes to `T00:00:00Z` as you said in your description that `T12:00:00Z` was a breaking change, as this breaking change is not obvious.
It does not seem that TZ switching is needed for it but I can be wrong.

fpaul-1A (approved):

Issue #640, opened 2023-08-08
[Feature]: Support of Open API Parameter Serialization in SDK requests


kpanot:
### Context

Today the serialization of the request parameters is defined at the @ama-sdk/core level with a static implementation.
The Open Api 3.0+ specification bring a way to customize the parameter serialization strategy via the 2 parameters `style` and `explode`.

### Proposal

Support of the `style` and `explode` parameter specification in Open Api specification.

**Links:**
- Open Api documentation: https://swagger.io/docs/specification/serialization/

PR #2865, opened 2025-02-18
[DO NOT REVIEW] feat: support exploded syntax in sdk

sdo-1A:
## Proposed change

Support of exploded syntax for query and path parameters: https://swagger.io/docs/specification/v3_0/serialization/

## Related issues

#640 

*- No issue associated -*

### DIFF
--- a/packages/@ama-sdk/client-angular/src/api-angular-client.ts
+++ b/packages/@ama-sdk/client-angular/src/api-angular-client.ts
@@ -2,6 +2,7 @@ import type {
   ApiClient,
   ApiTypes,
   BaseApiClientOptions,
+  ParamSerialization,
   PartialExcept,
   RequestOptions,
   RequestOptionsParameters,
@@ -12,8 +13,10 @@ import {
   EmptyResponseError,
   ExceptionReply,
   extractQueryParams,
+  filterUndefinedQueryParams,
   filterUndefinedValues,
   getResponseReviver,
+  preparePathParams,
   prepareUrl,
   processFormData,
   ReviverReply,
@@ -66,7 +69,17 @@ export class ApiAngularClient implements ApiClient {
   }
 
   /** @inheritdoc */
-  public extractQueryParams<T extends { [key: string]: any }>(data: T, names: (keyof T)[]): { [p in keyof T]: string; } {
+  public extractQueryParams<T extends { [key: string]: any }>(data: T, names: (keyof T)[]): { [p in keyof T]: string; };
+  /** @inheritdoc */
+  public extractQueryParams<T extends { [key: string]: any }>(data: T, names: { [K in keyof T]?: Omit<ParamSerialization, 'value'> }): { [p in keyof T]?: ParamSerialization; };
+  /** @inheritdoc */
+  public extractQueryParams<T extends { [key: string]: any }>(
+    data: T,
+    names: (keyof T)[] | { [K in keyof T]?: Omit<ParamSerialization, 'value'> }
+  ): { [p in keyof T]: string; } | { [p in keyof T]?: ParamSerialization; } {
+    if (Array.isArray(names)) {
+      return extractQueryParams(data, names);
+    }
     return extractQueryParams(data, names);
   }
 
@@ -75,7 +88,8 @@ export class ApiAngularClient implements ApiClient {
     let opts: RequestOptions = {
       ...requestOptionsParameters,
       headers: new Headers(filterUndefinedValues(requestOptionsParameters.headers)),
-      queryParams: filterUndefinedValues(requestOptionsParameters.queryParams)
+      queryParams: filterUndefinedValues(requestOptionsParameters.queryParams),
+      queryParameters: requestOptionsParameters.queryParameters ? filterUndefinedQueryParams(requestOptionsParameters.queryParameters) : undefined
     };
     if (this.options.requestPlugins) {
       for (const plugin of this.options.requestPlugins) {
@@ -90,7 +104,12 @@ export class ApiAngularClient implements ApiClient {
   }
 
   /** @inheritdoc */
-  public prepareUrl(url: string, queryParameters: { [key: string]: string | undefined } = {}) {
+  public preparePathParams<T extends { [key: string]: any }>(data: T, pathParameters: { [K in keyof T]?: Omit<ParamSerialization, 'value'> }): { [p in keyof T]: string } {
+    return preparePathParams(data, pathParameters);
+  }
+
+  /** @inheritdoc */
+  public prepareUrl(url: string, queryParameters: { [key: string]: string | undefined } | { [key: string]: ParamSerialization | undefined } = {}) {
     return prepareUrl(url, queryParameters);
   }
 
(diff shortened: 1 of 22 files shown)


packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/java/com/amadeus/codegen/ts/LambdaHelper.java
```diff
@@ -67,7 +67,7 @@ public UrlParamReplacerLambda() {}
 
       @Override
       public String formatFragment(String fragment) {
-          return fragment.replaceAll("\\{([\\w_-]+)\\}", "\\${data['$1']}");
+          return fragment.replaceAll("\\{([\\w_-]+)\\}", "\\${pathParameters['$1']}");
```
fpaul-1A, line 70: the doc of the class should be updated as well (same for the other one bellow)

packages/@ama-sdk/core/src/fwk/api.helpers.ts
```diff
@@ -52,6 +217,22 @@ export function filterUndefinedValues(object?: { [key: string]: string | undefin
...
+ * @returns an object without undefined values
+ */
+export function filterUndefinedQueryParams(object?: { [key: string]: ParamSerialization | undefined }): { [key: string]: ParamSerialization } {
+  return object
+    ? Object.keys(object)
+      .filter((objectKey) => !!object[objectKey])
+      .reduce<{ [key: string]: ParamSerialization }>((acc, objectKey) => {
+        acc[objectKey] = object[objectKey] as ParamSerialization;
+        return acc;
+      }, {})
```
fpaul-1A, line 232:
```suggestion
    ? Object.fromEntries(Object.entries(object).filter(([_key, value]) => typeof value !== 'undefined'))
```

packages/@ama-sdk/client-angular/src/api-angular-client.ts
```diff
@@ -75,7 +88,8 @@ export class ApiAngularClient implements ApiClient {
     let opts: RequestOptions = {
       ...requestOptionsParameters,
       headers: new Headers(filterUndefinedValues(requestOptionsParameters.headers)),
-      queryParams: filterUndefinedValues(requestOptionsParameters.queryParams)
+      queryParams: filterUndefinedValues(requestOptionsParameters.queryParams),
```
kpanot, line 91: I think you have a breaking change here for the plugins:
Regarding your template update, you always pass `{}` as value for `queryParams`.
Which means that the external plugins implemented to treat `queryParams` will not have any effect anymore.

There is also another point to handle: even if the correct `queryParams` are provided to the plugin, the update of it by a plugin will be ignore (as `queryParameters` has the priority).
A suggestion (not the best one :S) can be to raise an error if the ref to the `queryParams` is different that the original one and if `!!queryParameters`

packages/@ama-sdk/client-angular/src/api-angular-client.ts
```diff
@@ -66,7 +69,17 @@ export class ApiAngularClient implements ApiClient {
...
+  public extractQueryParams<T extends { [key: string]: any }>(data: T, names: { [K in keyof T]?: Omit<ParamSerialization, 'value'> }): { [p in keyof T]?: ParamSerialization; };
+  /** @inheritdoc */
+  public extractQueryParams<T extends { [key: string]: any }>(
+    data: T,
+    names: (keyof T)[] | { [K in keyof T]?: Omit<ParamSerialization, 'value'> }
+  ): { [p in keyof T]: string; } | { [p in keyof T]?: ParamSerialization; } {
+    if (Array.isArray(names)) {
+      return extractQueryParams(data, names);
+    }
     return extractQueryParams(data, names);
```
kpanot, line 83: Why is it different to 

```suggestion
    return extractQueryParams(data, names);
```
?

packages/@ama-sdk/core/src/fwk/api.helpers.ts
```diff
@@ -5,35 +5,200 @@ import type {
...
+  /** Style of the parameter */
+  style: string;
+}
+
+/**
+ * Check if the value is a record of type <string, string>.
+ * @param obj
+ */
+export function isRecordTypeString(obj: any): obj is { [key: string]: string } {
+  return Object.values(obj).every((item) => typeof item === 'string');
```
kpanot, line 26: This is `true` for any `type obj !== 'object'` maybe you should check:
```suggestion
  return  typeof obj === 'object' && Object.values(obj).every((item) => typeof item === 'string');
```

packages/@ama-sdk/core/src/fwk/api.helpers.ts
```diff
@@ -52,6 +217,22 @@ export function filterUndefinedValues(object?: { [key: string]: string | undefin
...
+ * Returns a filtered json object, removing all the undefined query parameters
+ * @param object JSON object to filter
+ * @returns an object without undefined values
+ */
+export function filterUndefinedQueryParams(object?: { [key: string]: ParamSerialization | undefined }): { [key: string]: ParamSerialization } {
+  return object
+    ? Object.keys(object)
+      .filter((objectKey) => !!object[objectKey])
+      .reduce<{ [key: string]: ParamSerialization }>((acc, objectKey) => {
+        acc[objectKey] = object[objectKey] as ParamSerialization;
```
kpanot, line 230:
```suggestion
        acc[objectKey] = object[objectKey]!;
```
?

packages/@ama-sdk/client-fetch/src/api-fetch-client.ts
```diff
@@ -63,12 +66,27 @@ export class ApiFetchClient implements ApiClient {
...
+  ): { [p in keyof T]: string; } | { [p in keyof T]?: ParamSerialization; } {
+    if (Array.isArray(names)) {
+      return extractQueryParams(data, names);
+    }
     return extractQueryParams(data, names);
   }
 
   /** @inheritdoc */
-  public tokenizeRequestOptions(url: string, queryParameters: { [key: string]: string }, piiParamTokens: { [key: string]: string }, data: any): TokenizedOptions | undefined {
+  public tokenizeRequestOptions(
```
kpanot, line 84: As other function you implemented, maybe an overload of this one to deprecate `{ [key: string]: string }` can be valuable?

packages/@ama-sdk/core/src/fwk/api.helpers.ts
kpanot: (you can even remove `_key`:
```suggestion
    ? Object.fromEntries(Object.entries(object).filter(([, value]) => typeof value !== 'undefined'))
```
)

kpanot:
You may want to add a `deprecate: ....` commit message to list your commit in deprecated warnings in release not

packages/@ama-sdk/client-angular/src/api-angular-client.ts
kpanot: But it should not change the return type of `extractQueryParams`.
Did you face type issue without the `Array.isArray`?
(I was actually triggered because adding JS process to handle TypeScript type issue is generally not the good practice :S)

sdo-1A:
Closed: Alternative solution implemented in #2896 to avoid breaking changes

PR #2871, opened 2025-02-19
Renovate presets: run `otter-ng-update` post upgrade commands in mode executionMode=update

malek-benazzouz:
## Proposed change

This mode makes the post upgrade command run for each dependency update, rather than once for all the updates in the branch. 
We need to set this mode because `ng update` does not allow the `--from=... --to=... --migrate-only` combination for many packages at the same time.

## Related issues

*- No issue associated -*

### DIFF
--- a/tools/renovate/tasks/otter-ng-update.json
+++ b/tools/renovate/tasks/otter-ng-update.json
@@ -4,9 +4,7 @@
   "packageRules": [
     {
       "matchPackageNames": [
-        "@o3r/core",
-        "@ama-sdk/core",
-        "@otter/core"
+        "@o3r/core"
       ],
       "postUpgradeTasks": {
         "commands": [
@@ -16,7 +14,7 @@
         "fileFilters": [
           "!**/.{npmrc,yarnrc*}"
         ],
-        "executionMode": "branch"
+        "executionMode": "update"
       }
     }
   ]


kpanot (approved):

tools/renovate/tasks/otter-ng-update.json
```diff
@@ -4,9 +4,7 @@
   "packageRules": [
     {
       "matchPackageNames": [
-        "@o3r/core",
-        "@ama-sdk/core",
```
malek-benazzouz, line 8: Removed `"@ama-sdk/core"` package because it exposes no `ng update` schematic, so it leads to the following issue:
```sh
Command failed: npm run ng update @ama-sdk/core -- --from=12.0.0-rc.0 --to=12.0.0-rc.11 --migrate-only --allow-dirty --force
[...]
Package does not provide migrations.
```

tools/renovate/tasks/otter-ng-update.json
```diff
@@ -4,9 +4,7 @@
...
-        "@otter/core"
```
malek-benazzouz, line 9: Removed `"@otter/core"` because it no longer exists.

tools/renovate/tasks/otter-ng-update.json
fpaul-1A: there are migration scripts on `@ama-sdk/schematics` which is included in the `ng-update/packageGroup` of `@o3r/core`
assuming there should be no repository with a dependency on `@ama-sdk/schematics` but no dependency on `@o3r/core`, we can keep the list with only `@o3r/core`, otherwise we should include `@ama-sdk/schematics` as well

fpaul-1A (approved):

PR #2881, opened 2025-02-20
feat: communication protocol

mrednic-1A:
## Related issues

*- No issue associated -*


packages/@ama-mfe/messages/README.md
```diff
@@ -0,0 +1,3 @@
+# Ama mfe messages package
+
+Contains interfaces for the common messages used by a communication protocol doing the dialog between iframes in a UI micro frontend architecture
```
cpaulve-1A, line 3: A bit more context with exemples or links to related package could be interested here

packages/@ama-mfe/ng-utils/README.md
```diff
@@ -0,0 +1,3 @@
+# Ama mfe angular utils
+
+This Angular project includes various helpers, wrappers, and services designed to facilitate communication within a microfrontend architecture based on iframes.
```
cpaulve-1A, line 3: This part should be improved with a guide on how to use the framework

cpaulve-1A (approved):

packages/@ama-mfe/messages/.compodocrc.json
```diff
@@ -0,0 +1,22 @@
+{
```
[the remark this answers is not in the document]
mrednic-1A: need to check that

packages/@ama-mfe/ng-utils/README.md
mrednic-1A: yes. 
also how to create a custom producer/consumer and register it in the com protocol

packages/@ama-mfe/messages/.compodocrc.json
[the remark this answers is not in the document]
matthieu-crouzet: On this monorepo we have only compodoc

packages/@ama-mfe/ng-utils/README.md
matthieu-crouzet: how to create could be handle through a schematic maybe ! :)

mrednic-1A: that could be a good idea. even if the code itself is not long because these services are very target specific (not too many common code), but at least the naming and best practices could be shared via a schematic.

packages/@ama-mfe/ng-utils/src/connect/connect.directive.ts
```diff
@@ -0,0 +1,81 @@
...
+
+      this.messageService.disconnect();
+      if (clientOrigin && moduleWindow) {
+        try {
+          await this.messageService.listen(connectId, {
+            window: moduleWindow,
+            origin: clientOrigin
+          });
+        } catch (e) {
+          // eslint-disable-next-line no-console -- log the error - replace this with a proper logger
```
matthieu-crouzet, line 68: Maybe set a TODO associated to an issue so we don't forget to do it

matthieu-crouzet: Could be as well associated with the TODO https://github.com/AmadeusITGroup/otter/pull/2881/files#diff-86c746b611675074b9596b1abc0830a4308cc69f58b3a066797f891dd4f50dc4R65

packages/@ama-mfe/ng-utils/src/navigation/navigation.consumer.service.ts
```diff
@@ -0,0 +1,110 @@
...
+      this.requestedUrl.next({ url: message.payload.url, channelId });
+      this.navigate(message.payload.url, channelId);
+    }
+  };
+
+  private readonly consumerManagerService = inject(ConsumerManagerService);
+
+  constructor() {
+    this.start();
+    inject(DestroyRef).onDestroy(() => this.stop());
```
matthieu-crouzet, line 72: We should decide a notation either `ngOnDestroy` either this one but not mixed both of them I think

packages/@ama-mfe/ng-utils/src/navigation/route-memorize/route-memorize.directive.ts
```diff
@@ -0,0 +1,76 @@
...
+    effect(() => {
+      const memorizeRoute = this.memorizeRoute();
+      const requested = requestedUrlSignal();
+      const id = this.memorizeRouteId() || this.connect();
+      if (memorizeRoute === false) {
+        return;
+      }
+      if (requested && id && requested.channelId === id) {
+        memory.memorizeRoute(id, requested.url, untracked(this.maxAge));
+      }
```
matthieu-crouzet, line 73:
```suggestion
      const memorizeRoute = this.memorizeRoute();
      if (memorizeRoute === false) {
        return;
      }
      const requested = requestedUrlSignal();
      const id = this.memorizeRouteId() || this.connect();
      if (requested && id && requested.channelId === id) {
        memory.memorizeRoute(id, requested.url, untracked(this.maxAge));
      }
```
So we avoid re-evaluate the effect while `memorizeRoute` is false

packages/@ama-mfe/ng-utils/README.md
mrednic-1A: https://github.com/AmadeusITGroup/otter/issues/2888

packages/@ama-mfe/ng-utils/src/messages/error/error.1.0.ts
```diff
@@ -0,0 +1,12 @@
+import type {
+  Message,
+} from '@amadeus-it-group/microfrontends';
+import type {
+  ErrorContent,
+} from './base';
+
+/** An error message with the version 1.0 */
+export interface ErrorMessageV10<S extends Message = Message> extends Message, ErrorContent<S> {
```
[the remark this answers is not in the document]
mrednic-1A: what about I let the name of the interface ErrorMessage in the file and I export it with an alias ?

packages/@ama-mfe/ng-utils/src/navigation/route-memorize/route-memorize.directive.ts
mrednic-1A: updated

packages/@ama-mfe/ng-utils/src/navigation/navigation.consumer.service.ts
mrednic-1A: In this package I kept the DestroyRef service in all cases.

packages/@ama-mfe/ng-utils/src/connect/connect.directive.ts
mrednic-1A: added the issue link everywhere

packages/@ama-mfe/messages/.compodocrc.json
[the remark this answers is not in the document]
mrednic-1A: so we keep compodoc for this one too

packages/@ama-mfe/messages/README.md
mrednic-1A: very good point.
this will be handled separately :)

matthieu-crouzet (dismissed):

packages/@ama-mfe/ng-utils/src/resize/resize.producer.service.ts
```diff
@@ -0,0 +1,79 @@
...
+  public startResizeObserver() {
+    this.resizeObserver = new ResizeObserver(() => {
+      if (document.body.scrollHeight !== this.actualHeight) {
+        this.actualHeight = document.body.scrollHeight;
+        const messageV10 = {
+          type: 'resize',
+          version: '1.0',
+          height: document.body.scrollHeight
+        } satisfies Resize;
+        // TODO: sendBest() is not implemented
```
fpaul-1A, line 72: do we want to open an issue to track this?

packages/@ama-mfe/ng-utils/src/theme/apply-theme.pipe.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+  });
+
+  it('should return undefined if url is undefined', () => {
+    expect(pipe.transform(undefined)).toBeUndefined();
+  });
+
+  it('should sanitize and return the url with theme if it is a string', () => {
+    const url = 'http://example.com';
+    const result = pipe.transform(url);
+    expect(sanitizer.sanitize).not.toHaveBeenCalled();
```
fpaul-1A, line 47: the title of the `it` doesn't seem to match the expectations

packages/@ama-mfe/ng-utils/src/theme/apply-theme.pipe.ts
```diff
@@ -0,0 +1,50 @@
...
+   * @returns The transformed SafeResourceUrl or undefined if the input URL is invalid.
+   */
+  public transform(url: string | SafeResourceUrl | undefined): SafeResourceUrl | undefined {
+    if (!url) {
+      return undefined;
+    }
+    const currentTheme = this.themeManagerService.currentTheme();
+    const urlString = typeof url === 'string'
+      ? url
+      : this.domSanitizer.sanitize(SecurityContext.RESOURCE_URL, url);
```
fpaul-1A, line 38: it looks weird to me that we sanitize the url only if it's a `SafeResourceUrl`, I would expect the opposite

packages/@ama-mfe/ng-utils/src/theme/theme.helpers.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/**
+ * Applies the given CSS theme as a stylesheet.
+ * @param themeValue - CSS as text containing a theme definition.
+ * @param cleanPrevious - Whether to remove previously applied stylesheets if no themeValue provided. Default is true.
+ */
+export function applyTheme(themeValue?: string, cleanPrevious = true): void {
+  if (themeValue !== undefined) {
+    const cssText = themeValue;
```
fpaul-1A, line 32: what is the point of introducing this variable?

packages/@ama-mfe/ng-utils/src/theme/theme.helpers.ts
```diff
@@ -0,0 +1,63 @@
...
+    cssHref = `${cssHref.endsWith('.css') ? cssHref : cssHref + THEME_URL_SUFFIX}`;
+    const themeRequest: Promise<void>[] = [];
+    themeRequest.push(getStyle(cssHref).then((styleToApply) => applyTheme(styleToApply, false)));
+    if (document.referrer) {
+      const url = new URL(document.referrer);
+      url.pathname += `${url.pathname.endsWith('/') ? '' : '/'}${cssHref}`;
+      themeRequest.push(getStyle(url.toString()).then((styleToApply) => applyTheme(styleToApply, false)));
+    }
+
+    await Promise.allSettled(themeRequest);
```
fpaul-1A, line 61: shouldn't the output promise of `applyInitialTheme` be rejected when one of the themeRequests fails?

packages/@ama-mfe/messages/src/messages/theme/index.ts
```diff
@@ -0,0 +1,9 @@
+import type {
+  Theme,
```
cpaulve-1A, line 2: Why not export them as Theme1_0 like you did with error message? 
What is the difference here?

packages/@ama-mfe/ng-utils/src/managers/interfaces.ts
```diff
@@ -0,0 +1,37 @@
...
+  /** The map of functions which will be executed for each supported version */
+  supportedVersions: {
+    [V in T['version']]: MessageCallback<T & { version: V }>;
+  };
+
+  start(): void;
+  stop(): void;
+}
+
+export interface MessageProducer<T extends Message = Message> {
```
kpanot, line 33: missing comments

packages/@ama-mfe/ng-utils/src/navigation/navigation.producer.service.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    inject(DestroyRef).onDestroy(() => {
+      producerManagerService.unregister(this);
+    });
+  }
+
+  /**
+   * Handles errors related to navigation messages.
+   * @param message - The error message content.
+   */
```
kpanot, line 74:
```suggestion
  /** @inheritDoc */
```

Would not do the job?

packages/@ama-mfe/ng-utils/src/navigation/navigation.producer.service.ts
```diff
@@ -0,0 +1,127 @@
...
+   * Handles embedded routing by listening to router events and sending navigation messages to the connected endpoints.
+   * It can be a parent window or another iframe
+   * @param options - Optional parameters to control the routing behavior {@see RoutingServiceOptions}.
+   */
+  public handleEmbeddedRouting(options?: RoutingServiceOptions): void {
+    const subRouteOnly = options?.subRouteOnly ?? false;
+    this.router.events.pipe(
+      this.takeUntilDestroyedOperator,
+      filter((event): event is NavigationStart => event instanceof NavigationStart),
+      switchMap((a) => {
```
kpanot, line 91:
```suggestion
      switchMap((event) => {
```

to keep a more meaningful variable

packages/@ama-mfe/ng-utils/src/navigation/navigation.producer.service.ts
```diff
@@ -0,0 +1,127 @@
...
+        );
+      })
+    ).subscribe(({ url, channelId }) => {
+      const messageV10 = {
+        type: 'navigation',
+        version: '1.0',
+        url
+      } satisfies Navigation;
+      // TODO: sendBest() is not implemented
+      if (document.referrer) {
```
kpanot, line 108: This logic can not be extracted to helpers as it seems standard and to be used in different producer.

packages/@ama-mfe/ng-utils/src/resize/resize.producer.service.ts
```diff
@@ -0,0 +1,79 @@
...
+ * When the height changes, it sends a resize message with the new height, to the connected peers
+ */
+@Injectable({
+  providedIn: 'root'
+})
+export class ResizeService implements MessageProducer<ResizeVersions> {
+  private actualHeight?: number;
+
+  private readonly messageService = inject(MessagePeerService<ResizeVersions>);
+  private resizeObserver?: ResizeObserver;
```
kpanot, line 33:
```suggestion
  private actualHeight?: number;
  private readonly messageService = inject(MessagePeerService<ResizeVersions>);
  private resizeObserver?: ResizeObserver;
```

packages/@ama-mfe/ng-utils/src/resize/resize.producer.service.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  /**
+   * The type of messages this service handles.
+   */
```
kpanot, line 37:
```suggestion
  /** @inheritDoc */
```

is not possible?

packages/@ama-mfe/ng-utils/src/theme/theme.helpers.ts
kpanot: Not necessary other you enforce the modules to provide CSS for all the themes.
The purpose here is to get the module theme override if they have one

packages/@ama-mfe/ng-utils/src/theme/theme.producer.service.ts
```diff
@@ -0,0 +1,134 @@
...
+        css: null
+      }
+      : undefined);
+    this.currentTheme = this.currentThemeSelection.asReadonly();
+
+    if (selectedThemeName) {
+      void this.changeTheme(selectedThemeName);
+    }
+
+    /** When the current theme changes, apply it to the current application */
```
kpanot, line 75:
```suggestion
    // When the current theme changes, apply it to the current application
```

packages/@ama-mfe/ng-utils/src/theme/theme.helpers.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Applies the given CSS theme as a stylesheet.
+ * @param themeValue - CSS as text containing a theme definition.
+ * @param cleanPrevious - Whether to remove previously applied stylesheets if no themeValue provided. Default is true.
+ */
+export function applyTheme(themeValue?: string, cleanPrevious = true): void {
+  if (themeValue !== undefined) {
+    const cssText = themeValue;
+    const sheet = new CSSStyleSheet();
+    sheet.replaceSync(cssText);
```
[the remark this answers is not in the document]
kpanot: Here the only 2 downloaded CSS files are the one hosted on the host application (the application containing the IFrame) and the one hosted on the module application (where the application on which the CSS should be applied).
I don't see additional risk relative to this method, did you think about something in particular I may not notice?

packages/@ama-mfe/ng-utils/README.md
mrednic-1A: yes, I have created a new issue to cover this too
https://github.com/AmadeusITGroup/otter/issues/2898

packages/@ama-mfe/messages/src/messages/theme/index.ts
mrednic-1A: just a mistake :)

packages/@ama-mfe/ng-utils/testing/babel.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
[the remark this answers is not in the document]
mrednic-1A: this is needed as the @amadeus-it-group/microfrontend does not expose commonjs, so I have to transform the esm in common js for the tests, and I do it with babel-jest

packages/@ama-mfe/ng-utils/package.json
```diff
@@ -0,0 +1,110 @@
+{
+  "name": "@ama-mfe/ng-utils",
+  "version": "0.0.0-placeholder",
+  "peerDependencies": {
+    "@ama-mfe/messages": "workspace:^",
+    "@amadeus-it-group/microfrontends": "^0.0.1",
+    "@amadeus-it-group/microfrontends-angular": "^0.0.1",
```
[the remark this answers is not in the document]
mrednic-1A: updated to support 0.x.x

packages/@ama-mfe/ng-utils/src/theme/apply-theme.pipe.ts
mrednic-1A: It is only to get the string value from the saferesource.

``
DomSanitizer.sanitize(context: SecurityContext, value: SafeValue | string | null): string | null
Gets a safe value from either a known safe value or a value with unknown safety.

If the given value is already a SafeValue, this method returns the unwrapped value. If the security context is HTML and the given value is a plain string, this method sanitizes the string, removing any potentially unsafe content. For any other security context, this method throws an error if provided with a plain string.
``

packages/@ama-mfe/ng-utils/src/theme/theme.helpers.ts
mrednic-1A: there was one at some point but not anymore. I removed it :)

packages/@ama-mfe/ng-utils/tsconfig.builders.json
```diff
@@ -0,0 +1,19 @@
+{
+  "extends": "../../../tsconfig.build",
+  "compilerOptions": {
+    "skipLibCheck": true,
```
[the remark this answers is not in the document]
mrednic-1A: removed. It was a copy paste most probably

packages/@ama-mfe/messages/.gitignore
```diff
@@ -0,0 +1,10 @@
+/fwk
+/helpers
+/index.*
+/bundles
+/test
+/dist-test
+/plugins
+/dist
+/utils
+/build
```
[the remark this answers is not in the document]
mrednic-1A: indeed

packages/@ama-mfe/messages/tsconfig.eslint.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "./tsconfig",
```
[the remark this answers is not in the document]
mrednic-1A: it is needed to lint the src files, no ?

packages/@ama-mfe/ng-utils/src/theme/apply-theme.pipe.ts
fpaul-1A: Understood for the SafeUrl treatment

However, I don't think we should return a `SafeResourceUrl` when the input was a `string` without sanitization
Maybe we can change the signature to be
```typescript
public transform<T extends string | SafeResourceUrl | undefined>(url: T): T {
```
and the `bypassSecurityTrustResourceUrl` should not be called from string input

packages/@ama-mfe/messages/src/messages/navigation/index.ts
```diff
@@ -0,0 +1,9 @@
+import type {
+  Navigation as NavigationV1_0,
+} from './navigation.1.0';
+
+export type { Navigation as NavigationV1_0 } from './navigation.1.0';
+export { MESSAGE_NAVIGATION_TYPE } from './base';
+
+/** The versions of navigation messages */
+export type NavigationVersions = NavigationV1_0;
```
matthieu-crouzet, line 9:
```suggestion
export type Navigation = NavigationV1_0;
```
or 
```suggestion
export type NavigationMessage = NavigationV1_0;
```

packages/@ama-mfe/messages/src/messages/theme/theme.1.0.ts
```diff
@@ -0,0 +1,21 @@
...
+export interface Theme extends Message {
+  /** The type of a theme message */
+  type: typeof MESSAGE_THEME_TYPE;
+  /** The version of this message */
+  version: '1.0';
+  /** The theme identification name */
+  name: string;
+  /** Theme definition: css variables plus specific style */
+  theme: string;
+}
```
matthieu-crouzet, line 21: Couldn't we have something like this ?
```suggestion
export interface Theme extends Message extends ThemeStructure {
  /** The type of a theme message */
  type: typeof MESSAGE_THEME_TYPE;
  /** The version of this message */
  version: '1.0';
}
```

packages/@ama-mfe/ng-utils/src/managers/interfaces.ts
```diff
@@ -0,0 +1,62 @@
...
+  [version: string]: MessageCallback<T>;
+}
+
+/** Description for the base message consumer */
+export interface BasicMessageConsumer {
+  /** Contains the message type */
+  type: string;
+  /** A map of messages versions and their callbacks */
+  supportedVersions: Record<string, MessageCallback<any>>;
+}
```
matthieu-crouzet, line 29: Shoudn't we have something like this ?
```suggestion
export interface BasicMessageConsumer<T extends Message = Message> {
  /** Contains the message type */
  type: T['type'];
  /** A map of messages versions and their callbacks */
  supportedVersions: Record<string, MessageCallback<T>>;
}
```

packages/@ama-mfe/ng-utils/src/navigation/navigation.producer.service.spec.ts
```diff
@@ -0,0 +1,134 @@
+/* eslint-disable no-undef -- document available in test env*/
```
matthieu-crouzet, line 1: You need to specify something like this instead 
https://github.com/AmadeusITGroup/otter/blob/main/eslint.shared.config.mjs#L96

packages/@ama-mfe/ng-utils/src/navigation/navigation.producer.service.ts
```diff
@@ -0,0 +1,130 @@
...
+})
+export class RoutingService implements MessageProducer<NavigationVersions> {
+  private readonly router = inject(Router);
+  private readonly activatedRoute = inject(ActivatedRoute);
+  private readonly messageService = inject(MessagePeerService<NavigationVersions>);
+
+  /**
+   * @inheritdoc
+   */
+  public readonly types = 'navigation';
```
matthieu-crouzet, line 62: Why not using `MESSAGE_NAVIGATION_TYPE` ?

packages/@ama-mfe/ng-utils/src/navigation/navigation.producer.service.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  private readonly takeUntilDestroyedOperator = takeUntilDestroyed();
+
+  constructor() {
+    const producerManagerService = inject(ProducerManagerService);
+    producerManagerService.register(this);
+
+    inject(DestroyRef).onDestroy(() => {
+      producerManagerService.unregister(this);
+    });
```
matthieu-crouzet, line 72: We may want to create an utilty function to reuse it in every producer
```
export const registerProducer = (producer: MessageProducer) => {
const producerManagerService = inject(ProducerManagerService);
    producerManagerService.register(producer);

    inject(DestroyRef).onDestroy(() => {
      producerManagerService.unregister(producer);
    });
}
```
and so it could be use like this 
```
constructor() {
  registerProducer(this);
}
```
so it will reduce the boilerplate in every producer

packages/@ama-mfe/ng-utils/src/resize/resize.producer.service.ts
```diff
@@ -0,0 +1,77 @@
...
+})
+export class ResizeService implements MessageProducer<ResizeVersions> {
+  private actualHeight?: number;
+  private readonly messageService = inject(MessagePeerService<ResizeVersions>);
+  private resizeObserver?: ResizeObserver;
+
+  /**
+   * @inheritdoc
+   */
+  public readonly types = 'resize';
```
matthieu-crouzet, line 37: We may use the exposed variable

packages/@ama-mfe/ng-utils/src/theme/apply-theme.pipe.ts
```diff
@@ -0,0 +1,50 @@
...
+import {
+  ThemeProducerService,
+} from './theme.producer.service';
+
+/**
+ * A pipe that applies the current theme from a theme manager service, to a given URL or SafeResourceUrl, as query param
+ */
+@Pipe({
+  name: 'applyTheme',
+  standalone: true
```
matthieu-crouzet, line 20: not sure you still have to specify it since we are using angular v19
```suggestion
```

packages/@ama-mfe/ng-utils/src/utils.ts
```diff
@@ -0,0 +1,28 @@
...
+  Message,
+  PeerConnectionOptions,
+} from '@amadeus-it-group/microfrontends';
+
+/**
+ * A constant array of known message types and their versions.
+ */
+export const KNOWN_MESSAGES = [
+  {
+    type: 'error',
```
matthieu-crouzet, line 11: isn't `error` exposed as a variable ?

packages/@ama-mfe/ng-utils/testing/babel.config.js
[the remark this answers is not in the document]
mrednic-1A: I think it is already taken into account by this issue 
https://github.com/AmadeusITGroup/microfrontends/issues/5

packages/@ama-mfe/messages/tsconfig.build.json
```diff
@@ -0,0 +1,27 @@
...
+    "incremental": true,
+    "composite": true,
+    "rootDir": ".",
+    "lib": [
+      "dom",
+      "dom.iterable",
+      "scripthost",
+      "es2017.object",
+      "esnext"
+    ],
```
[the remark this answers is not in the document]
mrednic-1A: no, removed :)

packages/@ama-mfe/messages/tsconfig.spec.json
```diff
@@ -0,0 +1,25 @@
...
+    "strictNullChecks": false,
+    "strictPropertyInitialization":false,
+    "noImplicitAny": false,
+    "incremental": true,
+    "composite": true,
+    "declarationMap": false,
+    "strict": false,
+    "module": "ES2020",
+    "target": "ES2020"
+  },
```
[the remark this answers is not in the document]
mrednic-1A: removed

packages/@ama-mfe/ng-utils/src/theme/theme.helpers.ts
kpanot: That's a bit different from the first comment but yes, it would make more sense to have a `return` instead of `await` (I did not notice it was not the case already).

packages/@ama-mfe/messages/tsconfig.eslint.json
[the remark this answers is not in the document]
kpanot: The purpose of this tsconfig is actually to lint the files that are not in the `src`.
You can add the extension to `tsconfig.base.json` (from the root folder) but I am not sure it is require neither :)

packages/@ama-mfe/ng-utils/testing/babel.config.js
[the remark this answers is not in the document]
kpanot: It is not exactly the description of the task linked (which is mainly to exposed minified and non-minified code).
We need explicitly ESM and CJS sources exported by the package.json

packages/@ama-mfe/ng-utils/src/theme/apply-theme.pipe.ts
mrednic-1A: this method has to return a SafeValue as the result is used as input to the src attribute of an iframe. 
I can;t change the signature of the input as angualr throws an error if I pass a string for the [src] input for an iframe. 
so we need to find a method to see that the string is a safe url

packages/@ama-mfe/ng-utils/src/theme/theme.helpers.ts
mrednic-1A: yes, it is better to return the value 
update done

packages/@ama-mfe/messages/package.json
```diff
@@ -0,0 +1,89 @@
...
+    "build:cjs": "swc src -d dist/cjs -C module.type=commonjs -q --strip-leading-paths",
+    "build:esm2020": "tsc -b tsconfig.build.json",
+    "postbuild": "yarn cpy './package.json' dist && patch-package-json-main",
+    "prepare:publish": "prepare-publish ./dist"
+  },
+  "dependencies": {
+    "tslib": "^2.6.2"
+  },
+  "peerDependencies": {
+    "@amadeus-it-group/microfrontends": "^0.0.2"
```
fpaul-1A, line 36:
```suggestion
    "@amadeus-it-group/microfrontends": "0.x.x"
```

packages/@ama-mfe/messages/tsconfig.eslint.json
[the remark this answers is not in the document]
mrednic-1A: indeed. there is no need for extension

packages/@ama-mfe/messages/src/messages/navigation/index.ts
mrednic-1A: I picked the second one as it is more explicit that it's for a message

packages/@ama-mfe/ng-utils/src/navigation/navigation.producer.service.ts
mrednic-1A: would DestroyRef be the good one as it is different and linked to the object in which is injected ?

packages/@ama-mfe/ng-utils/src/navigation/navigation.producer.service.ts
mrednic-1A: In fact this is a bit more complex story. 
document.referrer is not always reliable so it has to be replaced 

An issue has been opened in the scope https://github.com/AmadeusITGroup/otter/issues/2918

packages/@ama-mfe/messages/src/messages/navigation/navigation.versions.ts
```diff
@@ -0,0 +1,21 @@
...
+import type {
+  NAVIGATION_MESSAGE_TYPE,
+} from './base';
+
+/**
+ * The navigation message object sent via the communication protocol.
+ * Used in microfrontend architecture based on iframes where an iframe change its url and notify its parent with the url change
+ * The parent should update its own url with the information received in order to allow the refresh and the deep link share
+ */
+// eslint-disable-next-line @typescript-eslint/naming-convention -- version needs to be in interface name
```
matthieu-crouzet, line 13: We may customize @typescript-eslint/naming-convention to not complain on these files for pattern validating `V\d+_\d+$`
(Could done in another PR)

packages/@ama-mfe/messages/src/messages/theme/theme.versions.ts
```diff
@@ -0,0 +1,21 @@
...
+ * It contains the theme name and the theme deffinition
+ */
+// eslint-disable-next-line @typescript-eslint/naming-convention -- the version needs to be in interface name
+export interface ThemeV1_0 extends Message, ThemeStructure {
+  /** The type of a theme message */
+  type: typeof THEME_MESSAGE_TYPE;
+  /** The version of this message */
+  version: '1.0';
+  /** Theme definition: css variables plus specific style */
+  css: string;
```
matthieu-crouzet, line 20:
```suggestion
  css: NonNullable<ThemeStructure['css']>;
```

packages/@ama-mfe/ng-utils/src/navigation/navigation.producer.service.ts
matthieu-crouzet: Yes it will be the good one 
as it will be the only one available of the current injection context

matthieu-crouzet: It could be done in another PR

packages/@ama-mfe/messages/src/messages/theme/theme.versions.ts
matthieu-crouzet: Not mandatory as well

matthieu-crouzet (dismissed):

packages/@ama-mfe/ng-utils/src/theme/theme.helpers.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+      const themeValue2 = 'body { background-color: blue; }';
+      applyTheme(themeValue);
+      applyTheme(themeValue2, false);
+      expect(document.adoptedStyleSheets.length).toBe(2);
+      expect(document.adoptedStyleSheets[0].cssRules[0].cssText).toBe(themeValue);
+      expect(document.adoptedStyleSheets[1].cssRules[0].cssText).toBe(themeValue2);
+    });
+  });
+
+  describe('applyInitialTheme', () => {
```
kpanot, line 79: can you add a test with no `theme` query param?

packages/@ama-mfe/ng-utils/src/navigation/restore-route.pipe.ts
```diff
@@ -0,0 +1,95 @@
...
+  public transform(url: string, options?: Partial<RestoreRouteOptions>): string;
+  public transform(url: SafeResourceUrl, options?: Partial<RestoreRouteOptions>): SafeResourceUrl;
+  public transform(url: undefined, options?: Partial<RestoreRouteOptions>): undefined;
+  public transform(url: string | SafeResourceUrl | undefined, options?: Partial<RestoreRouteOptions>): string | SafeResourceUrl | undefined {
+    if (!url) {
+      return undefined;
+    }
+    const urlString = typeof url === 'string'
+      ? url
+      : this.domSanitizer.sanitize(SecurityContext.RESOURCE_URL, url);
```
kpanot, line 68:
```suggestion
    const urlString = typeof url === 'string'
      ? url
      : this.domSanitizer.sanitize(SecurityContext.RESOURCE_URL, url);
    if (!url) {
      return undefined;
    }
```

to avoid the falling `if` and final `return undefined`

packages/@ama-mfe/ng-utils/src/managers/consumer.manager.service.ts
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  /**
+   * Call the registered message callback(s) to consume the given message
+   * Handle error messages of internal communication protocol messages
+   * @param message message to consume
+   */
+  private async consumeAdditionalMessage(message: RoutedMessage<Message>) {
+    // not interested in service messages like 'connect' or 'disconnect'
+    if (isServiceMessage(message.payload)) {
```
kpanot, line 82: You may need to add a test to test invalid (or `undefined`) payload

packages/@ama-mfe/ng-utils/testing/babel.config.js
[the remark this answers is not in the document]
mrednic-1A: The `microfrontend` package will expose cjs in its 0.0.3 version

packages/@ama-mfe/ng-utils/src/navigation/navigation.producer.service.ts
mrednic-1A: Ref: 
https://github.com/AmadeusITGroup/otter/issues/2930

packages/@ama-mfe/ng-utils/src/theme/theme.helpers.ts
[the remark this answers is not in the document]
cpaulve-1A: I was trying to figure out if we could do a CSS injection from those css files (since we can send a postMessage with any content in the body to have it injected in the code source).
I have found a few CSS injection attack but the browsers seem to have patched them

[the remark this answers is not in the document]
mrednic-1A: I have added a sanitization step before applying the css

matthieu-crouzet (approved):

cpaulve-1A (approved):

packages/@ama-mfe/messages/tsconfig.eslint.json
```diff
@@ -0,0 +1,8 @@
+{
+  "include": [
+    "eslint*.config.mjs",
+    "jest.config.js",
+    "testing/*",
+    "scripts/*"
```
kpanot, line 6: not needed

packages/@ama-mfe/ng-utils/src/messages/error.sender.ts
```diff
@@ -0,0 +1,28 @@
...
+    version: '1.0',
+    ...content
+  } satisfies ErrorMessageV1_0);
+};
+
+/**
+ * Check if the given message is of type error and the error reson is present too
+ * @param message the message to be checked
+ */
+export const isErrorMessage = (message: any): message is Message & { type: typeof ERROR_MESSAGE_TYPE } & ErrorContent => (message && message.type === 'error' && !!message.reason);
```
kpanot, line 28:
```suggestion
export const isErrorMessage = (message: any): message is Message & { type: typeof ERROR_MESSAGE_TYPE } & ErrorContent => (message && typeof message === 'object' && message.type === 'error' && !!message.reason);
```

kpanot (approved):

matthieu-crouzet (approved):

PR #2807, opened 2025-02-06
feat(design): implementation for style-dictionary

kpanot:
## Proposed change

The purpose of this PR is to expose the `@o3r/design` features as [style-dictionary](https://styledictionary.com/) modules.

### Todo

- [x] Implement Metadata Format
- [x] Unit Tests
- [x] Integration Tests
- [x] Exposed code documentation
- [x] **Doc:** listing the modules
- [x] **Doc:** Get Started guide
- [x] **Doc:** Update Showcase Design Token Page
- [x] Angular builder options to switch implementation
- [x] `ng-add` schematics
- [x] Expose Otter [style transform-group](https://styledictionary.com/reference/hooks/transform-groups/)
- [x] Migration of Showcase

## Related issues

*- No issue associated -*

PR #2882, opened 2025-02-20
chore: migrate to Node LTS 22

kpanot:
## Proposed change

Migrate Node version to 22

## Related issues

* :rocket: Features
  * #2621
  * #2807

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -64,7 +64,7 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - uses: actions/setup-node@1d0ff469b7ec7b3cb9d8673fde0c81c44821de2a # v4.2.0
         with:
-          node-version: 20
+          node-version: 22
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -64,7 +64,7 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - uses: actions/setup-node@1d0ff469b7ec7b3cb9d8673fde0c81c44821de2a # v4.2.0
         with:
-          node-version: 20
+          node-version: 22
```
cpaulve-1A, line 67: Should not we ensure we still compatible with node 20 as it is still supported until next year?

vscaiceanu-1a (approved):

.github/workflows/main.yml
kpanot: Officially we are still supposed to be compatible with Node 16 but I am not sure that running an environment matrix with 3 versions of nodes would worse it :S

PR #2912, opened 2025-02-27
fix: o3r create with preset all and cms

### LINKED ISSUE #2907: [Bug]: o3r create fails with preset all
### Package name

create

### Package version

12.0.0-rc.0

### Reproduction steps

`npm create @o3r@~12.0.0-rc test-preset-all -- --preset=all`

Confirm all the prompts with default value.

For "Do you want to setup Playwright test framework for E2E?" you can choose yes or no, it doesn't make any difference.

### Current result

```shell
A merge conflicted on path "/jest.config.js".
npm ERR! code 3
```

### Expected result

Create command is successful. 

### Additional comments

The fix should be confirmed by a integration test in `o3r/create`.

vscaiceanu-1a:
## Related issues

* :bug: Fix resolves #2907

### DIFF
--- a/packages/@o3r/create/README.md
+++ b/packages/@o3r/create/README.md
@@ -56,3 +56,4 @@ On top of them, the following options can be provided to the initializer:
 - `--yarn`: Enforce `yarn` package manager. This option will be ignored if `--package-manager` is already specified.
 - `--yarn-version`: specify the version of yarn to use (default: `latest`)
 - `--exact-o3r-version`: use a pinned version for [Otter packages](https://github.com/AmadeusITGroup/otter/blob/main/docs/README.md).
+- `--preset <preset_name>`: the collection of Otter modules to install. If not provided it defaults to `basic`. The full list of presets is available in [@o3r/core](https://www.npmjs.com/package/@o3r/core#presets)
--- a/packages/@o3r/testing/schematics/ng-add/index.ts
+++ b/packages/@o3r/testing/schematics/ng-add/index.ts
@@ -147,33 +147,38 @@ function ngAddFn(options: NgAddSchematicsSchema): Rule {
         if (workingDirectory === undefined) {
           throw new O3rCliError(`Could not find working directory for project ${options.projectName || ''}`);
         } else {
-          const packageJsonFile = tree.readJson(`${workingDirectory}/package.json`) as PackageJson;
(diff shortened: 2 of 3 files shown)


packages/@o3r/create/README.md
```diff
@@ -56,3 +56,4 @@ On top of them, the following options can be provided to the initializer:
 - `--yarn`: Enforce `yarn` package manager. This option will be ignored if `--package-manager` is already specified.
 - `--yarn-version`: specify the version of yarn to use (default: `latest`)
 - `--exact-o3r-version`: use a pinned version for [Otter packages](https://github.com/AmadeusITGroup/otter/blob/main/docs/README.md).
+- `preset <preset_name>`: the collection of Otter modules to install. If not provided it defaults to `basic`. The full list of presets is available in [@o3r/core](https://www.npmjs.com/package/@o3r/core#presets)
```
fpaul-1A, line 59:
```suggestion
- `--preset <preset_name>`: the collection of Otter modules to install. If not provided it defaults to `basic`. The full list of presets is available in [@o3r/core](https://www.npmjs.com/package/@o3r/core#presets)
```
?

fpaul-1A (dismissed):

packages/@o3r/testing/schematics/ng-add/index.ts
```diff
@@ -169,7 +169,7 @@ function ngAddFn(options: NgAddSchematicsSchema): Rule {
             }),
             move(tree.root.path),
             renameTemplateFiles()
-          ]), MergeStrategy.Default);
+          ]), MergeStrategy.Overwrite);
```
matthieu-crouzet, line 172: Do we really want to override those files at the root folder if they exists ?

matthieu-crouzet: Here we are in `ng-add` not on the `create`

vscaiceanu-1a: Switched back to default to respect the following use case: you already have setup otter testing on your workspace but you create a new app/lib with otter testing. In this case the schematics should better not overwrite the files that were previously generated at the root.

fpaul-1A (approved):

mrednic-1A (approved):

PR #2913, opened 2025-02-27
feat: recommended preset for o3r core schematic

vscaiceanu-1a:
## Proposed change

Add a default `recommended` preset for `@o3r/core` ng add schematic.

- workspace schematic no longer brings eslint-config as it is brought by core for both project and workspace.
- it tests setup updated to use the basic preset in order to test the ng-add of eslint (and future packages part of recommended). We may consider having 2 IT setups (one for basic, one for recommended).
- workspace has flags to skip features (VSCode tools and renovate). For `basic` preset, the skip flags are passed by the core ng-add.


packages/@o3r/core/schematics/ng-add/index.ts
```diff
@@ -67,13 +67,17 @@ function ngAddFn(options: NgAddSchematicsSchema): Rule {
     };
 
     if (!options.projectName) {
+      const ngAddOptions = {
+        exactO3rVersion: options.exactO3rVersion,
+        ...(options.preset === 'basic' && { skipVscodeTools: true, skipRenovate: true })
```
fpaul-1A, line 72: can't this be set in the definition of the preset ? https://github.com/AmadeusITGroup/otter/blob/main/packages/%40o3r/core/schematics/shared/presets/basic.preset.ts

packages/@o3r/core/schematics/ng-add/utils/linter/index.spec.ts
```diff
@@ -2,38 +2,45 @@ import {
...
+    expect(isUsingLegacyConfig(tree)).toBe(false);
+  });
+
+  it('should return true for various legacy ESLint config file extensions', () => {
+    jest.mock(eslintPackageJson, () => ({}), { virtual: true });
+    const filenames = ['.eslintrc.json', '.eslintrc.js', '.eslintrc.yaml', '.eslintrc.yml'];
+    filenames.forEach((filename) => {
+      const tree = Tree.empty();
+      tree.create(filename, '');
+      expect(isUsingLegacyConfig(tree)).toBe(true);
```
fpaul-1A, line 43:
```suggestion
      expect(`${filename}:${isUsingLegacyConfig(tree)}`).toBe(`${filename}:true`);
```

vscaiceanu-1a: I think we will consider using a jest plugin for this kind of stuff. but in a different PR :)

packages/@o3r/core/README.md
```diff
@@ -104,7 +104,16 @@ Today the following presets are available.
...
+This preset does not add any additional modules.
+
+### Preset Recommended
+
+This preset is adding recommended Otter modules and features. This preset is the default one.
+
+* @o3r/eslint-config (or @o3r/eslint-config-otter for legacy configuration)
+* Renovate configuration
+* VSCode tooling
+
```
mrednic-1A, line 116: you can add a line for testing

fpaul-1A (approved):

packages/@o3r/core/README.md
```diff
@@ -104,7 +104,16 @@ Today the following presets are available.
...
 
-This preset is the default one, it does not add any additional modules.
+This preset does not add any additional modules.
+
+### Preset Recommended
+
+This preset is adding recommended Otter modules and features. This preset is the default one.
+
+* @o3r/eslint-config (or @o3r/eslint-config-otter for legacy configuration)
+* Renovate configuration
```
kpanot, line 115: maybe you can also add the links for the other packages

packages/@o3r/core/README.md
```diff
@@ -104,7 +104,16 @@ Today the following presets are available.
 
 ### Preset Basic
 
-This preset is the default one, it does not add any additional modules.
+This preset does not add any additional modules.
+
+### Preset Recommended
+
+This preset is adding recommended Otter modules and features. This preset is the default one.
```
kpanot, line 111: We do have only these packages that are recommended?
It may have been discussed already but why not adding: `logger`, `stylelint`, (`design-token`) which could make sense as well as recommendation? :S

packages/@o3r/core/README.md
vscaiceanu-1a: and for pre commit tools as well.

packages/@o3r/core/schematics/shared/presets/recommended.preset.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  const rule = defaultPresetRuleFactory(modules, options);
+
+  return {
+    modules,
+    rule: chain([
+      setupSchematicsParamsForProject({ '*:ng-add': { enableMetadataExtract: true } } as WorkspaceSchematics, options.projectName),
+      rule
+    ])
+  };
```
kpanot, line 33:
```suggestion
  return {
    modules,
    rule: chain([
      setupSchematicsParamsForProject({ '*:ng-add': { enableMetadataExtract: true } } as WorkspaceSchematics, options.projectName),
      defaultPresetRuleFactory(modules, options)
    ])
  };
```

packages/@o3r/core/README.md
vscaiceanu-1a: I think we should discuss and see if it makes sense to add more.
We have to make sure we are not heading into the wrong direction where many features were added by default.

fpaul-1A (approved):

mrednic-1A (dismissed):

packages/@o3r/core/schematics/shared/presets/recommended.preset.ts
```diff
@@ -0,0 +1,32 @@
...
+export function recommendedPreset(options: PresetOptions) {
+  const modules = [
+    options.isUsingEslintLegacy ? '@o3r/eslint-config-otter' : '@o3r/eslint-config',
+    '@o3r/testing'
+  ];
+
+  return {
+    modules,
+    rule: chain([
+      setupSchematicsParamsForProject({ '*:ng-add': { enableMetadataExtract: true } } as WorkspaceSchematics, options.projectName),
```
matthieu-crouzet, line 28: Shouldn't it done for the preset cms only ?

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #2945, opened 2025-03-05
fix: enhance response exception if response.text() fails

### LINKED ISSUE #2889: [Bug]: api client missing information in the emptyresponse error object
### Package name

core

### Package version

9.6

### Reproduction steps

When using the AngularApiFetch client to interract with the api, in the case when the call is rejected by a gateway  there is no spec to describe the error. So the AngularApiFetch client creates the object EmptyResponseError which not contain the information about the response (ex the status code).

### Current result

   EmptyResponseError object without data about the response.

### Expected result

Add the information from the response (at least the status code) when creating the EmptyResponseError object.

### Additional comments

Here is the part where the empty response error object is created.
https://github.com/AmadeusITGroup/otter/blob/a026fc1a70d600b3be5c296f23b50a4d8984510e/packages/%40ama-sdk/client-angular/src/api-angular-client.ts#L169

Do this implementation for the other clients not only the angular one.

cpaulve-1A:
## Proposed change

If a fetch response is available but the retrieval of its content fails, enhance the exception to provide status code.

## Related issues

* :bug: Fix #2889

### DIFF
--- a/packages/@ama-sdk/client-angular/src/api-angular-client.ts
+++ b/packages/@ama-sdk/client-angular/src/api-angular-client.ts
@@ -9,6 +9,7 @@ import type {
   TokenizedOptions,
 } from '@ama-sdk/core';
 import {
+  CanceledCallError,
   EmptyResponseError,
   ExceptionReply,
   extractQueryParams,
@@ -16,6 +17,7 @@ import {
   getResponseReviver,
   prepareUrl,
   processFormData,
+  RequestFailedError,
   ReviverReply,
   tokenizeRequestOptions,
 } from '@ama-sdk/core';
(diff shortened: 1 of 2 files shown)


matthieu-crouzet (dismissed):

mrednic-1A (approved):

mrednic-1A:
I think we should target a previous release, maybe the 10.4I guess it is the last supported)

packages/@ama-sdk/client-angular/src/api-angular-client.ts
```diff
@@ -166,7 +168,13 @@ export class ApiAngularClient implements ApiClient {
       response = await asyncResponse;
       root = response.body;
     } catch (e: any) {
-      exception = new EmptyResponseError(e.message || 'Fail to Fetch', undefined, { apiName, operationId, url, origin });
+      if (e instanceof CanceledCallError) {
+        exception = e;
+      } else if (response && !Number.isNaN(response.status)) {
+        exception = new RequestFailedError(e.message || 'Fail to Fetch', response.status, undefined, { apiName, operationId, url, origin });
```
fpaul-1A, line 174: not sure how big of a deal it is, but we are losing the stack of the original exception by creating a new one

fpaul-1A (dismissed):

packages/@ama-sdk/client-angular/src/api-angular-client.ts
cpaulve-1A: The original exception does not provide information on the status code.
I already pass the e.message, but the RequestFailedError does not allow to pass the stack itself.

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #2949, opened 2025-03-06
chore: remove node 18 from the supported versions

### LINKED ISSUE #2916: [Bug]: Otter v12 Node engine incompatibility
### Package name

- global -

### Package version

12.0.1

### Reproduction steps

- Install/enable Node.js 18.20.4
- Run ``npm create @o3r my-project``

### Current result

Dependencies installation fails due to engine Node version incompatibility:
```
error minimatch@10.0.1: The engine "node" is incompatible with this module. Expected version "20 || >=22". Got "18.20.4"
```

### Expected result

Package installation should work.

### Additional comments

Given its ``engines`` rule, Otter seems to still accept Node.js ``^18.19.1``, yet has dependencies which only accept Node.js 20+ (e.g. ``minimatch@10.0.1``, ``rimraf@6.0.1``, maybe others).

I think Otter should either:
- update its engines rule to make it clear it only supports Node.js 20+
- stop using dependencies requiring 20+

kpanot:
## Proposed change

chore: remove node 18 from the supported versions

## Related issues

* :rocket: Feature resolves #2916


apps/github-cascading-app/package.json
```diff
@@ -74,6 +74,6 @@
     "zone.js": "~0.15.0"
   },
   "engines": {
-    "node": "^18.19.1 || ^20.11.1 || >=22.0.0"
+    "node": "^20.11.1 || >=22.0.0"
```
fpaul-1A, line 77: this should probably target `release/12.0`

kpanot: cherry picked on #2956

sdo-1A (approved):

PR #2950, opened 2025-03-06
chore: upgrade @amadeus-it-group/microfrontends to v0.0.3

matthieu-crouzet:
## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@ama-mfe/messages/package.json
+++ b/packages/@ama-mfe/messages/package.json
@@ -33,7 +33,7 @@
     "tslib": "^2.6.2"
   },
   "peerDependencies": {
-    "@amadeus-it-group/microfrontends": "0.x.x"
+    "@amadeus-it-group/microfrontends": "0.0.3"
   },
   "peerDependenciesMeta": {
     "@amadeus-it-group/microfrontends": {
@@ -41,7 +41,7 @@
     }
   },
   "devDependencies": {
-    "@amadeus-it-group/microfrontends": "^0.0.2",
+    "@amadeus-it-group/microfrontends": "0.0.3",
     "@compodoc/compodoc": "^1.1.19",
     "@eslint-community/eslint-plugin-eslint-comments": "^4.4.0",
     "@nx/eslint-plugin": "~20.4.0",
--- a/packages/@ama-mfe/ng-utils/package.json
+++ b/packages/@ama-mfe/ng-utils/package.json
@@ -6,8 +6,8 @@
   },
   "peerDependencies": {
     "@ama-mfe/messages": "workspace:^",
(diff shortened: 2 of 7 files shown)


packages/@ama-mfe/messages/package.json
```diff
@@ -41,7 +41,7 @@
     }
   },
   "devDependencies": {
-    "@amadeus-it-group/microfrontends": "^0.0.2",
+    "@amadeus-it-group/microfrontends": "~0.0.3",
```
maxokorokov, line 44: Maybe better to stick with exact version `0.0.3` just in case, not even `~0.0.3`, as there might be breaking changes before `1.0.0`?

packages/@ama-mfe/ng-utils/package.json
```diff
@@ -6,8 +6,6 @@
   },
   "peerDependencies": {
     "@ama-mfe/messages": "workspace:^",
-    "@amadeus-it-group/microfrontends": "0.x.x",
```
mrednic-1A, line 9: The peer reps were added on purpose because if the user will add them we want to force to have the same version we have in the direct dependencies

packages/@ama-mfe/ng-utils/package.json
```diff
@@ -95,12 +95,6 @@
     "@ama-mfe/messages": {
       "optional": true
     },
-    "@amadeus-it-group/microfrontends": {
```
mrednic-1A, line 98: this should remain too

packages/@ama-mfe/ng-utils/package.json
```diff
@@ -6,8 +6,8 @@
   },
   "peerDependencies": {
     "@ama-mfe/messages": "workspace:^",
-    "@amadeus-it-group/microfrontends": "0.x.x",
-    "@amadeus-it-group/microfrontends-angular": "0.x.x",
```
mrednic-1A, line 10: we put 0.x.x to force a maximum 1 version, but if we could have breaking between 0.x.x then I think it's ok to force a fixed peer dep

packages/@ama-mfe/ng-utils/testing/jest.config.ut.js
```diff
@@ -14,9 +14,5 @@ module.exports = {
     '<rootDir>/builders/.*',
     '<rootDir>/schematics/.*'
   ],
-  transform: {
```
mrednic-1A, line 17: you can remove the babel.config file too, and the dependency to babel-jest

mrednic-1A (approved):

fpaul-1A (approved):

mrednic-1A (approved):

kpanot (approved):

PR #2955, opened 2025-03-06
fix: add warning log for unsupported config types

guillaumerochelle:
## Proposed change

Currently, any value (except arrays) in the default config is converted to text.
This can generate issues in the metadata when the value is not a number, string or boolean.
For example when a value is taken from another object.
```
const DEFAULT_VALUES = { foo: 'bar' };

interface MyConfig extends Configuration {
  myProperty: string;
};

const DEFAULT_CONFIG: MyConfig = {
  myProperty: DEFAULT_VALUES.foo // <-- extracted in the metadata as "DEFAULT_VALUES.foo", expected is "bar"
};
```

The proposed change is a warning logged when such cases are detected. It helps having metadata content which is CMS compliant.

## Related issues

*- No issue associated -*


kpanot (approved):

packages/@o3r/components/builders/component-extractor/helpers/component/component-config.extractor.ts
```diff
@@ -438,7 +438,11 @@ export class ComponentConfigExtractor {
                               property.type = typeReplacement;
                             }
                           } else {
-                            property.value = this.removeQuotationMarks(fieldNode.getText(this.source));
+                            if (ts.isStringLiteral(fieldNode) || ts.isNumericLiteral(fieldNode) || ts.isBooleanLiteral(fieldNode)) {
+                              property.value = this.removeQuotationMarks(fieldNode.getText(this.source));
+                            } else {
+                              this.logger.warn(`Unsupported type found will be ignored with kind = ${fieldNode.kind} and value = ${fieldNode.getText(this.source)}`);
```
matthieu-crouzet, line 444: Maybe we could add a unit-tests

matthieu-crouzet (approved):

cpaulve-1A (approved):

packages/@o3r/components/builders/component-extractor/helpers/component/component-config.extractor.ts
```diff
@@ -438,7 +438,11 @@ export class ComponentConfigExtractor {
                               property.type = typeReplacement;
                             }
                           } else {
-                            property.value = this.removeQuotationMarks(fieldNode.getText(this.source));
+                            if (ts.isStringLiteral(fieldNode) || ts.isNumericLiteral(fieldNode) || fieldNode.kind === ts.SyntaxKind.FalseKeyword || fieldNode.kind === ts.SyntaxKind.TrueKeyword) {
+                              property.value = this.removeQuotationMarks(fieldNode.getText(this.source));
+                            } else {
+                              this.logger.warn(`Unsupported type found will be ignored with kind = ${fieldNode.kind} and value = ${fieldNode.getText(this.source)}`);
```
kpanot, line 444: You can add this new case in UTs?

matthieu-crouzet: There are no UT for that extractor

kpanot: [packages/@o3r/components/builders/component-extractor/index.spec.ts](https://github.com/AmadeusITGroup/otter/blob/main/packages/%40o3r/components/builders/component-extractor/index.spec.ts) is covering 81% of the component extractor helpers.

mrednic-1A:
Please target the release/12.1.0-rc with this PR

packages/@o3r/components/builders/component-extractor/helpers/component/component-config.extractor.ts
guillaumerochelle: I added unit tests dedicated to the class ComponentConfigExtractor.

The other file you mentioned is using the showcase components. I don't want to add invalid configuration properties to the components for unit tests purposes.

fpaul-1A (approved):

kpanot (approved):

PR #2956, opened 2025-03-07
chore: remove node 18 from the supported versions

kpanot:
## Proposed change

Remove node 18 from the supported versions

## Related issues

*- No issue associated -*

* 🍒  pick of #2949
* :rocket: Feature #2916


package.json
```diff
@@ -128,8 +128,9 @@
   ],
   "resolutions": {
     "@sinonjs/fake-timers": "^14.0.0",
-    "@types/node": "^20.0.0",
+    "@types/node": "^20.11.1 || >=22.0.0",
```
fpaul-1A, line 131: Did you update this one on purpose?

fpaul-1A (approved):

package.json
kpanot: It was done by the version harmonize but I found the change relevante so I kept it

PR #2963, opened 2025-03-10
fix(palette-generator): add cache to palette generator application

kpanot:
## Proposed change

fix(palette-generator): add cache to palette generator application

## Related issues

*- No issue associated -*

### DIFF
--- a/apps/palette-generator/project.json
+++ b/apps/palette-generator/project.json
@@ -7,7 +7,9 @@
   "targets": {
     "build": {
       "executor": "nx:run-script",
-      "outputs": ["{projectRoot}/dist/package.json"],
+      "outputs": [
+        "{projectRoot}/dist/package.json"
+      ],
       "options": {
         "script": "postbuild"
       },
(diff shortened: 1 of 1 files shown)


apps/palette-generator/project.json
```diff
@@ -17,20 +19,34 @@
...
       "options": {
         "script": "build:ui"
       },
-      "outputs": ["{projectRoot}/dist/src"]
+      "inputs": [
+        "source",
+        "!{projectRoot}/src/plugin"
+      ],
+      "outputs": [
+        "{projectRoot}/dist/src"
```
fpaul-1A, line 48: I believe that if you change something only on `src/plugin`:
- the task `build-plugin` will run with no cache and update `dist/src/plugin`
- then `build-ui` will not detect changes in its inputs and restore its previous value of the `dist/src`
- we loose the changes done on `dist/src/plugin`

fpaul-1A: you might be right, I was not sure the `dependsOn` would be enough to invalidate the cache as none of the `inputs` have changed
can you double check if this is working as intended?

kpanot: Test done (with change only on plugin, only on ui and on both), it seems to work correctly in local.

matthieu-crouzet (approved):

PR #2966, opened 2025-03-10
fix: update theme doc for com protocol; remove unused deps for tests

mrednic-1A:
## Related issues

*- No issue associated -*


packages/@ama-mfe/ng-utils/src/theme/README.md
```diff
@@ -1,11 +1,21 @@
+# Initial theme load
```
kpanot, line 1: May be you can add a code example of its integration in a module?

kpanot (approved):

fpaul-1A (approved):

sdo-1A (approved):

packages/@ama-mfe/ng-utils/src/theme/README.md
mrednic-1A: the integration will be only the call to the helper in the main.ts file, but I could do it

PR #2968, opened 2025-03-10
fix(#2967): filtered out configuration without properties

### LINKED ISSUE #2967: [Bug]: Error when extracting component configuration for components with empty config
### Package name

configuration

### Package version

12.0.x

### Reproduction steps

- Configure the o3r CMS adapter for component configuration as explained in [CMS_ADAPTERS.md](https://github.com/AmadeusITGroup/otter/blob/main/docs/cms-adapters/CMS_ADAPTERS.md)
- Have an o3r component with empty configuration, for example:
```ts
import type {Configuration} from '@o3r/core';
import {computeItemIdentifier} from '@o3r/core';
// The config is empty
export interface MyCompConfig extends Configuration {}
export const MYCOMP_DEFAULT_CONFIG: MyCompConfig = {};
export const MYCOMP_CONFIG_ID = computeItemIdentifier('MyCompConfig', 'my-app');
```
- Run the components extractor: ``ng run my-app:extract-configurations``

### Current result

Extraction fails as below error occurs:
```
The output of configuration metadata is not valid regarding the json schema, please check the details below :
[] "does not meet minimum length of 1"
```

### Expected result

Component configuration should be extracted successfully

### Additional comments

The error occurs because the generated ``component.config.metadata.json`` contains a component configuration with empty ``properties``, which is against the ``"minItems": 1`` schema constraint introduced in [this PR](https://github.com/AmadeusITGroup/otter/pull/2641/commits/74f57c73a899cb15d6015fb7c0660e72c0f74d8a#diff-6cc05d1fb6a4373cf470cc1a90db8571e0500b07d8d6582342fe7d2a93499ac3)

matthieu-crouzet:
## Proposed change
Filtered out configuration without properties
## Related issues
* :bug: Fix resolves #2967


apps/showcase/src/app/configuration/configuration.config.ts
```diff
@@ -0,0 +1,15 @@
...
+import type {
+  Configuration,
+} from '@o3r/core';
+
+/**
+ * Empty configuration for testing purpose
+ */
+export interface ConfigurationConfig extends Configuration {}
+
+export const CONFIGURATION_DEFAULT_CONFIG: Readonly<ConfigurationConfig> = {} as const;
```
kpanot, line 13: missing comments to exported consts

kpanot (approved):

mrednic-1A (approved):

PR #2974, opened 2025-03-11
feat(ama-mfe-utils): integrate logger mechanism

### LINKED ISSUE #2887: [Feature]: Com protocol logger
### Context

Today the errors/warnings and info are dispatched via the browser console.

### Proposal

Implement a log mechanism using a logger or default on browser console.

- review the logs which are today dispatched via the console in order to not pollute the servers  (services which are exposing the same log for example - connection established should be logged once)
- the @amadeus-it-group/microfronted should have a mechanism to accept a logger object and to use it for logs (this is to be checked when implementing with the other project) ref: https://github.com/AmadeusITGroup/microfrontends/issues/4
- the logger in ama-mfe should be used to dispatch the logs at ama-mfe level and should be sent to the microfrontend package too (link with the step above)

kpanot:
## Proposed change

feat(ama-mfe-utils): integrate logger mechanism

## Related issues

* :rocket: Feature resolves #2887

### DIFF
--- a/packages/@ama-mfe/ng-utils/package.json
+++ b/packages/@ama-mfe/ng-utils/package.json
@@ -14,12 +14,14 @@
     "@angular/core": "^19.0.0",
     "@angular/platform-browser": "~19.2.0",
     "@angular/router": "~19.2.0",
+    "@o3r/logger": "workspace:^",
     "@o3r/schematics": "workspace:^",
     "rxjs": "^7.8.1"
   },
   "dependencies": {
     "@amadeus-it-group/microfrontends": "0.0.3",
     "@amadeus-it-group/microfrontends-angular": "0.0.3",
+    "@o3r/logger": "workspace:^",
     "tslib": "^2.6.2"
   },
   "sideEffects": false,
@@ -101,6 +103,9 @@
       "optional": true
     },
     "@angular-devkit/schematics": {
+      "optional": true
+    },
+    "@o3r/logger": {
       "optional": true
     },
     "@o3r/schematics": {
--- a/packages/@ama-mfe/ng-utils/src/connect/connect.directive.ts
+++ b/packages/@ama-mfe/ng-utils/src/connect/connect.directive.ts
@@ -16,6 +16,9 @@ import {
   DomSanitizer,
   SafeResourceUrl,
 } from '@angular/platform-browser';
+import {
+  LoggerService,
+} from '@o3r/logger';
 
 @Directive({
   selector: 'iframe[connect]',
@@ -51,6 +54,7 @@ export class ConnectDirective {
   });
 
   constructor() {
+    const logger = inject(LoggerService);
     // When the origin url or the peer id changes it will remake the connection with the new updates. The old connection is closed
     effect(async () => {
       const clientOrigin = this.clientOrigin();
@@ -65,9 +69,7 @@ export class ConnectDirective {
             origin: clientOrigin
           });
         } catch (e) {
-          // TODO https://github.com/AmadeusITGroup/otter/issues/2887 - proper logger
-          // eslint-disable-next-line no-console -- log the error - replace this with a proper logger
-          console.error(e);
+          logger.error(`Fail to connect to client (connection ID: ${connectId})`, e);
         }
       }
     });
--- a/packages/@ama-mfe/ng-utils/src/theme/theme.consumer.service.spec.ts
+++ b/packages/@ama-mfe/ng-utils/src/theme/theme.consumer.service.spec.ts
@@ -12,6 +12,9 @@ import {
   DomSanitizer,
 } from '@angular/platform-browser';
 import {
+  LoggerService,
+} from '@o3r/logger';
+import {
   ConsumerManagerService,
 } from '../managers/index';
 import {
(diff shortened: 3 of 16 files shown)


packages/@ama-mfe/ng-utils/package.json
```diff
@@ -14,12 +14,14 @@
...
     "@angular/platform-browser": "~19.2.0",
     "@angular/router": "~19.2.0",
+    "@o3r/logger": "workspace:^",
     "@o3r/schematics": "workspace:^",
     "rxjs": "^7.8.1"
   },
   "dependencies": {
     "@amadeus-it-group/microfrontends": "0.0.3",
     "@amadeus-it-group/microfrontends-angular": "0.0.3",
+    "@o3r/logger": "workspace:^",
```
matthieu-crouzet, line 24: why not in dep and peer instead of peer and devDep  ?

packages/@ama-mfe/ng-utils/src/connect/connect.directive.ts
```diff
@@ -51,6 +54,7 @@ export class ConnectDirective {
   });
 
   constructor() {
+    const logger = inject(LoggerService);
```
fpaul-1A, line 57: shouldn't this be optional?

packages/@ama-mfe/ng-utils/package.json
kpanot: Mainly because it can not be optional.
I did not want to enforce the application to provide `@o3r/logger` even if they do not use it on their side.

packages/@ama-mfe/ng-utils/src/connect/connect.directive.ts
kpanot: Not sure it bring anything :S.

If you are talking about:
```typescript
const logger = inject(LoggerService, { optional: true });
```
This will result to a type `LoggerService | undefined` but `LoggerService` is provided in root so always injected.

If you are talking about a dynImport with a `try catch`, it seems to bring too much complexity (especially in the constructor) for a simple service that do nothing by itself

packages/@ama-mfe/ng-utils/src/connect/connect.directive.ts
```diff
@@ -65,9 +69,7 @@ export class ConnectDirective {
             origin: clientOrigin
           });
         } catch (e) {
-          // TODO https://github.com/AmadeusITGroup/otter/issues/2887 - proper logger
-          // eslint-disable-next-line no-console -- log the error - replace this with a proper logger
-          console.error(e);
+          logger.error('Fail to connect to client', e);
```
mrednic-1A, line 72: clientId (connectId) could be added in the message

packages/@ama-mfe/ng-utils/src/theme/theme.helpers.ts
```diff
@@ -1,23 +1,34 @@
...
   try {
     const myRequest = new Request(cssPath);
     const response = await fetch(myRequest);
     const cssText = response.ok ? await response.text() : '';
     return cssText;
   } catch (error) {
-    // TODO https://github.com/AmadeusITGroup/otter/issues/2887 - proper logger
-    // eslint-disable-next-line no-console -- display the warning instead of failing silently
-    console.warn(`Failed to download style from: ${cssPath} with error: ${error?.toString()}`);
+    options?.logger?.warn(`Failed to download style from: ${cssPath} with error: ${error?.toString()}`);
```
matthieu-crouzet, line 31: shouldn't we fallback to console if none provided ?

packages/@ama-mfe/ng-utils/src/theme/theme.consumer.service.spec.ts
```diff
@@ -111,11 +121,10 @@ describe('ThemeConsumerService', () => {
       }
     };
     await themeHandlerService.supportedVersions['1.0'](themeMessage);
-    expect(downloadApplicationThemeCssSpy).toHaveBeenCalledWith('horizon');
+    expect(downloadApplicationThemeCssSpy).toHaveBeenCalledWith('horizon', expect.objectContaining({}));
     expect(applyThemeSpy).toHaveBeenCalledWith(themeMessage.payload.css);
     expect(applyThemeSpy).toHaveBeenCalledTimes(1);
-    // eslint-disable-next-line no-console -- checking that the console warn was called correctly
-    expect(console.warn).toHaveBeenCalledWith('no CSS variable for the theme horizon', 'no local css');
+    expect(loggerServiceMock.warn).toHaveBeenCalledWith(expect.stringMatching(/.+/), 'no local css');
```
fpaul-1A, line 127: any reason to no longer check the exact error message?

fpaul-1A (approved):

matthieu-crouzet (approved):

packages/@ama-mfe/ng-utils/src/theme/theme.consumer.service.spec.ts
fpaul-1A: it's just weird that you removed it for this one but you kept it for all the others, I have no opinion on the matter either

PR #2980, opened 2025-03-11
[draft] test: use dev drive on windows runners for it-tests

fpaul-1A:
## Proposed change

Desperate attempt to reduce the execution time of the it-tests on windows (especially with `npm` where the copy of `node_modules` takes forever, even when everything is cached)

## Related issues

*- No issue associated -*


fpaul-1A:
Not seeing any improvements in the execution of it-tests on windows.
It actually even seems worse somehow.

PR #2988, opened 2025-03-12
feat: exploded syntax in sdk - support dates

sdo-1A:
## Proposed change

Fix the newly created `PrimitiveType` to support dates (`utils.Date` and `utils.DateTime`)

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@ama-sdk/core/src/fwk/api.helpers.ts
+++ b/packages/@ama-sdk/core/src/fwk/api.helpers.ts
@@ -1,8 +1,9 @@
 import {
   TokenizedOptions,
 } from '../plugins/core/request-plugin';
-import type {
-  SupportedParamType,
+import {
+  isDateType,
+  type SupportedParamType,
 } from './param-serialization';
 import type {
   ReviverType,
@@ -76,7 +77,9 @@ export function stringifyQueryParams<T extends { [key: string]: SupportedParamTy
     .filter((name) => typeof queryParams[name] !== 'undefined' && queryParams[name] !== null)
     .reduce((acc, name) => {
       const prop = queryParams[name];
-      acc[name] = (typeof prop === 'object' && !Array.isArray(prop)) ? JSON.stringify(prop) : prop!.toString();
+      acc[name] = isDateType(prop)
+        ? prop.toJSON()
+        : ((typeof prop === 'object' && !Array.isArray(prop)) ? JSON.stringify(prop) : prop!.toString());
       return acc;
     }, {} as { [p in keyof T]: string });
 }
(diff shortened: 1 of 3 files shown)


mrednic-1A (approved):

packages/@ama-sdk/core/src/fwk/api.helpers.ts
```diff
@@ -76,7 +77,9 @@ export function stringifyQueryParams<T extends { [key: string]: SupportedParamTy
     .filter((name) => typeof queryParams[name] !== 'undefined' && queryParams[name] !== null)
     .reduce((acc, name) => {
       const prop = queryParams[name];
-      acc[name] = (typeof prop === 'object' && !Array.isArray(prop)) ? JSON.stringify(prop) : prop!.toString();
+      acc[name] = isDateType(prop)
+        ? prop.toJSON()
+        : ((typeof prop === 'object' && !Array.isArray(prop)) ? JSON.stringify(prop) : prop!.toString());
```
kpanot, line 82: Potentially `prop` can be `null` as `typeof null === 'object'`.
But as we never got the issue I guess it's fine :)

packages/@ama-sdk/core/src/fwk/param-serialization.ts
```diff
@@ -7,10 +11,18 @@ export interface ParamSerialization {
...
+export type PrimitiveType = string | number | boolean | utils.Date | utils.DateTime | undefined | null;
 /** Supported types for the operation parameters - primitives, primitive arrays, and simple non-nested objects */
 export type SupportedParamType = PrimitiveType | PrimitiveType[] | { [key: string]: PrimitiveType };
 
+/**
+ * Verify if property is of type utils.Date or utils.DateTime
+ * @param prop
+ */
+export function isDateType(prop: any): prop is utils.Date | utils.DateTime {
+  return typeof prop.toJSON === 'function';
```
kpanot, line 23: `toJSON` is also presents on native `Date`.
Would `return prop instanceof utils.Date || prop instanceof utils. DateTime` not do the job?

Otherwise, as it is mainly the function `toJSON` that is important for you, you can rename the function and provide `prop is {toJSON: () => string}`

fpaul-1A, line 23: isn't there a risk that this would be true for other objects not being Date type?
can't we use 
```typescript
return prop instanceof utils.Date || prop instanceof utils.DateTime;
```
?

packages/@ama-sdk/core/src/fwk/api.helpers.ts
sdo-1A: we shouldn't face this issue since we filtered for null values a couple of lines above, right?

kpanot: Indeed, I missed it

matthieu-crouzet (approved):

fpaul-1A (approved):

kpanot (approved):

PR #2989, opened 2025-03-12
fix(yarn-errors-reporter): prevent error when there are local changes

fpaul-1A:
## Proposed change

Renovate create pull request without using the `yarn postinstall` so we end up with local changes during the yarn errors report.

## Related issues

*- No issue associated -*


package.json
```diff
@@ -158,7 +158,7 @@
     "@angular/build": "~19.2.0",
     "@angular/cdk": "~19.2.0",
     "@angular/cli": "~19.2.0",
-    "@angular/common": "~19.2.0",
+    "@angular/common": "~19.0.0",
```
kpanot: why did you need this change?

tools/github-actions/yarn-errors-reporter/src/main.ts
```diff
@@ -50,9 +50,16 @@ async function run(): Promise<void> {
         }
       }
 
-      await exec('git', ['revert', '--no-commit', '-m', '1', 'HEAD']);
+      const hasLocalChanges = (await getExecOutput('git', ['status', '-s'], execOptions)).stdout.split('\n').length > 0;
```
kpanot, line 53: Does the output of `git` always `\n` for EOL or is it OS relative?

```suggestion
      const hasLocalChanges = (await getExecOutput('git', ['status', '-s'], execOptions)).stdout.split(EOL).length > 0;
```

fpaul-1A: it doesn't matter to count the number of lines

package.json
fpaul-1A: removed, this was for testing purposes

tools/github-actions/yarn-errors-reporter/src/main.ts
kpanot: good point

kpanot (approved):

matthieu-crouzet (approved):

PR #3001, opened 2025-03-14
chore: migrate jest config architecture to modular

kpanot:
## Proposed change

chore: migrate jest config architecture to modular

## Related issues

*- No issue associated -*


apps/palette-generator/package.json
```diff
@@ -25,6 +25,7 @@
     "@nx/jest": "~20.6.0",
     "@o3r/build-helpers": "workspace:^",
     "@o3r/eslint-plugin": "workspace:^",
+    "@o3r/test-helpers": "workspace:~",
```
fpaul-1A, line 28: I don't think it has any impact for a dev dep but why using `~` instead of `^`  as for the others?

kpanot: I think I followed an existing dep by copy pasting.
I will align them

packages/@o3r/test-helpers/src/jest-config/base.ts
```diff
@@ -0,0 +1,114 @@
...
+  return {
+    coverageReporters: ['cobertura'],
+    reporters: [
+      'default',
+      ['jest-junit', { outputDirectory: '<rootDir>/dist-test', outputName: 'junit.xml', classNameTemplate: '{filepath}' }],
+      'github-actions'
+    ],
+    testTimeout: 30_000,
+    testMatch: [
+      '<rootDir>/**/*.spec.ts'
```
fpaul-1A, line 108: do we really want that at root level when using `projects`?

packages/@o3r/test-helpers/testing/jest.config.ut.builders.js
fpaul-1A, line 1: this can be deleted I presume

packages/@ama-sdk/schematics/testing/jest.config.ut.builders.js
```diff
@@ -1,21 +1,22 @@
...
+  ...createDefaultPreset(getTsJestBaseConfig()),
+  ...getOtterJestBaseConfig(rootDir),
+  ...getJestUnitTestConfig({
+    testPathIgnorePatterns: [
+      '<rootDir>/src/.*'
+    ]
+  }),
+
   fakeTimers: {
-    enableGlobally: true,
```
fpaul-1A, line 13: why did you need to remove this?

packages/@o3r/extractors/testing/jest.config.ut.js
```diff
@@ -1,17 +1,19 @@
 const path = require('node:path');
-const getJestProjectConfig = require('../../../../jest.config.ut').getJestProjectConfig;
+const { getTsJestBaseConfig, getOtterJestBaseConfig, getJestUnitTestConfig } = require('@o3r/test-helpers');
+const { createCjsPreset } = require('jest-preset-angular/presets');
+// const { createDefaultPreset } = require('ts-jest');
```
fpaul-1A, line 4:
```suggestion
```

packages/@ama-sdk/schematics/testing/jest.config.ut.builders.js
kpanot: It was not needed, I actually aligned the different Jest Config without any fake timer in a first time and put it only to the ones requiring it.
It can be added back if you think it is important for this package

packages/@o3r/test-helpers/testing/jest.config.ut.builders.js
kpanot: indeed
I remove it

packages/@o3r/test-helpers/src/jest-config/base.ts
kpanot: Good point, probably not.

kpanot: Actually it is not supported to not have pattern:

image

If you have a better suggestion that the current one, I take it

fpaul-1A: I guess it's fine since we skip all the test just bellow but it looks more like a trick
what about `passWithNoTests: true`? this seems to be supported but not documented

kpanot: `passWithNoTests` produced other issue in the run of the VsCode plugin.
I finally made it work by ignoring `.*` for the projects which seems to please the plugin and the Jest CLI.

packages/@ama-sdk/schematics/testing/jest.config.ut.builders.js
kpanot: Remove both then

packages/@o3r/components/testing/jest.config.ut.builders.js
```diff
@@ -1,22 +1,21 @@
...
-  ]
+  ...createDefaultPreset(getTsJestBaseConfig()),
+  ...getOtterJestBaseConfig(rootDir),
+  ...getJestUnitTestConfig({
+    testPathIgnorePatterns: [
+      '<rootDir>/src/.*'
+    ],
+    fakeTimers: {
+      // This is needed to prevent timeout on builders tests
+      advanceTimers: true
```
fpaul-1A, line 17: this can be removed since we no longer have the fake timers enabled globally on every tests (same for core, localization, rules-engine, styling,... I am be missing some)

jest.config.js
```diff
@@ -1,19 +1,18 @@
+const { getJestProjectConfig } = require('@o3r/test-helpers');
 const { getJestProjects } = require('@o3r/workspace');
-const { getJestGlobalConfig } = require('./jest.config.ut');
+// const { createCjsPreset } = require('jest-preset-angular/presets');
```
fpaul-1A, line 3: this code is dead

kpanot: Long life to this code 👑 !

fpaul-1A (approved):

packages/@o3r/test-helpers/schematics/ng-add/index.ts
```diff
@@ -1,70 +0,0 @@
-import * as path from 'node:path';
```
cpaulve-1A, line 1: Why remove the ng-add for the @o3r/test-helpers?

packages/@o3r/eslint-plugin/schematics/ng-add/index.ts
```diff
@@ -1,14 +1,7 @@
...
-} from '@angular-devkit/schematics';
-
 /**
  * Add Otter eslint-plugin to an Angular Project
  */
-export function ngAdd(): Rule {
+export function ngAdd() {
   /* ng add rules */
-  return noop();
+  return () => {};
```
cpaulve-1A, line 6: Technically we should use it to import the " jsonc-eslint-parser" and "yaml-eslint-parser" if we consider this package as standalone

packages/@ama-sdk/schematics/testing/jest.config.ut.js
```diff
@@ -1,17 +1,18 @@
...
-    '<rootDir>/schematics/.*'
-  ]
+  ...createDefaultPreset(getTsJestBaseConfig()),
+  ...getOtterJestBaseConfig(rootDir),
+  ...getJestUnitTestConfig({
+    testPathIgnorePatterns: [
+      '<rootDir>/builders/.*',
+      '<rootDir>/schematics/.*'
+    ]
+  }),
```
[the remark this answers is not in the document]
kpanot: I prefered not for few reasons:
- It would force to do configuration to handle different cases (as it is done today)
- The purpose of the created helpers is to be modular to be picked and chose
- The currently there is no override to the output of the different helpers (and they are not conflicting together)
- The purpose was actually to select the different presets used at package level to remove unneeded dependencies (mainly angular) handle ESM packages

cpaulve-1A (approved):

PR #3002, opened 2025-03-14
ci: speed up execution of version harmonize

fpaul-1A:
## Proposed change

Eslint is very slow when using a pattern as input
Manually passing all the files make it much faster

## Related issues

*- No issue associated -*


matthieu-crouzet (approved):

package.json
```diff
@@ -31,7 +31,7 @@
     "update-yarn-sdks": "node -e \"'pnp' !== '$(yarn config get nodeLinker)' || process.exit(1)\" || yarn dlx @yarnpkg/sdks",
     "clear": "rimraf -g './{packages,tools,apps}/{@*/,}{amaterasu/,}*/{dist,build,dist-*}/'",
     "set:version": "yarn o3r-set-version --placeholder 0.0.0-placeholder --include '{apps,packages}/**/dist/{package,manifest}.json'",
-    "harmonize:version": "eslint '**/package.json*' '.yarnrc.yml' --quiet --fix",
+    "harmonize:version": "eslint --quiet --fix .yarnrc.yml $(glob \"**/package.json\" --ignore \"**/dist/**\")",
```
kpanot, line 34: do you really need `glob` here?
If you replace `'**/package.json'` by `"**/package.json"`, yarn was not doing the glob by itself?

fpaul-1A: if I let yarn handle it, it complains about the command line being too long
at least locally on windows

kpanot: So it does not scale :S.
I suppose that `glob` is working because the path is not absolute but one day an additional package.json will break it

fpaul-1A: I don't know if there is an easy way to shard the execution, I was initially considering piping the glob with `xargs -d\n -n 30` to limit to 30 files per instance of eslint execution but it wouldn't work on windows (at least not on default CLI)

cpaulve-1A, line 34: Should we update a schematic? Is there anyplace where we set this script up for our users?

cpaulve-1A (approved):

PR #3003, opened 2025-03-14
ci: fix eslint cache

fpaul-1A:
## Related issues

*- No issue associated -*

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -165,7 +165,7 @@
       },
       "configurations": {
         "ci": {
-          "command": "yarn eslint \"**/*.{m,c,}{j,t}s\" \"**/*.json\" --quiet --cache-location .cache/eslint"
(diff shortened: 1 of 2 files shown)


nx.json
```diff
@@ -165,7 +165,7 @@
       },
       "configurations": {
         "ci": {
-          "command": "yarn eslint \"**/*.{m,c,}{j,t}s\" \"**/*.json\" --quiet --cache-location .cache/eslint"
+          "command": "yarn eslint \"**/*.{m,c,}{j,t}s\" \"**/*.json\" --quiet --cache --cache-location .cache/eslint"
```
matthieu-crouzet, line 168: shouldn't we add `--cache --cache-location .cache/eslint` for the version harmonize at the root package.json too ?

matthieu-crouzet (dismissed):

nx.json
fpaul-1A: this can be done separately, here I am just fixing the cache used for the `lint` check

matthieu-crouzet (approved):

PR #3036, opened 2025-03-21
fix: training sdk exercise is broken

cpaulve-1A:
## Proposed change

bump showcase training version to use sdk-training new methods

## Related issues

*- No issue associated -*


apps/showcase/src/assets/trainings/sdk/shared/monorepo-template.json
fpaul-1A, line 1: any reason not to reformat the JSON? everything is on one line here
also, some files seems to be using CRLF (`\r\n`) we should probably put LF everywhere

cpaulve-1A: I will review the linebreak, I may need to review the script to serialize a folder and enforce the \n end of line.

kpanot (dismissed):

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #3044, opened 2025-03-24
fix: adjust path replacement script to work with node 20.19.0

guillaumerochelle:
## Proposed change

adjustPath method relies on the fact that node throws an error when requiring an ESM module. In the catch block, the code module is transformed to CJS.

This method works until Node 20.18.3.
With Node 20.19.0, [require(esm) is now enabled by default](https://nodejs.org/fr/blog/release/v20.19.0#requireesm-is-now-enabled-by-default) and therefore, no errors are thrown and ESM module are not transformed to CJS anymore.

This PR is a candidate fix for this issue.
Before requiring the module, the type will be inferred. If ESM is detected, the module will be converted to CJS.

## Related issues

Fixes 🐛 #3047

### DIFF
--- a/packages/@o3r/testing/src/tools/path-replacement/path-replacement.ts
+++ b/packages/@o3r/testing/src/tools/path-replacement/path-replacement.ts
@@ -10,6 +10,9 @@ import {
 import {
   requireFromString,
 } from 'module-from-string';
+
+type ModuleType = 'ESM' | 'CJS' | 'unknown';
+
 /**
  * Switch to the needed implementation of core testing, when running e2e tests
  * transforms ESM into CJS when needed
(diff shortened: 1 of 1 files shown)


packages/@o3r/testing/src/tools/path-replacement/path-replacement.ts
```diff
@@ -18,51 +21,111 @@ import {
...
+      type = 'CJS';
+
+      break;
+    }
+    case '.js': {
+    // Find recursively the package.json file to check if it is an ESM module
+      const packageJsonPath = findPackageJson(modulePath);
+      if (packageJsonPath !== null) {
+        // eslint-disable-next-line @typescript-eslint/no-require-imports -- Needed to require the package.json file
+        const packageJson = require(packageJsonPath);
```
fpaul-1A, line 106: why not using `fs.readFile`? the file was found using file system and not require

packages/@o3r/testing/src/tools/path-replacement/path-replacement.ts
```diff
@@ -18,51 +21,111 @@ import {
...
+    const filePath = require.resolve(newId, { paths });
 
+    const moduleType = checkModuleType(filePath, moduleTypesCache);
+    if (moduleType === 'ESM') {
+      return convertESMToCJS(newId, filePath, modulesCache, customTransformOptions);
+    } else {
+      try {
+        return Reflect.apply(originalRequire, this, [newId]);
+      } catch {
+        return convertESMToCJS(newId, filePath, modulesCache, customTransformOptions);
```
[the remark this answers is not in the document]
cpaulve-1A: If Reflect.apply fails on ESM files, this logic is the same as before. If not, why do we fallback to the conversion in the catch?

packages/@o3r/testing/src/tools/path-replacement/path-replacement.ts
```diff
@@ -18,51 +21,111 @@ import {
...
 export function adjustPath(frameworkName: 'playwright' | 'protractor', customTransformOptions: TransformOptions = {}) {
   const modulesCache: Record<string, any> = {};
+  const moduleTypesCache: Record<string, ModuleType> = {};
   // eslint-disable-next-line @typescript-eslint/unbound-method -- No need to bind the method, we are using `apply`
   const originalRequire = Module.prototype.require;
   const regex = new RegExp(`@o3r/testing/core(?!/${frameworkName})(.*)`);
   Module.prototype.require = function (this: NodeJS.Module, id: string) {
     const newId = id.replace(regex, `@o3r/testing/core/${frameworkName}$1`);
+    const paths = ([] as string[])
+      .concat(this.paths, this.paths.map((i) => i.replace(/[/\\]node_modules$/, '')));
```
kpanot, line 31:
```suggestion
    const paths = [
      ...this.paths,
      ...this.paths.map((i) => i.replace(/[/\\]node_modules$/, ''))
    ];
```

packages/@o3r/testing/src/tools/path-replacement/path-replacement.ts
```diff
@@ -18,51 +21,111 @@ import {
...
+    const trans = transformSync(fileContent.toString(), {
+      loader: 'js',
+      format: 'cjs',
+      target: 'es2016',
+      sourcesContent: true,
+      sourceRoot: cwd,
+      ...customTransformOptions
+    });
+
+    /**
```
kpanot, line 62:
```suggestion
    /*
```

packages/@o3r/testing/src/tools/path-replacement/path-replacement.ts
```diff
@@ -18,51 +21,111 @@ import {
...
+        return convertESMToCJS(newId, filePath, modulesCache, customTransformOptions);
+      }
+    }
+  } as NodeJS.Require;
+}
+
+function convertESMToCJS(newId: string, filePath: string, modulesCache: Record<string, any>, customTransformOptions: TransformOptions = {}): any {
+  if (!modulesCache[newId]) {
+    const fileContent = fs.readFileSync(filePath);
+    const cwd = path.resolve(process.cwd(), 'src');
```
kpanot, line 51: (not needed to be fixed in this PR as it is existing code)

Not sure about using `process.cwd()` in this context.
If it is launch from the project root path it would make the job but when launch from the workspace root it may cause issue.
Potentially it could be safer to determinate it from the filePath.

packages/@o3r/testing/src/tools/path-replacement/path-replacement.ts
```diff
@@ -18,51 +21,111 @@ import {
...
+    }
+  }
+
+  cache[modulePath] = type;
+  return type;
+}
+
+function findPackageJson(filePath: string): string | null {
+  const packageJsonPath = path.join(filePath, 'package.json');
+  if (fs.existsSync(packageJsonPath)) {
```
kpanot, line 123: You should prefer `(() => { try { require.resolve(packageJsonPath); return true } catch { return false; } })()` to support no node_module based package managers

packages/@o3r/testing/src/tools/path-replacement/path-replacement.ts
guillaumerochelle: Done

packages/@o3r/testing/src/tools/path-replacement/path-replacement.ts
guillaumerochelle: Done
Note: No node_modules based package managers were not supported before and still aren't.

packages/@o3r/testing/src/tools/path-replacement/path-replacement.ts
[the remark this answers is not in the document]
guillaumerochelle: I think the catch was redundant now that we detect ESM modules.
ESM are converted to CJS.
Unknown and CJS modules are required with Reflect.apply

fpaul-1A (dismissed):

packages/@o3r/testing/src/tools/path-replacement/path-replacement.ts
```diff
@@ -18,51 +21,119 @@ import {
...
+    default: {
+      type = 'unknown';
+    }
+  }
+
+  cache[modulePath] = type;
+  return type;
+}
+
+function findPackageJson(filePath: string): string | null {
```
kpanot, line 119: The name is a bit misleading, it is more a `dirPath` than a `filePath` :)

guillaumerochelle: Adjusted the parameter and added some jsdoc for good measure.

packages/@o3r/testing/src/tools/path-replacement/path-replacement.ts
guillaumerochelle: I won't fix it in this PR. It works for now.

matthieu-crouzet (approved):

PR #3049, opened 2025-03-25
chore: prevent issues when switching to node-modules linker

fpaul-1A:
## Related issues

*- No issue associated -*


matthieu-crouzet (dismissed):

packages/@o3r/components/tsconfig.build.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../../tsconfig.build",
   "compilerOptions": {
+    "types": [],
```
kpanot, line 4: Would not make sense to be at `tsconfig.base` level and request the different tsconfig to list the types they requires?

package.json
```diff
@@ -31,7 +31,7 @@
     "update-yarn-sdks": "node -e \"'pnp' !== '$(yarn config get nodeLinker)' || process.exit(1)\" || yarn dlx @yarnpkg/sdks",
     "clear": "rimraf -g './{packages,tools,apps}/{@*/,}{amaterasu/,}*/{dist,build,dist-*}/'",
     "set:version": "yarn o3r-set-version --placeholder 0.0.0-placeholder --include '{apps,packages}/**/dist/{package,manifest}.json'",
-    "harmonize:version": "eslint --quiet --fix .yarnrc.yml $(glob \"**/package.json\" --ignore \"**/dist/**\")",
```
kpanot, line 34: You can then maybe remove `glob` from the deps.

packages/@o3r/configuration/tsconfig.build.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../../tsconfig.build",
   "compilerOptions": {
+    "types": null,
```
kpanot, line 4: Why not listing the required types in that build?
(Same question for the other similar cases)

packages/@o3r/rules-engine/package.json
```diff
@@ -54,6 +54,7 @@
     "@schematics/angular": "^19.0.0",
     "globby": "^11.1.0",
     "jasmine": "^5.0.0",
+    "json-schema": "0.4.0",
```
kpanot, line 57: Why do you need to pin this dep version ?

kpanot (dismissed):

kpanot (approved):

PR #3051, opened 2025-03-25
refactor(ama-mfe-ng-utils): simplify resize directive

maxokorokov:
## Proposed change

This is for discussion first, I haven't modified tests yet.

Simplify resize directive and message consumer. Three changes:

1. Using `[style.height.px]` simplifies things greatly
2. `connectId` is handled at the consumer level, not at the directive level. Also I don't see the point in providing it in root as it is tightly coupled with `connectId` and directive instance.
3. I'm not sure I understand why there are both `[connect]` and `[scalable]` inputs? Can we just remove `scalable` input and use connect? I don't really see the point in specifying BOTH.

WDYT?

### DIFF
--- a/packages/@ama-mfe/ng-utils/src/resize/resize.consumer.service.ts
+++ b/packages/@ama-mfe/ng-utils/src/resize/resize.consumer.service.ts
@@ -22,21 +22,21 @@ import {
 /**
  * This service listens for resize messages and updates the height of elements based on the received messages.
  */
-@Injectable({
-  providedIn: 'root'
-})
+@Injectable()
 export class ResizeConsumerService implements MessageConsumer<ResizeMessage> {
-  private readonly newHeight = signal<{ height: number; channelId: string } | undefined>(undefined);
+  private readonly heightPxSignal = signal<number | undefined>(undefined);
 
   /**
    * A readonly signal that provides the new height information from the channel.
    */
-  public readonly newHeightFromChannel = this.newHeight.asReadonly();
+  public readonly heightPx = this.heightPxSignal.asReadonly();
 
   /**
    * The type of messages this service handles ('resize').
    */
   public readonly type = RESIZE_MESSAGE_TYPE;
+
+  public from = '';
 
   /**
    * The supported versions of resize messages and their handlers.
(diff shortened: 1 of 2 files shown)


matthieu-crouzet:
LGTM

packages/@ama-mfe/ng-utils/src/resize/resize.consumer.service.ts
```diff
@@ -22,22 +22,22 @@ import {
...
    */
-  public readonly newHeightFromChannel = this.newHeight.asReadonly();
+  public readonly heightPx = this.heightPxSignal.asReadonly();
 
   /**
    * The type of messages this service handles ('resize').
    */
   public readonly type = RESIZE_MESSAGE_TYPE;
 
+  public from = '';
```
kpanot, line 39: missing comments

mrednic-1A:
nice work :) it simplifies a lot the things. ok for points 1 and 2.
For 3 still ok, and as you changed the selector of the directive there is no need of scalable input :)

packages/@ama-mfe/ng-utils/src/resize/resize.consumer.service.ts
maxokorokov: We've decided to remove this notion of 'who sent resize message' completely for now and to introduce it later if necessary, but higher at the `ConsumerManagerService` level.

maxokorokov:
@mrednic-1A, @kpanot, @matthieu-crouzet please take another look.

We've decided with @mrednic-1A to simplify it even more and remove completely the notion of 'who was resize message sent from'. We might introduce it later if necessary at the higher `ConsumerManager` level.

packages/@ama-mfe/ng-utils/src/resize/resize.consumer.service.ts
```diff
@@ -46,7 +44,7 @@ export class ResizeConsumerService implements MessageConsumer<ResizeMessage> {
      * Use the message paylod to compute a new height and emit it via the public signal
      * @param message message to consume
      */
-    '1.0': (message: RoutedMessage<ResizeV1_0>) => this.newHeight.set({ height: message.payload.height, channelId: message.from })
+    '1.0': (message: RoutedMessage<ResizeV1_0>) => this.heightPxSignal.set(message.payload.height)
```
kpanot, line 47: If I understand well, with this change we will then limit the application to a single scalable iframe per app without strange behavior.
Not sure we want to bring this limitation at this stage to simplify the code.

maxokorokov: Hey, yes, of course one day we might need a mechanism to support multiple things (ex. resize and subroute navigation). It's just today's consumer/producer model doesn't support this, because they're not configurable and register via DI.

I'd say passing and `appId` as a part of a data model every time is not the best way to do it ;) ex. if we have other app-specific messages?

We need something like, even though it is not probably the best way to do it:

```ts
consumer.configure({
  from: ['appId']
})

// or make `.start()` configurable:
consumer.start({
  from: ['appId']
})
```

I'd also argue that it is not up to `ResizeDirectve` to configure the consumer, but to one who knows the `appId`. I can go further and argue that `ResizeDirective` shouldn't even exist just to do `[style.height]` binding, but it should be done by the iframe wrapper.

I can add configuration in a separate PR, or later if necessary, I just don't think resize directive should even know about 'appId'.

WDYT?

PR #3053, opened 2025-03-26
feat(ama-mfe): get the url of the host using query parameter

### LINKED ISSUE #2918: [Feature]: Communication protocol access to parent url
### Context

Currently on the communication protocol `document.referrer ` is used to identify if the application is embedded in an iframe or to get and use the parent url in the application embedded.

document.referrer is not always reliable because in case of a redirection to the application in cause, document.referrer will return the url before redirection, even for a top window.

### Proposal

Replace document.referrer with a mechanism which gives the possibility to get the parent url inside the application embedded. This url is needed in producers/consumers/connectivity used inside the embedded app.

For the simple usecase when it is needed to know if the app is embedded or not, an equality of window objects could be enough.

fpaul-1A:
## Proposed change

We consider 3 ways to get the origin of the host
- `document.referrer` works on all browsers but fails on internal redirection (login for example)
- `location.ancestorOrigins` supports redirections but doesn't work on Firefox
- use a query parameter works in all scenarios but require an additional action to modify the URL of the iframe

## Related issues

* :rocket: Feature resolves #2918

### DIFF
--- a/packages/@ama-mfe/ng-utils/src/utils.ts
+++ b/packages/@ama-mfe/ng-utils/src/utils.ts
@@ -17,13 +17,26 @@ export const KNOWN_MESSAGES = [
 ] as const satisfies Message[];
 
 /**
+ * Search parameter to add to the URL when embedding an iframe containing the URL of the host.
+ * This is needed to support Firefox (on which {@link https://developer.mozilla.org/docs/Web/API/Location/ancestorOrigins | location.ancestorOrigins} is not currently supported)
+ * in case of redirection inside the iframe.
+ */
+export const MFE_HOST_URL_PARAM = 'ama-mfe-host-url';
+
+/**
  * Returns the default options for starting a client endpoint peer connection.
- * As origin it will take the parent origin and the window will be the parent window
+ * As `origin`, it will take the parent origin and the `window` will be the parent window.
+ * The parent origin will use the first found among:
+ * - look for the search parameter {@link MFE_HOST_URL_PARAM} in the URL of the iframe
+ * - use the first item in `location.ancestorOrigins` (currently not supported on Firefox)
+ * - use `document.referrer` (will only work if called before any redirection in the iframe)
  */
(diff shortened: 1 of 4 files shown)


matthieu-crouzet (approved):

packages/@ama-mfe/ng-utils/src/utils.ts
```diff
@@ -16,14 +16,26 @@ export const KNOWN_MESSAGES = [
...
+ */
+export const MFE_HOST_URL_PARAM = 'ama-mfe-host-url';
+
 /**
  * Returns the default options for starting a client endpoint peer connection.
- * As origin it will take the parent origin and the window will be the parent window
+ * As `origin`, it will take the parent origin and the `window` will be the parent window.
+ * The parent origin will use the first found among:
+ * - look for the search parameter {@link MFE_HOST_URL_PARAM} in the URL of the iframe
+ * - use the first item in `location.ancestorOrigins` (not supported on Firefox)
```
kpanot, line 30:
```suggestion
 * - use the first item in `location.ancestorOrigins` (currently not supported on Firefox)
```

packages/@ama-mfe/ng-utils/src/utils.ts
```diff
@@ -16,14 +16,26 @@ export const KNOWN_MESSAGES = [
   }
 ] as const satisfies Message[];
 
+/**
+ * Search parameter to add to the URL when embedding an iframe containing the URL of the host.
+ * This is needed to support Firefox (on which `location.ancestorOrigins` is not supported) in case of redirection inside the iframe.
```
kpanot, line 21:
```suggestion
 * This is needed to support Firefox (on which {@link https://developer.mozilla.org/docs/Web/API/Location/ancestorOrigins | location.ancestorOrigins} is currently not supported) in case of redirection inside the iframe.
```

kpanot (approved):

kpanot (approved):

matthieu-crouzet (approved):

PR #3055, opened 2025-03-27
fix: allow to get api with a particular config

matthieu-crouzet:
## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@o3r/apis-manager/src/apis-manager/api-manager.ts
+++ b/packages/@o3r/apis-manager/src/apis-manager/api-manager.ts
@@ -1,6 +1,6 @@
 import type {
-  Api,
   ApiClient,
+  ApiName,
 } from '@ama-sdk/core';
 
 /**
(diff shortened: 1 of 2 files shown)


sdo-1A (approved):

fpaul-1A (approved):

packages/@o3r/apis-manager/src/apis-manager/api-manager.ts
```diff
@@ -35,7 +35,14 @@ export class ApiManager {
    * @param api API to get the configuration for
    */
   public getConfiguration(api?: string | Api): ApiClient {
-    return (api && this.apiConfigurations[typeof api === 'string' ? api : api.apiName]) || this.defaultConfiguration;
+    if (typeof api === 'string') {
+      if (this.apiConfigurations[api]) {
+        return this.apiConfigurations[api];
+      } else {
+        throw new Error(`Unknown API configuration: ${api}\nKnown API configurations: ${Object.keys(this.apiConfigurations).join(', ')}`);
```
kpanot, line 42: This is changing a bit the behavior.
Why not returning `this.defaultConfiguration` as it is today when a `ApiClass` does not have dedicated config?
(with potentially a warning logged)

fpaul-1A: I actually asked for this change, in my opinion if the `customApiName` is explicitly passed and does not exists on the configuration map, it's a real error and we should not fall back to the default config

mrednic-1A: could we add in the error message the fact that we could not find the config for a custom api name ?

fpaul-1A (approved):

kpanot (approved):

Issue #2532, opened 2024-11-29
[Feature]: Component highlight based on the HTML prefix name


AurelienCURTI:
### Context

In order to visualize where all the UI components are placed on a webpage, it would be nice to be able to have a way to visualize it globally by adding a highlight at component level. The purpose is to identify where components from different libraries are used into the page. For example:

 Those components will have an highlight purple:
-  my-bespoke-component-test-1
-  my-bespoke-component-test-2

 Those components will have a green highlight:
-  lib-component-test-1
-  lib-component-test-2

### Proposal

- [x] Highlight components based on a list of HTML tag prefix name.
- [ ] If a component is a presenter and a direct child of a container of the same component, only highlight the presenter.
- [x] Ignore components with no width (we can consider to highlight component with a min width and min height of 15px for example).
- [x] If two differents component are on the same screen position (overlay), we should be able to visually understand which one overlay the other.
- [x] Being able to modify the depth of the highlight to define how deep we want to highlight childs components.
- [x] Having the name of the component by hovering it with the mice and have an easy way to copy the name into the clipboard.
- [x] Make it more configurable as possible (providing highlight colors, desired thresholds, should we ignore container in case it contain a direct presenter, switch on/off the live refresh etc).

PR #3057, opened 2025-03-28
feat: group and highlight components by tag name

MrPa2a:
### Related issues
🚀 Feature https://github.com/AmadeusITGroup/otter/issues/2532

### Use `postMessage` to play with the configs (chrome extension part is not done yet)

- Set up groups (same message for the following configurations elementMinWidth, elementMinHeight, throttleInterval, maxDepth, chipsOpacity and autoRefresh)

```
window.postMessage({
  type: 'otter',
  to: 'app',
  content: {
    dataType: 'changeHighlightConfiguration',
    groupsInfo: {
      otter: {
        backgroundColor: '#f4dac6',
        color: 'black',
        regexp: '^o3r',
        displayName: 'o3r'
      },
      designFactory: {
        backgroundColor: '#000835',
        regexp: '^df',
        displayName: 'df'
      },
      ngBootstrap: {
        backgroundColor: '#0d6efd',
        regexp: '^ngb',
        displayName: 'ngb'
      }
    },
    maxDepth: 6,
    chipsOpacity: 0.8,
    autoRefresh: true
  }
});
```

- Activate highlight

```
window.postMessage({
  type: 'otter',
  to: 'app',
  content: {
    dataType: 'toggleHighlight',
    isRunning: true
  }
});
```
- You can also try to hover the chips and play with the opacity to have a better view on the highlighted elements

### DIFF
--- a/packages/@o3r/components/src/devkit/highlight/helpers.ts
+++ b/packages/@o3r/components/src/devkit/highlight/helpers.ts
@@ -1,0 +1,201 @@
+import {
+  HIGHLIGHT_CHIP_CLASS,
+  HIGHLIGHT_OVERLAY_CLASS,
+  HIGHLIGHT_WRAPPER_CLASS,
+} from './constants';
+import type {
+  ElementWithGroupInfo,
+  GroupInfo,
+} from './models';
+
+/**
+ * Retrieve the identifier of the element
+ * @param element
+ */
+export function getIdentifier(element: ElementWithGroupInfo): string {
+  const { tagName, attributes, classList } = element.htmlElement;
+  const regexp = new RegExp(element.regexp, 'i');
+  if (!regexp.test(tagName)) {
+    const attribute = Array.from(attributes).find((attr) => regexp.test(attr.name));
+    if (attribute) {
+      return `${attribute.name}${attribute.value ? `="${attribute.value}"` : ''}`;
+    }
+    const className = Array.from(classList).find((cName) => regexp.test(cName));
+    if (className) {
+      return className;
+    }
+  }
+  return tagName;
+}
+
+/**
+ * Filters a list of HTML elements and returns those that match specific group information.
+ *
+ * Each element is checked against a set of criteria:
+ * - The element's dimensions must meet the minimum height and width requirements.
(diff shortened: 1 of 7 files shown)


fpaul-1A (dismissed):

packages/@o3r/components/src/devkit/highlight/helpers.ts
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * Throttle {@link fn} with a {@link delay}
+ * @param fn method to run
+ * @param delay given in ms
+ */
+export function throttle<T extends (...args: any[]) => any>(fn: T, delay: number): (...args: Parameters<T>) => void {
+  let timerFlag: ReturnType<typeof setTimeout> | null = null;
+
+  const throttleFn = (...args: Parameters<T>) => {
+    if (timerFlag === null) {
```
cpaulve-1A, line 82: You always call the fn twice even if it has only been called once. Maybe we should add a flag to run the fn() after the timeout only if it has been called during the throttle (in an else block for instance)

matthieu-crouzet: you can remove L85 
it's from my code sorry :)

packages/@o3r/components/src/devkit/highlight/highlight.service.ts
```diff
@@ -0,0 +1,226 @@
...
+  runRefreshIfNeeded,
+  throttle,
+} from './helpers';
+import type {
+  ElementWithGroupInfo,
+  ElementWithGroupInfoAndDepth,
+  GroupInfo,
+} from './models';
+
+export class HighlightService {
```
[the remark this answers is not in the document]
matthieu-crouzet: At the beginning, it was done on the same model of `OtterInspectorService`.
But from what I can see we are using `inject` in this service so indeed it should be an injectable now.

You can add 
```typescript
@Injectable({
  providedIn: 'root'
})
```
and use this service like that `private readonly highlightService = inject(HighlightService)`

fpaul-1A (approved):

kpanot (approved):

PR #3064, opened 2025-03-31
fix(sdk-generator): to support the module resolution bundler

### LINKED ISSUE #3022: [Bug]: TS7016 error when using an o3r sdk in an app with `moduleResolution: bundler`
### Package name

core

### Package version

12.0.9

### Reproduction steps

This issue occurs with a new Angular generated application:

```sh
ng new app
cd app
npm install @scope/sdk
```

In `apps/cockpit/src/app/app.component.ts`, add the following code:
```ts
// ...
import { ScopeApi } from '@scope/sdk';

// ...
constructor() {
    const scopeApi = new ScopeApi(undefined);
}
```

### Current result

If we try to run `ng serve`, the following error is raised:
```sh
X [ERROR] TS7016: Could not find a declaration file for module '@dapi/sdk'. 'C:/git_clones/test/ng19new/node_modules/@scope/sdk/esm2020/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/scope__sdk` if it exists or add a new declaration (.d.ts) file containing `declare module '@scope/sdk';` [plugin angular-compiler]

    src/app/app.component.ts:3:29:
      3 │ import { ScopeApi } from '@scope/sdk';
        ╵                              ~~~~~~~~~~~
```

### Expected result

No error is raised.

### Additional comments

Note: this is due to the following field in `tsconfig.json`:
```json
"moduleResolution": "bundler",
```
Because switching to `"moduleResolution": "node"` fixes the issue.

However, `bundler` is the value that is defined by default when generating an app with `ng new`, so I am not sure switching to `node` is a proper solution.

This issue seems to be due to the way the sdk is packaged, which is why I am raising it in `@ama-sdk` package (I may by wrong though, as I am not familiar with this process).

kpanot:
## Proposed change

fix(sdk-generator): to support the module resolution bundler

Includes:

- [x] Update script
- [x] Update repository SDKs
- [x] Migrate showcase to bundler module resolution

## Related issues

* 🐛 Fix resolves #3022

### DIFF
--- a/packages/@ama-sdk/schematics/migration.json
+++ b/packages/@ama-sdk/schematics/migration.json
@@ -25,6 +25,11 @@
       "version": "11.4.0-alpha.0",
       "description": "Updates of @ama-sdk/schematics to v11.4.*",
       "factory": "./schematics/ng-update/index#updateV11_4"
+    },
+    "migration-v12_1_3": {
+      "version": "12.1.3",
+      "description": "Updates of @ama-sdk/schematics to v12.1.3",
+      "factory": "./schematics/ng-update/index#updateV12_1_3"
     }
   }
 }
--- a/apps/showcase/tsconfig.app.json
+++ b/apps/showcase/tsconfig.app.json
@@ -21,7 +21,7 @@
     "downlevelIteration": true,
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
     "resolveJsonModule": true,
     "importHelpers": true,
     "target": "ES2022",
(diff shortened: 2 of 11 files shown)


packages/@ama-sdk/schematics/migration.json
```diff
@@ -25,6 +25,11 @@
       "version": "11.4.0-alpha.0",
       "description": "Updates of @ama-sdk/schematics to v11.4.*",
       "factory": "./schematics/ng-update/index#updateV11_4"
+    },
+    "migration-v12_2": {
+      "version": "12.2.0-alpha.0",
```
fpaul-1A, line 30: why not `12.1` ?

fpaul-1A (approved):

cpaulve-1A (approved):

PR #3076, opened 2025-04-02
fix: sdk exploded syntax - take into account request plugin options

sdo-1A:
## Proposed change

When we previously implemented the exploded syntax for SDK parameter serialization, we removed the possibility of taking into account the query parameters that can be updated in the Request Plugins - which needs to be fixed. Here is a possible fix, which prepares the URL using the request plugins options (which could update the query parameters). 

Also, the parameter `queryParamSerialization` has been added to `RequestOptionsParameters`, which can be used in order to serialize the query parameter modifications that are potentially made. In another PR, I will add a method allowing to deserialize the parameters, which facilitates the modifications in the Request Plugins.

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@ama-sdk/core/src/fwk/param-serialization.ts
+++ b/packages/@ama-sdk/core/src/fwk/param-serialization.ts
@@ -15,12 +15,46 @@ export type PrimitiveType = string | number | boolean | Date | utils.Date | util
 /** Supported types for the operation parameters - primitives, primitive arrays, and simple non-nested objects */
 export type SupportedParamType = PrimitiveType | PrimitiveType[] | { [key: string]: PrimitiveType };
 
+/** URL encoding of space character, delimiter for spaceDelimited style */
+const SPACE_URL_CODE = encodeURIComponent(' ');
+/** URL encoding of pipe character, delimiter for pipeDelimited style */
+const PIPE_URL_CODE = encodeURIComponent('|');
+/** URL encoding of opening square bracket, used in deepObject style */
+const OPENING_SQUARE_BRACKET_URL_CODE = encodeURIComponent('[');
+/** URL encoding of closing square bracket, used in deepObject style */
+const CLOSING_SQUARE_BRACKET_URL_CODE = encodeURIComponent(']');
+
 /**
  * Verify if property is of type utils.Date or utils.DateTime
  * @param prop
  */
 export function isDateType(prop: any): prop is Date | utils.Date | utils.DateTime {
   return prop instanceof Date || prop instanceof utils.Date || prop instanceof utils.DateTime;
+}
+
+/**
+ * Check if the parameter is a record of type <string, string>.
+ * @param param
+ */
+export function isParamValueRecord(param: any): param is { [key: string]: string } {
+  return typeof param === 'object' && Object.values(param).every((item) => typeof item === 'string');
+}
+
+/** Query parameter value and serialization */
+export type QueryParamValueSerialization = { value: SupportedParamType } & ParamSerialization;
+
+/**
+ * Serialize query parameters of request plugins
+ * @param queryParams
+ */
+export function serializeRequestPluginQueryParams(queryParams: { [key: string]: QueryParamValueSerialization }) {
+  const queryParamsValues: { [key: string]: SupportedParamType } = {};
+  const queryParamSerialization: { [key: string]: ParamSerialization } = {};
+  Object.entries(queryParams).forEach(([paramKey, paramValue]) => {
+    queryParamsValues[paramKey] = paramValue.value;
+    queryParamSerialization[paramKey] = { explode: paramValue.explode, style: paramValue.style };
+  });
+  return serializeQueryParams(queryParamsValues, queryParamSerialization);
 }
 
 /**
@@ -46,13 +80,13 @@ function serializeArrayQueryParams(queryParamName: string, queryParamValue: Prim
         if (emptyArray) {
           break;
         }
-        return encodeURIComponent(queryParamName) + '=' + filteredArray.map((v) => isDateType(v) ? v.toJSON() : encodeURIComponent(v.toString())).join('%20');
+        return encodeURIComponent(queryParamName) + '=' + filteredArray.map((v) => isDateType(v) ? v.toJSON() : encodeURIComponent(v.toString())).join(SPACE_URL_CODE);
       }
       case 'pipeDelimited': {
         if (emptyArray) {
           break;
         }
-        return encodeURIComponent(queryParamName) + '=' + filteredArray.map((v) => isDateType(v) ? v.toJSON() : encodeURIComponent(v.toString())).join('%7C');
+        return encodeURIComponent(queryParamName) + '=' + filteredArray.map((v) => isDateType(v) ? v.toJSON() : encodeURIComponent(v.toString())).join(PIPE_URL_CODE);
       }
     }
   }
@@ -81,15 +115,16 @@ function serializeObjectQueryParams(queryParamName: string, queryParamValue: { [
         encodeURIComponent(propName) + ',' + (isDateType(propValue) ? propValue.toJSON() : encodeURIComponent(propValue.toString()))).join(',');
   } else if (paramSerialization.style === 'spaceDelimited' && !paramSerialization.explode && !emptyObject) {
     return encodeURIComponent(queryParamName) + '=' + Object.entries(filteredObject).map(([propName, propValue]) =>
-      encodeURIComponent(propName) + '%20' + (isDateType(propValue) ? propValue.toJSON() : encodeURIComponent(propValue.toString()))
-    ).join('%20');
+      encodeURIComponent(propName) + SPACE_URL_CODE + (isDateType(propValue) ? propValue.toJSON() : encodeURIComponent(propValue.toString()))
+    ).join(SPACE_URL_CODE);
   } else if (paramSerialization.style === 'pipeDelimited' && !paramSerialization.explode && !emptyObject) {
     return encodeURIComponent(queryParamName) + '=' + Object.entries(filteredObject).map(([propName, propValue]) =>
-      encodeURIComponent(propName) + '%7C' + (isDateType(propValue) ? propValue.toJSON() : encodeURIComponent(propValue.toString()))
-    ).join('%7C');
+      encodeURIComponent(propName) + PIPE_URL_CODE + (isDateType(propValue) ? propValue.toJSON() : encodeURIComponent(propValue.toString()))
+    ).join(PIPE_URL_CODE);
   } else if (paramSerialization.style === 'deepObject' && paramSerialization.explode && !emptyObject) {
     return Object.entries(filteredObject).map(([propName, propValue]) =>
-      encodeURIComponent(queryParamName) + '%5B' + encodeURIComponent(propName) + '%5D=' + (isDateType(propValue) ? propValue.toJSON() : encodeURIComponent(propValue.toString()))
+      encodeURIComponent(queryParamName) + OPENING_SQUARE_BRACKET_URL_CODE + encodeURIComponent(propName) + CLOSING_SQUARE_BRACKET_URL_CODE + '='
+      + (isDateType(propValue) ? propValue.toJSON() : encodeURIComponent(propValue.toString()))
     ).join('&');
   }
 }
@@ -110,9 +145,8 @@ export function serializeQueryParams<T extends { [key: string]: SupportedParamTy
       } else if (typeof queryParamValue === 'object' && !isDateType(queryParamValue)) {
         serializedValue = serializeObjectQueryParams(queryParamName, queryParamValue, paramSerialization);
       } else {
-        if (paramSerialization.style === 'form') {
-          serializedValue = encodeURIComponent(queryParamName) + '=' + (isDateType(queryParamValue) ? queryParamValue.toJSON() : encodeURIComponent(queryParamValue.toString()));
-        }
+        // NOTE: 'form' style is the default value for primitive types
+        serializedValue = encodeURIComponent(queryParamName) + '=' + (isDateType(queryParamValue) ? queryParamValue.toJSON() : encodeURIComponent(queryParamValue.toString()));
       }
       if (serializedValue) {
         acc[queryParamName as keyof T] = serializedValue;
--- a/packages/@ama-sdk/core/src/plugins/additional-params/additional-params-sync.request.ts
+++ b/packages/@ama-sdk/core/src/plugins/additional-params/additional-params-sync.request.ts
@@ -1,7 +1,16 @@
 import {
+  stringifyQueryParams,
+} from '../../fwk/api.helpers';
+import {
+  isParamValueRecord,
+  type QueryParamValueSerialization,
+  serializeRequestPluginQueryParams,
+} from '../../fwk/param-serialization';
+import {
   PluginSyncRunner,
   RequestOptions,
   RequestPlugin,
+  RequestPluginContext,
 } from '../core';
 import {
   isStringOrUndefined,
@@ -11,7 +20,8 @@ export interface AdditionalParametersSync {
   /** Additional headers */
   headers?: { [key: string]: string } | ((headers: Headers) => { [key: string]: string });
   /** Additional query params */
-  queryParams?: { [key: string]: string } | ((defaultValues?: { [key: string]: string }) => { [key: string]: string });
+  queryParams?: { [key: string]: string } | { [key: string]: QueryParamValueSerialization }
+    | ((defaultValues?: { [key: string]: string } | { [key: string]: QueryParamValueSerialization }) => { [key: string]: string } | { [key: string]: QueryParamValueSerialization });
   /** Additional body params */
   body?: (defaultValues?: string) => string | null;
 }
@@ -30,7 +40,7 @@ export class AdditionalParamsSyncRequest implements RequestPlugin {
     this.additionalParams = additionalParams;
   }
 
-  public load(): PluginSyncRunner<RequestOptions, RequestOptions> {
+  public load(context?: RequestPluginContext): PluginSyncRunner<RequestOptions, RequestOptions> {
     return {
       transform: (data: RequestOptions) => {
         const queryParams = typeof this.additionalParams.queryParams === 'function' ? this.additionalParams.queryParams(data.queryParams) : this.additionalParams.queryParams;
@@ -38,7 +48,21 @@ export class AdditionalParamsSyncRequest implements RequestPlugin {
         const body = this.additionalParams.body && isStringOrUndefined(data.body) ? this.additionalParams.body(data.body) : undefined;
 
         if (queryParams) {
-          data.queryParams = { ...data.queryParams, ...queryParams };
+          if (data.paramSerializationOptions?.enableParameterSerialization) {
+            if (isParamValueRecord(queryParams)) {
+              throw new Error('It is not possible to serialize additional query parameters without their serialization properties `value`, `explode`, and `style`.');
+            } else {
+              data.queryParams = { ...data.queryParams, ...serializeRequestPluginQueryParams(queryParams) };
+            }
+          } else {
+            if (isParamValueRecord(queryParams)) {
+              data.queryParams = { ...data.queryParams, ...queryParams };
+            } else {
+              const queryParamsValues = Object.fromEntries(Object.entries(queryParams).map(([key, value]) => [key, value.value]));
+              data.queryParams = { ...data.queryParams, ...stringifyQueryParams(queryParamsValues) };
+              (context?.logger || console).log('The serialization of additional query parameters has been ignored since parameter serialization is not enabled.');
+            }
+          }
         }
 
         if (body !== undefined) {
--- a/packages/@ama-sdk/schematics/README.md
+++ b/packages/@ama-sdk/schematics/README.md
@@ -119,9 +119,27 @@ For more information, check out OpenAPI's documentation on [parameter serializat
 It is important to note that, as in OpenAPI 3.1, we only support simple arrays and simple non-nested objects in path and query parameters. 
 The parameter types that we support are stored in `SupportedParamType` in the package `@ama-sdk/core`.
 
+To enable the parameter serialization within your API, you can set the option `enableParameterSerialization` to `true` (its current default value is `false`) in the constructor. For example:
+```typescript
+const apiConfig: ApiClient = new ApiFetchClient(
+  {
+    basePath: 'https://petstore3.swagger.io/api/v3',
+    enableParameterSerialization: true
+  }
+);
+```
+
+We provide the methods `serializeQueryParams` and `serializePathParams` to serialize the values of query and path parameters. However, it is also possible to pass
+your own serialization methods if the ones provided do not meet your requirements. These custom methods can be passed as a parameter to the API client constructor.
+
 > [!NOTE]
-> We provide the methods `serializeQueryParams` and `serializePathParams` to serialize the values of query and path parameters. However, it is also possible to pass
-> your own serialization methods if the ones provided do not meet your requirements. These custom methods can be passed as a parameter to the API client constructor.
+> If you have enabled the serialization mechanism and want to update the query parameters within a `RequestPlugin`, these must be serialized before being returned to the
+> API to prepare the URL. You can do so by using the serialization method that we provide (`serializeQueryParams`) or your own serialization method. The value of the query
+> parameters returned by the `RequestPlugin` will be forwarded to the next plugin and the last value will be directly added to the URL.
+
+> [!NOTE]
+> It is important to note that special characters have to be encoded, as required by RFC6570 and RFC3986. Please take this into account if you choose to use your own
+> serialization methods.
 
 #### Light SDK
 
--- a/packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/java/com/amadeus/codegen/ts/LambdaHelper.java
+++ b/packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/java/com/amadeus/codegen/ts/LambdaHelper.java
@@ -59,11 +59,28 @@ public class LambdaHelper {
    * Replaces placeholders in the URL such as
    * /carts/{cartId}/travelers
    * to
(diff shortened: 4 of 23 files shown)


packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/resources/typescriptFetch/api/api.mustache
```diff
@@ -112,21 +117,16 @@ export class {{classname}} implements Api {
       headers,
       method: '{{httpMethod}}',
       basePath,
-      queryParams,
+      queryParams{{#hasQueryParams}}: serializedQueryParams,
```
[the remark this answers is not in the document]
sdo-1A: because the variable `serializedQueryParams` is only created if there are query params (it is also surrounded by `hasQueryParams`)

cpaulve-1A (approved):

packages/@ama-sdk/schematics/README.md
```diff
@@ -123,6 +123,11 @@ The parameter types that we support are stored in `SupportedParamType` in the pa
 > We provide the methods `serializeQueryParams` and `serializePathParams` to serialize the values of query and path parameters. However, it is also possible to pass
 > your own serialization methods if the ones provided do not meet your requirements. These custom methods can be passed as a parameter to the API client constructor.
 
+> [!NOTE]
+> If you update the query parameters within a `RequestPlugin`, these must be serialized before being returned to the API to prepare the URL. You can do so by using the
```
kpanot, line 127: I would then expect to have this serialization in our own plugin (such as `AdditionalParamsRequest`).
Did I misunderstand something?

packages/@ama-sdk/schematics/README.md
```diff
@@ -123,6 +123,11 @@ The parameter types that we support are stored in `SupportedParamType` in the pa
 > We provide the methods `serializeQueryParams` and `serializePathParams` to serialize the values of query and path parameters. However, it is also possible to pass
 > your own serialization methods if the ones provided do not meet your requirements. These custom methods can be passed as a parameter to the API client constructor.
 
+> [!NOTE]
+> If you update the query parameters within a `RequestPlugin`, these must be serialized before being returned to the API to prepare the URL. You can do so by using the 
+> serialization method that we provide (`serializeQueryParams`) or your own serialization method. The value of the query parameters returned by the `RequestPlugin` will
+> be directly added to the URL.
```
fpaul-1A, line 129:
```suggestion
> serialization method that we provide (`serializeQueryParams`) or your own serialization method. The value of the query parameters returned by a `RequestPlugin` will
> be forwarded to the next plugin and the last value will be directly added to the URL.
```

fpaul-1A (approved):

packages/@ama-sdk/core/src/fwk/param-serialization.ts
```diff
@@ -15,6 +15,18 @@ export type PrimitiveType = string | number | boolean | Date | utils.Date | util
...
+export type ParamTypeForDeserialization = 'primitive' | 'array' | 'object';
+
+/** URL encoding of space character, delimiter for spaceDelimited style */
+const SPACE_URL_CODE = '%20';
+/** URL encoding of pipe character, delimiter for pipeDelimited style */
+const PIPE_URL_CODE = '%7C';
+/** URL encoding of opening square bracket, used in deepObject style */
+const OPENING_SQUARE_BRACKET_URL_CODE = '%5B';
+/** URL encoding of closing square bracket, used in deepObject style */
+const CLOSING_SQUARE_BRACKET_URL_CODE = '%5D';
```
fpaul-1A, line 28:
```suggestion
/** URL encoding of space character, delimiter for spaceDelimited style */
const SPACE_URL_CODE = encodeURIComponent(' ');
/** URL encoding of pipe character, delimiter for pipeDelimited style */
const PIPE_URL_CODE = encodeURIComponent('|');
/** URL encoding of opening square bracket, used in deepObject style */
const OPENING_SQUARE_BRACKET_URL_CODE = encodeURIComponent('[');
/** URL encoding of closing square bracket, used in deepObject style */
const CLOSING_SQUARE_BRACKET_URL_CODE = encodeURIComponent(']');
```

packages/@ama-sdk/core/src/fwk/param-serialization.ts
```diff
@@ -23,6 +35,37 @@ export function isDateType(prop: any): prop is Date | utils.Date | utils.DateTim
...
+/** Query parameter value and serialization */
+export type QueryParamValueSerialization = { value: SupportedParamType } & ParamSerialization;
+
+/**
+ * Serialize query parameters of request plugins
+ * @param queryParams
+ */
+export function serializeRequestPluginQueryParams(queryParams: { [key: string]: string } | { [key: string]: QueryParamValueSerialization }) {
+  if (isRecordTypeString(queryParams)) {
+    return Object.entries(queryParams).reduce((acc, [paramKey, paramValue]) => {
```
cpaulve-1A, line 55: Why handle it as a special use case ?

packages/@ama-sdk/core/src/fwk/param-serialization.ts
```diff
@@ -246,3 +289,172 @@ export function serializePathParams<T extends { [key: string]: SupportedParamTyp
...
+ * OpenAPI Parameter Serialization {@link https://swagger.io/specification | documentation}
+ * @param serializedParamValue serialized query parameter value
+ * @param paramSerialization parameter serialization
+ */
+export function deserializeObjectQueryParams(serializedParamValue: string, paramSerialization: ParamSerialization) {
+  if (paramSerialization.explode && (paramSerialization.style === 'form' || paramSerialization.style === 'deepObject')) {
+    return serializedParamValue.split('&').reduce((obj: { [key: string]: PrimitiveType }, currentValue) => {
+      const [key, value] = currentValue.split('=');
+      // NOTE: The key of an object in deepObject style is surrounded by opening and closing square brackets
+      obj[paramSerialization.style === 'deepObject' ? key.split(OPENING_SQUARE_BRACKET_URL_CODE)[1].split(CLOSING_SQUARE_BRACKET_URL_CODE)[0] : key] = value;
```
cpaulve-1A, line 326: We should probably consider also if the character is not encoded or at least document it

cpaulve-1A:
```suggestion
      obj[paramSerialization.style === 'deepObject' ? key.match(`^${OPENING_SQUARE_BRACKET_URL_CODE})([^${CLOSING_SQUARE_BRACKET_URL_CODE}]+)`)[1]` : key] = value;
```

packages/@ama-sdk/core/src/fwk/param-serialization.ts
```diff
@@ -246,3 +289,172 @@ export function serializePathParams<T extends { [key: string]: SupportedParamTyp
...
+  }
+  if (paramSerialization.style !== 'deepObject' && !paramSerialization.explode) {
+    // NOTE: The deserialized value will exist since these exploded styles are supported for arrays
+    return deserializeArrayQueryParams(serializedParamValue, paramSerialization)!.reduce((obj: { [key: string]: PrimitiveType }, currentValue, index, array) => {
+      if (index % 2 === 0) {
+        obj[currentValue] = array[index + 1];
+      }
+      return obj;
+    }, {} as { [key: string]: PrimitiveType });
+  }
```
cpaulve-1A, line 338: You should log something if the combination exploded/style is not supported

packages/@ama-sdk/core/src/fwk/param-serialization.ts
sdo-1A: According to the OpenAPI documentation, special characters should be encoded since this is required by RFC6570 and RFC3986

packages/@ama-sdk/core/src/fwk/param-serialization.ts
```diff
@@ -246,3 +289,172 @@ export function serializePathParams<T extends { [key: string]: SupportedParamTyp
...
+    let deserializedValue: SupportedParamType;
+    if (paramSerialization.paramType === 'array') {
+      deserializedValue = deserializeArrayQueryParams(serializedParamValue, paramSerialization);
+    } else if (paramSerialization.paramType === 'object') {
+      deserializedValue = deserializeObjectQueryParams(serializedParamValue, paramSerialization);
+    } else {
+      deserializedValue = serializedParamValue.split('=')[1];
+    }
+    if (deserializedValue) {
+      acc[queryParamName as keyof T] = deserializedValue;
```
cpaulve-1A, line 362: why do you need  to cast it?

packages/@ama-sdk/core/src/fwk/param-serialization.spec.ts
```diff
@@ -215,3 +217,62 @@ describe('Serialize parameters', () => {
     expect(serializePathParams(mockUtilsDateTimeParams, { idUtilsDateTime: { explode: true, style: 'simple' } })).toEqual({ idUtilsDateTime: '2025-01-01T00:00:00.000' });
   });
 });
+
+describe('Deserialize parameters', () => {
+  it('should correctly deserialize query parameters', () => {
+    // value = primitive
+    expect(deserializeQueryParams({ idPrimitive: 'idPrimitive=5' }, { idPrimitive: { explode: true, style: 'form', paramType: 'primitive' } })).toEqual({ idPrimitive: '5' });
```
cpaulve-1A, line 224: We could split them to identify more easily what broken. Here all of them will fail with the same description

packages/@ama-sdk/core/src/fwk/param-serialization.ts
```diff
@@ -246,3 +289,172 @@ export function serializePathParams<T extends { [key: string]: SupportedParamTyp
...
+      obj[key] = value;
+      return obj;
+    }, {} as { [key: string]: PrimitiveType });
+  }
+  return splitObject.reduce((obj: { [key: string]: PrimitiveType }, currentValue, index, array) => {
+    if (index % 2 === 0) {
+      obj[currentValue] = array[index + 1];
+    }
+    return obj;
+  }, {} as { [key: string]: PrimitiveType });
```
fpaul-1A, line 432: since this is used several times we could extract the code in a dedicated function
```typescript
export function objectFromPairwiseArray<T>(inputArray: T[]): Record<string, T> {
  return inputArray.reduce((obj: { [key: string]: T}, currentValue, index, array) => {
    if (index % 2 === 0) {
      obj[currentValue] = array[index + 1];
    }
    return obj;
  }, []);
}
```
```suggestion
  return objectFromPairwiseArray(splitObject);
```
or something like that

packages/@ama-sdk/core/src/fwk/param-serialization.ts
```diff
@@ -246,3 +289,172 @@ export function serializePathParams<T extends { [key: string]: SupportedParamTyp
...
+/**
+ * Deserialize path parameters of type object
+ * OpenAPI Parameter Serialization {@link https://swagger.io/specification | documentation}
+ * @param serializedParamValue serialized path parameter value
+ * @param paramSerialization parameter serialization
+ */
+export function deserializeObjectPathParams(serializedParamValue: string, paramSerialization: ParamSerialization) {
+  const splitObject: string[] = paramSerialization.style === 'matrix' && paramSerialization.explode
+    ? serializedParamValue.substring(1).split(';')
+    : deserializeArrayPathParams(serializedParamValue, paramSerialization)!;
```
cpaulve-1A, line 419: Why do you use the ArrayPathParams method to deserialize an object?

packages/@ama-sdk/core/src/fwk/param-serialization.ts
sdo-1A: otherwise I get the following error: 
`TS2862: Type { [p in keyof T]: SupportedParamType; } is generic and can only be indexed for reading.`

packages/@ama-sdk/core/src/fwk/param-serialization.ts
cpaulve-1A: Even if this is a similar logic, it seems confusing when reading the code.

sdo-1A: Since the delimiter of the serialized object parameter is the same as the delimiter of array parameters:
https://swagger.io/specification/#style-examples

packages/@ama-sdk/core/src/fwk/param-serialization.ts
```diff
@@ -246,3 +289,172 @@ export function serializePathParams<T extends { [key: string]: SupportedParamTyp
...
+  if (paramSerialization.explode) {
+    return splitObject.reduce((obj: { [key: string]: PrimitiveType }, currentValue) => {
+      const [key, value] = currentValue.split('=');
+      obj[key] = value;
+      return obj;
+    }, {} as { [key: string]: PrimitiveType });
+  }
+  return splitObject.reduce((obj: { [key: string]: PrimitiveType }, currentValue, index, array) => {
```
cpaulve-1A, line 427: We could have a comment here to explain the logic between the alternative key/value pairs

packages/@ama-sdk/core/src/fwk/param-serialization.ts
```diff
@@ -246,3 +289,172 @@ export function serializePathParams<T extends { [key: string]: SupportedParamTyp
...
+ * @param serializedParamValue serialized path parameter value
+ * @param paramSerialization parameter serialization
+ */
+export function deserializeArrayPathParams(serializedParamValue: string, paramSerialization: ParamSerialization) {
+  switch (paramSerialization.style) {
+    case 'simple': {
+      return serializedParamValue.split(',');
+    }
+    case 'label': {
+      return paramSerialization.explode ? serializedParamValue.substring(1).split('.') : serializedParamValue.substring(1).split(',');
```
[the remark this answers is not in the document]
sdo-1A: I linked the OpenAPI documentation in the function description, I can add comments if you find it easier otherwise the documentation provides examples

packages/@ama-sdk/core/src/fwk/param-serialization.ts
sdo-1A: Same as the comment below, I added the link to the OpenAPI documentation in the function description but I can add description comments if this is not enough

packages/@ama-sdk/core/src/fwk/param-serialization.ts
sdo-1A: this is used only twice (deserialization of object query and path parameters), I am not sure if it is worth creating another function in the code but if it makes it easier to read/understand then yes I can make the update

packages/@ama-sdk/core/src/fwk/param-serialization.ts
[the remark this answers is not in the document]
cpaulve-1A: Or just an intermediate variable:
` // All label param value are prefixed with a '.'
const labelParamValue = serializedParamValue.substring(1)`

Or you could have something like 
`export function deserializeArrayPathParams(serializedParamValue: string, paramSerialization: ParamSerialization) {
  let prefixParamValue;
  let delimiter
  switch (paramSerialization.style) {
       case 'matrix': {
          prefixParamValue = paramSerialization.match('^(;.+=)')[1];
          delimiter = paramSerialization.explode ? ';' : prefixParamValue;
      }
 case 'simple': {
      prefixParamValue = '';
      delimiter = ',';
      break;
    }
    case 'label': {
        prefixParamValue = '.';
        delimiter = paramSerialization.explode ? '.' : ',';
        break;
    }
    }
    const value = serializedParamValue.replace(`^${}`, '');
    return serializedParamValue.split(delimiter);
  }
}

packages/@ama-sdk/core/src/fwk/param-serialization.ts
cpaulve-1A:
```typescript
    myParam.matchAll(`${outerDelimiter}?[^${innerDelimiter}]+${innerDelimiter}[^${outerDelimiter}]+${outerDelimiter}?`).reduce(obj, serializedProperty => {
    const splitProperty = serializedProperty.match('^${outerDelimiter}?([^${innerDelimiter}]+)${innerDelimiter}([^${outerDelimiter}]+)${outerDelimiter}?
});
   obj[splitProperty[1]] = splitProperty[2];
```
But I am not sure it is clearer (it works for all the use case I think if you remove the prefix)

packages/@ama-sdk/core/src/fwk/param-serialization.ts
sdo-1A: I added a comment to explain, hoping this makes it a bit clearer, what do you think?

packages/@ama-sdk/core/src/fwk/param-serialization.ts
sdo-1A: @cpaulve-1A let me know if the added comments make it a bit easier to understand the logic

packages/@ama-sdk/schematics/README.md
sdo-1A: the serialization has been added to the request plugins

packages/@ama-sdk/core/src/fwk/param-serialization.ts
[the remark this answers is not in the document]
sdo-1A: @cpaulve-1A let me know if the added comments make it a bit easier to understand the logic

packages/@ama-sdk/core/src/fwk/param-serialization.ts
cpaulve-1A: Please document it just in case

packages/@ama-sdk/core/src/fwk/param-serialization.ts
cpaulve-1A: My bad, I just did not get you had the process if the explode and style property were provided or not.
Maybe we can rephrase the inference method to something
isParamValueRecord, isSerializedFormat or reverse the logic with a isParamWithSerializationInfo.
I am not sure these names are better though.

packages/@ama-sdk/core/src/fwk/param-serialization.ts
cpaulve-1A: I would rename the method to avoid confusion or duplicate code.

packages/@ama-sdk/core/src/fwk/param-serialization.ts
```diff
@@ -246,3 +289,192 @@ export function serializePathParams<T extends { [key: string]: SupportedParamTyp
...
+ * @param paramSerialization parameter serialization
+ */
+export function deserializeObjectPathParams(serializedParamValue: string, paramSerialization: ParamSerialization) {
+  // NOTE: Object path parameters can be split by the same delimiters as the path parameters of type array
+  // NOTE: There is an exception for path parameters of exploded style 'matrix' which is not serialized like its corresponding array
+  // This exception is serialized like this (prefixed by a ';'): ';prop1=value1;prop2=value2'
+  const splitObject: string[] = paramSerialization.style === 'matrix' && paramSerialization.explode
+    ? serializedParamValue.substring(1).split(';').map((value) => decodeURIComponent(value))
+    : deserializeArrayPathParams(serializedParamValue, paramSerialization)!;
+  // NOTE: Object path parameters that are non-exploded are serialized as 'prop=value'
```
cpaulve-1A, line 438: I don't understand why we provide the serialization for the non-exploded syntax above the exploded one

packages/@ama-sdk/core/src/fwk/param-serialization.ts
```diff
@@ -246,3 +289,192 @@ export function serializePathParams<T extends { [key: string]: SupportedParamTyp
...
+  if (paramSerialization.explode) {
+    return splitObject.reduce((obj: { [key: string]: PrimitiveType }, currentValue) => {
```
cpaulve-1A, line 440:
```suggestion
    return splitObject.reduce((obj: { [key: string]: PrimitiveType }, serializedProperty) => {
```

packages/@ama-sdk/core/src/fwk/param-serialization.ts
cpaulve-1A: You could rename it something like splitParameterElements

packages/@ama-sdk/core/src/fwk/param-serialization.ts
sdo-1A: After discussion, we decided to add these logs in another PR if necessary

packages/@ama-sdk/core/src/fwk/param-serialization.ts
sdo-1A: oops this is a typo, should be `exploded` instead of `non-exploded`

sdo-1A:
After reflection, we have decided to implement the deserialization of the parameters in a separate PR since these functions are beneficial but non-essential. This will allow further time for discussions on the implementation.
Also, this deserialization feature does not need to target `release/12.1` necessarily so this new PR will target the `main` branch.

cpaulve-1A (dismissed):

packages/@ama-sdk/schematics/README.md
```diff
@@ -119,9 +119,17 @@ For more information, check out OpenAPI's documentation on [parameter serializat
 It is important to note that, as in OpenAPI 3.1, we only support simple arrays and simple non-nested objects in path and query parameters. 
 The parameter types that we support are stored in `SupportedParamType` in the package `@ama-sdk/core`.
 
+We provide the methods `serializeQueryParams` and `serializePathParams` to serialize the values of query and path parameters. However, it is also possible to pass
```
cpaulve-1A, line 122: We should document as well the new flag to enable the serialization.

packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/java/com/amadeus/codegen/ts/LambdaHelper.java
```diff
@@ -76,12 +93,28 @@ public String formatFragment(String fragment) {
...
    * to
-   * /carts/${this.piiParamTokens['$1'] || serializedPathParams['$1']}/travelers
+   * /carts/${this.piiParamTokens['$1'] || data['$1']}/travelers
    */
   public static class TokenizedUrlParamReplacerLambda extends CustomLambda {
 
       public TokenizedUrlParamReplacerLambda() {}
 
+      @Override
+      public String formatFragment(String fragment) {
```
cpaulve-1A, line 103: Do you still need it?

packages/@ama-sdk/core/src/plugins/additional-params/additional-params-sync.request.ts
```diff
@@ -38,7 +44,11 @@ export class AdditionalParamsSyncRequest implements RequestPlugin {
         const body = this.additionalParams.body && isStringOrUndefined(data.body) ? this.additionalParams.body(data.body) : undefined;
 
         if (queryParams) {
-          data.queryParams = { ...data.queryParams, ...queryParams };
+          if (data.paramSerializationOptions?.enableParameterSerialization && !isParamValueRecord(queryParams)) {
+            data.queryParams = { ...data.queryParams, ...serializeRequestPluginQueryParams(queryParams) };
+          } else if (isParamValueRecord(queryParams)) {
```
fpaul-1A, line 49: should we log/throw an error in case both `data.paramSerializationOptions?.enableParameterSerialization` and `isParamValueRecord(queryParams)` are false?
I assume here the params are silently ignored

packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/resources/typescriptFetch/api/api.mustache
```diff
@@ -100,33 +98,65 @@ export class {{classname}} implements Api {
       body = data['{{baseName}}'] as any;
     }
     {{/bodyParam}}
+
```
sdo-1A, line 101: To facilitate the review of this file, I will explain my logic here (small tip: the file is easier to read in **unified** mode):
- All of the query / path parameter logic is grouped together since these params are not needed for the headers or body
- `queryParams` will always be created and initialized as an empty object `{}`
- `paramSerializationOptions` will always be created with the value of `this.client.options.enableParameterSerialization` since this is passed to the request options
- If there are path parameters, we will create the `if` statement to allow path parameter serialization
  - Within this `if` statement, we check if there are query params as well to add their serialization logic if needed
- If there are no path parameters, the variables `basePath` and `tokenizedUrl` are initialized as they were before
  - We check if there are query parameters to add the same logic as before (an `if` statement with the serialization logic)

packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/java/com/amadeus/codegen/ts/LambdaHelper.java
sdo-1A: yes in case there are serialized path parameters

sdo-1A: It's not clear since the history of these files is not shown but I have reverted `TokenizedUrlParamReplacerLambda` and `UrlParamReplacerLambda` to their previous values (using `data`) and created new classes for each of them to use `serializedPathParams` instead of `data`

packages/@ama-sdk/core/src/plugins/additional-params/additional-params-sync.request.ts
sdo-1A: I could throw an error yes if you find this helpful, I was not sure when to add error logs or not

sdo-1A: updated

cpaulve-1A (approved):

packages/@o3r-training/training-sdk/src/api/dummy/dummy-api.ts
fpaul-1A, line 1: this one hasn't been re-generated?

sdo-1A: yes it has been re-generated, but there are no query params and no path params

fpaul-1A (approved):

cpaulve-1A (approved):

PR #3084, opened 2025-04-03
fix: codecov config to reduce false positive and add palette generator

kpanot:
## Proposed change

fix: codecov config to reduce false positive and add palette generator

## Related issues

*- No issue associated -*

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -26,6 +26,11 @@ ignore:
   - 'tools/@*/*/dist'
   - 'apps/@*/*/dist'
 
+  # builder cache folders
+  - 'packages/@*/*/build'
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -26,6 +26,11 @@ ignore:
   - 'tools/@*/*/dist'
   - 'apps/@*/*/dist'
 
+  # builder cache folders
+  - 'packages/@*/*/build'
```
fpaul-1A, line 30: still wishing it could just read our `.gitignore` :(

fpaul-1A (approved):

cpaulve-1A (approved):

PR #3091, opened 2025-04-07
chore(ama-sdk-schematics): ignore coverage for generated files

kpanot:
## Proposed change

chore(ama-sdk-schematics): ignore coverage for generated files

## Related issues

*- No issue associated -*


packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/resources/typescriptFetch/api/api.mustache
```diff
@@ -1,3 +1,5 @@
+/* istanbul ignore next -- generated file */
```
matthieu-crouzet, line 1: This comment will be generated at the beginning of all generated files 
But what if the project doesn't use istanbul for code coverage
I would expect to have a configuration file in which we specify folder to exclude
as we have done in our case 
https://github.com/AmadeusITGroup/otter/pull/3091/files#diff-3e878d9bdc47f29a0ab909a7db939acf08d2d2f1aaba81e6f897b32361fa40dbR34-R40

kpanot: I think I guess what disturb you but your recommendation can not be applied :S.
We recommend Jest (which come with Istanbul) as we recommend EsLint.
In contrary, we do not recommend CodeCov.
The way to ignore code from Istanbul coverage is the added comment (cf: https://stackoverflow.com/questions/35147449/exclude-pattern-or-folder-in-istanbul-coverage).

These added comments do not affect the final built JS, but, and I think it is the reason of our message, add a line in a TS code.
We can add an option to the generator to disable this additional line but I see more advantage to keep it to promote coverage in our generated SDKs.

matthieu-crouzet: Why not using the istanbul configuration property [exclude](https://github.com/istanbuljs/babel-plugin-istanbul#ignoring-files) to not pollute every files with the same comment ?

kpanot: Unfortunately this is not Istanbul configuration but `babel-plugin-istanbul` one.
It does not seem applicable to Jest without adding a dependency to this plugin :S.

matthieu-crouzet: https://jestjs.io/docs/configuration#coveragepathignorepatterns-arraystring
sorry I think this one is the one I wanted to talk about 😄

packages/@ama-sdk/schematics/package.json
```diff
@@ -71,7 +71,8 @@
     "@openapitools/openapi-generator-cli": "^2.15.0",
     "openapi-types": "^12.0.0",
     "ts-node": "~10.9.2",
-    "type-fest": "^4.30.1"
+    "type-fest": "^4.30.1",
+    "typescript": "~5.8.2"
```
mrednic-1A, line 75: It is not an optional one as it is used only in schematics ?

packages/@ama-sdk/schematics/schematics/ng-update/typescript/v12.3/coverage-ignore.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+} from 'node:path';
+import {
+  Tree,
+} from '@angular-devkit/schematics';
+import {
+  updateJestConfigCoveragePathIgnorePatterns,
+} from './coverage-ignore';
+
+describe('updateJestConfigCoveragePathIgnorePatterns', () => {
+  const mock = readFileSync(resolve(__dirname, '..', '..', '..', '..', 'testing', 'mocks', 'jest-config', 'mock.jest.config.js'), { encoding: 'utf8' });
```
mrednic-1A, line 15: I cannot see these 2 mock files in the repo, neither pushed in this PR. I think I mis something :S
https://github.com/AmadeusITGroup/otter/tree/main/packages/%40ama-sdk/schematics/testing/mocks

packages/@ama-sdk/schematics/testing/mocks/jest-config/result.jest.config.js
```diff
@@ -0,0 +1,17 @@
...
+module.exports = {
+  displayName: require('./package.json').name,
+  preset: 'ts-jest',
+  rootDir: '.',
+  roots: ['<rootDir>/src/'],
+  reporters: [
+    'default',
+    ['jest-junit', { outputDirectory: '<rootDir>/dist-test', suiteName: '<% if (projectName) { %>@<%=projectName%>/<% } %><%=projectPackageName%> unit tests' }]
+  ],
+  coveragePathIgnorePatterns: [
```
matthieu-crouzet, line 11: In our repo we use codecov to declare the files to ignore.
But if it take the information from the jest config,
shouldn't we set it in the jest config and remove it from codecov?
So framework user will be able to have example in our repo as well.

matthieu-crouzet (dismissed):

matthieu-crouzet (dismissed):

packages/@ama-sdk/schematics/schematics/ng-update/typescript/v12.3/coverage-ignore.ts
```diff
@@ -0,0 +1,83 @@
...
+    (ctx) => (rootNode) => {
+      const { factory } = ctx;
+      return ts.visitNode(rootNode, (node) => {
+        return ts.visitEachChild(node, (statement) => {
+          if (
+            ts.isExpressionStatement(statement)
+            && ts.isBinaryExpression(statement.expression)
+            && /module\.exports/.test(statement.expression.left.getText(rootNode))
+            && ts.isObjectLiteralExpression(statement.expression.right)
+            && !statement.expression.right.properties.some((prop) => ts.isPropertyAssignment(prop) && prop.name.getText(rootNode) === 'coveragePathIgnorePatterns')
```
cpaulve-1A, line 42: Should we try to complete missing files if the coveragePathIgnorePatterns is already defined?

kpanot: I would say that discovering intersection between globs is a bit too much :D.

fpaul-1A (dismissed):

matthieu-crouzet (approved):

PR #3093, opened 2025-04-07
chore: install with no scripts after postinstall

vscaiceanu-1a:
## Proposed change

Run `yarn install` after the `postinstall` scripts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "test:affected": "yarn nx affected --target=test",
     "test-e2e": "yarn nx run-many --target=test-e2e",
     "test-int": "yarn nx run-many --target=test-int",
-    "postinstall": "husky && yarn build:lint && yarn harmonize:version && yarn update-yarn-sdks",
+    "postinstall": "husky && yarn build:lint && yarn harmonize:version && yarn update-yarn-sdks && yarn install --mode=skip-build",
     "update-yarn-sdks": "node -e \"'pnp' !== '$(yarn config get nodeLinker)' || process.exit(1)\" || yarn sdks",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -27,7 +27,7 @@
     "test:affected": "yarn nx affected --target=test",
     "test-e2e": "yarn nx run-many --target=test-e2e",
     "test-int": "yarn nx run-many --target=test-int",
-    "postinstall": "husky && yarn build:lint && yarn harmonize:version && yarn update-yarn-sdks",
+    "postinstall": "husky && yarn build:lint && yarn harmonize:version && yarn update-yarn-sdks && cross-env YARN_ENABLE_SCRIPTS=false yarn install",
```
cpaulve-1A, line 30: This should be part of the yarnrc

cpaulve-1A: We should never allow the scripts from our packages

cpaulve-1A: Is it linked to https://github.com/LavaMoat/LavaMoat/pull/1218 ?
If so, please add a reference in a ticket or in the code

vscaiceanu-1a: `YARN_ENABLE_SCRIPTS=false` doesn't execute `postinstall`. The equivalent of `npm install --ignore-scripts`.

We want to skip the `postinstall` scripts for `yarn install` only in this case, right?

kpanot, line 30:
```suggestion
    "postinstall": "husky && yarn build:lint && yarn harmonize:version && yarn update-yarn-sdks && yarn install --mode=skip-build",
```
was not doing the job?

vscaiceanu-1a: it seems to be doing a better job.

kpanot (approved):

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #3105, opened 2025-04-09
chore(mfe): upgrade microfrontends packages + fix typings issue

matthieu-crouzet:
## Related issues

*- No issue associated -*

### DIFF
--- a/.renovaterc.json5
+++ b/.renovaterc.json5
@@ -2,6 +2,7 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
     "github>AmadeusITGroup/otter//tools/renovate/base",
+    "github>AmadeusITGroup/otter//tools/renovate/group/mfe",
     "github>AmadeusITGroup/otter//tools/renovate/group/design-factory",
     "github>AmadeusITGroup/otter//tools/renovate/branching-strategy/release-branches",
     ":labels(dependencies)"
--- a/nx.json
+++ b/nx.json
@@ -488,7 +488,7 @@
     "prepare": {
       "executor": "nx:run-script",
       "inputs": [
-        "base"
+        "project"
       ],
       "outputs": [
         "{projectRoot}/src/package.json"
--- a/packages/@ama-mfe/messages/package.json
+++ b/packages/@ama-mfe/messages/package.json
@@ -32,7 +32,7 @@
     "tslib": "^2.6.2"
   },
   "peerDependencies": {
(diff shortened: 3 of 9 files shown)


tools/renovate/group/otter.json
```diff
@@ -14,6 +14,13 @@
         "@ama-mfe"
       ]
     },
+    {
```
kpanot, line 17: It is a bit misleading to have this Otter dedicated rules file.
Why not creating another `group/mfe.json` file which can be referred in the `otter-project.json`?

tools/renovate/group/otter.json
```diff
@@ -5,13 +5,14 @@
...
-        "@o3r-training",
-        "@ama-mfe"
+      "matchPackagePatterns": [
+        "^@otter",
+        "^@o3r",
+        "^@ama-sdk",
+        "^@ama-terasu",
+        "^@o3r-training",
+        "^@ama-mfe",
+        "^@amadeus-it-group/microfrontends"
```
kpanot, line 15: Not sure it should be in this group.
This group is currently matching the packages released by Otter is a similar release cycle. It is not the case of this package :S

matthieu-crouzet: don't we want to group `@ama-mfe` and `@amadeus-it-group/microfrontends` ?

fpaul-1A (dismissed):

.renovaterc.json5
```diff
@@ -44,6 +44,13 @@
         ]
       }
     },
+    {
```
kpanot, line 47: Maybe you can directly extend `github>AmadeusITGroup/otter//tools/renovate/group/mfe`

packages/@ama-mfe/ng-utils/project.json
```diff
@@ -21,7 +21,11 @@
         "tsConfig": "packages/@ama-mfe/ng-utils/tsconfig.build.json"
       }
     },
-    "prepare": {},
+    "prepare": {
+      "inputs": [
+        "project"
```
kpanot, line 26: Why is it needed?
It seems to be standard and execute `cp-package-json`, does it mean that `project` should be used instead od `base` in the other `prepare` action as well?

matthieu-crouzet: Indeed, I didn't check that it was the exact same script for all `prepare`

kpanot (approved):

fpaul-1A (approved):

vscaiceanu-1a (approved):

PR #3108, opened 2025-04-09
chore(renovate): use matchPackageNames instead of matchPackagePrefixes and matchPackagePatterns

matthieu-crouzet:
## Related issues

*- No issue associated -*

### DIFF
--- a/.renovaterc.json5
+++ b/.renovaterc.json5
@@ -85,8 +85,8 @@
     {
       // This rule disable the upgrade of the gaurav-nelson/github-action-markdown-link-check
       // TODO: re-activate the upgrade when the following issue is fixed: gaurav-nelson/github-action-markdown-link-check#200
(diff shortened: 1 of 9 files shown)


.renovaterc.json5
```diff
@@ -85,8 +85,8 @@
     {
       // This rule disable the upgrade of the gaurav-nelson/github-action-markdown-link-check
       // TODO: re-activate the upgrade when the following issue is fixed: gaurav-nelson/github-action-markdown-link-check#200
-      "matchPackagePatterns": [
-        ".*markdown-link-check"
+      "matchPackageNames": [
+        "/.*markdown-link-check/"
```
fpaul-1A, line 89: we should document in the readme that the presets only support Renovate >=38

fpaul-1A (approved):

PR #3116, opened 2025-04-10
feat: parameter deserialization

sdo-1A:
## Proposed change

Provide deserialization functions as a tool to deserialize query and path parameters.
This PR was created as a split of PR #3076 

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@ama-sdk/core/src/fwk/param-deserialization.spec.ts
+++ b/packages/@ama-sdk/core/src/fwk/param-deserialization.spec.ts
@@ -1,0 +1,74 @@
+import {
+  deserializePathParams,
+  deserializeQueryParams,
+} from './param-deserialization';
+
+describe('Deserialize parameters', () => {
+  it('should correctly deserialize query parameters of type primitive', () => {
+    expect(deserializeQueryParams({ idPrimitive: 'idPrimitive=5' }, { idPrimitive: { explode: true, style: 'form', paramType: 'primitive' } })).toEqual({ idPrimitive: '5' });
+    expect(deserializeQueryParams({ idPrimitive: 'idPrimitive=5' }, { idPrimitive: { explode: false, style: 'form', paramType: 'primitive' } })).toEqual({ idPrimitive: '5' });
+  });
+
+  it('should correctly deserialize query parameters of type array', () => {
+    expect(deserializeQueryParams({ idArray: 'idArray=3&idArray=4&idArray=5' }, { idArray: { explode: true, style: 'form', paramType: 'array' } })).toEqual({ idArray: ['3', '4', '5'] });
+    expect(deserializeQueryParams({ idArray: 'idArray=3,4,5' }, { idArray: { explode: false, style: 'form', paramType: 'array' } })).toEqual({ idArray: ['3', '4', '5'] });
+    expect(() => deserializeQueryParams({ idArray: 'should-not-work' }, { idArray: { explode: true, style: 'spaceDelimited', paramType: 'array' } }))
+      .toThrow(`Unable to deserialize query parameter idArray since the combination explode=true and style='spaceDelimited' is not supported.`);
+    expect(deserializeQueryParams({ idArray: 'idArray=3%204%205' }, { idArray: { explode: false, style: 'spaceDelimited', paramType: 'array' } })).toEqual({ idArray: ['3', '4', '5'] });
+    expect(() => deserializeQueryParams({ idArray: 'should-not-work' }, { idArray: { explode: true, style: 'pipeDelimited', paramType: 'array' } }))
+      .toThrow(`Unable to deserialize query parameter idArray since the combination explode=true and style='pipeDelimited' is not supported.`);
+    expect(deserializeQueryParams({ idArray: 'idArray=3%7C4%7C5' }, { idArray: { explode: false, style: 'pipeDelimited', paramType: 'array' } })).toEqual({ idArray: ['3', '4', '5'] });
+  });
+
+  it('should correctly deserialize query parameters of type object', () => {
+    expect(deserializeQueryParams({ idObject: 'role=admin&firstName=Alex' }, { idObject: { explode: true, style: 'form', paramType: 'object' } }))
+      .toEqual({ idObject: { role: 'admin', firstName: 'Alex' } });
+    expect(deserializeQueryParams({ idObject: 'idObject=role,admin,firstName,Alex' }, { idObject: { explode: false, style: 'form', paramType: 'object' } }))
+      .toEqual({ idObject: { role: 'admin', firstName: 'Alex' } });
+    expect(() => deserializeQueryParams({ idObject: 'should-not-work' }, { idObject: { explode: true, style: 'spaceDelimited', paramType: 'object' } }))
+      .toThrow(`Unable to deserialize query parameter idObject since the combination explode=true and style='spaceDelimited' is not supported.`);
+    expect(deserializeQueryParams({ idObject: 'idObject=role%20admin%20firstName%20Alex' }, { idObject: { explode: false, style: 'spaceDelimited', paramType: 'object' } }))
+      .toEqual({ idObject: { role: 'admin', firstName: 'Alex' } });
+    expect(() => deserializeQueryParams({ idObject: 'should-not-work' }, { idObject: { explode: true, style: 'pipeDelimited', paramType: 'object' } }))
+      .toThrow(`Unable to deserialize query parameter idObject since the combination explode=true and style='pipeDelimited' is not supported.`);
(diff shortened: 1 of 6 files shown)


packages/@ama-sdk/core/src/fwk/param-deserialization.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+describe('Deserialize parameters', () => {
+  it('should correctly deserialize query parameters of type primitive', () => {
+    expect(deserializeQueryParams({ idPrimitive: 'idPrimitive=5' }, { idPrimitive: { explode: true, style: 'form', paramType: 'primitive' } })).toEqual({ idPrimitive: '5' });
+    expect(deserializeQueryParams({ idPrimitive: 'idPrimitive=5' }, { idPrimitive: { explode: false, style: 'form', paramType: 'primitive' } })).toEqual({ idPrimitive: '5' });
+  });
+
+  it('should correctly deserialize query parameters of type array', () => {
+    expect(deserializeQueryParams({ idArray: 'idArray=3&idArray=4&idArray=5' }, { idArray: { explode: true, style: 'form', paramType: 'array' } })).toEqual({ idArray: ['3', '4', '5'] });
+    expect(deserializeQueryParams({ idArray: 'idArray=3,4,5' }, { idArray: { explode: false, style: 'form', paramType: 'array' } })).toEqual({ idArray: ['3', '4', '5'] });
+    expect(deserializeQueryParams({ idArray: 'should-not-work' }, { idArray: { explode: true, style: 'spaceDelimited', paramType: 'array' } })).toEqual({ idArray: undefined });
```
fpaul-1A, line 15: should this return `undefined` or throw an error?

mrednic-1A (dismissed):

packages/@ama-sdk/core/src/fwk/param-deserialization.ts
```diff
@@ -0,0 +1,212 @@
+import {
+  CLOSING_SQUARE_BRACKET_URL_CODE,
+  OPENING_SQUARE_BRACKET_URL_CODE,
+  ParamSerialization,
```
kpanot, line 4:
```suggestion
  type ParamSerialization,
```
?

packages/@ama-sdk/core/src/fwk/param-deserialization.ts
```diff
@@ -0,0 +1,212 @@
...
+      deserializedValue = deserializeObjectQueryParams(serializedParamValue, paramSerialization);
+    } else {
+      // NOTE: Query parameters of type primitive are prefixed with the parameter name like this: 'paramName=value'
+      deserializedValue = decodeURIComponent(serializedParamValue.split('=')[1]);
+    }
+    if (deserializedValue) {
+      acc[queryParamName as keyof T] = deserializedValue;
+    }
+    return acc;
+  }, {} as { [p in keyof T]: SupportedParamType });
```
kpanot, line 127: the type is a bit incorrect as this stage (accept if `T` is `{}`).
Maybe it would be more something like:
```suggestion
  }, {} as Record<string, SupportedParamType>) as { [p in keyof T]: SupportedParamType };
```

packages/@ama-sdk/core/src/fwk/param-deserialization.ts
```diff
@@ -0,0 +1,212 @@
...
+    } else if (paramSerialization.paramType === 'object') {
+      deserializedValue = deserializeObjectPathParams(serializedParamValue, paramSerialization);
+    } else {
+      deserializedValue = deserializePrimitivePathParams(serializedParamValue, paramSerialization);
+    }
+    if (deserializedValue) {
+      acc[pathParamName as keyof T] = deserializedValue;
+    }
+    return acc;
+  }, {} as { [p in keyof T]: SupportedParamType });
```
kpanot, line 211: Same comment as above:
```suggestion
  }, {} as Record<string, SupportedParamType>) as { [p in keyof T]: SupportedParamType };
```

packages/@ama-sdk/core/src/fwk/param-deserialization.ts
sdo-1A: Isn't it fine to accept `{}` ? Because we have conditions in the functions to avoid unsupported use cases
If I change the type to `Record<string, SupportedParamType>`, we could not return an `{}`

packages/@ama-sdk/core/src/fwk/param-deserialization.spec.ts
sdo-1A: this should return `undefined`, we don't throw errors for non-supported use cases

fpaul-1A: I would expect a function like this to throw when the format is not recognized, but I'm ok to return `undefined`

packages/@ama-sdk/core/src/fwk/param-deserialization.ts
kpanot: `{ [p in keyof T]: SupportedParamType }` means that all fields of the object `T` should be present in your object. So if `T` has more than 1 field, the code `{} as { [p in keyof T]: SupportedParamType }` is wrong.

The purpose is not to change the return type (which remains `{ [p in keyof T]: SupportedParamType }`) but the `acc` type to avoid wrong typing.

sdo-1A: As discussed by message, we decided to use `Partial<...>` objects in order to correctly type the `acc` in the reducers.
This update was applied in both the deserialization and serialization functions.

packages/@ama-sdk/client-beacon/src/api-beacon-client.ts
```diff
@@ -110,15 +110,15 @@ export class ApiBeaconClient implements ApiClient {
...
+  public serializeQueryParams<T extends { [key: string]: SupportedParamType }>(queryParams: T, queryParamSerialization: { [p in keyof T]: ParamSerialization }): Partial<{ [p in keyof T]: string }> {
     if (this.options.serializeQueryParams) {
       return this.options.serializeQueryParams(queryParams, queryParamSerialization);
     }
     return serializeQueryParams(queryParams, queryParamSerialization);
   }
 
   /** @inheritdoc */
-  public serializePathParams<T extends { [key: string]: SupportedParamType }>(pathParams: T, pathParamSerialization: { [p in keyof T]: ParamSerialization }): { [p in keyof T]: string } {
+  public serializePathParams<T extends { [key: string]: SupportedParamType }>(pathParams: T, pathParamSerialization: { [p in keyof T]: ParamSerialization }): Partial<{ [p in keyof T]: string }> {
```
cpaulve-1A, line 121: I don't think you need to put them as Partial. You are not supposed to remove keys from `pathParams`. Also, returning Partial here would technically be considered as a breaking change (true for the whole file)

packages/@ama-sdk/core/src/fwk/param-deserialization.ts
```diff
@@ -0,0 +1,212 @@
...
+    case 'label': {
+      // NOTE: Path parameters of style label are prefixed with a '.'
+      return serializedParamValue.substring(1).split(paramSerialization.explode ? '.' : ',').map((value) => decodeURIComponent(value));
+    }
+    case 'matrix': {
+      // NOTE: Path parameters of style matrix and exploded true are prefixed with a ';paramName='
+      // NOTE: Path parameters of style matrix and exploded false are written like this: ';paramName=value1;paramName=value2'
+      return paramSerialization.explode
+        ? serializedParamValue.substring(1).split(';').map((value) => decodeURIComponent(value.split('=')[1]))
+        : serializedParamValue.split('=')[1].split(',').map((value) => decodeURIComponent(value));
```
cpaulve-1A, line 33: From the comment, I expected .split(';') instead of ','

cpaulve-1A (approved):

packages/@ama-sdk/client-beacon/src/api-beacon-client.ts
sdo-1A: **Update:** we discussed with the team and decided to throw errors in non-supported use cases in order to return a non-partial object and avoid this breaking change

packages/@ama-sdk/core/src/fwk/param-deserialization.ts
sdo-1A: As mentioned in another comment below, this previous message can be ignored.

packages/@ama-sdk/core/src/fwk/param-deserialization.spec.ts
sdo-1A: **Update:** we modified in order to throw errors for non-supported use cases

kpanot (approved):

PR #3133, opened 2025-04-16
fix(rules-engine): debug perfs linked to fact snapshots

fpaul-1A:
## Proposed change

Currently, every time the value of a fact changes, a full snapshot is sent.
To avoid a big overload, I propose to send it at max once every second.

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@o3r/rules-engine/src/engine/debug/engine.debug.ts
+++ b/packages/@o3r/rules-engine/src/engine/debug/engine.debug.ts
@@ -1,12 +1,15 @@
 import {
+  debounceTime,
   Observable,
   ReplaySubject,
+  Subject,
 } from 'rxjs';
 import {
   concatMap,
   share,
   shareReplay,
   startWith,
+  switchMap,
   tap,
   withLatestFrom,
 } from 'rxjs/operators';
@@ -58,6 +61,8 @@ export class EngineDebugger {
 
   private performanceMeasures$!: Observable<PerformanceMeasure[]>;
 
+  private readonly requestFactsSnapshot = new Subject<void>();
+
   /** Stream emitting a debug event when is fired; timeline is kept */
   public readonly debugEvents$: Observable<DebugEvent>;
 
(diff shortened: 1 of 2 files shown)


packages/@o3r/rules-engine/src/engine/debug/engine.debug.ts
```diff
@@ -303,15 +320,7 @@ export class EngineDebugger {
...
-    factValue$.subscribe(async () => {
-      const timestamp = Date.now();
-      const facts = await this.getFactsSnapshot(this.registeredRuleEngine!.getRegisteredFactsNames());
-      this.debugEventsSubject$.next(() => ({
-        timestamp,
-        type: 'AvailableFactsSnapshot',
-        facts
-      }));
-    });
+    factValue$.subscribe(() => this.requestFactsSnapshot.next());
```
matthieu-crouzet, line 323: Why not ?
```suggestion
    factValue$.pipe(debounceTime(1000)).subscribe(async () => {
      const timestamp = Date.now();
      const facts = await this.getFactsSnapshot(this.registeredRuleEngine!.getRegisteredFactsNames());
      this.debugEventsSubject$.next(() => ({
        timestamp,
        type: 'AvailableFactsSnapshot',
        facts
      }));
    }));
```
Do we want to be able to trigger `requestFactsSnapshot` in another case ?

packages/@o3r/rules-engine/src/engine/debug/engine.debug.ts
```diff
@@ -99,6 +104,18 @@ export class EngineDebugger {
...
+      switchMap(async () => {
+        const timestamp = Date.now();
+        const facts = await this.getFactsSnapshot(this.registeredRuleEngine!.getRegisteredFactsNames());
+        this.debugEventsSubject$.next(() => ({
+          timestamp,
+          type: 'AvailableFactsSnapshot',
+          facts
+        }));
+      })
+    ).subscribe();
```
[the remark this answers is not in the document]
fpaul-1A: are you sure the `takeUntilDestroyed` would work here? this is not an injectable, there is no `DestroyRef`

packages/@o3r/rules-engine/src/engine/debug/engine.debug.ts
matthieu-crouzet: indeed ! 😃

packages/@o3r/rules-engine/src/engine/debug/engine.debug.ts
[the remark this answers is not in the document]
matthieu-crouzet: then we should expose a `stop` method that unsubscribe it and call it in the ondestroy of the injectable where we use this class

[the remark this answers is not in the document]
fpaul-1A: updated to be subscribed/unsubscribed with `debugEvents$`

packages/@o3r/rules-engine/src/engine/debug/engine.debug.ts
```diff
@@ -73,7 +78,7 @@ export class EngineDebugger {
   constructor(options?: EngineDebuggerOptions) {
     this.debugEventsSubject$ = new ReplaySubject<() => (Promise<DebugEvent> | DebugEvent)>(options?.eventsStackLimit);
     this.initializePerformanceObserver();
-    this.debugEvents$ = this.debugEventsSubject$.pipe(
+    const debugEvents = this.debugEventsSubject$.pipe(
```
cpaulve-1A, line 81: Why remove the $ when it remains an observable ?

matthieu-crouzet (approved):

packages/@o3r/rules-engine/src/engine/debug/engine.debug.ts
```diff
@@ -73,32 +78,50 @@ export class EngineDebugger {
...
+          const timestamp = Date.now();
+          const facts = await this.getFactsSnapshot(this.registeredRuleEngine!.getRegisteredFactsNames());
+          this.debugEventsSubject$.next(() => ({
+            timestamp,
+            type: 'AvailableFactsSnapshot',
+            facts
+          }));
+        })
+      ).subscribe();
+      const debugEventsSubscription = this.debugEventsSubject$.pipe(
```
cpaulve-1A, line 94: Should we also have a debounce here?

cpaulve-1A (approved):

PR #3141, opened 2025-04-17
feat(testing): rework e2e testing

fpaul-1A:
## Proposed change

Simplify the generated e2e setup
- Remove the sanity vs scenario
- Remove the generators

Should we include a basic setup for visual-testing / mocking with Kassette (optional of course) ?

## Related issues

*- No issue associated -*


tsconfig.jest.json
```diff
@@ -7,5 +7,6 @@
     "composite": true,
     "esModuleInterop": true
   },
-  "include": []
+  "include": [],
+  "exclude": ["**/*.e2e.spec.ts"]
```
kpanot, line 11: Would not make more sense to not use `.spec.ts` for e2e scenarii instead?
Is there a reason for this renaming?

vscaiceanu-1a: Today we are using `e2e.spec.ts`. Are you suggesting to use `e2e.ts` for e2e scenarii instead of `e2e.spec.ts`?

vscaiceanu-1a: Agree, but it involves a rename of our existing naming convention. I prefer doing it in a separate PR for clarity. 
Again,`e2e.spec.ts` is what we generate today and what we use in the Otter project.

cpaulve-1A (approved):

packages/@o3r/testing/schematics/index.it.spec.ts
```diff
@@ -43,7 +51,10 @@ describe('ng add testing', () => {
     const { applicationPath, workspacePath, appName, isInWorkspace, o3rVersion, untouchedProjectsPaths, libraryPath } = o3rEnvironment.testEnvironment;
     const execAppOptions = { ...getDefaultExecSyncOptions(), cwd: workspacePath };
     const relativeApplicationPath = path.relative(workspacePath, applicationPath);
-    packageManagerExec({ script: 'ng', args: ['add', `@o3r/testing@${o3rVersion}`, '--testingFramework', 'jest', '--skip-confirmation', '--project-name', appName] }, execAppOptions);
+    packageManagerExec({
+      script: 'ng',
+      args: ['add', `@o3r/testing@${o3rVersion}`, '--enablePlaywright', '--testingFramework', 'jest', '--skip-confirmation', '--project-name', appName]
```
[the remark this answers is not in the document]
matthieu-crouzet: It could but it will slow down the test as it will have to create another environment 
I dont see conflict between playwright and jest and why we should create 2 differents environments
I would expect that both work together not only separately

kpanot (approved):

vscaiceanu-1a (approved):

kpanot (approved):

matthieu-crouzet (approved):

cpaulve-1A (approved):

PR #3142, opened 2025-04-18
feat: allow ng-add outside of Otter project

kpanot:
## Proposed change

feat: allow ng-add outside of Otter project

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@ama-mfe/ng-utils/package.json
+++ b/packages/@ama-mfe/ng-utils/package.json
@@ -22,6 +22,7 @@
     "@amadeus-it-group/microfrontends": "0.0.7",
     "@amadeus-it-group/microfrontends-angular": "0.0.7",
     "@o3r/logger": "workspace:^",
+    "@o3r/schematics": "workspace:^",
     "tslib": "^2.6.2"
   },
   "sideEffects": false,
@@ -52,7 +53,6 @@
     "@o3r/build-helpers": "workspace:^",
     "@o3r/eslint-config": "workspace:^",
     "@o3r/eslint-plugin": "workspace:^",
-    "@o3r/schematics": "workspace:^",
     "@o3r/test-helpers": "workspace:^",
     "@schematics/angular": "~19.2.0",
     "@types/jest": "~29.5.2",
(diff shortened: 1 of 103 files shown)


packages/@ama-mfe/ng-utils/package.json
```diff
@@ -22,6 +22,7 @@
     "@amadeus-it-group/microfrontends": "0.0.7",
     "@amadeus-it-group/microfrontends-angular": "0.0.7",
     "@o3r/logger": "workspace:^",
+    "@o3r/schematics": "workspace:^",
```
fpaul-1A, line 25: won't that be trouble when we try to target a previous minor of @o3r?
say we want to install `@ama-mfe/ng-utils@~12.1.0` but it will install the `@o3r/schematics@12.2`

matthieu-crouzet (approved):
Great it simplifies the code a lot !

packages/@o3r/eslint-config/tsconfig.build.json
```diff
@@ -4,6 +4,7 @@
     "incremental": true,
     "composite": true,
     "module": "CommonJS",
+    "allowJs": true,
```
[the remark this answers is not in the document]
kpanot: Not part of the good PR, it should be on #3144.
This file is not used to compile the `*.cjs` files but only to provide auto-completion to the IDE.
VsCode is nice and cover JS (or `cjs`) even if the `tsconfig` does not allow it but `eslint` is not as gentle as vscode and require explicitly `allowJs`.

packages/@ama-mfe/ng-utils/package.json
fpaul-1A: This is what we currently have when installing @o3r/core
image

I understand that according to our peer dependencies this should not be a problem, but still I find it a bit weird
(not linked to your PR but we are generalizing it to all the packages in a way)

packages/@o3r/core/schematics/ng-add-create/templates/schematics/ng-add/index.ts.template
```diff
@@ -64,9 +70,4 @@ function ngAddFn(options: NgAddSchematicsSchema): Rule {
  * Add module to an Angular Project
  * @param options ng add options
  */
-export const ngAdd = (options: NgAddSchematicsSchema): Rule => async (_, { logger }) => {
-  const {
-    createOtterSchematic
-  } = await import('@o3r/schematics').catch(reportMissingSchematicsDep(logger));
```
mrednic-1A, line 70: you can remove the code of `reportMissingSchematicsDep` few lines above

cpaulve-1A (approved):

fpaul-1A (approved):

PR #3144, opened 2025-04-18
fix: unlock eslint pre-commit process on schematics

kpanot:
## Proposed change

fix: unlock eslint pre-commit process on schematics

## Related issues

*- No issue associated -*

### DIFF
--- a/eslint.local.config.mjs
+++ b/eslint.local.config.mjs
@@ -8,10 +8,7 @@ export default defineConfig([
   {
     name: '@o3r/framework/projects',
     languageOptions: {
-      sourceType: 'module',
-      parserOptions: {
-        projectService: true
-      }
+      sourceType: 'module'
     }
   },
   {
--- a/packages/@ama-sdk/core/tsconfig.json
+++ b/packages/@ama-sdk/core/tsconfig.json
@@ -1,4 +1,4 @@
-/* IDE usage only */
+/* For IDE and EsLint usage */
 {
   "extends": "../../../tsconfig.base",
   "references": [
@@ -6,7 +6,13 @@
       "path": "./tsconfig.spec.json"
     },
     {
+      "path": "./tsconfig.builders.json"
+    },
+    {
       "path": "./tsconfig.build.json"
+    },
+    {
+      "path": "./tsconfig.eslint.json"
     }
   ]
 }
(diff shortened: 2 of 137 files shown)


eslint.local.config.mjs
```diff
@@ -10,7 +10,10 @@ export default defineConfig([
     languageOptions: {
       sourceType: 'module',
       parserOptions: {
-        projectService: true
```
matthieu-crouzet, line 13: Could you check with @malek-benazzouz why we moved to `projectService: true` please ?

packages/@ama-sdk/core/tsconfig.json
```diff
@@ -5,6 +5,9 @@
     {
       "path": "./tsconfig.spec.json"
     },
+    {
+      "path": "./tsconfig.builders.json"
```
matthieu-crouzet, line 9: no eslint config for this package ?

eslint.local.config.mjs
kpanot: `projectService: true` is equivalent to `projectService: ['./**/tsconfig.json']` (with `allowJs: true` on top of the config).
The problem we have is that for `ncc` compilation we use the `tsconfig.json` for compilation (not only for listing the different references as done in the other projects). So we cannot have a unique `tsconfig.json` to cover all the files of the project, we need to add the `'./**/tsconfig.eslint.json'`.

packages/@ama-sdk/core/tsconfig.json
kpanot: It's a miss.
I added it

eslint.local.config.mjs
malek-benazzouz: My change (in an other repository) was to move from `project` to `projectService`, because:
1. it eases linting configuration and aligns better with what the IDE lints. It is now the recommended setup for typed linting as of typescript-eslint v8.
2. It was fixing some IDE import errors in unit test files in our case.

matthieu-crouzet: You remove the `projectService: true`
and from what I understood we should have it everywhere not only at the root

cpaulve-1A (approved):

fpaul-1A (approved):

PR #3152, opened 2025-04-21
feat: add o3r prefix to generators

### LINKED ISSUE #3012: [Feature]: Add differentiator in logs for o3r vs ng generators
### Context

When listing the full list of generators (angular schematics generators available) there is no easy way to know which generator is an angular base one and which one is o3r.

### Proposal

Add a differentiator for otter generators from the angular ones (with a link to the doc if possible).

Bonus issue: Warning on npm workspace not installed, even is installed (2 times if we answer yes for telemetry)

kpanot:
## Proposed change

feat: add o3r prefix to generators

## Related issues

* :rocket: Feature resolves #3012

### DIFF
--- a/packages/@ama-sdk/schematics/collection.json
+++ b/packages/@ama-sdk/schematics/collection.json
@@ -8,29 +8,44 @@
       "aliases": ["install", "i"]
     },
     "migrate": {
-      "description": "Execute migration scripts between 2 versions.",
+      "description": "[ama-sdk] Execute migration scripts between 2 versions.",
       "factory": "./schematics/migrate/index#migrate",
-      "schema": "./schematics/migrate/schema.json"
+      "schema": "./schematics/migrate/schema.json",
+      "aliases": [
+        "ama-migrate"
+      ]
     },
     "typescript-mock": {
-      "description": "Generate an api mock into the project.",
+      "description": "[ama-sdk] Generate an api mock into the project.",
       "factory": "./schematics/typescript/mock/index#ngGenerateMock",
-      "schema": "./schematics/typescript/mock/schema.json"
+      "schema": "./schematics/typescript/mock/schema.json",
+      "aliases": [
+        "ama-typescript-mock"
+      ]
     },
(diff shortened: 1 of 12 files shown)


packages/@ama-sdk/schematics/collection.json
```diff
@@ -7,30 +7,45 @@
       "schema": "./schematics/ng-add/schema.json",
       "aliases": ["install", "i"]
     },
-    "migrate": {
+    "ama-migrate": {
```
matthieu-crouzet, line 10: It's not really the way we discuss it with the team 
We said that we should have something `[o3r] xxx` for the description 
not to change the generator name 
because in this case you will have some with the prefix and some without if we want to override the one from Angular or from another library

kpanot: They are still replacing the Angular's ones thanks to the `alias` but indeed both Otter and original ones will appear in the `--help`, one of the advances is that they are grouped in the display and all accessible generated will appear.

I can of course update the PR and change to comments if it is the preferred choice from the team.

---

As general comment: if a decision refining a issue is taken in an internal discussion, the issue should be updated with the output.

matthieu-crouzet (approved):

packages/@ama-sdk/schematics/collection.json
```diff
@@ -8,29 +8,44 @@
       "aliases": ["install", "i"]
     },
     "migrate": {
-      "description": "Execute migration scripts between 2 versions.",
+      "description": "[ama] Execute migration scripts between 2 versions.",
```
fpaul-1A, line 11: Should we use `[ama]` or `[ama-sdk]`?

fpaul-1A (approved):

fpaul-1A (approved):

vscaiceanu-1a (approved):

matthieu-crouzet (approved):

vscaiceanu-1a (approved):

PR #3155, opened 2025-04-21
fix: yarn warning reporting

kpanot:
## Proposed change

fix: yarn warning reporting

See example on: https://github.com/kpanot/otter/pull/1

## Related issues

*- No issue associated -*


tools/github-actions/yarn-errors-reporter/src/main.ts
```diff
@@ -92,8 +127,9 @@ async function run(): Promise<void> {
     const yarnLockPath = path.resolve(cwd, 'yarn.lock');
     const execOptions = {
       cwd,
-      ignoreReturnCode: true
-    };
+      ignoreReturnCode: true,
+      silent: true
```
matthieu-crouzet, line 131: That's not really cool when we have to debug

tools/github-actions/yarn-errors-reporter/src/main.ts
```diff
@@ -71,14 +82,38 @@ async function removeErrorComment() {
...
+}
+
+function formatComment(output: YarnInstallOutputLine[]) {
+  return [
+    '| Error | Description |',
+    '| -- | -- |',
+
+    ...output.map((line) => `| ${line.displayName} | ${line.data.replace(/\s/g, ' ')
+      // eslint-disable-next-line no-control-regex -- use to remove ansi color char
+      .replace(/\u001B\[[0-9;]+m(.+?)\u001B\[39m/g, '**$1**')
```
matthieu-crouzet, line 112: Could we have a comment to show an example of these transformation, as the regexp is quite complex here

.github/workflows/main.yml
```diff
@@ -39,6 +44,8 @@ jobs:
       - name: Yarn errors reporter
         if: github.event_name == 'pull_request'
         uses: ./tools/github-actions/yarn-errors-reporter
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
matthieu-crouzet, line 48: Why do you need that ?
Isn't this variable accessible by default ?

kpanot: Nop, it is a secret, not a environment variable, an action can not access to a secret.
We use the same methods to the different action requiring the token

tools/github-actions/yarn-errors-reporter/src/main.ts
kpanot: The `debug mode` from the Github Action disable it so it should not be an issue.
I think In contrary, it was polluting the output and the warning reporting was drowned in it

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #3173, opened 2025-04-28
test: cross os relative paths for int tests

vscaiceanu-1a:
## Proposed change

Compute the relative path correctly for int test for all OS

### DIFF
--- a/packages/@o3r/analytics/schematics/index.it.spec.ts
+++ b/packages/@o3r/analytics/schematics/index.it.spec.ts
@@ -40,7 +40,7 @@ describe('ng add analytics', () => {
     ].sort());
 
     [libraryPath, ...untouchedProjectsPaths].forEach((untouchedProject) => {
-      expect(diff.all.some((file) => file.startsWith(path.posix.relative(workspacePath, untouchedProject)))).toBe(false);
+      expect(diff.all.some((file) => file.startsWith(path.relative(workspacePath, untouchedProject).replace(/\\+/g, '/')))).toBe(false);
     });
     expect(() => packageManagerInstall(execAppOptions)).not.toThrow();
     expect(() => packageManagerRunOnProject(appName, isInWorkspace, { script: 'build' }, execAppOptions)).not.toThrow();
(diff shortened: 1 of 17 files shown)


packages/@o3r/test-helpers/src/utilities/fs.ts
fpaul-1A, line 1: this could be name `path.ts` instead of `fs.ts`?

fpaul-1A (approved):

packages/@o3r/test-helpers/src/utilities/path.ts
```diff
@@ -0,0 +1,12 @@
+import * as path from 'node:path';
+
+const normalizeToPosixPath = (p: string) => p.split(path.sep).join(path.posix.sep);
+
+/**
+ * Get a cross OS relative path
```
kpanot, line 6: This is not clear because `path.relative` is cross OS.

Why a simple:
```typescript
return path.relative(from, to).replace(/\\+/g, '/');
```
would not do the job ?

vscaiceanu-1a: The problem is not with `path.relative` itself but with the inputs that we have to normalize before.
Otherwise I was ending up with `../D:\a\otter\it-tests\test-app-eslint-config2\libs\untouched-lib`.

kpanot: Sorry but it is not clear.
Not sure how the `path.relative(from, to).replace(/\\+/g, '/')` can result to `../D:\a\otter\it-tests\test-app-eslint-config2\libs\untouched-lib`.

The original usage of `path.posix.relative` can be problematic on Windows OS but a replacement post `path.relative` does not seem to be.

vscaiceanu-1a:
```typescript
const workspacePath = 'D:\\a\\otter\\it-tests\\test-app-eslint-config2';
const untouchedProject = 'D:\\a\\otter\\it-tests\\test-app-eslint-config2\\apps\\untouched-app';
path.relative(workspacePath, untouchedProject).replace(/\\+/g, '/')
// '../D:/a/otter/it-tests/test-app-eslint-config2/apps/untouched-app'
```
what I am looking for is 

```typescript
path.posix.relative(workspacePath.replace(/\\+/g, '/'), untouchedProject.replace(/\\+/g, '/'))
// 'apps/untouched-app'
```
I thought that the split + join on the separator was more elegant than the replace.

kpanot: > ```ts
> const workspacePath = 'D:\\a\\otter\\it-tests\\test-app-eslint-config2';
> const untouchedProject = 'D:\\a\\otter\\it-tests\\test-app-eslint-config2\\apps\\untouched-app';
> path.relative(workspacePath, untouchedProject).replace(/\\+/g, '/')
> // '../D:/a/otter/it-tests/test-app-eslint-config2/apps/untouched-app'
> ```

This is incorrect, on Windows OS it results to `apps/untouched-app`, what you expected.
(You can test locally with `path.win32.relative` on MacOS).

I don't mind to use `.join(path.sep).join(path.posix.sep)` instead `.replace(/\\+/g, '/')` (even if it is longer to write and less performant :D) as far this transformation is done after the relative path calculation.
(I am even not sure a new function is actually needed for a simple `path.relative(from, to).replace(/\\+/g, '/')` but ...)

kpanot (approved):

PR #3157, opened 2025-04-22
feature: testing generator starts local server for pw

vscaiceanu-1a:
## Proposed change

When adding Playwright via the Otter Testing generator, we should also start the local server

PR #3175, opened 2025-04-28
fix: do not start server for pw in int tests

vscaiceanu-1a:
## Proposed change

The Playwright server is already started via the config (https://github.com/AmadeusITGroup/otter/pull/3157) 
No need to start another server in the integration test.

### DIFF
--- a/packages/@o3r/core/schematics/index.it.spec.ts
+++ b/packages/@o3r/core/schematics/index.it.spec.ts
@@ -5,10 +5,6 @@
  */
 const o3rEnvironment = globalThis.o3rEnvironment;
 
-import {
-  execSync,
-  spawn,
-} from 'node:child_process';
 import * as path from 'node:path';
 import {
   addImportToAppModule,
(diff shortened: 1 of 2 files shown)


packages/@o3r/core/schematics/index.it.spec.ts
```diff
@@ -19,9 +15,7 @@ import {
   packageManagerInstall,
   packageManagerRunOnProject,
 } from '@o3r/test-helpers';
-import getPidFromPort from 'pid-from-port';
```
kpanot, line 22: Can you now remove this dep from almost everywhere?

vscaiceanu-1a: It's still used in https://github.com/AmadeusITGroup/otter/blob/main/packages/%40o3r/test-helpers/src/jest-environment/index.ts which is used in all int tests

fpaul-1A (approved):

packages/@o3r/core/schematics/index.it.spec.ts
vscaiceanu-1a: but I'll check in another PR if we can remove it from core at least. This PR is kind of urgent because the int tests are failing on main

mrednic-1A (approved):

PR #3182, opened 2025-04-30
feat: support a computed id for the memorize route in com protocol

mrednic-1A:
## Proposed change

This is just a proposal to have a computed id for the route to memorize (the moduleId + routeId)

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@ama-mfe/ng-utils/src/navigation/route-memorize/route-memorize.directive.ts
+++ b/packages/@ama-mfe/ng-utils/src/navigation/route-memorize/route-memorize.directive.ts
@@ -68,7 +68,7 @@ export class RouteMemorizeDirective {
       }
       const requested = requestedUrlSignal();
       const id = this.memorizeRouteId() || this.connect();
-      if (requested && id && requested.channelId === id) {
+      if (requested && id && requested.channelId && id.startsWith(requested.channelId)) {
         memory.memorizeRoute(id, requested.url, untracked(this.maxAge));
       }
     });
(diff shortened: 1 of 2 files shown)


packages/@ama-mfe/ng-utils/src/navigation/route-memorize/route-memorize.directive.ts
```diff
@@ -68,7 +68,7 @@ export class RouteMemorizeDirective {
       }
       const requested = requestedUrlSignal();
       const id = this.memorizeRouteId() || this.connect();
-      if (requested && id && requested.channelId === id) {
+      if (requested && id && requested.channelId && id.startsWith(requested.channelId)) {
```
cpaulve-1A, line 71: Could there be an overlap with the following:
-channelId1
- channelId12 ?

mrednic-1A: yes it can. I could not find a better way to be safer on the check because on the navigation message we have the channelId sending the message and the new url to navigate (without the base path ). I am open to ideas

mrednic-1A: the use-case I want to support is to have the same module app instance in 2 iframes (displaying 2 modules - a route for each module). So as this is one instance of the app, the channelId (moduleId) will be unique, so the identifier for the route memorized will be the same. that's why I tried to add a second part for the identifier (the base route)
like that I will have,
id1 = moduleId + module1BasePath 
id2 = moduleId + module2BasePath
and as I have only the channelId received in the message in this directive, I just did the startswith which I agree is not the best :)

mrednic-1A: thanks for the ideas. 
after internal discussions we manage to solve the usecase by having different module ids even if an instance of an app is used to serve 2 modules. 
here is the PR which covers that
https://github.com/AmadeusITGroup/otter/pull/3205

Issue #3047, opened 2025-03-25
[Bug]: Path replacement script breaking with node 20.19.0


cpaulve-1A:
### Package name

testing

### Package version

12.1

### Reproduction steps

`adjustPath` method from `packages/@o3r/testing/src/tools/path-replacement/path-replacement.ts` is failing after migration to node 20.19.0. 

This may be linked to a change introduce by node 20.19.0 around esm: https://nodejs.org/fr/blog/release/v20.19.0#requireesm-is-now-enabled-by-default.

### Current result

.

### Expected result

.

PR #3189, opened 2025-05-06
fix: adjust path replacement script to work with node 20.19.0

guillaumerochelle:
## Proposed change

Follow-up PR after it-tests breaking due to pnp infinite loop and maximum call stack exceeded

## Related issues

Fixes 🐛  #3047

### DIFF
--- a/.github/workflows/it-tests.yml
+++ b/.github/workflows/it-tests.yml
@@ -23,7 +23,7 @@ on:
 env:
   NX_PARALLEL: ${{ vars.NX_PARALLEL }}
   NX_TASK_TARGET_CONFIGURATION: ci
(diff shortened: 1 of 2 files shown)


.github/workflows/it-tests.yml
```diff
@@ -23,7 +23,7 @@ on:
 env:
   NX_PARALLEL: ${{ vars.NX_PARALLEL }}
   NX_TASK_TARGET_CONFIGURATION: ci
-  NX_SKIP_NX_CACHE: ${{ inputs.skipNxCache }}
+  NX_SKIP_NX_CACHE: true
```
kpanot, line 26: Why did you need this change?

PR #3194, opened 2025-05-08
feat(style-dictionary): add better sorting for variables

### LINKED ISSUE #3192: [Feature]: Better sorting for Style-Dictionary
### Context

The current Style-Dictionary CSS formatter provided by the `@o3r/style-dictionary` package is sorting the variable name based on an alphabetic order.

### Proposal

As we do use palettes with names following the pattern `<name>-<grade>`, where `<grade>` is a number, a better sorting based on both numeric and string should be applied.

#### Current result

```css
:root {
  --my-var-10: #000;
  --my-var-00: #000;
}
```

#### Expected result

```css
:root {
  --my-var-00: #000;
  --my-var-10: #000;
}
```

kpanot:
## Proposed change

feat(style-dictionary): add better sorting for variables

## Related issues

* :rocket: Feature resolves #3192

### DIFF
--- a/packages/@o3r/style-dictionary/src/formats/css.format.mts
+++ b/packages/@o3r/style-dictionary/src/formats/css.format.mts
@@ -9,6 +9,9 @@ import {
   OTTER_NAME_PREFIX,
 } from '../constants.mjs';
 import {
+  sortByPath,
+} from '../helpers/sort-by-path.sort.helpers.mjs';
+import {
   getDefaultCssFormatter,
 } from './css-formatters/default.formatter.mjs';
 import type {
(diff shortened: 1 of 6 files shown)


packages/@o3r/style-dictionary/src/formats/css.format.mts
```diff
@@ -44,17 +47,12 @@ export const cssFormat: Format = {
...
-      if (outputReferences) {
-        allTokens = [...allTokens].sort(
-          sortByReference(tokens, { unfilteredTokens: dictionary.unfilteredTokens, usesDtcg })
-        );
-      }
-
+      const { allTokens, tokens, unfilteredTokens } = dictionary;
       const propertyFormatter = getDefaultCssFormatter(baseFormatterOptions);
 
       return allTokens
```
cpaulve-1A, line 53: Is it okay to modify the initial token list ?

kpanot: I did not see issue but indeed, let be secure and keep the original copy

mrednic-1A (approved):

vscaiceanu-1a (approved):

cpaulve-1A (approved):

PR #3200, opened 2025-05-12
fix: temporary rollback of @swc/cli due to error since version 0.7.4

sdo-1A:
## Proposed change

Fallback the version of @swc/cli since it started failing in v0.7.5
This is a temporary fix while waiting for the fix on @swc/cli

## Related issues

https://github.com/swc-project/pkgs/issues/103
Follow-up on our side once the fix is implemented in @swc/cli: https://github.com/AmadeusITGroup/otter/issues/3201

*- No issue associated -*


vscaiceanu-1a (approved):

kpanot (approved):

kpanot:
Can you add a issue on our side to follow it and unpin the dependency when fixed?

sdo-1A:
@kpanot here is the issue on our side: https://github.com/AmadeusITGroup/otter/issues/3201

PR #3205, opened 2025-05-13
feat: embedded module id handled by the host app

mrednic-1A:
## Proposed change

Allow the host app to provide the moduleId for a module to embed, in this way the module does not need to provide its id for the communication protocol.

## Related issues

*- No issue associated -*


packages/@ama-mfe/ng-utils/src/connect/connect.providers.ts
```diff
@@ -23,12 +24,12 @@ import {
  * Provide the communication protocol connection configuration
  * @param connectionConfig The identifier of the application in the communication protocol ecosystem plus the types of messages able to exchange
  */
-export function provideConnection(connectionConfig: ConnectionConfig) {
+export function provideConnection(connectionConfig?: ConnectionConfig) {
   persistHostInfo();
   const config: MessagePeerConfig = {
-    id: connectionConfig.id,
+    id: connectionConfig?.id || getHostInfo().moduleApplicationId,
```
kpanot, line 30: It should not be the opposite and be something like `getHostInfo().moduleApplicationId || connectionConfig?.id` because otherwise, if the application decide of its name, the set at host level will never be used :S.

packages/@ama-mfe/ng-utils/src/host-info/host-info.ts
```diff
@@ -24,6 +29,11 @@ export interface MFEHostInformation {
    * ID of the host application
    */
   hostApplicationId?: string;
+
+  /**
+   * ID of the module to embed defined at host level
+   */
+  moduleApplicationId: string;
```
kpanot, line 36: Why is it mandatory? I would expect to provide module name only when required to differentiate modules when provided to a single app

packages/@ama-mfe/ng-utils/src/host-info/host-info.pipe.ts
```diff
@@ -28,10 +29,10 @@ export class HostInfoPipe implements PipeTransform {
...
    * @returns - The transformed SafeResourceUrl or undefined if the input URL is invalid.
    */
-  public transform(url: string, applicationId: string): string;
-  public transform(url: SafeResourceUrl, applicationId: string): SafeResourceUrl;
-  public transform(url: undefined, applicationId: string): undefined;
-  public transform(url: string | SafeResourceUrl | undefined, applicationId: string): string | SafeResourceUrl | undefined {
+  public transform(url: string, options: { hostId: string; moduleId: string }): string;
+  public transform(url: SafeResourceUrl, options: { hostId: string; moduleId: string }): SafeResourceUrl;
+  public transform(url: undefined, options: { hostId: string; moduleId: string }): undefined;
+  public transform(url: string | SafeResourceUrl | undefined, options: { hostId: string; moduleId: string }): string | SafeResourceUrl | undefined {
```
kpanot, line 35: This is a breaking change and turn the `moduleId` to mandatory.
I would expect a parameter of type `options: string | { hostId: string; moduleId?: string }`

packages/@ama-mfe/ng-utils/src/host-info/host-info.ts
mrednic-1A: I will change it to optional and do a check in the provide connection to see that it is provided.
The way I have seen it is that all the apps will provide the module name like this, because it will be done automatically by the hostInfo pipe. If you dont use the pipe then you can pass it when providing the connection.

packages/@ama-mfe/ng-utils/src/host-info/host-info.pipe.ts
mrednic-1A: It was added in this version 12.3.0-prerelease. we are not breaking (we are but between prereleases)

packages/@ama-mfe/ng-utils/src/connect/connect.providers.ts
mrednic-1A: I see, ok for me

packages/@ama-mfe/ng-utils/src/host-info/host-info.pipe.ts
mrednic-1A: I changed the moduleId to optional as it is not necessary to be provided. If the host will not provide it, it will be the module which will provide it. I agree to not force the host to provide the moduleId.

packages/@ama-mfe/ng-utils/src/navigation/navigation.consumer.service.ts
```diff
@@ -95,7 +96,7 @@ export class NavigationConsumerService implements MessageConsumer<NavigationMess
   private navigate(url: string, channelId?: string) {
     const { paths, queryParams } = this.parseUrl(url);
     // No need to keep these in the URL
-    [MFE_HOST_URL_PARAM, MFE_HOST_APPLICATION_ID_PARAM].forEach((key) => delete queryParams[key]);
+    [MFE_HOST_URL_PARAM, MFE_HOST_APPLICATION_ID_PARAM, MFE_MODULE_APPLICATION_ID_PARAM].forEach((key) => delete queryParams[key]);
```
kpanot, line 99: Could make sense to move this array to `host-info` to not forget to include the next one.

kpanot (approved):

vscaiceanu-1a (approved):

sdo-1A (approved):

PR #3221, opened 2025-05-20
feat: make o3r testing optional for o3r application

### LINKED ISSUE #2668: [Bug]: @o3r/testing should not be installed by default for generated application
### Package name

workspace

### Package version

11.*

### Reproduction steps

Add @o3r/workspace to a project and generate an application with `ng g application`

### Current result

A prompt informs the user `@o3r/testing` will be installed and asks the user if they want a jest setup.

### Expected result

This dependency should not be mandatory as it is only used to activate the visual testing masks from the chrome extension.. 
`@o3r/testing` should be optional in `@o3r/application` and required at runtime.

### Additional comments

_No response_

vscaiceanu-1a:
## Proposed change

Non breaking change alternative to #3220
Testing as optional dependency for application and dynamic import for testing in the Message Service.

## Related issues

* :bug: Fix resolves #2668


packages/@o3r/core/schematics/index.it.spec.ts
```diff
@@ -118,9 +113,5 @@ describe('new otter application', () => {
     expect(diff.modified).toContainEqual(expect.stringMatching(new RegExp(path.posix.join(relativeApplicationPath, 'src/app/app.routes.ts').replace(/[/\\]+/g, '[\\\\/]'))));
 
     expect(() => packageManagerRunOnProject(appName, isInWorkspace, { script: 'build' }, execAppOptions)).not.toThrow();
-
-    packageManagerExecOnProject(appName, isInWorkspace, { script: 'playwright', args: ['install', '--with-deps'] }, execAppOptions);
-    expect(() => packageManagerRunOnProject(appName, isInWorkspace, { script: 'test:playwright' }, execAppOptions)).not.toThrow();
-    expect(() => packageManagerRunOnProject(appName, isInWorkspace, { script: 'test:playwright:sanity' }, execAppOptions)).not.toThrow();
```
vscaiceanu-1a, line 124: first, these shouldn't be in `core` and second, these generators will be removed in https://github.com/AmadeusITGroup/otter/pull/3141

sdo-1A (approved):

PR #3229, opened 2025-05-21
feat(ama-mfe): add `provideDisableHistoryWrites()` provider

maxokorokov:
A provider that turns off common History modification for an app running inside an iframe


packages/@ama-mfe/ng-utils/src/history/history.providers.ts
```diff
@@ -0,0 +1,22 @@
...
+export function noop() {}
+
+/**
+ * If an iframe is not sandboxed or is of the same origin, navigation inside it will mess up the main window history.
+ * This disables writing to and overriding history API from inside the iframe to prevent this.
+ * Theoretically, this might break applications that rely on reading history API from inside the iframe.
+ *
+ * This should also allow having `CustomPathLocationStrategy` in the iframe if necessary.
+ */
+export function provideDisableHistoryWrites() {
```
kpanot, line 17: I think you may need to add documentation in the readme regarding it as well.

packages/@ama-mfe/ng-utils/src/history/history.providers.ts
```diff
@@ -0,0 +1,22 @@
...
+  provideAppInitializer,
+} from '@angular/core';
+
+/**
+ * No operation function used to override the history API methods.
+ */
+export function noop() {}
+
+/**
+ * If an iframe is not sandboxed or is of the same origin, navigation inside it will mess up the main window history.
```
kpanot, line 11: Can these conditions be detected from inside the iframe?
If so, do you think it could make sense to disable history write systematically?

Because I imagine that an application can not always, at implementation time, if it needs to disable it or not :S

packages/@ama-mfe/ng-utils/src/history/history.providers.ts
```diff
@@ -0,0 +1,22 @@
+import {
+  provideAppInitializer,
+} from '@angular/core';
+
+/**
+ * No operation function used to override the history API methods.
+ */
+export function noop() {}
```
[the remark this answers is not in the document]
maxokorokov: It is not exported publicly, just for unit testing

packages/@ama-mfe/ng-utils/src/history/history.providers.ts
maxokorokov: They can't know from the inside - we tell them they're running in embedded mode.
It is up to us externally to decide - if iframe is `allow-same-origin`, we have to disable their history. If not - we don't care frankly.

Anyway this logic probably should be outside / configurable for me.

packages/@ama-mfe/ng-utils/src/history/history.providers.ts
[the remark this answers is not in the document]
kpanot: The export is propagated via `index.ts` and exposed outside by `public_api.ts`.
If I understand correctly, it may need to not be exported by one of them

packages/@ama-mfe/ng-utils/src/history/history.providers.ts
matthieu-crouzet, line 17: Can't we have something more generic to be future proof ?
```typescript
export type HistoryWritesOptions = Partial<Keep<History, 'pushState' | 'replaceState'>>;
export function provideHistoryWrites(options: HistoryWritesOptions = {}) {
  return provideAppInitializer(() => {
    Object.defineProperty(history, 'pushState', options?.pushState ?? history.pushState);
    Object.defineProperty(history, 'replaceState', options?.replaceState ?? history.replaceState);
  });
}
```
we may want to expose
```typescript 
export const withDisableOptions = () => {
  pushState: { value: noop, writable: false, configurable: false },
  replaceState: { value: noop, writable: false, configurable: false }
} as const satisfies HistoryWritesOptions;
```
so it can be used like this 
```typescript
provideHistoryWrites(withDisableOptions());
```
> [!NOTE]
> You can change the naming if you find something more relevant

mrednic-1A, line 17: could we call this function in the `provideConnection()`, to avoid making the modules to do it separately ?
https://github.com/AmadeusITGroup/otter/blob/51b628e914d1857fabd81f3d2ac74decf3584ef9/packages/%40ama-mfe/ng-utils/src/connect/connect.providers.ts#L38

maxokorokov: I didn't put it there just because it is not related to "connection", but it would be easier I guess to have 1 thing for modules to do

maxokorokov: I don't know. Personally not a fan of making things generic, especially when this most likely be hidden from end user by some setup logic. They probably shouldn't call this directly as mentioned here indeed → https://github.com/AmadeusITGroup/otter/pull/3229#discussion_r2101784516

Maybe just not make it public at all for now, stick it into `provideConnection` and do configurable thing if we ever need to configure this?

maxokorokov: In the end added it to `provideConnection()` as it is the only exising API where you could add app level initalizer

kpanot (approved):

packages/@ama-mfe/ng-utils/src/history/index.ts
```diff
@@ -0,0 +1 @@
+export { provideDisableHistoryWrites } from './history.providers';
```
kpanot, line 1: This file should not be exported by `public_api`?

maxokorokov: no, why? `provideDisabledHistoryWrites` stays internal API for now

PR #3230, opened 2025-05-21
feat(ama-mfe): handle `NavigationMessage` consumption by `RoutingService`

maxokorokov:
`RoutingService` used by embedded modules now both produces and consumes Navigation messages:

- when a module receives a message - it triggers internal Routing navigation
- when internal Routing navigation is finished - it sends a message to notify others

### DIFF
--- a/packages/@ama-mfe/ng-utils/eslint.local.config.mjs
+++ b/packages/@ama-mfe/ng-utils/eslint.local.config.mjs
@@ -32,7 +32,7 @@ export default [
   },
   {
     name: '@ama-mfe/ng-utils/messages-definition',
-    files: ['**/*.versions.ts', '**/*.consumer.service.ts'],
+    files: ['**/*.versions.ts', '**/*.consumer.service.ts', '**/routing.service.ts'],
     rules: {
       '@typescript-eslint/naming-convention': [
         'error',
--- a/packages/@ama-mfe/ng-utils/src/managers/utils.ts
+++ b/packages/@ama-mfe/ng-utils/src/managers/utils.ts
@@ -2,7 +2,11 @@ import {
   DestroyRef,
   inject,
 } from '@angular/core';
+import {
+  ConsumerManagerService,
+} from './consumer.manager.service';
 import type {
+  MessageConsumer,
   MessageProducer,
 } from './interfaces';
 import {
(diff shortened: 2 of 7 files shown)


packages/@ama-mfe/ng-utils/src/managers/utils.ts
```diff
@@ -22,3 +26,17 @@ export const registerProducer = (producer: MessageProducer) => {
     producerManagerService.unregister(producer);
   });
 };
+
+/**
+ * Method to call in the constructor of a consumer
+ * @note should be used in injection context
+ * @param consumer
+ */
+export const registerConsumer = (consumer: MessageConsumer) => {
```
kpanot, line 35: *[Not a blocking question]*

As this file is importing both `ConsumerManagerService` and `ProducerManagerService` now. Is the tree-shaker clever enough to not bring one (or the other) if only one of these both *register* function is called?

packages/@ama-mfe/ng-utils/eslint.local.config.mjs
```diff
@@ -32,7 +32,7 @@ export default [
   },
   {
     name: '@ama-mfe/ng-utils/messages-definition',
-    files: ['**/*.versions.ts', '**/*.consumer.service.ts'],
+    files: ['**/*.versions.ts', '**/*.consumer.service.ts', '**/routing.service.ts'],
```
kpanot, line 35: *[open question]*
Should we have a naming convention specific for services extending both interfaces?

packages/@ama-mfe/ng-utils/src/navigation/routing.service.ts
```diff
@@ -68,10 +73,36 @@ export class RoutingService implements MessageProducer<NavigationMessage> {
...
+   */
+  public readonly type = 'navigation';
+
+  /**
+   * Use the message payload to navigate to the specified URL.
+   * @param message message to consume
+   */
+  public readonly supportedVersions = {
+    '1.0': (message: RoutedMessage<any>) => {
+      void this.router.navigateByUrl(message.payload.url);
```
kpanot, line 87: why voiding it instead of `await` (or `return`) ?

matthieu-crouzet: I'm not sure we want to be waiting for the execution of a message no ?
We just want a message to trigger its logic and continue its execution no ?

kpanot: I can agree but I think it is not directly related.
The promise returned by the function is not awaited to run something else (or potentiellement logging in the future as guess).
I am not sure that voiding any async function called in this callback is a good practice :S.

packages/@ama-mfe/ng-utils/eslint.local.config.mjs
maxokorokov: too much config for me. No idea, `prosumer` maybe? haha :D

packages/@ama-mfe/ng-utils/src/managers/utils.ts
maxokorokov: I think they should be smart enough, plus if this as an Angular lib, it should be distributed in FESM format, so it's a one big file in the end + `sideEffects: false` in package json usually. 

Ex: https://unpkg.com/@ama-mfe/ng-utils@12.2.5/fesm2022/ama-mfe-ng-utils.mjs

matthieu-crouzet (approved):

kpanot (approved):

PR #3243, opened 2025-05-29
feat: caching for e2e tests

kpanot:
## Proposed change

feat: caching for e2e tests

## Related issues

*- No issue associated -*

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -36,8 +36,8 @@
       "{projectRoot}/tsconfig.build.json",
       "{projectRoot}/tsconfig.build.*.json",
       "{projectRoot}/src/**/*",
-      "!{projectRoot}/**/*.spec.ts",
-      "!{projectRoot}/**/*.spec.scss"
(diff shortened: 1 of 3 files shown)


matthieu-crouzet (approved):

nx.json
```diff
@@ -240,6 +240,19 @@
...
+    "test-e2e": {
+      "executor": "nx:run-script",
+      "options": {
+        "script": "test:playwright"
+      },
+      "inputs": [
+        "source",
+        "schematics",
+        "assets",
+        "^schematics"
```
fpaul-1A, line 252: we may want to add a new `namedInput` for e2e, not all the files inside `e2e-playwright` folder are matched with the current list set up here (kassette config, screenshots, mocks)
also I think the `"!{projectRoot}/**/*.spec.ts"` in the `source` will exclude all the spec files, even if added in an other group

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #3250, opened 2025-06-04
fix: added deps in int tests

vscaiceanu-1a:
## Proposed change
Adding `generatorDependencies` not to rely on peerdeps that, in some cases, provoke miss-alignment.
```json
    "@angular/cdk": "~19.0.0",
    "@angular/common": "~19.2.0",
    "@angular/compiler": "~19.2.0",
    "@angular/core": "~19.2.0",
    "@angular/forms": "~19.2.0",
    "@angular/material": "~19.0.0",
```
This leads to  install failures when adding o3r packages.


kpanot:
Not sure it is a change you can do on a `patch`.
It will break the application remained on Angular `>= 19.0.0 < 19.2.0`.

vscaiceanu-1a:
> Not sure it is a change you can do on a `patch`. It will break the application remained on Angular `>= 19.0.0 < 19.2.0`.

In this case I will try going through `generatorDependencies` to fix the problem

PR #3268, opened 2025-06-13
feat: in range time operator

vscaiceanu-1a:
## Related issues

*- No issue associated -*


kpanot (dismissed):

packages/@o3r/rules-engine/src/engine/operator/operators/time-based.operators.ts
```diff
@@ -0,0 +1,34 @@
+import {
+  isValidTimeInput,
+  isValidTimeRange,
+} from '../operator.helpers';
+import {
+  Operator,
+} from '../operator.interface';
+
+/**
+ * Check if the current time is between the specified start and end times
```
cpaulve-1A, line 10: Current time may be misleading, I would expect the currentDateTime basefact to be used here when actually any date can be checked

packages/@o3r/rules-engine/src/engine/operator/operators/time-based.operators.ts
```diff
@@ -0,0 +1,34 @@
...
+    const parseTime = (timeStr: string) => {
+      const [hours, minutes] = timeStr.split(':').map(Number);
+      return hours * 60 + minutes;
+    };
+
+    const currentMinutes = parseTime(currentTime);
+    const startMinutes = parseTime(from);
+    const endMinutes = parseTime(to);
+
+    return startMinutes <= endMinutes ? currentMinutes >= startMinutes && currentMinutes <= endMinutes : currentMinutes >= startMinutes || currentMinutes <= endMinutes;
```
cpaulve-1A, line 25: In the description, we could mention we support the use case where the time range crosses midnight?

cpaulve-1A, line 25: In the description, we could mention we support the use case where the time range crosses midnight?

cpaulve-1A (approved):

cpaulve-1A (approved):

PR #3284, opened 2025-06-19
chore(deps): update ts-jest dependency and add jest-util dependency

sdo-1A:
## Proposed change

add dependency to `jest-util` to fix the update of `ts-jest` to version `~29.4.0` in #3274 
https://github.com/kulshekhar/ts-jest/blob/v29.4.0/package.json

UPDATE:
updated to PR to update `ts-jest` to version `~29.4.0` and add dependency to `jest-util`
note: `jest` does not need to be updated since it is already set to the latest minor of v29 (`~29.7.0`) - the update of the major dependencies will be done in v13 of otter
this will close PR #3274 

## Related issues

*- No issue associated -*


apps/chrome-devtools/package.json
```diff
@@ -75,6 +75,7 @@
     "jest-environment-jsdom": "~29.7.0",
     "jest-junit": "~16.0.0",
     "jest-preset-angular": "~14.5.0",
+    "jest-util": "~30.0.0",
```
fpaul-1A, line 78: is it ok to use `jest-util@30` while keeping `jest@29` ?

sdo-1A: I think so since `ts-jest` v29.4.0 uses `jest@30`
https://github.com/kulshekhar/ts-jest/blob/v29.4.0/package.json#L132

sdo-1A: the goal of this PR was to fix the renovate PR #3274 
should we ignore it and update all 3 packages in this PR ?

fpaul-1A: I agree with Kilian, I think it's better to upgrade all the jest dependencies at once even if the peer dependencies seem to accept the discrepancies.
Renovate won't get mad and will kindly close the PR.

kpanot (dismissed):

kpanot (approved):

kpanot (approved):

matthieu-crouzet (approved):

PR #3291, opened 2025-06-24
feat: redirect if failed fetch plugin

mrednic-1A:
## Proposed change

Plugin to redirect a fetch call to a fallback url in case of original call fails.

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@ama-sdk/client-fetch/src/plugins/redirect-if-failed/readme.md
+++ b/packages/@ama-sdk/client-fetch/src/plugins/redirect-if-failed/readme.md
@@ -1,0 +1,4 @@
+## Redirect If Failed
+
+Plugin to redirect a fetch call to a fallback url in case of original call fails. The redirection is done via a GET request and only the headers are forwarded from the context. The fetch plugins existing in the context are applied too, on the fallback call.
+
--- a/packages/@ama-sdk/client-fetch/src/plugins/redirect-if-failed/redirect-if-failed.fetch.ts
+++ b/packages/@ama-sdk/client-fetch/src/plugins/redirect-if-failed/redirect-if-failed.fetch.ts
@@ -1,0 +1,63 @@
+import type {
+  FetchCall,
+  FetchPlugin,
+  FetchPluginContext,
+} from '../../fetch-plugin';
+
+export type RedirectIfFailedPluginContext = FetchPluginContext & { redirectPlugins?: Set<RedirectIfFailedFetch> };
+
+/**
+ * Plugin to Redirect to another fetch call
+ */
+export class RedirectIfFailedFetch implements FetchPlugin {
+  /** The url to fallback when initial call fails */
+  public readonly fallbackUrl: string;
+
+  /**
+   * RedirectIfFailed Fetch plugin
+   * @param fallbackUrl
+   */
+  constructor(fallbackUrl: string) {
+    this.fallbackUrl = fallbackUrl;
+  }
+
+  /**
+   * Launch a redirect using the fallback url from context
+   * @param context
+   */
+  private redirect(context: RedirectIfFailedPluginContext) {
+    if (context.redirectPlugins?.has(this)) {
+      throw new Error(`Redirection loop detected: ${this.fallbackUrl} already visited.`);
+    }
+    context.redirectPlugins ||= new Set();
+    context.redirectPlugins.add(this);
+
+    return context.fetchPlugins
+      .reduce((asyncResponse, plugin) => plugin.transform(asyncResponse), fetch(context.url, { headers: context.options?.headers }));
+  }
+
+  /** @inheritDoc */
+  public load(context: RedirectIfFailedPluginContext) {
+    return {
+      transform: async (fetchCall: FetchCall) => {
+        try {
+          const result = await fetchCall;
+          if (!result.ok && !context.redirectPlugins?.has(this)) {
+            context.url = this.fallbackUrl;
+            return this.redirect(context);
+          }
+          return result;
+        } catch (error) {
+          if (!context.redirectPlugins?.has(this)) {
+            context.url = this.fallbackUrl;
+            return this.redirect(context);
+          }
+          if (context.redirectPlugins?.has(this)) {
+            throw new Error(`Redirection loop detected: ${this.fallbackUrl} already visited.`);
+          }
+          throw error;
+        }
+      }
+    };
+  }
+}
--- a/packages/@ama-sdk/client-fetch/src/plugins/index.ts
+++ b/packages/@ama-sdk/client-fetch/src/plugins/index.ts
@@ -6,3 +6,4 @@ export * from './perf-metric/index';
 export * from './retry/index';
 export * from './timeout/index';
 export * from './wait-for/index';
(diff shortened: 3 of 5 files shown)


packages/@ama-sdk/client-fetch/src/plugins/redirect-if-failed/redirect-if-failed.fetch.ts
```diff
@@ -0,0 +1,55 @@
...
+        } catch (error) {
+          if (context.url !== this.fallbackUrl) {
+            context.url = this.fallbackUrl;
+            return this.redirect(context);
+          }
+          throw error;
+        }
+      }
+    };
+  }
```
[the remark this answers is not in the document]
mrednic-1A: j'ai fait les updates :)

fpaul-1A (approved):

packages/@ama-sdk/client-fetch/src/plugins/redirect-if-failed/readme.md
```diff
@@ -0,0 +1,5 @@
+## Redirect If Failed
+
+Plugin to redirect a fetch call to a fallback url in case of original call fails.
+
+> ⚠️ **Warning:** If two or more instances of the plugin are loaded simultaneously on the same API, with different fallback URLs, the plugin may enter an infinite loop of fallback URL requests if the fallback URLs calls fail.
```
kpanot, line 5:
```suggestion
> [!WARNING]
> ⚠️ If two or more instances of the plugin are loaded simultaneously on the same API, with different fallback URLs, the plugin may enter an infinite loop of fallback URL requests if the fallback URLs calls fail.
```

packages/@ama-sdk/client-fetch/src/plugins/redirect-if-failed/readme.md
```diff
@@ -0,0 +1,5 @@
+## Redirect If Failed
+
+Plugin to redirect a fetch call to a fallback url in case of original call fails.
```
kpanot, line 3: You may need to specify that the redirection is done via a `GET` request?

packages/@ama-sdk/client-fetch/src/plugins/redirect-if-failed/readme.md
kpanot, line 5: I you want, there is a way to avoid infinite loop:
You can add a `X-....` header when doing the redirection,
This header will contain the list of the stack trace of the redirections and if you do not fallback twice to the same URL (throw an error instead)

mrednic-1A: I've done an update :)

mrednic-1A: I removed the warning as I implemented the other suggestion to avoid the infinite loop

packages/@ama-sdk/client-fetch/src/plugins/redirect-if-failed/redirect-if-failed.fetch.ts
```diff
@@ -0,0 +1,64 @@
...
+      context.redirectPlugins.push(this);
+    }
+
+    return context.fetchPlugins
+      .reduce((asyncResponse, plugin) => plugin.transform(asyncResponse), fetch(context.url, { headers: context.options?.headers }));
+  }
+
+  /** @inheritDoc */
+  public load(context: FetchPluginContext & { redirectPlugins?: RedirectIfFailedFetch[] }) {
+    const loadedPlugin = {
```
kpanot, line 38:
```suggestion
    return {
```

packages/@ama-sdk/client-fetch/src/plugins/redirect-if-failed/redirect-if-failed.fetch.ts
```diff
@@ -0,0 +1,64 @@
...
+  constructor(fallbackUrl: string) {
+    this.fallbackUrl = fallbackUrl;
+  }
+
+  /**
+   * Launch a redirect using the fallback url from context
+   * @param context
+   */
+  private redirect(context: FetchPluginContext & { redirectPlugins?: RedirectIfFailedFetch[] }) {
+    context.redirectPlugins ||= [];
```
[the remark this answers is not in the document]
mrednic-1A: no need to throw in redirect as the detection of redoing a call is handled in the transform under certain conditions :)

packages/@ama-sdk/client-fetch/src/plugins/redirect-if-failed/redirect-if-failed.fetch.ts
```diff
@@ -0,0 +1,64 @@
...
+          if (!result.ok && !context.redirectPlugins?.includes(this)) {
+            if (context.redirectPlugins?.includes(this)) {
+              throw new Error(`Redirection loop detected: ${this.fallbackUrl} already visited.`);
+            }
+            context.url = this.fallbackUrl;
+            return this.redirect(context);
+          }
+          return result;
+        } catch (error) {
+          if (!context.redirectPlugins?.includes(this)) {
```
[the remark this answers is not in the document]
mrednic-1A: same as above

packages/@ama-sdk/client-fetch/src/plugins/redirect-if-failed/redirect-if-failed.fetch.ts
```diff
@@ -0,0 +1,63 @@
...
+            return this.redirect(context);
+          }
+          return result;
+        } catch (error) {
+          if (!context.redirectPlugins?.has(this)) {
+            context.url = this.fallbackUrl;
+            return this.redirect(context);
+          }
+          if (context.redirectPlugins?.has(this)) {
+            throw new Error(`Redirection loop detected: ${this.fallbackUrl} already visited.`);
```
cpaulve-1A, line 56: Isn't it already handled in this.redirect?

packages/@ama-sdk/client-fetch/src/plugins/redirect-if-failed/redirect-if-failed.fetch.ts
```diff
@@ -0,0 +1,63 @@
...
+} from '../../fetch-plugin';
+
+export type RedirectIfFailedPluginContext = FetchPluginContext & { redirectPlugins?: Set<RedirectIfFailedFetch> };
+
+/**
+ * Plugin to Redirect to another fetch call
+ */
+export class RedirectIfFailedFetch implements FetchPlugin {
+  /** The url to fallback when initial call fails */
+  public readonly fallbackUrl: string;
```
cpaulve-1A, line 14: The plugin is applied to all the endpoint of an API, is it okay to have a single fallbackUrl? Should we define a map between endpoints and their own fallback (with * to apply to all the endpoints)?

cpaulve-1A: Today you apply the plugin to all the endpoint of the API so you can't filter out POST calls even if it is specified it is not supported

PR #3348, opened 2025-07-18
fix: harmonize script failing on windows

cpaulve-1A:
Remove single quotes not supported by windows

We do not use the git ls-files command as bash script are not well supported on npm

## Related issues

*- No issue associated -*


packages/@o3r/eslint-config/schematics/ng-add/index.ts
```diff
@@ -109,7 +109,7 @@ const addHarmonizeScript = (): Rule => {
       return tree;
     }
     const postInstall = rootPackageJsonObject.scripts.postinstall;
-    rootPackageJsonObject.scripts.harmonize = `eslint '**/package.json' ${isYarnPackageManager ? '.yarnrc.yml ' : ''}--quiet --fix --no-error-on-unmatched-pattern`;
+    rootPackageJsonObject.scripts.harmonize = `eslint "**/package.json" ${isYarnPackageManager ? '.yarnrc.yml ' : ''}--quiet --fix --no-error-on-unmatched-pattern`;
```
fpaul-1A, line 112: don't you need to escape the doublequotes ?
```suggestion
    rootPackageJsonObject.scripts.harmonize = `eslint \\"**/package.json\\" ${isYarnPackageManager ? '.yarnrc.yml ' : ''}--quiet --fix --no-error-on-unmatched-pattern`;
```

kpanot: It is not already escaped by the `JSON.stringify()`?

PR #3385, opened 2025-07-30
fix: align tsconfig moduleResolution to bundler

### LINKED ISSUE #3340: [Feature]: Set the module resolution to `bundler` in the global `tsconfig.build.json`
### Context

In the update of the major versions, many `moduleResolution` properties in the files `tsconfig.build.json` of the packages were updated from `node` to `bundler`. Here is the documentation on the module resolutions: https://www.typescriptlang.org/tsconfig/#moduleResolution

### Proposal

Set the `moduleResolution` to `bundler` in the global `tsconfig.build.json` and for exceptions (like packages that require `"module": "CommonJS"`), set the `moduleResolution` to `node`.

cpaulve-1A:
## Proposed change

Only use other moduleResolution if module is node or commonjs:

## Related issues

* :bug: Fix resolves #3340

### DIFF
--- a/apps/github-cascading-app/tsconfig.build.json
+++ b/apps/github-cascading-app/tsconfig.build.json
@@ -8,6 +8,7 @@
     "composite": true,
     "incremental": true,
     "module": "commonjs",
+    "moduleResolution": "node10",
     "tsBuildInfoFile": "build/tsconfig.build.tsbuildinfo"
   },
   "include": [
--- a/tools/github-actions/audit/tsconfig.build.json
+++ b/tools/github-actions/audit/tsconfig.build.json
@@ -4,6 +4,7 @@
     "types": ["node"],
     "target": "ES2022",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */
     "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
+    "moduleResolution": "node10",
     "outDir": "./tmp",                        /* Redirect output structure to the directory. */
     "rootDir": "./src",                       /* Specify the root directory of input files. Use to control the output directory structure with --outDir. */
     "strict": true,                           /* Enable all strict type-checking options. */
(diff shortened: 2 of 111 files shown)


apps/github-cascading-app/tsconfig.build.json
```diff
@@ -8,6 +8,7 @@
     "composite": true,
     "incremental": true,
     "module": "commonjs",
+    "moduleResolution": "node",
```
kpanot, line 11: Not sure this one is required

package.json
```diff
@@ -301,6 +301,9 @@
     "monaco-editor": {
       "unplugged": true
     },
+    "ng-packagr@20.0.1": {
+      "unplugged": true
```
kpanot, line 305: Why is it needed?

tools/github-actions/audit/tsconfig.build.json
```diff
@@ -4,6 +4,7 @@
     "types": ["node"],
     "target": "ES2022",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */
     "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
+    "moduleResolution": "node",
```
kpanot, line 7: is it required for ncc compilation?

package.json
cpaulve-1A: Not needed, this is a leftover of an investigation on ng-packagr issue.

tools/github-actions/audit/tsconfig.build.json
cpaulve-1A: image
Yes as the module is commonjs.

apps/github-cascading-app/tsconfig.build.json
cpaulve-1A: image
It is as this is commonjs. If we are to review the module strategy and remove non es modules, we probably need to do it in another user story as a feature.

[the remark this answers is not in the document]
matthieu-crouzet: Why not moving to node16 or nodenext ?

kpanot (approved):

PR #3394, opened 2025-08-01
fix(localization): issue when a translation is confure() call

kpanot:
## Proposed change

fix(localization): issue when a translation is listen before the first bundle download

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@o3r/localization/src/tools/localization.service.ts
+++ b/packages/@o3r/localization/src/tools/localization.service.ts
@@ -16,6 +16,7 @@ import {
 import {
   BehaviorSubject,
   combineLatest,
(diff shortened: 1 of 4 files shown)


packages/@o3r/localization/src/tools/localization.service.ts
```diff
@@ -155,11 +158,11 @@ export class LocalizationService {
   /**
    * Configures TranslateService and registers locales. This method is called from the application level.
    */
-  public configure() {
+  public async configure() {
```
matthieu-crouzet, line 161: Isn't it a breaking change to return a Promise now ?

PR #3415, opened 2025-08-14
fix(mcp): fix mcp js import

matthieu-crouzet:


tools/@o3r/mcp/src/cli/index.ts
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 import {
   StdioServerTransport,
-} from '@modelcontextprotocol/sdk/server/stdio';
+} from '@modelcontextprotocol/sdk/server/stdio.js';
```
kpanot, line 4: Does it requires the `allowJs` option to tsconfig?

kpanot (approved):

PR #3416, opened 2025-08-14
feat: remove deprecated o3r/eslint-config-otter package

cpaulve-1A:
## Proposed change
Stylistic and unicorn dropped the support of legacy config in favor of flat config
For this reason, we drop the support for the otter package based on legacy config

## Related issues

*- No issue associated -*


packages/@o3r/core/README.md
```diff
@@ -114,7 +114,7 @@ This preset does not add any additional modules.
 
 This preset is adding recommended Otter modules and features. This preset is the default one.
 
-* [@o3r/eslint-config](https://www.npmjs.com/package/@o3r/eslint-config) (or [@o3r/eslint-config-otter](https://www.npmjs.com/package/@o3r/eslint-config-otter) for legacy configuration)
+* [@o3r/eslint-config](https://www.npmjs.com/package/@o3r/eslint-config) (or [@o3r/eslint-config](https://www.npmjs.com/package/@o3r/eslint-config-otter) for legacy configuration)
```
matthieu-crouzet, line 117:
```suggestion
* [@o3r/eslint-config](https://www.npmjs.com/package/@o3r/eslint-config) 
```

packages/@o3r/rules-engine/package.json
```diff
@@ -198,7 +198,7 @@
       "@o3r/components",
       "@o3r/configuration",
       "@o3r/dynamic-content",
-      "@o3r/eslint-config-otter",
+      "@o3r/eslint-config",
```
matthieu-crouzet, line 201: Why do we have a packageGroup on @o3r/rules-engine ?
Maybe @kpanot you know why

kpanot: It was decided with the argument that the Rules Engine can be used in non-otter application (it was a request by a team, do not remember who...). As the Rule Engines has modular action handlers from other packages from Otter, the argument was that for the people who install the Rules Engines outside from Otter Project and try to update it, we align the update of packages exposing action handlers.

(The decision taken does not cover `@o3r/eslint-config` package which does not seem to have reason to be part of them)

fpaul-1A (approved):

PR #3445, opened 2025-08-25
fix: add check for navigation entry if load event was fired prematurely

### LINKED ISSUE #3444: [Bug]: First load in analytics not triggered
### Package name

analytics

### Package version

12.2.x

### Reproduction steps

When chaining the Angular bootstrap process with something else (like a config fetch that is required to start the app), the first `'load'` event is never registered. This has a cascading effect on other services that rely on the performance markers.

In the `src/index.ts` of Angular precede the bootstrap process with a process that takes a couple of milliseconds:
```typescript
performFetchOfConfig()
  .then(() => platformBrowserDynamic().bootstrapModule(AppModule, { ngZoneEventCoalescing: true }))
```
This leads to an load event that is triggered outside of the bootstrap process of Angular.

### Current result

No `'load'` event is emitted during or after the bootstrap process, thereby no `firstLoad` property is emitted.

### Expected result

Missed `'load'` event should be taken into account.

### Additional comments

_No response_

ntrotner:
## Proposed change
When a page is loaded without having Angular initialized, the load event is missed by the `EventTrackService`. To fix the case of missed `load` event the navigation entry from the `performance` API is used.

`loadEventStart` and `loadEventEnd` are the most reliable markers that indicate that the load is completed, when their value is not 0. The alias for these attributes is `duration`:
Screenshot 2025-08-25 at 15 21 15

For further reference check out
[Processing Model of Navigation Timing](https://w3c.github.io/navigation-timing/#processing-model)

### Case: Angular Initialized Instantly (listening for load)
Screenshot 2025-08-26 at 13 30 02

### Case: Angular Initialization Postponed (check for existing load)
Screenshot 2025-08-26 at 13 26 09

## Related issues

* :bug: Fix resolves #3444

### DIFF
--- a/packages/@o3r/analytics/src/services/event-track/event-track.helpers.spec.ts
+++ b/packages/@o3r/analytics/src/services/event-track/event-track.helpers.spec.ts
@@ -1,0 +1,29 @@
+import {
+  isPerformanceNavigationEntry,
+} from './event-track.helpers';
+
+describe('isPerformanceNavigationEntry', () => {
+  test('should return true if entry is navigation', () => {
+    const performanceEntry: PerformanceEntry = {
+      duration: 0,
+      name: '',
+      startTime: 0,
+      entryType: 'navigation',
+      toJSON: () => ({})
+    };
+
+    expect(isPerformanceNavigationEntry(performanceEntry)).toEqual(true);
+  });
+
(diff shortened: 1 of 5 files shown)


ntrotner:
closing PR in favor of https://github.com/AmadeusITGroup/otter/pull/3453

PR #3486, opened 2025-09-04
Fix/no nav message when no browser history update

mrednic-1A:
When an embedded module does a navigate without updating the browser history, to not send a navigation message to the host app. In this way both history will match.

## Related issues

*- No issue associated -*


packages/@ama-mfe/ng-utils/src/navigation/routing.service.ts
```diff
@@ -121,6 +121,7 @@ export class RoutingService implements MessageProducer<NavigationMessage>, Messa
     this.router.events.pipe(
       takeUntilDestroyed(),
       filter((event): event is NavigationEnd => event instanceof NavigationEnd),
+      filter((_event) => !this.router.getCurrentNavigation()?.extras?.skipLocationChange),
```
fpaul-1A, line 124: weird to have a filter that is not pure, but avoiding it might make it overcomplicated, so I'm ok with that

PR #3537, opened 2025-09-19
feat: remove component suffix from app files

sdo-1A:
## Proposed change

The latest major version of Angular (v20) has a style guide update consisting of removing suffixes in the generation of components, directives, services, and pipes. 
https://blog.angular.dev/announcing-angular-v20-b5c9c06cf301#49c0

A workaround to enable the suffix generation has been provided, but this option does not apply to app files. Hence we have updated our app files by removing the `.component` suffixes and renaming the classes from `AppComponent` to `App`. We still support both ways of naming files where necessary.

## Related issues

*- No issue associated -*


packages/@o3r/test-helpers/src/utilities/angular.ts
```diff
@@ -15,8 +15,10 @@ import * as path from 'node:path';
...
  */
 export async function addImportToAppModule(appFolderPath: string, moduleName: string, modulePath: string) {
-  let appModuleFilePath = path.join(appFolderPath, 'src/app/app.module.ts');
-  if (!existsSync(appModuleFilePath)) {
+  const appModuleFilePath1 = path.join(appFolderPath, 'src/app/app-module.ts');
+  const appModuleFilePath2 = path.join(appFolderPath, 'src/app/app.module.ts');
+  let appModuleFilePath = existsSync(appModuleFilePath1) ? appModuleFilePath1 : (existsSync(appModuleFilePath2) ? appModuleFilePath2 : undefined);
+  if (!appModuleFilePath) {
     // assume standalone component
     const appComponentFilePath = path.join(appFolderPath, 'src/app/app.component.ts');
```
mrednic-1A, line 23: Don't we need to check here for the component for both cases ?
`src/app/app.component.ts` or `src/app/app.ts` ? or even `src/app/app-component.ts`

sdo-1A: I check right below in the line: 
```
    appModuleFilePath = existsSync(appComponentFilePath) ? appComponentFilePath : path.join(appFolderPath, 'src/app/app.ts');
```
(this was done in a previous PR during the update of major dependencies for v13)
If this is not clear enough I can use the same method as above for the app module file path

mrednic-1A: indeed, somehow I missed it :)

PR #3574, opened 2025-10-06
fix: revive dictionaries deep

mrednic-1A:
## Proposed change

Possibility to revive fields from a dictionary which has a field to be revived.

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/resources/typescriptFetch/model/model.mustache
+++ b/packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/resources/typescriptFetch/model/model.mustache
@@ -54,8 +54,10 @@ export interface {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{
       {{#vars}}
         {{#vendorExtensions}}
           {{#x-field-name}}
+            {{^x-field-exists}}
   /** Dictionary extraction for {{baseName}} */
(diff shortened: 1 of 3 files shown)


packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/resources/typescriptFetch/model/model.mustache
```diff
@@ -54,8 +54,13 @@ export interface {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{
       {{#vars}}
         {{#vendorExtensions}}
           {{#x-field-name}}
+            {{#x-field-exists}}
+  /** Property {{x-field-name}} is backed up as a dictionary extraction for {{baseName}} */
```
fpaul-1A, line 58: won't we have double comments if there already is a description for the field in the spec?

kpanot: Then is the double /** */ supported by TsDoc, for a signle property?

mrednic-1A: the comment is not above the property because there is no property. I put it like that to not lose this info. Ideally it should be appended to the ts doc of the existing property.

PR #3594, opened 2025-10-13
feat(mcp): resources for release notes

matthieu-crouzet:
## Proposed change

image

After some check it seems that for now VSCode GitHub Copilot chat is not reading the content of the resource.
So we need to provide a tool for each resource to access its content.

I used [`ResourceTemplate`](https://github.com/modelcontextprotocol/typescript-sdk?tab=readme-ov-file#resources) for release notes as we have several resources associated.

I created a ResourceRegistry to be able to access easily to the content.

I hope at some point that copilot will be able to read the resource_link provided without having to provide a text with the content of it.

## Related issues

*- No issue associated -*


tools/@o3r/mcp/src/release-notes/index.ts
```diff
@@ -0,0 +1,114 @@
...
+    {
+      title: 'Get Otter release notes',
+      description: 'Get the release notes for the latest Otter releases',
+      annotations: {
+        readOnlyHint: true,
+        openWorldHint: false
+      },
+      inputSchema: {
+        version: z.string().optional()
+          .describe('The version of the release notes to get, if not provided the all release notes will be returned, format to use is vX.Y.0, e.g. v13.1.0')
```
vscaiceanu-1a:
```suggestion
          .describe('The version of the release notes to get, if not provided, all release notes will be returned. Format to use is vX.Y.0, e.g. v13.1.0')
```

PR #3618, opened 2025-10-20
feat(ama-mfe): enable history navigation from an embedded module

fpaul-1A:
## Related issues

*- No issue associated -*


vscaiceanu-1a (approved):

maxokorokov (commented):
Hey,

my only fear with this is that instead of forbidding the embedded module to mess with parent's history we do the opposite.

ex. if there are 2 embedded modules running at the same time - it's up to the parent to decide what to do with the history, not allow the embedded module to `history.replaceState()`

I think the true solution for this for me would be having `sandbox="allow-scripts"` only, not messing with the shared state.

It's LGTM for the fix though, just a bit worried about long-term :)

matthieu-crouzet (approved):

fpaul-1A:
> Hey,
> 
> my only fear with this is that instead of forbidding the embedded module to mess with parent's history we do the opposite.
> 
> ex. if there are 2 embedded modules running at the same time - it's up to the parent to decide what to do with the history, not allow the embedded module to `history.replaceState()`
> 
> I think the true solution for this for me would be having `sandbox="allow-scripts"` only, not messing with the shared state.
> 
> It's LGTM for the fix though, just a bit worried about long-term :)

I've opened issue https://github.com/AmadeusITGroup/otter/issues/3623 to follow-up on this
I will merge this PR in the meantime to fix the current issues with navigation

PR #3635, opened 2025-10-24
fix(mcp): support more inputs for metadata retrieval

matthieu-crouzet:
## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@ama-mcp/otter/src/metadata-per-release/index.ts
+++ b/packages/@ama-mcp/otter/src/metadata-per-release/index.ts
@@ -32,6 +32,9 @@ import {
   McpServer,
   ResourceTemplate,
 } from '@modelcontextprotocol/sdk/server/mcp.js';
+import type {
+  CallToolResult,
+} from '@modelcontextprotocol/sdk/types.d.ts';
 import {
   tgz,
 } from 'compressing';
@@ -44,6 +47,7 @@ import {
 
 /**
  * Options for the tool metadata per release
+ * @experimental
  */
 export interface MetadataPerReleaseOptions extends ToolDefinition, ResourceToolOptions, CacheToolOptions {
   /**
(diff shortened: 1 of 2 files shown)


packages/@ama-mcp/otter/src/metadata-per-release/index.ts
```diff
@@ -162,10 +166,52 @@ const retrieveAllMetadata = async (
...
 };
 
+const matchPackageName = (key: string, packageName: string) => {
+  const packageNameKey = getPackageAndTagFromKey(key)[0]?.toLowerCase();
+  const packageNameLowerCase = packageName.toLowerCase();
+  return packageNameKey.includes(packageNameLowerCase) || packageName.includes(packageNameKey);
+};
+
+const matchTagName = (key: string, tagName: string) => {
+  const tagNameKey = getPackageAndTagFromKey(key)[1]?.toLowerCase().replace(/^v/, '');
```
kpanot:
```suggestion
  const tagNameKey = getPackageAndTagFromKey(key)[1]?.toLowerCase().replace(/^v/i, '');
```
(just in case)

Or `semver.valid()` if you have semver helpers

matthieu-crouzet: At first, I didn't want to add `semver` but maybe at some point it make sense to add it.
I fixed the regexp as you proposed, we may want to open an issue to use `semver` in the different tools link to version already created.

fpaul-1A: I'm not sure to understand how the `/i` helps since there is a `toLowerCase()` just before

matthieu-crouzet: Indeed :D
I didn't think about it

fpaul-1A (approved):

PR #3641, opened 2025-10-27
fix: style guide updates - schematics - support type

sdo-1A:
## Proposed change

Fix of style guide schematics previously updated in order to support the 'type' option

## Related issues

*- No issue associated -*


packages/@o3r/analytics/schematics/analytics-to-component/index.ts
```diff
@@ -83,16 +84,19 @@ const checkAnalytics = (componentPath: string, tree: Tree, baseFileName: string)
 export function ngAddAnalyticsFn(options: NgAddAnalyticsSchematicsSchema): Rule {
   return async (tree: Tree, context: SchematicContext) => {
     try {
-      const baseFileName = basename(options.path, '.ts');
+      const baseFileName = basename(options.path, `${options.type ? '.' + options.type : ''}.ts`);
```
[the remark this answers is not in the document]
fpaul-1A: **Solution 1:**
I don't really like to have a new parameter that doesn't really seem related to the generator, even if I understand the purpose here.
It might be acceptable if the value is set once in the `angular.json` and the user don't have to pass it everytime
We may want to change the code to work if the type is not present for whatever reason though
```suggestion
      const baseFileName = basename(basename(options.path, '.ts'), options.type ? '.' + options.type : '');
```

**Solution 2:**
This might work in most cases, but it might result in some weird namings if the users are using dots in the filename for whatever reason without `type` option


Overall, I'll go with solution 1 but it's a though call

packages/@o3r/core/schematics/service/templates/base/__name__/__featureName__/__name__-__featureName__.spec.ts.template
```diff
@@ -1,6 +1,6 @@
 import {getTestBed, TestBed} from '@angular/core/testing';
 import {BrowserDynamicTestingModule, platformBrowserDynamicTesting} from '@angular/platform-browser-dynamic/testing';
-import {<%= serviceName %>} from './<%= name %>-<%= featureName %>';
+import {<%= serviceName %>} from './<%= name %>-<%= featureName %><% if (type) { %>.<%= type %><% } %>';
```
matthieu-crouzet: Do we want to handle the `type` option on service if we don't call the Angular schematic to generate service ?

packages/@o3r/core/schematics/page/index.ts
```diff
@@ -156,7 +157,8 @@ function ngGeneratePageFn(options: NgGeneratePageSchematicsSchema): Rule {
         template({
           ...strings,
           ...options,
-          pageName
+          pageName,
+          type: options.type
```
fpaul-1A: isn't this is already included in `...options`?

matthieu-crouzet (approved):

PR #3653, opened 2025-10-30
fix(vscode-ext): do not re-run publish dependencies before publishing extension

matthieu-crouzet:
## Related issues

*- No issue associated -*

### DIFF
--- a/apps/vscode-extension/project.json
+++ b/apps/vscode-extension/project.json
@@ -40,8 +40,7 @@
       "options": {
         "command": "npm install && npx @vscode/vsce publish",
         "cwd": "apps/vscode-extension/dist"
-      },
-      "dependsOn": ["prepare-publish"]
+      }
     }
   },
   "tags": []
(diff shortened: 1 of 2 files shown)


sdo-1A (dismissed):

apps/vscode-extension/project.json
```diff
@@ -33,7 +33,8 @@
           "node tools/@o3r/build-helpers/scripts/prepare-publish.mjs --root={projectRoot}",
           "node tools/@o3r/build-helpers/scripts/clear-package-json-dev-dependencies.mjs --root={projectRoot}"
         ]
-      }
+      },
+      "dependsOn": []
```
kpanot: Not sure about this change, per default it depend on `build` because it needs the build to be completed to publish.
The cache should ensure to not rerun a build if it has been run previously, as we do to other packages.
Why removing the dependency for this package specifically?

fpaul-1A (dismissed):

kpanot (approved):

fpaul-1A (approved):

PR #3682, opened 2025-11-04
ci: skip it-tests for pull-requests targetting main

fpaul-1A:
## Related issues

*- No issue associated -*

### DIFF
--- a/.github/workflows/it-tests.yml
+++ b/.github/workflows/it-tests.yml
@@ -35,6 +35,7 @@ permissions:
 jobs:
   prepare-verdaccio:
     runs-on: ubuntu-latest
+    if: ${{ github.event_name != 'pull_request' || github.base_ref != 'main' || contains(github.event.pull_request.labels.*.name, 'runAllTests') }}
     env:
       NODE_COMPILE_CACHE: ~/.cache/node-compile-cache
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/it-tests.yml
```diff
@@ -75,7 +76,7 @@ jobs:
         packageManager: [yarn, npm]
         testEnvironment: [o3r-project-with-app]
         fastRun:
-          - ${{ github.event_name == 'pull_request' && github.base_ref == 'main' && !contains(github.event.pull_request.labels.*.name, 'runAllTests')}}
+          - ${{ github.event_name == 'pull_request' && github.base_ref == 'main' && contains(github.event.pull_request.labels.*.name, 'runFastTests')}}
```
vscaiceanu-1a: not sure if we have this internal feature documented already. If it is the case we may want to update it with the new label.

.github/workflows/it-tests.yml
```diff
@@ -35,6 +35,7 @@ permissions:
 jobs:
   prepare-verdaccio:
     runs-on: ubuntu-latest
+    if: ${{ github.event_name != 'pull_request' || github.base_ref != 'main' || contains(github.event.pull_request.labels.*.name, 'runAllTests') || contains(github.event.pull_request.labels.*.name, 'runFastTests')}}
```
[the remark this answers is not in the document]
fpaul-1A: the labels were not really opposite, but I've removed the `fastRun`

.github/workflows/it-tests.yml
fpaul-1A: what feature?
(I've removed it)

PR #3692, opened 2025-11-05
chore: add label to skip cache

cpaulve-1A:
## Related issues

*- No issue associated -*


tools/github-actions/setup/action.yml
```diff
@@ -1,5 +1,10 @@
 name: 'Setup'
 description: 'Setup for Otter Library'
+inputs:
```
kpanot: Why not passing it as NX_SKIP_REMOTE_CACHE `env` instead of `inputs`?
It seems to be more natural as any other environment variable used during the installation

Issue #3609, opened 2025-10-17
[Feature] Migrate to signal based viewChild


matthieu-crouzet:
### Context

In our showcase / training application, we still use the decorator for ViewChild / ViewChildren (maybe still the case for Input and Output).
We should migrate to the new notation

### Proposal

We should migrate to the new notation based on signal.

PR #3706, opened 2025-11-10
feat: migrate to signal-based viewChild and viewChildren

sdo-1A:
## Proposed change

Migration to signal-based `viewChild` and `viewChildren`

## Related issues

* :rocket: Feature #3609

### DIFF
--- a/apps/showcase/src/app/component-replacement/component-replacement.spec.ts
+++ b/apps/showcase/src/app/component-replacement/component-replacement.spec.ts
@@ -9,6 +9,9 @@ import {
   RouterModule,
 } from '@angular/router';
 import {
+  NgbScrollSpyService,
+} from '@ng-bootstrap/ng-bootstrap';
+import {
   provideMarkdown,
 } from 'ngx-markdown';
 import {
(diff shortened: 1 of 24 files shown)


apps/showcase/src/app/component-replacement/component-replacement.component.ts
```diff
@@ -50,13 +49,11 @@ import {
 })
 export class ComponentReplacementComponent implements AfterViewInit {
   private readonly inPageNavPresService = inject(InPageNavPresService);
-
-  @ViewChildren(InPageNavLinkDirective)
-  private readonly inPageNavLinkDirectives!: QueryList<InPageNavLink>;
+  private readonly inPageNavLinkDirectives = viewChildren<InPageNavLink>(InPageNavLinkDirective);
```
fpaul-1A: Can't we use something like this to have the equivalent as before ?
```suggestion
  private readonly inPageNavLinkDirectives = viewChildren.required<InPageNavLink>(InPageNavLinkDirective);
```

apps/showcase/src/app/component-replacement/component-replacement.component.ts
```diff
@@ -50,13 +49,11 @@ import {
...
 
   public links$ = this.inPageNavPresService.links$;
 
   public ngAfterViewInit() {
-    this.inPageNavPresService.initialize(this.inPageNavLinkDirectives);
+    this.inPageNavPresService.initialize([...this.inPageNavLinkDirectives()]);
```
fpaul-1A: Why do we need a shallow copy here?

fpaul-1A (approved):

apps/showcase/src/app/component-replacement/component-replacement.html
```diff
@@ -3,7 +3,7 @@ <h1>Component Replacement</h1>
   <div class="right-nav order-1 order-lg-2 col-12 col-lg-2 sticky-lg-top pt-5 pt-lg-0">
     <o3r-in-page-nav-pres
       id="component-replacement-nav"
-      [links]="links$ | async"
+      [links]="(links$ | async) ?? []"
```
matthieu-crouzet: Can't it be the default value of the input to not have to duplicate the `?? []` in every pages?

matthieu-crouzet (approved):

PR #3710, opened 2025-11-12
[1/3] feat(ama-openapi): add core feature of the openapi build strategy

kpanot:
## Proposed change

feat(ama-openapi): add core feature of the openapi build strategy

PR 1 / 3

## Related issues

*- No issue associated -*


packages/@ama-openapi/core/README.md
```diff
@@ -0,0 +1,618 @@
...
+    - [Shareable specification external references](#shareable-specification-external-references)
+  - [Technical information](#technical-information)
+    - [Schema Validation](#schema-validation)
+    - [Renovate configuration](#renovate-configuration)
+    - [Custom installer](#custom-installer)
+  - [Integration](#integration)
+
+## Overview
+
+The `@ama-openapi/core` package provides a dependency management system for [OpenAPI specifications](https://redocly.com/). It allows you to:
```
vscaiceanu-1a: is this link correct? https://redocly.com/ ?

packages/@ama-openapi/core/README.md
```diff
@@ -0,0 +1,618 @@
...
+### NPM package dependency
+
+The usage of NodeJs greatly simplify the retrieval and versioning of specification.
+Thanks to it, a [Renovate configuration](https://docs.renovatebot.com/configuration-options/) can be set to automatize the upgrade of specifications.
+
+To depend of an OpenAPI package, the dependency can be added via the command:
+
+```shell
+npm install @my/dep-spec
+npm install
```
vscaiceanu-1a: why do you need to `npm install` after adding the package?

packages/@ama-openapi/core/src/core/file-system/parse-file.mts
matthieu-crouzet:
```suggestion
 * Parse a file content based on its extension
```

packages/@ama-openapi/core/README.md
```diff
@@ -0,0 +1,618 @@
...
+
+```mermaid
+flowchart LR
+spec2["spec2@1.2.0"]
+spec1 --> spec2
+spec1 --> spec3 --> spec2
+```
+
+> [!IMPORTANT]
+> The NPM dependency allow accessing to the models but do not allow the models to be referred in your specifications.\
```
vscaiceanu-1a:
```suggestion
> The NPM dependency allows access to the models but do not allow the models to be referred in your specifications.\
```

Also, how does this help if I cannot reference it? Should we rephrase saying that on top of adding the dep, we should also reference it in the manifest?

packages/@ama-openapi/core/src/core/file-system/parse-file.mts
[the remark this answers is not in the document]
kpanot: It is not supported by `mts` because a reserved keyword.
The trailing `,` is mandatory.

The TS associated error is:
```
this syntax is reserved in files with the .mts or .cts extension. Add a trailing comma or explicit constraint. ts(7060)
```

packages/@ama-openapi/core/README.md
```diff
@@ -0,0 +1,618 @@
...
+| `mask` | Mask to apply to the model as [defined in stoplight](https://meta.stoplight.io/docs/platform/2nebi9gb2ankj-override-model-properties) |
+
+### Mask Features
+
+The mask feature offers 4 main capabilities:
+
+- Filter the fields of a model object `properties`
+- Override object property's value
+- Add object property
+- Be apply on referenced object
```
vscaiceanu-1a:
```suggestion
- Be applied on a referenced object
```
or 
```suggestion
- Be applied on referenced objects
```

packages/@ama-openapi/core/README.md
```diff
@@ -0,0 +1,618 @@
...
+</details>
+
+## Extracted model details
+
+As described in the [dependency resolution section](#dependency-resolution-concept), the external dependency models are extracted in the `models_external` directory to be referred by your local specification files.
+
+The purposes of this mechanism are the following:
+
+- Be agnostic of NodeJs package manager mechanism (and support `plug'n'play` module resolution)
+- Allow the the application of transform on top of definitions
```
vscaiceanu-1a:
```suggestion
- Allow the application of [transforms](#transforms) on top of definitions
```

packages/@ama-openapi/core/src/core/manifest/extract-dependency-models.mts
matthieu-crouzet:
```suggestion
  model: Model & Require<Pick<Model, 'path'>>;
```

packages/@ama-openapi/core/README.md
```diff
@@ -0,0 +1,618 @@
...
+
+</details>
+
+## Shareable specification
+
+As the dependency system is based on NPM package, a shareable specification should follow the requirements for an NPM package:
+
+1. A `package.json` should be provided and exposed the files accessible by the consumer
+2. The package should be published to an NPM registry thanks to the command `npm publish`
+3. The references of the different models should not break.
```
vscaiceanu-1a: What do you mean by 'should not break'? Should respect semver?

packages/@ama-openapi/core/README.md
```diff
@@ -0,0 +1,618 @@
...
+    },
+  }
+}
+```
+
+</details>
+
+### Shareable specification external references
+
+To ensure the external references to be still resolvable by the [Redocly bundle](https://redocly.com/docs/cli/commands/bundle) process when the consumer will reference one of your models, the following steps are required:
```
vscaiceanu-1a:
```suggestion
To ensure that the external references will be resolvable by the [Redocly bundle](https://redocly.com/docs/cli/commands/bundle) process when the consumer will reference one of your models, the following steps are required:
```

packages/@ama-openapi/core/README.md
```diff
@@ -0,0 +1,618 @@
...
+```
+
+You can use this schema in your IDE or build tools to get validation and autocomplete support for manifest files.
+
+> [!TIP]
+> The project generator [@ama-openapi/create](https://www.npmjs.com/package/@ama-openapi/create) prepare the manifest configuration file (and `package.json`) to refer the `manifest.schema.json`.
+
+### Renovate configuration
+
+The package expose Renovate preset to help to upgrade and mutualize the dependencies relative to the Ama Openapi ecosystem at:
```
vscaiceanu-1a:
```suggestion
The package exposes a Renovate preset to help upgrade and consolidate dependencies related to the Ama OpenAPI ecosystem at:
```

packages/@ama-openapi/core/src/core/transforms/rename.mts
matthieu-crouzet: I would expect S to extends `{ title?: string }`

eslint.shared.config.mjs
```diff
@@ -194,7 +194,7 @@ export default defineConfig([
   },
   {
     name: '@o3r/framework/mjs-files',
-    files: ['**/*.mjs'],
+    files: ['**/*.m{j,t}s'],
```
cpaulve-1A: What's the link with your PR?

nx.json
```diff
@@ -331,6 +343,19 @@
       ],
       "cache": true
     },
+    "expose-renovate": {
```
cpaulve-1A: why do we touch the renovate configuration here?

packages/@ama-openapi/core/src/core/transforms/transform.mts
matthieu-crouzet: It seems strange to have this helper in this file as it's not linked to the model we declare here

packages/@ama-openapi/core/src/core/file-system/parse-file.mts
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Check if the file is a JSON file
+ * @param filePath
+ */
+export const isJsonFile = (filePath: string): boolean => {
+  return filePath.toLowerCase().endsWith('.json');
+};
+
+/**
+ *n Parse a file content based on its extension
```
vscaiceanu-1a:
```suggestion
 * Parse a file content based on its extension
```

packages/@ama-openapi/core/README.md
```diff
@@ -0,0 +1,618 @@
...
+flowchart LR
+spec2-2["spec2@~1.2.0"]
+spec2-1["spec2@~1.1.0"]
+spec1 --> spec2-2
+spec1 --> spec3 --> spec2-1
+```
+
+Will bring the 2 versions of the `spec2` in the final bundle generated by [Redocly](https://redocly.com/) (has there is no intersection between `~1.1.0` and `~1.2.0`).
+
+If needed, NPM offers the feature to enforce the version over the dependencies via [overrides field](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides):
```
cpaulve-1A:
```suggestion
If needed, NPM offers the feature to enforce the version over the dependencies via the [overrides field](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides):
```

packages/@o3r/eslint-config/src/rules/typescript/stylistic.cjs
matthieu-crouzet: Why do we need to change that ?

packages/@ama-openapi/core/README.md
```diff
@@ -0,0 +1,618 @@
...
+
+## Extracted model details
+
+As described in the [dependency resolution section](#dependency-resolution-concept), the external dependency models are extracted in the `models_external` directory to be referred by your local specification files.
+
+The purposes of this mechanism are the following:
+
+- Be agnostic of NodeJs package manager mechanism (and support `plug'n'play` module resolution)
+- Allow the the application of transform on top of definitions
+- Support to link the `$ref` to the referred definition (thanks to [VsCode OpenApi extension](https://marketplace.visualstudio.com/items?itemName=42Crunch.vscode-openapi))
```
cpaulve-1A:
```suggestion
- Support linking the `$ref` to the referred definition (thanks to [VsCode OpenApi extension](https://marketplace.visualstudio.com/items?itemName=42Crunch.vscode-openapi))
```

packages/@ama-openapi/core/README.md
```diff
@@ -0,0 +1,618 @@
...
+- Support of version override when exposed as dependency.
+
+### Reference only
+
+When there is not transform apply to the external model, the only purpose of the model file generated in `models_external` is to redirect to the correct dependency path (concretely in `node_module`).\
```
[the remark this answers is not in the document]
cpaulve-1A:
```suggestion
When there is no `transform` applied to the external model, the only purpose of the model file generated in `models_external` is to redirect to the correct dependency path (concretely in `node_modules`).\
```

packages/@ama-openapi/core/README.md
```diff
@@ -0,0 +1,618 @@
...
+
+- `ama-openapi.manifest.json`
+- `ama-openapi.manifest.yaml`
+- `ama-openapi.manifest.yml`
+- `openapi.manifest.json`
+- `openapi.manifest.yaml`
+- `openapi.manifest.yml`
+
+> [!TIP]
+> In case no valid file from previous the list is found, the mechanism will parse the `package.json` to find manifest definition.
```
cpaulve-1A:
```suggestion
> In case no valid file from previous the list is found, the mechanism will parse the `package.json` to find manifest definitions.
```

packages/@o3r/eslint-config/src/rules/typescript/stylistic.cjs
kpanot: It is linked to the comment https://github.com/AmadeusITGroup/otter/pull/3710/files#r2517781137 .
We have currently a conflict between the trailing `,` required by TS on `mts` and `cts`, and this rule

packages/@ama-openapi/core/eslint.local.config.mjs
```diff
@@ -0,0 +1,41 @@
...
+      globals: {
+        ...globals.node,
+        NodeJS: true
+      }
+    }
+  },
+  {
+    name: '@ama-openapi/core/overrides',
+    rules: {
+      'import/no-cycle': 'off'
```
cpaulve-1A: Why? If this is mandatory we should only disable it on the line with an issue and either a proper explanation or a follow up task

eslint.shared.config.mjs
kpanot: Because without this rule, the `__dirname` was forbidden for `mts` files.
The EsLint rule naming may not be aligned with the files list now.
We can then:
- duplicate the rule to target `mts`
- rename the rule

codecov.yml
```diff
@@ -7,19 +7,19 @@ coverage:
     patch:
       default:
         target: auto
-        threshold: 5%
+        threshold: 0.3%
```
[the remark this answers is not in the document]
kpanot: Fare enough, it will be move to dedicated PR

packages/@ama-openapi/core/src/core/file-system/cleaner.spec.ts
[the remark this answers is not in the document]
kpanot: Not necessarily, the tests are currently passing.
But I can move it if you prefer

nx.json
kpanot: Because a Renovate file (`/renovate/default.json`) is exposed in the `dist/`.
The name of the task is inspired from `expose-schemas` task

packages/@ama-openapi/core/eslint.local.config.mjs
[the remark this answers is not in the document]
kpanot: This is relative to the comment https://github.com/AmadeusITGroup/otter/pull/3710#discussion_r2518458540 .
The rule does not seem to be disableable via comment in this code.

packages/@o3r/artifactory-tools/eslint.local.config.mjs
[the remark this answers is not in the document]
kpanot: Required by the IDE linter check.
It can be moved to dedicated PR indeed

packages/@ama-openapi/core/src/core/file-system/parse-file.mts
```diff
@@ -0,0 +1,25 @@
...
+ * @param filePath
+ */
+export const parseFile = async <S,>(filePath: string): Promise<S> => {
```
cpaulve-1A: Why `<S,>` and not `<S>`?

packages/@ama-openapi/core/src/core/file-system/write-model.mts
vscaiceanu-1a: isn't it better to check for the directory existence before trying to create it?

packages/@ama-openapi/core/src/core/file-system/parse-file.mts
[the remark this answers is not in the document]
matthieu-crouzet: I found an [issue](https://github.com/dprint/dprint-plugin-typescript/issues/378) where they mention to use `<S extends unknown> or `<S,>`
I would prefer the `extends unknown` because we are used to it and are not used to trailing comma

packages/@ama-openapi/core/src/core/manifest/extract-dependency-models.mts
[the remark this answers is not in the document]
matthieu-crouzet: Can't we configure typescript to work with something like 
`import { anything } from './anywhere';`
instead of
`import { anything } from './anywhere.mjs'`
as we don't have mjs file in the source files but only in the dist

packages/@ama-openapi/core/src/core/file-system/parse-file.mts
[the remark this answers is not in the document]
kpanot: Image

`S extends unknown` would require a eslint rule disabling, I think `s,` seems to be less constraining

packages/@ama-openapi/core/src/core/manifest/extract-dependency-models.mts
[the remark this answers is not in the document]
kpanot: I do not know if it is possible and if it is a good practice but why not. 
Anyway this is not a comment for this package only anymore, I create a task for it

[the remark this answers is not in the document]
kpanot: Not really actually, only the following lines:

```typescript
  const artifactPackageJson = require.resolve(`${artifactName}/package.json`);
  const artifactBasePath = dirname(artifactPackageJson);
  const packageJsonContent = await fs.readFile(artifactPackageJson, { encoding: 'utf8' });
  const version = (JSON.parse(packageJsonContent) as PackageJson).version || 'latest';
```

Does it really worse a function for it?

packages/@ama-openapi/core/README.md
kpanot: Nop, fixed it. Well spotted

packages/@ama-openapi/core/README.md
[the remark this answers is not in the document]
kpanot: Not supported feature in this version.
I will add it

packages/@ama-openapi/core/src/core/transforms/apply-mask.mts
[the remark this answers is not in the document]
kpanot: The alternative is to pass `retrievedModel` and `context` to all the functions, even if they do not use them, because some called functions required them deeper in the recursive function.
Not sure it is more readable.

packages/@ama-openapi/core/README.md
```diff
@@ -0,0 +1,618 @@
...
+
+> [!TIP]
+> The project generator [@ama-openapi/create](https://www.npmjs.com/package/@ama-openapi/create) prepare the manifest configuration file (and `package.json`) to refer the `manifest.schema.json`.
+
+### Renovate configuration
+
+The package expose Renovate preset to help to upgrade and mutualize the dependencies relative to the Ama Openapi ecosystem at:
+
+```text
+@ama-openapi/core/renovate/default.json
```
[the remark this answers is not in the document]
kpanot: I did the change but I am not change I want a link here.
The purpose is, for the user to refer the schema in its files, a link is not really convenient to copy/past

packages/@ama-openapi/core/src/core/transforms/rename.mts
kpanot: The return type is not allowed and should be `S & { title?: string | undefined }` which is equivalent to `S`.
I did the first update.

packages/@ama-openapi/core/tsconfig.build.json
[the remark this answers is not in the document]
kpanot: VsCode auto updates....

packages/@o3r/create/ama-openapi-cli/package.json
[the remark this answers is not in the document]
kpanot: Not sure how you see the file, it is removed in the second commit of the PR :S

packages/@ama-openapi/core/src/core/file-system/parse-file.mts
```diff
@@ -0,0 +1,25 @@
...
+import {
+  load,
+} from 'js-yaml';
+
+/**
+ * Check if the file is a JSON file
+ * @param filePath
+ */
+export const isJsonFile = (filePath: string): boolean => {
+  return filePath.toLowerCase().endsWith('.json');
```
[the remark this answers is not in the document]
kpanot: Nop, redocly does not support it yet for specifications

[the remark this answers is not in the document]
kpanot: Possible evolution but not for this version, I will create a task

packages/@ama-openapi/core/src/core/transforms/to-reference.mts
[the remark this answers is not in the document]
matthieu-crouzet: It's easier when debugging as you know that you are already went through that code no ?

eslint.shared.config.mjs
kpanot: Finally I renamed the rule, feel free to re-open this comment if needed.

packages/@ama-openapi/core/src/core/transforms/to-reference.mts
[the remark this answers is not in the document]
kpanot: This is the purpose of the 2 debug logs.
I do not think that the argument is enough strong to justify adding an optional annotation.
Please also notice that the annotation `GENERATED_REF_PROPERTY_KEY` is valid only when associated to a `$ref` (as indicated in the doc regarding `x-internal-reference-generated`)

PR #3715, opened 2025-11-12
fix: missing package group for ama-mfe

fpaul-1A:
## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@o3r/core/package.json
+++ b/packages/@o3r/core/package.json
@@ -200,6 +200,8 @@
   "ng-update": {
     "migrations": "./migration.json",
     "packageGroup": [
+      "@ama-mfe/messages",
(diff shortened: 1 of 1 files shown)


kpanot (approved):

packages/@o3r/core/package.json
```diff
@@ -200,6 +200,8 @@
   "ng-update": {
     "migrations": "./migration.json",
     "packageGroup": [
+      "@ama-mfe/messages",
+      "@ama-mfe/ng-utils",
```
vscaiceanu-1a: should add them to the issue template as well but maybe in a different PR on main.

vscaiceanu-1a (approved):

PR #3721, opened 2025-11-13
feat(sdk-schematics): allow variable in server URL

kpanot:
Add the support of the [Server properties](https://swagger.io/docs/specification/v3_0/api-host-and-base-path/) from OpenApi 3


packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/resources/typescriptFetch/constants/servers.mustache
```diff
@@ -0,0 +1,22 @@
+import type {
+  Server
+} from '@ama-sdk/core';
+
+/** List of servers defined in the OpenAPI document */
+export const SDK_SERVERS = [
+{{#servers}}
```
matthieu-crouzet: Can we create a mustache file to reuse this template part as it's the same as the one you are adding in `packages/@ama-sdk/schematics/schematics/typescript/core/openapi-codegen-typescript/src/main/resources/typescriptFetch/api/api.mustache`

packages/@o3r-training/showcase-sdk/open-api.yaml
```diff
@@ -35,6 +44,8 @@ tags:
 paths:
   /pet:
     post:
+      servers:
+        - url: '/v3'
```
matthieu-crouzet: Why do we have twice the `/v3` in the definitions ?

packages/@ama-sdk/core/src/fwk/server.ts
```diff
@@ -0,0 +1,80 @@
...
+  if (basePath) {
+    logger?.debug?.('Using basePath from client configuration');
+    return basePath;
+  }
+
+  let matchingServer: Server | undefined;
+
+  if (!server) {
+    if (servers && servers.length > 0) {
+      matchingServer = servers[0];
```
fpaul-1A: this might be redundant as it will already be handled by line 53

fpaul-1A (dismissed):

matthieu-crouzet (dismissed):

fpaul-1A (approved):

PR #3736, opened 2025-11-18
chore: reduce coverage requirement for PR content

kpanot:
## Related issues

*- No issue associated -*

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -4,17 +4,17 @@ codecov:
 
 coverage:
   status:
-    patch:
-      default:
-        target: auto
-        threshold: 0.3%
-        informational: true
-
     default_rules:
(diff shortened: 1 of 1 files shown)


fpaul-1A (approved):

codecov.yml
```diff
@@ -4,17 +4,17 @@ codecov:
...
     default_rules:
       flag_coverage_not_uploaded_behavior: exclude
       target: auto
       threshold: 0.3%
       informational: true
+      
+    patch:
+      default:
+        target: 75%
+        threshold: 100%
```
fpaul-1A: if I understand it well, it means that we accept any change in the coverage as long as the global coverage is at least 75%?

PR #3819, opened 2025-12-16
feat: add windsurf rules

matthieu-crouzet:
## Related issues

*- No issue associated -*

### DIFF
--- a/.windsurf/rules/code-style-typescript.md
+++ b/.windsurf/rules/code-style-typescript.md
@@ -1,0 +1,31 @@
+---
+trigger: glob
+globs: **/*.ts
+---
+
+# Otter Framework - TypeScript/JavaScript Code Style
+
+## General Formatting
+
+All formatting rules (charset, indentation, line endings, etc.) are defined in `.editorconfig` at the repository root. Refer to that file for the configuration.
+
+## TypeScript / JavaScript Conventions
+
+- **Always write JSDoc comments** for methods and properties using `/** ... */` pattern
+- **Properties must have the most restricted type possible**:
+  ```typescript
+  // Bad
+  private type: string;
+
+  // Good
+  private type: 'A' | 'B';
+  ```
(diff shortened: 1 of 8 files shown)


.windsurf/rules/project-structure.md
```diff
@@ -0,0 +1,38 @@
...
+├── docs/                              # Documentation
+└── migration-guides/                  # Version migration guides
+```
+
+## Package Scopes
+
+| Scope | Purpose | Angular |
+|-------|---------|---------|
+| `@o3r/*` | Core Otter framework libraries | Yes |
+| `@ama-sdk/*` | SDK generation and client utilities | Partial |
```
fpaul-1A: we may want to explicit what `Partial` means here
I suppose it means only `@angular-cli` ? or is it linked to `client-angular` ?

matthieu-crouzet: I think it's link to the client-angular 
either all scopes would have Partial because of ng-add schematics

matthieu-crouzet: All have schematics so I didn't mention them,
Partial is because `@ama-sdk/client-angular` is in Angular

.windsurf/rules/code-style-typescript.md
```diff
@@ -0,0 +1,31 @@
...
+  ```typescript
+  // Bad
+  private type: string;
+
+  // Good
+  private type: 'A' | 'B';
+  ```
+- **Imports must be at the top of the file** - never in the middle of code
+- **Follow existing ESLint rules** defined in `packages/@o3r/eslint-config`
+- **Use `const` by default**, `let` only when reassignment is needed, never `var`
```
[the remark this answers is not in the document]
matthieu-crouzet: Maybe, the goal of this PR is just to do some test, we can later improve by removing or adding stuff in it :)

PR #3823, opened 2025-12-17
fix: use array.tosorted instead of array.sort

fpaul-1A:
## Proposed change

Preserve immutability of arrays by using `Array.toSorted` instead of `Array.sort` so a copy is created instead of modifying the input

## Related issues

*- No issue associated -*


kpanot (approved):

matthieu-crouzet (approved):

matthieu-crouzet:
Can't we set up an ESLint rule for that ?

PR #3828, opened 2025-12-18
feat: add user activity tracking for embedded apps

mrednic-1A:
## Description

This PR introduces a User Activity Tracking feature for micro-frontend architectures, enabling host applications (shells) to monitor user interactions across embedded modules.

### Features

- **ActivityProducerService**: Listens for DOM events (click, keydown, scroll, touchstart, focus) and:
  - Exposes a `localActivity` signal for consumers within the same application
  - Sends throttled activity messages via the communication protocol to connected peers

- **ActivityConsumerService**: Receives activity messages from connected peers and exposes them via the `latestReceivedActivity` signal

- **Nested Iframe Tracking**: Detects when focus moves to an iframe and simulates activity by emitting periodic `iframeinteraction` events, solving the problem of cross-origin restrictions and focus isolation

- **Configurable Throttling**: Prevents flooding the communication channel with configurable throttle intervals

- **Event Filtering**: `shouldBroadcast` option allows filtering which events trigger activity messages (e.g., exclude iframe events in shell since embedded module activity comes via the communication protocol)

### Configuration Options

| Option | Default | Description |
|--------|---------|-------------|
| `throttleMs` | `1000` | Minimum interval between activity messages |
| `trackNestedIframes` | `false` | Enable tracking of nested iframes |
| `nestedIframePollIntervalMs` | `1000` | Polling interval for iframe focus detection |
| `nestedIframeActivityEmitIntervalMs` | `30000` | Activity signal interval while iframe has focus |
| `highFrequencyThrottleMs` | `300` | Throttle time for high-frequency events (scroll) |
| `shouldBroadcast` | - | Optional filter function for events |

### Use Cases

- Session timeout functionality
- User analytics
- Any feature requiring awareness of user activity across micro-frontends

### Changes

- Added [ActivityProducerService](cci:2://file:///c:/git_clones/github/opensource/otter/packages/@ama-mfe/ng-utils/src/user-activity/activity-producer.service.ts:42:0-229:1) and [ActivityConsumerService](cci:2://file:///c:/git_clones/github/opensource/otter/packages/@ama-mfe/ng-utils/src/user-activity/activity-consumer.service.ts:24:0-73:1)
- Added [IframeActivityTrackerService](cci:2://file:///c:/git_clones/github/opensource/otter/packages/@ama-mfe/ng-utils/src/user-activity/iframe-activity-tracker.service.ts:18:0-124:1) for nested iframe detection
- Added `DEFAULT_ACTIVITY_PRODUCER_CONFIG` with sensible defaults
- Added comprehensive tests
- Added documentation in README.md

### DIFF
--- a/packages/@ama-mfe/messages/src/messages/user-activity/base.ts
+++ b/packages/@ama-mfe/messages/src/messages/user-activity/base.ts
@@ -1,0 +1,14 @@
+/** The user activity message type used for communication between shell and iframes */
+export const USER_ACTIVITY_MESSAGE_TYPE = 'user_activity';
+
+/**
+ * Types of user activity events that can be tracked
+ */
+export type UserActivityEventType =
+  | 'click'
+  | 'keydown'
+  | 'scroll'
+  | 'touchstart'
+  | 'focus'
+  | 'visibilitychange'
+  | 'iframeinteraction';
--- a/packages/@ama-mfe/ng-utils/README.md
+++ b/packages/@ama-mfe/ng-utils/README.md
@@ -295,3 +295,157 @@ The host information is stored in session storage so it won't be lost when navig
 When using iframes to embed applications, the browser history might be shared by the main page and the embedded iframe. For example `<iframe src="https://example.com" sandbox="allow-same-origin">` will share the same history as the main page. This can lead to unexpected behavior when using browser 'back' and 'forward' buttons.
 
 To avoid this, the `@ama-mfe/ng-utils` will forbid the application running in the iframe to alter the browser history. It will happen when connection is configured using `provideConection()` function. This will prevent the iframe to be able to use the `history.pushState` and `history.replaceState` methods.
+
+### User Activity Tracking
+
+The User Activity Tracking feature allows a host application (shell) to monitor user interactions across embedded micro-frontends. This is useful for implementing session timeout functionality, analytics, or any feature that needs to know when users are actively interacting with the application.
+
+#### How it works
+
+- **Producer**: The `ActivityProducerService` listens for DOM events (click, keydown, scroll, touchstart, focus) and:
+  - Exposes a `localActivity` signal for consumers within the same application (not throttled for local detection).
+  - Sends throttled activity messages via the communication protocol to connected peers.
+- **Consumer**: The `ActivityConsumerService` receives activity messages from connected peers via the communication protocol and exposes them via the `latestReceivedActivity` signal.
+
+Both services can be used in either the shell (host) or embedded applications depending on your use case. For example:
+- A shell can produce activity signals to notify embedded modules of user interactions in the host.
+- An embedded module can consume activity signals from the shell.
+- An application can use the producer's `localActivity` signal to detect activity locally.
+
+The service automatically throttles messages sent to peers to prevent flooding the communication channel. High-frequency events like `scroll` have additional throttling.
+
+#### Producer Configuration
+
+Start the `ActivityProducerService` to send activity signals to connected peers:
+
+```typescript
+import { inject, runInInjectionContext } from '@angular/core';
+import { bootstrapApplication } from '@angular/platform-browser';
+import { ConnectionService, ActivityProducerService } from '@ama-mfe/ng-utils';
+
+bootstrapApplication(App, appConfig)
+  .then((m) => {
+    runInInjectionContext(m.injector, () => {
+      if (window.top !== window.self) {
+        inject(ConnectionService).connect('hostUniqueID');
+        // Start activity tracking with custom throttle
+        inject(ActivityProducerService).start({
+          throttleMs: 5000 // Send at most one message every 5 seconds
+        });
+      }
+    });
+  });
+```
+
+##### Configuration Options
+
+| Option | Type | Default | Description |
+|--------|------|---------|-------------|
+| `throttleMs` | `number` | `1000` | Minimum interval between activity messages sent to the host |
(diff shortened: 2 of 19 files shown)


packages/@ama-mfe/ng-utils/src/messages/user-activity.ts
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Type guard to check if a message is a user activity message
+ * @param message The message to check
+ */
+export function isUserActivityMessage(message: unknown): message is UserActivityMessage {
+  return (
+    typeof message === 'object'
+    && message !== null
+    && 'type' in message
+    && (message as VersionedMessage).type === USER_ACTIVITY_MESSAGE_TYPE
```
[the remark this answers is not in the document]
mrednic-1A: the cast is not mandatory indeed

packages/@ama-mfe/ng-utils/README.md
```diff
@@ -295,3 +295,157 @@ The host information is stored in session storage so it won't be lost when navig
...
+##### Tracking Nested Iframes
+
+When a user interacts with content inside an iframe (e.g., a third-party widget, payment form, or embedded content), the parent application cannot detect those interactions directly. This is because:
+
+1. **Cross-origin restrictions**: DOM events inside cross-origin iframes do not bubble up to the parent document.
+2. **Focus isolation**: When an iframe has focus, the parent document stops receiving keyboard and mouse events.
+
+This creates a problem for detection of user interactions: a user could be actively filling out a form inside an iframe, but the host application would see no activity and might incorrectly trigger a session timeout.
+
+**Solution**: When `trackNestedIframes` is enabled, the `ActivityProducerService` polls `document.activeElement` to detect when an iframe gains focus. While an iframe has focus, the service simulates activity by periodically emitting `iframeinteraction` events. This ensures the host application knows the user is still active, even though it cannot see the actual interactions inside the iframe.
```
vscaiceanu-1a: what about when the user has focus on the inner iframe but it is idle (no activity within the inner iframe). It will still prevent the timeout, right?

packages/@ama-mfe/ng-utils/src/user-activity/activity-producer.service.ts
```diff
@@ -0,0 +1,232 @@
...
+export class ActivityProducerService implements MessageProducer<UserActivityMessage> {
+  private readonly messageService = inject(ConnectionService);
+  private readonly destroyRef = inject(DestroyRef);
+  private readonly iframeActivityTracker = inject(IframeActivityTrackerService);
+  private readonly logger = inject(LoggerService);
+
+  /**
+   * Current configuration
+   */
+  private config?: ActivityProducerConfig;
```
kpanot: This is used only in `start` and `onActivityThrottled` which is called by `start`.
I think it would be safer to keep it in `start` and pass it to `onActivityThrottled` via parameters

mrednic-1A: done

packages/@ama-mfe/ng-utils/src/user-activity/config.ts
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * DOM events that indicate user activity
+ */
+export const ACTIVITY_EVENTS: Exclude<UserActivityEventType, 'visibilitychange'>[] = [
+  'click',
+  'keydown',
+  'scroll',
+  'touchstart',
+  'focus'
+];
```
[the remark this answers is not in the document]
mrednic-1A: same as the other

packages/@ama-mfe/messages/src/messages/user-activity/base.ts
```diff
@@ -0,0 +1,14 @@
+/** The user activity message type used for communication between shell and iframes */
+export const USER_ACTIVITY_MESSAGE_TYPE = 'user_activity';
+
+/**
+ * Types of user activity events that can be tracked
+ */
+export type UserActivityEventType =
+  | 'click'
+  | 'keydown'
```
cpaulve-1A: I would prefer the use of key up because key down will trigger as long as the key is pressed:

Image

Image

packages/@ama-mfe/ng-utils/src/user-activity/iframe-activity-tracker.service.ts
```diff
@@ -0,0 +1,124 @@
...
+    return this.activityIntervalId !== undefined;
+  }
+
+  /**
+   * Polls document.activeElement to detect iframe focus changes.
+   * When iframe gains focus: emit immediately and start activity interval.
+   * When iframe loses focus: stop activity interval.
+   */
+  private checkActiveElement(): void {
+    const activeElement = document.activeElement;
```
cpaulve-1A: Are we sure it will take into account a change of tab? If the user focused the iframe and switch window/tab, what does happen?

packages/@ama-mfe/messages/src/messages/user-activity/base.ts
mrednic-1A: Indeed. I have added a filter to do nothing if it is a _keydown_ repeat event.

packages/@ama-mfe/ng-utils/src/user-activity/iframe-activity-tracker.service.ts
mrednic-1A: yes I tested this.
image
we still get events that the user is inside an iframe, even if we change the tab.
'iframeInteraction' is fired when the browser tab is not active.

vscaiceanu-1a (approved):

PR #3856, opened 2026-01-12
feat: clean up deprecated - v14

sdo-1A:
## Related issues

*- No issue associated -*


packages/@o3r/dynamic-content/src/services/styling/style-lazy-loader-module.ts
```diff
@@ -1,11 +1,6 @@
...
   NgModule,
 } from '@angular/core';
-import {
-  DynamicContentModule,
-} from '../dynamic-content/index';
 
-@NgModule({
-  imports: [DynamicContentModule]
-})
+@NgModule({})
```
matthieu-crouzet: Should we keep this module as it does not import anything anymore

kpanot: It will actually break the application of someone importing the module without warning, even if removing it won't have any effect.
For the most smooth migration we can keep it, deprecate it to v15 and add a ngUpdate rule to remove it in v14.

packages/@o3r/dynamic-content/package.json
```diff
@@ -41,7 +41,8 @@
     "express-interceptor": "^1.2.0",
     "jasmine": "^5.0.0",
     "rxjs": "^7.8.1",
-    "type-fest": "^5.3.1"
+    "type-fest": "^5.3.1",
+    "typescript": "~5.9.2"
```
fpaul-1A: this should be aligned with the other peer dependencies of the monorepo
```suggestion
    "typescript": "^5.9.0"
```

PR #3873, opened 2026-01-16
feat: use prompt registry validate collections github action

matthieu-crouzet:
## Related issues

*- No issue associated -*


tools/@o3r/workspace-helpers/package.json
fpaul-1A: should we also remove the script line 21?
```
 "prompt-registry:validate-collections": "./scripts/prompt-registry/validate-collections.mjs",
```

PR #3882, opened 2026-01-20
fix: #3755 Remove dependency on module-from-string

### LINKED ISSUE #3755: [Feature]: Remove dependency on module-from-string
### Context

module-from-string is a no longer maintained package with several issues.

### Proposal

Remove the dependency to avoid future issues.

cpaulve-1A:
## Proposed change

module-from-string suffers from a prototype pollution vulnerability.
It has not been maintained for over a year and does not offer a non vulnerable version

## Related issues

* :bug: Fix resolves #3755


packages/@o3r/testing/src/tools/path-replacement/path-replacement.ts
```diff
@@ -7,17 +7,30 @@ import {
...
-} from 'module-from-string';
+
+/**
+ * Load a CommonJS module from a string.
+ * @param code The module code to execute
+ * @param dirname The directory to use as the module's dirname
+ */
+function requireFromString(code: string, dirname: string): unknown {
+  const filename = path.join(dirname, '_virtual.js');
+  const module = new Module(filename);
```
kpanot: Should be cleaner to not use `module` to not override global one

kpanot (approved):

fpaul-1A (approved):

PR #4118, opened 2026-04-01
feat: transloco package - tools and tests

sdo-1A:
## Proposed change

Migration of tools from ngx-translate to Transloco

`src/tools/` directory — the main runtime integration.
Each file has its corresponding `*.spec.ts` test file (using `TranslocoTestingModule` / `provideTranslocoTesting`)

| File | Changes |
|------|---------|
| `localization-service.ts` | Replace `TranslateService` → `TranslocoService` |
| `localization-module.ts` | Replace `TranslateModule` → `TranslocoModule` / `provideTransloco()` |
| `localization-provider.ts` | Replace `TranslateService` → `TranslocoService` |
| `translations-loader.ts` | Implement `TranslocoLoader` instead of `TranslateLoader` |
| `localization-translate-pipe.ts` | Wrap `TranslocoPipe` or `TranslocoService` |
| `localization-translate-directive.ts` | Wrap `TranslocoDirective` or `TranslocoService` |
| `text-direction-service.ts` | Replace `onLangChange` → `langChanges$` |
| `localization-token.ts` | Update injection tokens |
| Localized pipes | Adapt locale-change subscription |

`src/core/translate-message-format-lazy-compiler.ts` — replace `TranslateCompiler` with `TranslocoTranspiler`

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@o3r/transloco/src/tools/localization-helpers.ts
+++ b/packages/@o3r/transloco/src/tools/localization-helpers.ts
@@ -1,0 +1,9 @@
+/**
+ * Formats a debug key string combining the translation key and its value.
+ * Used across the localization service, pipe, and directive when debugMode is enabled.
+ * @param key The translation key
+ * @param value The translated value
+ */
+export function getDebugKey(key: string, value: string): string {
+  return `${key} - ${value}`;
+}
--- a/packages/@o3r/transloco/src/tools/localization-module.ts
+++ b/packages/@o3r/transloco/src/tools/localization-module.ts
@@ -1,0 +1,110 @@
+import {
+  BidiModule,
+} from '@angular/cdk/bidi';
+import {
+  CommonModule,
+  CurrencyPipe,
+  DatePipe,
+  DecimalPipe,
+} from '@angular/common';
+import {
+  InjectionToken,
+  LOCALE_ID,
+  ModuleWithProviders,
+  NgModule,
+  Optional,
+} from '@angular/core';
+import {
+  TranslocoModule,
+} from '@jsverse/transloco';
+import {
+  DEFAULT_LOCALIZATION_CONFIGURATION,
+  LocalizationConfiguration,
+} from '../core';
+import {
+  LocalizationService,
+} from './localization-service';
+import {
+  LOCALIZATION_CONFIGURATION_TOKEN,
+} from './localization-token';
+import {
+  LocalizationTranslateDirective,
+} from './localization-translate-directive';
+import {
+  O3rLocalizationTranslatePipe,
+} from './localization-translate-pipe';
+import {
+  LocalizedCurrencyPipe,
+} from './localized-currency-pipe';
+import {
+  LocalizedDatePipe,
+} from './localized-date-pipe';
+import {
+  LocalizedDecimalPipe,
+} from './localized-decimal-pipe';
+import {
+  TextDirectionService,
+} from './text-direction-service';
+
+/**
+ * creates LocalizationConfiguration, which is used if the application
+ * @param configuration Localization configuration
+ */
+export function createLocalizationConfiguration(configuration?: Partial<LocalizationConfiguration>): LocalizationConfiguration {
+  return {
+    ...DEFAULT_LOCALIZATION_CONFIGURATION,
+    ...configuration
+  };
+}
+
+/**
+ * Factory to inject the LOCALE_ID token with the current language into Angular context
+ * @param localizationService Localization service
+ */
+export function localeIdNgBridge(localizationService: LocalizationService) {
+  return localizationService.getCurrentLanguage();
+}
+
+/** Custom Localization Configuration Token to override default localization configuration */
+export const CUSTOM_LOCALIZATION_CONFIGURATION_TOKEN = new InjectionToken<Partial<LocalizationConfiguration>>('Partial Localization configuration');
+
+@NgModule({
+  declarations: [O3rLocalizationTranslatePipe, LocalizationTranslateDirective, LocalizedDatePipe, LocalizedDecimalPipe, LocalizedCurrencyPipe],
+  imports: [TranslocoModule, BidiModule, CommonModule],
+  exports: [TranslocoModule, O3rLocalizationTranslatePipe, LocalizationTranslateDirective, LocalizedDatePipe, LocalizedDecimalPipe, LocalizedCurrencyPipe],
+  providers: [
+    { provide: LOCALIZATION_CONFIGURATION_TOKEN, useFactory: createLocalizationConfiguration, deps: [[new Optional(), CUSTOM_LOCALIZATION_CONFIGURATION_TOKEN]] },
+    { provide: LOCALE_ID, useFactory: localeIdNgBridge, deps: [LocalizationService] },
+    { provide: DatePipe, useClass: LocalizedDatePipe },
+    { provide: DecimalPipe, useClass: LocalizedDecimalPipe },
+    { provide: CurrencyPipe, useClass: LocalizedCurrencyPipe },
+    TextDirectionService
+  ]
+})
+export class LocalizationModule {
+  /**
+   * forRoot method should be called only once from the application index.ts
+   * It will do several things:
+   * - provide the configuration for the whole application
+   * - register all locales specified in the LocalizationConfiguration
+   * - configure TranslocoService
+   * - inject LOCALE_ID token
+   * @param configuration LocalizationConfiguration
+   */
+  public static forRoot(
+    configuration?: () => Partial<LocalizationConfiguration>
+  ): ModuleWithProviders<LocalizationModule> {
+    return {
+      ngModule: LocalizationModule,
+      providers: [
+        LocalizationService,
+        ...(configuration
+          ? [{
+            provide: CUSTOM_LOCALIZATION_CONFIGURATION_TOKEN,
+            useFactory: configuration
+          }]
+          : [])
+      ]
+    };
+  }
+}
--- a/packages/@o3r/transloco/src/tools/localization-service.ts
+++ b/packages/@o3r/transloco/src/tools/localization-service.ts
@@ -1,0 +1,257 @@
+import {
+  inject,
+  Injectable,
+  signal,
+} from '@angular/core';
+import {
+  toObservable,
+} from '@angular/core/rxjs-interop';
+import {
+  TranslocoService,
+} from '@jsverse/transloco';
+import {
+  select,
+  Store,
+} from '@ngrx/store';
+import {
+  LoggerService,
+} from '@o3r/logger';
+import {
+  combineLatest,
+  distinctUntilChanged,
+  firstValueFrom,
+  map,
+  Observable,
+  of,
+  shareReplay,
+  startWith,
+  switchMap,
+} from 'rxjs';
+import {
+  LocalizationConfiguration,
+} from '../core/localization-configuration';
+import {
+  LocalizationOverrideStore,
+  selectLocalizationOverride,
+} from '../stores/index';
+import {
+  getDebugKey,
+} from './localization-helpers';
+import {
+  LOCALIZATION_CONFIGURATION_TOKEN,
+} from './localization-token';
+
+/**
+ * Service which is wrapping the configuration logic of TranslocoService from JSVerse Transloco.
+ * Any application willing to use localization just needs to inject LocalizationService
+ * in the root component and call its configure() method.
+ */
+@Injectable()
+export class LocalizationService {
+  private readonly translateService = inject(TranslocoService);
+  private readonly logger = inject(LoggerService, { optional: true });
+  private readonly configuration = inject<LocalizationConfiguration>(LOCALIZATION_CONFIGURATION_TOKEN);
+  private readonly store = inject<Store<LocalizationOverrideStore>>(Store, { optional: true });
+
+  private readonly localeSplitIdentifier = '-';
+
+  /**
+   * Internal signal that we use to track changes between keys only and translation mode
+   */
+  private readonly _showKeys = signal(false);
+
+  /**
+   * Map of localization keys to replace a key to another
+   */
+  private readonly keyMapping$?: Observable<Record<string, any>>;
+
+  /**
+   * _showKeys exposed as an Observable
+   */
+  public readonly showKeys$ = toObservable(this._showKeys);
+
+  /**
+   * Return the current value of debug show/hide translation keys.
+   */
+  public readonly showKeys = this._showKeys.asReadonly();
+
+  constructor() {
+    this.configure().catch((err) => {
+      this.logger?.error(`Failed to configure LocalizationService: ${err}`);
+    });
+    if (this.store) {
+      this.keyMapping$ = this.store.pipe(
+        select(selectLocalizationOverride)
+      );
+    } else {
+      this.logger?.debug('Store not available: localization key overrides via rules engine will not be active.');
+    }
+  }
+
+  /**
+   * This will handle the fallback language hierarchy to find out fallback language.
+   * supportedLocales language has highest priority, next priority goes to fallbackLocalesMap and default would be
+   * fallbackLanguage.
+   * @param language Selected language.
+   * @returns selected language if supported, fallback language otherwise.
+   */
+  private checkFallbackLocalesMap<T extends string | undefined>(language: T) {
+    if (language && !this.configuration.supportedLocales.includes(language)) {
+      const closestSupportedLanguageCode = this.getFirstClosestSupportedLanguageCode(language);
+      const fallbackForLanguage = this.getFallbackMapLangCode(language);
+      const fallbackStrategyDebug = (fallbackForLanguage && ' associated fallback language ')
+        || (closestSupportedLanguageCode && ' closest supported language ')
+        || (this.configuration.fallbackLanguage && ' configured default language ');
+      const fallbackLang = fallbackForLanguage || closestSupportedLanguageCode || this.configuration.fallbackLanguage || language;
+      if (language !== fallbackLang) {
+        this.logger?.debug(`Non supported languages ${language} will fallback to ${fallbackStrategyDebug} ${fallbackLang}`);
+      }
+      return fallbackLang;
+    } else if (!language) {
+      this.logger?.debug('Language is not defined');
+    }
+    return language;
+  }
+
+  /**
+   * This function checks if fallback language can be provided from fallbackLocalesMap.
+   * supportedLocales: ['en-GB', 'en-US', 'fr-FR'], fallbackLocalesMap: {'en-CA': 'en-US', 'de': 'fr-FR'}
+   * translate to en-CA -> fallback to en-US, translate to de-DE -> fallback to fr-FR
+   * translate to en-NZ -> fallback to en-GB
+   * @param language Selected language.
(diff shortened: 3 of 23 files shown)


packages/@o3r/transloco/src/tools/localization-service.ts
```diff
@@ -0,0 +1,248 @@
...
+
+  /**
+   * Is the translation deactivation enabled
+   */
+  public isTranslationDeactivationEnabled() {
+    return this.configuration.enableTranslationDeactivation;
+  }
+
+  /**
+   * Wrapper to call the TranslocoService method getAvailableLangs().
```
matthieu-crouzet: **[Critical] `useLanguage()` returns immediately — behavioral regression**

`setActiveLang()` is synchronous (returns `string`). Wrapping it in `of()` means this observable completes **before** the translation bundle is loaded. In the old ngx-translate, `translateService.use()` returned an Observable that completed *after* the bundle loaded.

This makes `await firstValueFrom(this.useLanguage(language))` in `configure()` resolve before translations are ready — consumers may read untranslated keys.

**Suggested fix:** Use `this.translocoService.load(language)` to return an Observable that waits for the bundle, or combine `setActiveLang` with a load mechanism:
```ts
public useLanguage(language: string): Observable<Translation> {
  language = this.checkFallbackLocalesMap(language);
  this.translocoService.setActiveLang(language);
  return this.translocoService.load(language);
}
```

See [Transloco API docs](https://jsverse.gitbook.io/transloco) for `load()` behavior.

packages/@o3r/transloco/src/tools/localization-service.ts
```diff
@@ -0,0 +1,248 @@
...
+   */
+  private readonly keyMapping$?: Observable<Record<string, any>>;
+
+  /**
+   * _showKeys$ exposed as an Observable
+   */
+  public showKeys$ = this._showKeys$.asObservable();
+
+  constructor() {
+    void this.configure();
```
matthieu-crouzet: **[Critical] Fire-and-forget `void this.configure()` — unhandled promise rejection**

`configure()` is `async` and calls `await firstValueFrom(...)`. If it rejects (invalid config, TranslocoService failure), the error is completely unhandled — `void` discards the promise. This produces cryptic "Unhandled promise rejection" errors and leaves the service in an indeterminate state.

**Suggested fix:** Either add a `.catch()` handler:
```ts
constructor() {
  this.configure().catch((err) => {
    this.logger?.error?.(`Failed to configure LocalizationService: ${err}`);
  });
  // ...
}
```
Or better: remove the constructor call entirely and use Angular's `APP_INITIALIZER` so callers can properly await initialization.

packages/@o3r/transloco/src/tools/translations-loader.ts
```diff
@@ -0,0 +1,141 @@
...
+   *2. if 1 fails try to load fallback lang but only if it's different from lang in 1
+   * @param lang - language of the bundle
+   * @param fallbackLanguage - fallback language in case bundle in language not found
+   */
+  public getTranslationFromLocal(lang: string, fallbackLanguage: string): Observable<Translation> {
+    const pathPrefix: string = this.localizationConfiguration.bundlesOutputPath;
+    return this.downloadLanguageBundle$(pathPrefix + lang + JSON_EXT).pipe(
+      catchError(() => {
+        if (lang === fallbackLanguage) {
+          this.logger?.warn(`Failed to load ${lang} from ${pathPrefix + lang + JSON_EXT}.`);
```
matthieu-crouzet: **[Critical] Silent empty fallback — no logging on innermost `catchError`**

When both the requested language AND the fallback language fail to load, this `catchError(() => of({}))` silently returns an empty translation dictionary with **zero logging**. The application renders raw translation keys everywhere with no indication of failure.

This is the last line of defense — it should at minimum log an error:
```ts
catchError((err) => {
  this.logger?.error?.(
    `Failed to load fallback language ${fallbackLanguage} from ${pathPrefix + fallbackLanguage + JSON_EXT}. Application will have NO translations.`
  );
  return of({});
})
```

packages/@o3r/transloco/src/core/transloco-message-format-lazy-transpiler.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * Clear the cache of the compiled translations
+   */
+  public clearCache() {
+    this.cache = {};
+  }
+
+  /** @inheritDoc */
+  public transpile({ value, params = {}, key }: TranspileParams): any {
```
matthieu-crouzet: **[Important] No try-catch — `IntlMessageFormat` can throw and crash the view**

`new IntlMessageFormat(value, lang, ...)` throws `SyntaxError` for invalid ICU syntax (e.g., missing closing brace), and `.format(params)` throws for missing required parameters. Since this is called from pipes/directives during change detection, a single bad translation key can crash the entire component tree.

**Suggested fix:**
```ts
public transpile({ value, params = {}, key }: TranspileParams): any {
  if (typeof value !== 'string') {
    return value;
  }
  try {
    // ... existing cache/compile logic ...
  } catch (err) {
    console.error(`[TranslocoMessageFormatLazyTranspiler] Failed to transpile key "${key}": ${err}`);
    return value; // Return raw value as graceful fallback
  }
}
```

packages/@o3r/transloco/src/core/transloco-message-format-lazy-transpiler.ts
```diff
@@ -0,0 +1,95 @@
...
+    if (typeof value !== 'string') {
+      return value;
+    }
+
+    const lang = this.translocoService.getActiveLang();
+    const cacheKey = `${lang}_${key}`;
```
matthieu-crouzet: **[Important] Stale cache — key doesn't include translation value**

The cache key is `` `${lang}_${key}` `` but the cached `IntlMessageFormat` is compiled from `value`. If the translation value for a key changes at runtime (e.g., store override, dynamic content reload), the cache returns the stale compiled result.

**Suggested fix:** Include `value` in the cache key:
```ts
const cacheKey = `${lang}_${key}_${value}`;
```
Or validate the cached entry was compiled from the same value.

matthieu-crouzet (commented):
## Automated Review — New Findings

10 new findings from a multi-agent code review (code quality, error handling, test coverage, type design). Items already raised by @matthieu-crouzet are excluded.

### Suggestions (lower priority)

- **Inconsistent barrel exports** (`tools/index.ts`): Mixes `export *` and `export { X }` with no clear rule. Consider using one style consistently.
- **`LocalizedDatePipe` has `showKeys` behavior but `LocalizedCurrencyPipe` / `LocalizedDecimalPipe` do not** — inconsistency in debug mode behavior across localized pipes.
- **Duplicated `ChangeDetectorRefFixture`** across 3 spec files — consider extracting to a shared testing utility.

### Test Coverage Gaps (Codecov: 67.1% patch coverage)

| File | Coverage | Missing |
|------|----------|---------|
| `localization-translate-directive.ts` | 9.6% | 47 lines |
| `transloco-message-format-lazy-transpiler.ts` | 23.1% | 20 lines |
| `text-direction-service.ts` | 26.3% | 14 lines |
| `text-directionality.ts` | 26.7% | 11 lines |

The transpiler and directive are the highest-priority untested files — both contain complex logic (caching, dual-strategy rendering) that is easy to break silently.

### Useful references
- Transloco repo: https://github.com/jsverse/transloco
- Transloco docs: https://jsverse.gitbook.io/transloco

package.json
```diff
@@ -198,6 +198,7 @@
     "@o3r/rules-engine": "workspace:~",
     "@o3r/telemetry": "workspace:~",
     "@o3r/test-helpers": "workspace:~",
+    "@o3r/transloco": "workspace:~",
```
kpanot: Why is it needed at workspace root level dep?

packages/@o3r/transloco/src/tools/localization-module.ts
```diff
@@ -0,0 +1,110 @@
...
+  /**
+   * forRoot method should be called only once from the application index.ts
+   * It will do several things:
+   * - provide the configuration for the whole application
+   * - register all locales specified in the LocalizationConfiguration
+   * - configure TranslocoService
+   * - inject LOCALE_ID token
+   * @param configuration LocalizationConfiguration
+   */
+  public static forRoot(
```
kpanot: Do we want to keep the module pattern or switch to provider?

packages/@o3r/transloco/src/tools/localization-service.ts
```diff
@@ -0,0 +1,253 @@
...
+  private readonly logger = inject(LoggerService, { optional: true });
+  private readonly configuration = inject<LocalizationConfiguration>(LOCALIZATION_CONFIGURATION_TOKEN);
+  private readonly store = inject<Store<LocalizationOverrideStore>>(Store, { optional: true });
+
+  private readonly localeSplitIdentifier: string = '-';
+
+  /**
+   * Internal subject that we use to track changes between keys only and translation mode
+   */
+  private readonly _showKeys$ = new BehaviorSubject(false);
```
kpanot: Could use signal

packages/@o3r/transloco/src/tools/localization-translate-directive.ts
```diff
@@ -0,0 +1,160 @@
...
+   */
+  protected override getTranslateFn(lang: string, prefix: string | undefined): (key: string, params?: Record<string, unknown>) => any {
+    const parentTranslateFn = super.getTranslateFn(lang, prefix);
+    return (key: string, params?: Record<string, unknown>) => {
+      if (this.showKeys) {
+        return key;
+      }
+      const value = parentTranslateFn(key, params);
+      if (this.localizationConfig.debugMode) {
+        return `${key} - ${value as string}`;
```
kpanot: As it is used in 2 places, maybe you can extract it as function:

```typescript
private getDebugKey = (key: string, value: string) => `${key} - ${value}`
```

packages/@o3r/transloco/src/tools/localization-service.ts
```diff
@@ -0,0 +1,254 @@
...
+  private checkFallbackLocalesMap<T extends string | undefined>(language: T) {
+    if (language && !this.configuration.supportedLocales.includes(language)) {
+      const closestSupportedLanguageCode = this.getFirstClosestSupportedLanguageCode(language);
+      const fallbackForLanguage = this.getFallbackMapLangCode(language);
+      const fallbackStrategyDebug = (fallbackForLanguage && ' associated fallback language ')
+        || (closestSupportedLanguageCode && ' closest supported language ')
+        || (this.configuration.fallbackLanguage && ' configured default language ');
+      const fallbackLang = fallbackForLanguage || closestSupportedLanguageCode || this.configuration.fallbackLanguage || language;
+      if (language !== fallbackLang) {
+        this.logger?.debug(`Non supported languages ${language} will fallback to ${fallbackStrategyDebug} ${fallbackLang}`);
```
kpanot:
```suggestion
        this.logger?.debug?.(`Non supported languages ${language} will fallback to ${fallbackStrategyDebug} ${fallbackLang}`);
```

packages/@o3r/transloco/src/tools/localization-helpers.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Formats a debug key string combining the translation key and its value.
+ * Used across the localization service, pipe, and directive when debugMode is enabled.
+ * @param key The translation key
+ * @param value The translated value
+ */
+export function getDebugKey(key: string, value: string): string {
```
kpanot, line 7: Can you add simple UTs?

packages/@o3r/transloco/src/tools/translations-loader.ts
```diff
@@ -0,0 +1,141 @@
...
```
kpanot: Do we want to keep `warn` for this case or is it more an `error`?

packages/@o3r/transloco/src/tools/localized-decimal-pipe.ts
```diff
@@ -0,0 +1,50 @@
...
+export class LocalizedDecimalPipe extends DecimalPipe implements OnDestroy, PipeTransform {
+  private readonly onLangChange: Subscription;
+  private readonly localizationService = inject(LocalizationService);
+  private readonly changeDetectorRef = inject(ChangeDetectorRef);
+
+  constructor() {
+    super(inject(LocalizationService).getCurrentLanguage());
+    this.onLangChange = this.localizationService.getTranslateService().langChanges$.subscribe(() =>
+      this.changeDetectorRef.markForCheck()
+    );
```
kpanot:
```suggestion
    this.localizationService.getTranslateService().langChanges$
      .pipe(takeUntilDestroy())
      .subscribe(() =>
        this.changeDetectorRef.markForCheck()
      );
```

to avoid unsubscription management

packages/@o3r/transloco/src/tools/localized-date-pipe.ts
```diff
@@ -0,0 +1,56 @@
...
+export class LocalizedDatePipe extends DatePipe implements OnDestroy, PipeTransform {
+  private readonly onLangChange: Subscription;
+  private readonly localizationService = inject(LocalizationService);
+  private readonly changeDetectorRef = inject(ChangeDetectorRef);
+
+  constructor() {
+    super(inject(LocalizationService).getCurrentLanguage());
+    this.onLangChange = this.localizationService.getTranslateService().langChanges$.subscribe(() =>
+      this.changeDetectorRef.markForCheck()
+    );
```
kpanot:
```suggestion
    this.localizationService.getTranslateService().langChanges$
      .pipe(takeUntilDestroy())
      .subscribe(() =>
        this.changeDetectorRef.markForCheck()
      );
```

packages/@o3r/transloco/src/tools/localized-currency-pipe.ts
```diff
@@ -0,0 +1,54 @@
...
+export class LocalizedCurrencyPipe extends CurrencyPipe implements OnDestroy, PipeTransform {
+  private readonly onLangChange: Subscription;
+  private readonly localizationService = inject(LocalizationService);
+  private readonly changeDetectorRef = inject(ChangeDetectorRef);
+
+  constructor() {
+    super(inject(LocalizationService).getCurrentLanguage());
+    this.onLangChange = this.localizationService.getTranslateService().langChanges$.subscribe(() =>
+      this.changeDetectorRef.markForCheck()
+    );
```
kpanot:
```suggestion
    this.localizationService.getTranslateService().langChanges$
      .pipe(takeUntilDestroy())
      .subscribe(() =>
        this.changeDetectorRef.markForCheck()
      );
```

packages/@o3r/transloco/src/tools/localization-translate-pipe.ts
```diff
@@ -0,0 +1,112 @@
...
+  /** last key queried */
+  protected lastQueryKey?: string;
+
+  /** last key resolved */
+  protected lastResolvedKey?: string;
+
+  constructor() {
+    super(
+      inject(TranslocoService),
+      inject(TRANSLOCO_SCOPE, { optional: true }) ?? undefined,
```
kpanot, line 70: Is the `?? undefined` mandatory ?
```suggestion
      inject(TRANSLOCO_SCOPE, { optional: true }),
```

sdo-1A: Yes it is required since the `TranslocoPipe` constructor signature expects `OrArray<TranslocoScope> | undefined` and `string | undefined` for the scope and lang parameters, but Angular's `inject(TOKEN, { optional: true })` returns `T | null` so the `?? undefined` is required to match the expected types

kpanot (approved):

packages/@o3r/transloco/src/tools/localization-translate-directive.ts
```diff
@@ -0,0 +1,151 @@
...
+  standalone: false
+})
+export class LocalizationTranslateDirective extends TranslocoDirective implements OnInit, OnChanges {
+  private readonly localizationService = inject(LocalizationService);
+  private readonly localizationConfig = inject<LocalizationConfiguration>(LOCALIZATION_CONFIGURATION_TOKEN);
+  private readonly changeDetectorRef = inject(ChangeDetectorRef);
+  private readonly hostElement = inject(ElementRef);
+  private readonly hostRenderer = inject(Renderer2);
+  private readonly isAttributeStrategy = inject(TemplateRef, { optional: true }) === null;
+  private readonly o3rDestroyRef = inject(DestroyRef);
```
fpaul-1A, line 64: any reason to use a prefix here?
```suggestion
  private readonly destroyRef = inject(DestroyRef);
```

sdo-1A: I had to rename this to `o3rDestroyRef` since the `TranslocoDirective` has a private `destroyRef` property so there is a name collision between the two

packages/@o3r/transloco/src/tools/localization-translate-directive.ts
```diff
@@ -0,0 +1,151 @@
...
+   * Internal subscription to the LocalizationService key mapping
+   */
+  private onKeyChange?: Subscription;
+
+  /** @inheritdoc */
+  @Input()
+  public set transloco(key: string) {
+    if (key && key !== this.key) {
+      this.onKeyChange?.unsubscribe();
+      this.onKeyChange = this.localizationService.getKey(key)
```
fpaul-1A, line 81: __nitpick__ this seems a bit anti-pattern for reactive programing I think
Instead of `unsubscribe`/`subscribe`, I would consider using a subject for the key and the `switchMap` operator
The setter would use:
```typescript
this.translocoSubject.next(key);
```
And the subscription could be done once in the constructor with something like:
```typescript
this.translocoSubject.pipe(
  takeUntilDestroyed(),
  switchMap((key) => this.localizationService.getKey(key)),
  subscribe((newKey) => {
      const previousKey = this.key;
      this.key = newKey;
      this.changeDetectorRef.markForCheck();
      super.ngOnChanges({ transloco: new SimpleChange(previousKey, newKey, !previousKey) });
  });
```` 

maybe it's not the right PR for this kind of refactor though

File: packages/@o3r/transloco/src/tools/localization-translate-pipe.ts [hunk 19]
```diff
@@ -0,0 +1,112 @@
...
+   */
+  public transform(query: string, ...args: any[]): any {
+    if (!query || this.localizationService.showKeys()) {
+      return query;
+    }
+
+    if (query !== this.lastQueryKey) {
+      this.lastQueryKey = query;
+      if (this.onKeyChange) {
+        this.onKeyChange.unsubscribe();
```
fpaul-1A, line 94: same as previous comment, we could probably use `switchMap` to avoid these  manual `unsubscribe`

fpaul-1A (approved):

packages/@o3r/transloco/src/tools/localization-translate-directive.ts
sdo-1A: As discussed, I will refactor this in a separate PR

packages/@o3r/transloco/src/tools/localization-translate-pipe.ts
sdo-1A: Same as above, as discussed, I will refactor this in a separate PR

PR #4132, opened 2026-04-09
feat: ignore files based on git attributes

matthieu-crouzet:
## Related issues

*- No issue associated -*


sdo-1A (dismissed):

.gitattributes
```diff
@@ -12,6 +12,9 @@
...
 # Generated files
-tools/github-actions/*/packaged-action/** linguist-generated
-packages/@o3r-training/showcase-sdk/src/api/** linguist-generated
-packages/@o3r-training/showcase-sdk/src/models/base/** linguist-generated
+**/packaged-action/** linguist-generated
+**/src/api/** linguist-generated
+**/src/models/base/** linguist-generated
+**/src/spec/api-mock.ts linguist-generated
+**/src/spec/operation-adapter.ts linguist-generated
+**/src/constants/** linguist-generated
```
kpanot: Seems a bit too generic, why adding this pattern? Should be only applied to SDKs?

.gitattributes
```diff
@@ -12,6 +12,9 @@
...
 
 # Generated files
-tools/github-actions/*/packaged-action/** linguist-generated
-packages/@o3r-training/showcase-sdk/src/api/** linguist-generated
-packages/@o3r-training/showcase-sdk/src/models/base/** linguist-generated
+**/packaged-action/** linguist-generated
+**/src/api/** linguist-generated
+**/src/models/base/** linguist-generated
+**/src/spec/api-mock.ts linguist-generated
+**/src/spec/operation-adapter.ts linguist-generated
```
kpanot: Prefixing with `*sdk` seems to make it more restrictive

```suggestion
**/*sdk/src/api/** linguist-generated
**/*sdk/src/models/base/** linguist-generated
**/*sdk/src/spec/api-mock.ts linguist-generated
**/*sdk/src/spec/operation-adapter.ts linguist-generated
```

matthieu-crouzet: Indeed but then path is too restrictive to be used in the shared eslint config so I need to duplicate the same logic for each sdk package :/

packages/@ama-styling/figma-sdk/.gitattributes
```diff
@@ -1,8 +1,6 @@
...
-src/spec/api-mocks.ts linguist-generated
-src/spec/operation-adapter.ts linguist-generated
 src/api/** linguist-generated
 src/models/base/** linguist-generated
-
-# in case of core models, the `index.ts` file should be ignored as follows:
-# !src/models/base/<model>/index.ts
+src/spec/api-mock.ts linguist-generated
+src/spec/operation-adapter.ts linguist-generated
+src/constants/** linguist-generated
```
kpanot, line 6: I think it is missing in `packages/@ama-sdk/schematics/schematics/typescript/shell/templates/base/.gitattributes.template` as well, can you add it if missing?

matthieu-crouzet: it already exists :)

fpaul-1A (dismissed):

packages/@ama-styling/figma-sdk/.gitattributes
kpanot: Are you sure? I do not see this line on the target branch:
https://github.com/AmadeusITGroup/otter/blob/main/packages/%40ama-sdk/schematics/schematics/typescript/shell/templates/base/.gitattributes.template

matthieu-crouzet: I though you were speaking about the file itself, as it was not part of the PR

fpaul-1A (approved):

sdo-1A (approved):

kpanot (approved):

kpanot (approved):

fpaul-1A (approved):

PR #4134, opened 2026-04-10
fix: plugin name should not contains space

matthieu-crouzet:
## Related issues

*- No issue associated -*

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -6,7 +6,7 @@
   },
   "plugins": [
     {
-      "name": "Otter MCP servers recommendations",
+      "name": "otter",
       "source": "./tools/llm/plugins/otter",
       "tags": [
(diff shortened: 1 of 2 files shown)


mrednic-1A (approved):

sdo-1A (approved):

.claude-plugin/marketplace.json
```diff
@@ -6,7 +6,7 @@
   },
   "plugins": [
     {
-      "name": "Otter MCP servers recommendations",
```
kpanot, line 9: Maybe we can add it as `description`?

kpanot (approved):

.claude-plugin/marketplace.json
matthieu-crouzet: In fact the description is already present the plugin definition itself

PR #4140, opened 2026-04-13
feat: transloco package - devkit

sdo-1A:
## Proposed change

- Copy `src/devkit/` — replace `TranslateService` references with `TranslocoService` and `@ngx-translate/core` imports with `@jsverse/transloco`.
- Copy corresponding `*.spec.ts` files for devtools.

Copy `src/core/translate-message-format-lazy-compiler.ts` → `src/core/translate-message-format-lazy-transpiler.ts`:
- The old `TranslateMessageFormatLazyCompiler` (101 lines) was a full reimplementation of lazy ICU compilation because ngx-translate's ecosystem only offered eager compilation. It manually managed caching via a `Map<string, IntlMessageFormat>`, wrapped each key in a closure for lazy parsing, and directly depended on `intl-messageformat`.
- The new `TranslateMessageFormatLazyTranspiler` (30 lines) extends `MessageFormatTranspiler` from `@jsverse/transloco-messageformat`, which already provides lazy transpilation, built-in caching, and automatic locale management out of the box.
- The only addition is a `clearCache()` method (needed by devtools to force-reload translations at runtime), which calls the inherited `setLocale(null)` to recreate the internal `MessageFormat` instance and discard its cache.
- The file re-exports `MessageformatConfig` and `TRANSLOCO_MESSAGE_FORMAT_CONFIG` from `@jsverse/transloco-messageformat` so that consumers of `@o3r/transloco` can configure the transpiler without adding a direct dependency on `@jsverse/transloco-messageformat`.

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@o3r/transloco/src/devkit/localization-devkit-interface.ts
+++ b/packages/@o3r/transloco/src/devkit/localization-devkit-interface.ts
@@ -1,0 +1,142 @@
+import type {
+  ConnectContentMessage,
+  ContextualizationDevtoolsCommonOptions,
+  DevtoolsCommonOptions,
+  MessageDataTypes,
+  OtterMessageContent,
+  RequestMessagesContentMessage,
+} from '@o3r/core';
+import type {
+  LocalizationMetadata,
+} from '../core';
+
+/**
+ * Configuration options for the Localization Devtools service
+ */
+export interface LocalizationDevtoolsServiceOptions extends DevtoolsCommonOptions, ContextualizationDevtoolsCommonOptions {
+  /** Path to the localization metadata file */
+  metadataFilePath: string;
+}
+
+/** Localizations message content */
+export interface LocalizationsContentMessage extends OtterMessageContent<'localizations'> {
+  /** Localizations metadata */
+  localizations: LocalizationMetadata;
+}
+
+/** Languages message content */
+export interface LanguagesContentMessage extends OtterMessageContent<'languages'> {
+  /** Languages available */
+  languages: string[];
+}
+
+/** Switch languages message content */
+export interface SwitchLanguageContentMessage extends OtterMessageContent<'switchLanguage'> {
+  /** Language */
+  language: string;
+}
+
+/** Display localization key message content */
+export interface DisplayLocalizationKeysContentMessage extends OtterMessageContent<'displayLocalizationKeys'> {
+  /** Toggle the display of the localization keys */
+  toggle?: boolean;
+}
+
+/** Update localization message content */
+export interface UpdateLocalizationContentMessage extends OtterMessageContent<'updateLocalization'> {
+  /** Localization key */
+  key: string;
+  /** Localization value */
+  value: string;
+  /** Lang */
+  lang?: string;
+}
+
+/** Reload localization Keys message content */
+export interface ReloadLocalizationKeysContentMessage extends OtterMessageContent<'reloadLocalizationKeys'> {
+  /** Lang */
+  lang?: string;
+}
+
+/** Is translation deactivation enabled message content */
+export interface IsTranslationDeactivationEnabledContentMessage extends OtterMessageContent<'isTranslationDeactivationEnabled'> {
+  /** Whether translation deactivation is enabled */
+  enabled: boolean;
+}
+
+/** Get translation values message content */
+export interface GetTranslationValuesContentMessage extends OtterMessageContent<'getTranslationValuesContentMessage'> {
+  /** Translation key-value pairs */
+  translations: { [localizationKey: string]: string };
+}
+
+/** Union type of all localization message contents */
+type LocalizationMessageContents = LanguagesContentMessage
+  | ReloadLocalizationKeysContentMessage
+  | SwitchLanguageContentMessage
+  | LocalizationsContentMessage
+  | DisplayLocalizationKeysContentMessage
+  | UpdateLocalizationContentMessage
+  | IsTranslationDeactivationEnabledContentMessage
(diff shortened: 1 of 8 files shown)


packages/@o3r/transloco/src/devkit/localization-devtools.ts
```diff
@@ -0,0 +1,105 @@
...
+    this.appRef.tick();
+  }
+
+  /**
+   * Reload a language from the language file
+   * @param language language to reload
+   */
+  public async reloadLocalizationKeys(language?: string) {
+    const lang = language || this.getCurrentLanguage();
+    if ((this.translateTranspiler as TranslateMessageFormatLazyTranspiler).clearCache) {
```
mrednic-1A: can't we call `this.translateTranspiler.setLocale(null)` directly, like this we don;t need anymore this class _TranslateMessageFormatLazyTranspiler_

packages/@o3r/transloco/package.json
```diff
@@ -71,6 +75,8 @@
     "@compodoc/compodoc": "^1.1.32",
     "@eslint-community/eslint-plugin-eslint-comments": "^4.4.0",
     "@jsverse/transloco": "~8.2.1",
+    "@jsverse/transloco-messageformat": "~8.2.1",
+    "@jsverse/utils": "1.0.0-beta.5",
```
fpaul-1A: do we really need that? I can't find where this is being used

packages/@o3r/transloco/src/devkit/localization-devtools.ts
```diff
@@ -0,0 +1,105 @@
...
+   * @param fn called when the language is changed in the app
+   */
+  public onLanguageChange(fn: (language: string) => any): Subscription {
+    return this.localizationService
+      .getTranslateService()
+      .langChanges$
+      .subscribe((lang) => {
+        fn(lang);
+      });
+  }
```
fpaul-1A: we have no way of unregistering this callback?
this can lead to memory leaks in case the callback contains a reference to a component
when said component will be destroyed if won't be garbage collected because the reference will still be present in this service

I understand that this is probably never used in production mode? but it still looks like bad practice

packages/@o3r/transloco/package.json
sdo-1A: It is a dependency for `@jsverse/transloco`

sdo-1A: For information I plan on creating a PR to fix all root files (package.json, project.json, tsconfig)
It's difficult to split the PRs regarding the root files so I plan to fix these files at the end when each sub-part is merged

packages/@o3r/transloco/src/devkit/localization-devkit-interface.ts
```diff
@@ -0,0 +1,130 @@
...
+  RequestMessagesContentMessage,
+} from '@o3r/core';
+import {
+  Subscription,
+} from 'rxjs';
+import type {
+  LocalizationMetadata,
+} from '../core';
+
+export interface LocalizationDevtoolsServiceOptions extends DevtoolsCommonOptions, ContextualizationDevtoolsCommonOptions {
```
kpanot, line 16: missing comments

packages/@o3r/transloco/package.json
fpaul-1A: > It is a dependency for `@jsverse/transloco`

why do we need to provide it if it is a `dependency`?

fpaul-1A (approved):

packages/@o3r/transloco/package.json
sdo-1A: Nevermind I got confused, it was added to solve this comment from nx-cloud (above): 
https://github.com/AmadeusITGroup/otter/pull/4140#pullrequestreview-4104245748
But since we removed the dependency on `@jsverse/transloco-messageformat` we can remove the dependency

packages/@o3r/transloco/src/devkit/localization-devtools.ts
```diff
@@ -0,0 +1,125 @@
...
+        subscription.unsubscribe();
+        this.languageChangeSubscriptions.delete(subscription);
+      }
+    };
+  }
+
+  /**
+   * Clear all language change listeners
+   */
+  public clearLanguageChangeListeners(): void {
```
kpanot, line 79: Should not be called on the `onDestroy` of this injectable as well (as it is not a singleton)?

sdo-1A: I think this injectable is a singleton since it's provided at the module level in LocalizationDevtoolsModule so there's only one instance per application - this means that it lives for the entire application lifetime so ngOnDestroy would only be called when the app shuts down, so I don't think it's necessary for cleanup

sdo-1A: `clearLanguageChangeListeners()` is a utility for manual cleanup if needed

kpanot: It is not `provided: 'root'` so it is probable injected once per app but there is no technical limitation to be provided more than one.

I understand that it is a manual cleanup function but if this module is destroy, there is no solution to close the subscriptions in `languageChangeSubscriptions`

sdo-1A: Good point I will update by adding an ngOnDestroy and calling this function

fpaul-1A (approved):

mrednic-1A (approved):

PR #4141, opened 2026-04-13
feat: support Angular Strict template

kpanot:
## Proposed change

feat: support Angular Strict template

## Related issues

*- No issue associated -*


apps/chrome-devtools/src/app-devtools/localization-panel/localization-panel-pres.html
```diff
@@ -55,7 +55,7 @@ <h3 ngbAccordionHeader>
                 <ng-template>
                   <div class="form-group">
                     <div class="d-flex gap-4 align-items-center position-relative">
-                      <ng-container [formGroupName]="form.value.lang">
+                      <ng-container [formGroupName]="form.value.lang || null">
```
matthieu-crouzet: should it be null or should it be wrapped in an if-block?

apps/showcase/src/components/training/code-editor-view/code-editor-view.html
```diff
@@ -14,7 +14,7 @@
           @if (editorOptions()) {
             <ngx-monaco-editor class="h-100 position-relative"
                                [options]="editorOptions()"
-                               [model]="model$ | async"
+                               [model]="(model$ | async)!"
```
matthieu-crouzet, line 17: instead of `!` we can wrapped it in an if-block

apps/showcase/src/components/training/training-step/training-step-pres.html
```diff
@@ -12,8 +12,8 @@ <h2>{{title()}}</h2>
       <as-split-area [size]="instructions() ? 50 : 100">
         <div class="h-100 overflow-hidden">
           <code-editor-view
-            [project]="project()"
-            [editorMode]="editorMode()"
+            [project]="project()!"
```
matthieu-crouzet: same

apps/showcase/src/components/training/training.html
```diff
@@ -50,7 +50,7 @@ <h1 class="m-0 flex-grow-1">{{ title() }}</h1>
       [class.constrained]="!!(currentStep.dynamicContent.solutionProject() || currentStep.dynamicContent.project())"
       [instructions]="currentStep.dynamicContent.htmlContent()"
       [title]="currentStep.description.stepTitle"
-      [project]="showSolution() ? currentStep.dynamicContent.solutionProject() : currentStep.dynamicContent.project()"
+      [project]="(showSolution() ? currentStep.dynamicContent.solutionProject() : currentStep.dynamicContent.project())!"
```
matthieu-crouzet: why do we need the `!`, we have the hand on this component

apps/showcase/src/components/utilities/in-page-nav/in-page-nav-pres.ts
```diff
@@ -73,7 +73,7 @@ export class InPageNavPres implements OnChanges, OnDestroy {
 
   /** List of links */
   @Input()
-  public links: readonly InPageNavLink[] = [];
+  public links?: readonly InPageNavLink[] | null = [];
```
matthieu-crouzet: shouldn't we move to input-signal ?

packages/@o3r/rules-engine/src/components/rules-engine/rule-key-value/rule-key-value-pres-component.ts
```diff
@@ -49,13 +49,13 @@ export class RuleKeyValuePresComponent implements OnChanges {
    * Current value of the object
    */
   @Input()
-  public value!: string;
+  public value!: any;
```
matthieu-crouzet, line 52: Do we really support any or only primitive type ?
I fear that we will show [object Object] in the UI if we allow any instead of string

packages/@o3r/core/src/core/interfaces/context.ts
```diff
@@ -30,7 +31,7 @@ export type EventEmitterify<T extends BaseContextOutput> = { [P in keyof T]: Eve
 /**
  * Context of the component
  */
-export type Context<T extends ContextInput = {}, U extends BaseContextOutput = {}> = { [P in keyof T]: T[P] } & { [P in keyof U]: EventEmitterify<U>[P] };
+export type Context<T extends ContextInput = {}, U extends BaseContextOutput = {}> = { [P in keyof T]: (T[P] | InputSignal<T[P]>) } & EventEmitterify<U>;
```
matthieu-crouzet, line 34: Shouldn't the support of InputSignal be part of a dedicated PR with UT? :)

apps/chrome-devtools/src/app-devtools/app.html
```diff
@@ -14,7 +14,7 @@
             <li [ngbNavItem]="2.1">
               <a ngbNavLink>Facts</a>
               <ng-template ngbNavContent>
-                <o3r-facts-snapshot [facts]="facts$ | async" />
+                <o3r-facts-snapshot [facts]="(facts$ | async) || []" />
```
fpaul-1A: might be better to use a signal here to avoid creating a new instance of empty array each time

packages/@o3r/rules-engine/src/components/rules-engine/ruleset-history/ruleset-history-pres-component.html
```diff
@@ -108,9 +110,9 @@
                       <ng-container [ngTemplateOutlet]="triggers"
                       [ngTemplateOutletContext]="{triggers: (ruleEvaluation.triggers[ruleEvaluation.rule.id])}" />
                       <o3r-rule-actions-pres
-                        [actions]="ruleEvaluation.outputActions"
-                        [temporaryFacts]="ruleEvaluation.temporaryFacts"
-                        [runtimeOutputs]="execution.rulesetInformation?.rules[index]?.outputRuntimeFacts" />
+                        [actions]="ruleEvaluation.outputActions || []"
+                        [temporaryFacts]="ruleEvaluation.temporaryFacts || {}"
+                        [runtimeOutputs]="rule?.outputRuntimeFacts || []" />
```
fpaul-1A: why don't we keep the undefined values? creating new empty instances can mess with the change detection

apps/chrome-devtools/src/app-devtools/app.html
kpanot: The `|| []` is because the `| async` can return `undefined`, not the observable.
I add a `@if` instead

packages/@o3r/core/src/core/interfaces/context.ts
kpanot: Created the issue #4143
This change need to remain in this PR to unlock blocking type issue.

packages/@o3r/rules-engine/src/components/rules-engine/rule-key-value/rule-key-value-pres-component.ts
kpanot: We do use it with `any` today, `JSON.stringify` is used to display object (see below of this file).
The `string` type seems to be an issue which has not been fixed before because there was not type check on the template

packages/@o3r/rules-engine/src/components/rules-engine/ruleset-history/ruleset-history-pres-component.html
kpanot: `RuleActionsPresComponent` does not allow `undefined` (neither `null`) as input.
I do the change in `RuleActionsPresComponent` to support `undefined` if you prefer but it is similar.

fpaul-1A: it's probably better to change the component to support `undefined`
`[]` and `{}` will always be considered as different from previous value (`[] !=== []`), so it will trick Angular into thinking that the input has changed even if it's not the case

apps/showcase/src/components/training/code-editor-view/code-editor-view.html
kpanot: Due to issue on `ngx-monaco-editor` it can not be done properly.
It needs to be initialized with a value, even `null` (without a type allowing it on their side).

fpaul-1A (commented):
~~Those should be removed as well~~

packages/@o3r/rules-engine/src/components/rules-engine/ruleset-history/ruleset-history-pres-component.html
```diff
@@ -78,12 +79,12 @@
...
                                   inputs: execution.inputFacts,
-                                  runtimeInputs: execution.rulesetInformation?.rules[index]?.inputRuntimeFacts
+                                  runtimeInputs: rule?.inputRuntimeFacts
                                 }" />
                         @if (!ruleEvaluation.error) {
                           <o3r-rule-actions-pres
-                            [temporaryFacts]="ruleEvaluation.temporaryFacts"
-                            [runtimeOutputs]="execution.rulesetInformation?.rules[index]?.outputRuntimeFacts"
+                            [temporaryFacts]="ruleEvaluation.temporaryFacts || {}"
+                            [runtimeOutputs]="rule?.outputRuntimeFacts || []"
```
fpaul-1A: These should be removed as well

fpaul-1A (dismissed):

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #4156, opened 2026-04-16
feat: renovate post upgrade install from private reg

mrednic-1A:
## Proposed change

Add `-with-auth` variants of the existing Renovate task presets so that post-upgrade tasks (`npm install`, `yarn install`, `ng update`, SDK regeneration, spec upgrade) can run against private npm registries.

When the registry args are provided, the task configures auth on the Renovate runner before running the install:
- **npm / Yarn 1** — sets the user-level npm config (`npm config set --location=user registry …` and `…:_auth …`).
- **Yarn 2+** — sets `npmRegistryServer` and `npmRegistries[…].npmAuthIdent` via `yarn config set -H` (under `tools/renovate/yarn2/`).

When the args are empty, no auth is configured and the existing install behavior is preserved, so the new variants are drop-in replacements for repos that don't need a private registry.

The token is supplied through an env var name (e.g. `NPM_AUTH_BASE64`) that the consuming repo exposes via Renovate's `customEnvVariables`, so no credential is ever stored in the preset.

Documentation: usage, arg conventions, and Yarn 2 vs Yarn 1 differences are documented in `tools/renovate/README.md`.

## Related issues

*- No issue associated -*


tools/renovate/tasks/base.json
```diff
@@ -11,7 +11,7 @@
       ],
       "postUpgradeTasks": {
         "commands": [
-          "npm install"
+          "{{#if (equals '{{arg0}}' '')}}npm install{{else}}NPM_CONFIG_USERCONFIG={{arg0}} npm install{{/if}}"
```
kpanot: Should not be surrounded by `'` to support space in paths?
```suggestion
          "{{#if (equals '{{arg0}}' '')}}npm install{{else}}NPM_CONFIG_USERCONFIG='{{arg0}}' npm install{{/if}}"
```

mrednic-1A: done

tools/renovate/tasks/base-private-reg.json
```diff
@@ -0,0 +1,47 @@
...
+    {
+      "matchDepTypes": [
+        "packageManager"
+      ],
+      "matchPackageNames": [
+        "yarn"
+      ],
+      "postUpgradeTasks": {
+        "commands": [
+          "NPM_CONFIG_USERCONFIG='{{arg0}}' yarn install"
```
fpaul-1A: is that truly doing anything for yarn? probably not yarn berry at least

mrednic-1A: yes, yarn 1 is keeping a 'backward compatibility' with npm via this npm_config_*
https://classic.yarnpkg.com/en/docs/envvars#npm_config

for yarn berry this env var will do nothing, and auth the setup is different and based only on host_rules.

I have added explanations in the readme

tools/renovate/tasks/base-private-reg.json
```diff
@@ -0,0 +1,47 @@
+{
```
kpanot: Maybe you should give a name more explicit to this task file

tools/renovate/tasks/base-private-reg.json
kpanot: Why `yarn` is not part of the argument like the other tasks?
Maybe you can define both `NPM_CONFIG_USERCONFIG` and `YARN_RC_FILENAME` to cover the 2 cases (when 1 is used the other is ignored so no conflict).

tools/renovate/tasks/otter-ng-update-private-reg.json
```diff
@@ -0,0 +1,21 @@
...
+      "matchPackageNames": [
+        "@o3r/core"
+      ],
+      "postUpgradeTasks": {
+        "commands": [
+          "NPM_CONFIG_USERCONFIG='{{arg1}}' {{arg0}} install",
+          "{{arg0}} run ng update {{{depName}}} {{#if (equals '{{arg0}}' 'npm')}}-- {{/if}}--from={{{currentVersion}}} --to={{{newVersion}}} --migrate-only --allow-dirty --force"
+        ],
+        "fileFilters": [
+          "!**/.{npmrc,yarnrc*}"
```
kpanot: you may need to add the `{{arg1}}` which may have a complet different name

tools/renovate/sdk-private-reg.json
```diff
@@ -0,0 +1,11 @@
+{
```
kpanot: I do not think you should have name "private" here which is not meaningful.

tools/renovate/tasks/base-private-reg.json
mrednic-1A: There is no strong constraint about why we do not have arguments for this. I duplicated what we already have, separated configs. https://github.com/AmadeusITGroup/otter/blob/main/tools/renovate/tasks/base.json

Having YARN_RC_FILENAME is redundant, as yarn 2+ is capable of creating its registries (even for post upgrade tasks) from the host rules. 
https://docs.renovatebot.com/getting-started/private-packages/#yarn-2

tools/renovate/sdk-private-reg.json
mrednic-1A: it's true it is not really meaningful, because we need this if the project itself is using private registries so the 'install' need auth.

tools/renovate/yarn2/tasks/base-with-auth.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "description": "Trigger post install scripts when updating package manager, with private registry authentication for Yarn Berry (Yarn >= 2). Configures Yarn home config (~/.yarnrc.yml) on the Renovate runner so 'yarn install' can authenticate against a private registry, then runs the install. arg0: scheme-less registry URL with trailing slash (e.g. '//npm.my-company.example/api/npm/npm/'); used to derive 'https:{{arg0}}' for npmRegistryServer and as the key for 'npmRegistries[...].npmAuthIdent'. arg1: name of the environment variable holding the base64-encoded 'user:password' token (e.g. 'NPM_AUTH_BASE64'); must be exposed to Renovate via customEnvVariables. If your registry uses a bearer token instead, replace npmAuthIdent with npmAuthToken in this preset. References: https://yarnpkg.com/configuration/yarnrc ; https://yarnpkg.com/cli/config/set ; https://docs.renovatebot.com/getting-started/private-packages/#yarn-2",
+  "packageRules": [
+    {
+      "matchDepTypes": [
```
kpanot: You can add [MatchCurrentVersion](https://docs.renovatebot.com/configuration-options/#packagerulesmatchcurrentversion):
```json
"matchCurrentVersion": ">2.0.0"
```
to restrain to yarn 2.

When done, you can extend `tools/renovate/tasks/base-with-auth.json({{arg0}}, {{arg1}})` and expose only this file (as it will handle `npm`, `yarn` and `yarn berry`.

tools/renovate/base.json
```diff
@@ -11,10 +11,10 @@
     "group:linters",
     "helpers:pinGitHubActionDigestsToSemver",
     ":automergeMinor",
-    "github>AmadeusITGroup/otter//tools/renovate/group/angular",
-    "github>AmadeusITGroup/otter//tools/renovate/group/typescript",
-    "github>AmadeusITGroup/otter//tools/renovate/managers/github-node-version.json5",
-    "github>AmadeusITGroup/otter//tools/renovate/managers/mcp-servers.json5"
+    "github>AmadeusITGroup/otter//tools/renovate/group/angular#feat/renovate-install-private-reg",
```
kpanot: to be rollback

tools/renovate/otter-project.json
```diff
@@ -2,12 +2,12 @@
...
   "description": "Group all otter dependencies and trigger ng update scripts with yarn",
   "extends": [
-    "github>AmadeusITGroup/otter//tools/renovate/group/otter",
-    "github>AmadeusITGroup/otter//tools/renovate/group/mfe",
-    "github>AmadeusITGroup/otter//tools/renovate/tasks/base",
-    "github>AmadeusITGroup/otter//tools/renovate/branching-strategy/release-branches",
-    "github>AmadeusITGroup/otter//tools/renovate/tasks/otter-ng-update(yarn)",
-    "github>AmadeusITGroup/otter//tools/renovate/tasks/yarn-pnp",
-    "github>AmadeusITGroup/otter//tools/renovate/managers/renovate-otter-presets.json5"
+    "github>AmadeusITGroup/otter//tools/renovate/group/otter#feat/renovate-install-private-reg",
```
kpanot: to be rollback

tools/renovate/README.md
```diff
@@ -108,3 +115,110 @@ If you are using npm package manager instead of yarn:
...
 }
 ```
+
+## Private registry support
+
+If your project uses a private npm registry, the post-upgrade install commands need access to your registry credentials. The approach differs depending on the package manager.
+
+### npm and Yarn 1 (Classic)
+
+Create a dedicated `.npmrc` file (e.g. `.npmrc-renovate`) in your repository with the registry configuration:
```
kpanot:
```suggestion
Create a dedicated [.npmrc](https://docs.npmjs.com/cli/configuring-npm/npmrc) file (e.g. `.npmrc-renovate`) in your repository with the registry configuration:
```

tools/renovate/tasks/base-private-reg.json
mrednic-1A: done

tools/renovate/sdk-private-reg.json
mrednic-1A: done

tools/renovate/yarn2/tasks/base-with-auth.json
mrednic-1A: done

kpanot (approved):

tools/renovate/tasks/base-with-auth.json5
```diff
@@ -0,0 +1,59 @@
...
+      "matchPackageNames": [
+        "npm",
+        "yarn"
+      ],
+      "postUpgradeTasks": {
+        // TODO: In v15, remove the {{else}}{{{depName}}} install part from the conditional commands to avoid redundant execution time.
+        // The else branch should be empty (or use 'true' or ';' as no-op). Renovate bots will need to update their allowed commands configuration
+        // to include an empty/no-op command. Details to be included in v15 migration guide.
+        "commands": [
+          "{{#unless (equals '{{arg0}}' '')}}npm config set --location=user registry https:{{arg0}}{{else}}{{{depName}}} install{{/unless}}",
```
kpanot, line 18:
```suggestion
          "{{#unless (equals '{{arg0}}' '')}}npm config set --location=user registry \"https:{{arg0}}\"{{else}}{{{depName}}} install{{/unless}}",
```

fpaul-1A (approved):

tools/renovate/tasks/base-with-auth.json5
mrednic-1A: the registry url is normally not containing special characters. 
I don;t think it is really necessary. Also on line 20 I am using the same arg0 without escape

tools/renovate/tasks/otter-ng-update-private-reg.json
mrednic-1A: outdated, the arg1 was the name of the npmrc custom file, but that solution is dropped

PR #4195, opened 2026-05-05
fix(ama-styling): fix metadata formatter edge case when token name ha…

rglearns:
### Summary
This fixes metadata generation when token paths like `{input.color.foreground.value}` reference tokens nested within a structural parent object. In DTCG mode, `.value` is treated as a literal path segment rather than a token-value accessor, causing path resolution to return the parent object instead of the leaf token.

### Root Cause
When building metadata for a token like `input-color-foreground-label` that references `{input.color.foreground.value}`:

1. In DTCG mode, the `.value` suffix is NOT stripped (only `.$value` is stripped by Style Dictionary's getReferences)
2. `getReferences` tries to resolve the full path `input.color.foreground.value`
3. The token hierarchy has `foreground` as a parent containing both `value` and `label` as sibling tokens:
   ```json
   "foreground": {
     "value": { "$type": "color", "$value": "#545352" },
     "label": { "$type": "color", "$value": "{input.color.foreground.value}" }
   }
4.Style Dictionary's getValueByPath traverses to foreground and returns that parent object instead of drilling into value
5.The metadata formatter receives { value: <actual token>, label: <another token> } (no name property on the wrapper)
6.It tries to read properties like ref.name and ref.original from this parent wrapper and fails with unnamed-ref  error

### DIFF
--- a/packages/@ama-styling/style-dictionary/src/formats/metadata-format.mts
+++ b/packages/@ama-styling/style-dictionary/src/formats/metadata-format.mts
@@ -21,6 +21,17 @@ import type {
   FormatterOptions,
 } from './css-formatters/interface-formatter.mjs';
 
+const unwrapReferenceToken = (ref: TransformedToken | { value: TransformedToken }): TransformedToken => {
+  // Unwrap when getReferences returns a parent structural node instead of the leaf token.
+  // This occurs when token path resolution yields a parent object with child tokens,
+  // e.g., { label: token, value: token } instead of the token itself.
+  // Safe to unwrap only when outer has no name (not a token) but inner.value has name (is a token).
+  if (ref && typeof ref === 'object' && 'value' in ref && !('name' in ref) && ref.value?.name) {
+    return ref.value;
+  }
+  return ref as TransformedToken;
+};
+
 export const metadataFormat: Format = {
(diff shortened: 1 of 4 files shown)


packages/@ama-styling/style-dictionary/src/formats/metadata-format.mts
```diff
@@ -63,15 +63,29 @@ export const metadataFormat: Format = {
     const getValueFromCssVariable = (strValue: string) =>
       strValue.replace(new RegExp(`^.*?${separator}(.*?)[ ;]*$`), '$1').trim();
 
+    const unwrapReferenceToken = (ref: TransformedToken | { value: TransformedToken }): TransformedToken => {
```
matthieu-crouzet: I don't think `unwrapReferenceToken` should be defined in the scope of this function as it does not use any context.

rglearns: Indeed , moved out 
I let you review and resolve the conversation if fine for you

kpanot (dismissed):

kpanot (approved):

mrednic-1A (approved):

PR #4220, opened 2026-05-19
feat(ama-openapi): new preprocessor to enforce discriminator to be req

kpanot:
## Proposed change

feat(ama-openapi): new preprocessor to enforce discriminator to be required

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@ama-openapi/redocly-plugin/src/plugins/decorators/common/enforce-discriminator-mandatory/enforce-discriminator-mandatory.decorator.mts
+++ b/packages/@ama-openapi/redocly-plugin/src/plugins/decorators/common/enforce-discriminator-mandatory/enforce-discriminator-mandatory.decorator.mts
@@ -1,0 +1,75 @@
+import type {
+  Async2Preprocessor,
+  Async3Preprocessor,
+  Oas2Preprocessor,
+  Oas3Preprocessor,
+} from '@redocly/openapi-core';
+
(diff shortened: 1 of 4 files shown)


mrednic-1A (approved):

packages/@ama-openapi/redocly-plugin/src/plugins/decorators/common/enforce-discriminator-mandatory/enforce-discriminator-mandatory.decorator.mts
```diff
@@ -0,0 +1,75 @@
...
+ *       ama-openapi/enforce-discriminator-mandatory: on
+ * ```
+ */
+export const enforceDiscriminatorMandatoryDecorator: Preprocessor = (options: EnforceDiscriminatorMandatoryDecoratorOptions) => {
+  return {
+    Schema: {
+      // use `any` type because schema OAS 2.0 is not exposed in the core, and we want to support both OAS 2.0 and 3.0 schemas
+      leave: (node: any) => {
+        // Check if schema has a discriminator
+        if (!node.discriminator) {
```
vscaiceanu-1a, line 47: why not doing this in `skip()`?

kpanot: To not skip the sub-schema in case it needs to go deeper.
(I did a `skip` first and got the issue so I rollback to a simple `if`)

vscaiceanu-1a (approved):

PR #4228, opened 2026-05-21
fix: o3r-set-version for wildcards in files

mrednic-1A:
### Problem
The wildcard version regex `/"\*(.+)"(,?)$/gm` was too broad and incorrectly matched file glob patterns in the `files` array, transforming `"**.metadata.json"` → `"=*.metadata.json"`.

### Solution
Changed regex to `/"\*([0-9]+\.[0-9]+\.[0-9]+[^"]*)"(,?)$/gm` to only match actual semver versions (e.g., `"*1.2.3"` → `"=1.2.3"`), while preserving glob patterns like `"**.metadata.json"`, `"*.js"`, etc.

### Changes
- **Fixed**: Wildcard version regex to require semver pattern
- **Extracted**: Regexes to `set-version-regexes.ts` for shared use
- **Added**: Comprehensive test suite (66 tests) covering:
  - Placeholder version replacement (main feature)
  - Workspace protocol transformations
  - Wildcard version transformations (bug fix)
  - Private field removal
  - Full transformation chains
  - Edge cases

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@o3r/workspace/src/cli/set-version-regexes.ts
+++ b/packages/@o3r/workspace/src/cli/set-version-regexes.ts
@@ -1,0 +1,35 @@
+/**
+ * Escapes special regex characters in a string to create a literal match pattern.
+ * @param str - The string to escape (e.g., "0.0.0(-placeholder)?")
+ * @returns The escaped string safe for use inside a RegExp constructor
+ */
+export function escapeForRegex(str: string): string {
+  return str.replace(/[$()*+./?[\\\]^{|}-]/g, '\\$&');
+}
+
+/**
+ * Creates a regex to match placeholder versions (e.g., "0.0.0", "0.0.0-placeholder")
+ * @param placeholder - The placeholder pattern (default: "0.0.0(-placeholder)?")
+ * @returns RegExp that captures the version prefix (~, ^) in $1
+ */
+export function createPlaceholderRegex(placeholder: string): RegExp {
+  return new RegExp('"([~^]?)' + escapeForRegex(placeholder) + '"', 'g');
+}
+
+/**
+ * Regex to match workspace protocol versions (e.g., "workspace:*", "workspace:~1.0.0")
+ * Captures: $1 = prefix (~, ^, or *), $2 = trailing comma if present
+ */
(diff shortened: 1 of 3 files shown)


packages/@o3r/workspace/src/cli/set-version-regexes.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * Creates a regex to match placeholder versions (e.g., "0.0.0", "0.0.0-placeholder")
+ * @param placeholder - The placeholder pattern (default: "0.0.0(-placeholder)?")
+ * @returns RegExp that captures the version prefix (~, ^) in $1
+ */
+export function createPlaceholderRegex(placeholder: string): RegExp {
+  return new RegExp('"([~^]?)' + placeholder
+    .replace(/[$()+.?[\\\]^{|}]/g, '\\$&')
```
kpanot: `escapeRegExp` would not do the job?

packages/@o3r/workspace/src/cli/set-version-regexes.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * Creates a regex to match placeholder versions (e.g., "0.0.0", "0.0.0-placeholder")
+ * @param placeholder - The placeholder pattern (default: "0.0.0(-placeholder)?")
+ * @returns RegExp that captures the version prefix (~, ^) in $1
+ */
+export function createPlaceholderRegex(placeholder: string): RegExp {
```
kpanot, line 15: Maybe you should add UT for this function as it is not simple to read

packages/@o3r/workspace/src/cli/set-version-regexes.ts
mrednic-1A: I just extracted the one we had in the main file, without any change. I'll do a test for the escape

mrednic-1A: updated

packages/@o3r/workspace/src/cli/set-version-regexes.ts
mrednic-1A: updated

kpanot (approved):

matthieu-crouzet (approved):

PR #4229, opened 2026-05-22
fix(ama-openapi): schema $id to make it readable by claude

kpanot:
## Proposed change

fix(ama-openapi): schema $id to make it readable by claude
fix(ama-openapi): removed vscode config allowed only at user level
docs(ama-openapi): to give instruction to user to setup copilot/claude
fix(ama-openapi): duplicate claude instruction to be used by github agents

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@ama-openapi/create/templates/.vscode/settings.json.template
+++ b/packages/@ama-openapi/create/templates/.vscode/settings.json.template
@@ -1,8 +1,5 @@
 {
   "chat.plugins.enabled": true,
-  "chat.plugins.marketplaces": [
-    "AmadeusITGroup/otter"
-  ],
   "editor.defaultFormatter": "Redocly.openapi-vs-code",
   "files.readonlyFromPermissions": true,
   "files.readonlyInclude": {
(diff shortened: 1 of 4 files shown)


packages/@ama-openapi/create/templates/.vscode/settings.json.template
```diff
@@ -1,8 +1,5 @@
 {
   "chat.plugins.enabled": true,
-  "chat.plugins.marketplaces": [
```
mrednic-1A, line 3: we do not want to have this in a new generated project ?
we want them to have it at user settings ?

mrednic-1A (approved):

packages/@ama-openapi/create/templates/.vscode/settings.json.template
kpanot: This settings is not allowed by VsCode at project level but it is also not required. VsCode can determine the additional marketplace based on the `.claude/settings.json` file.

vscaiceanu-1a (approved):

PR #4255, opened 2026-06-05
fix: localization & transloco optional in components package

mrednic-1A:
## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@o3r/components/src/rules-engine/placeholder-action-handler.ts
+++ b/packages/@o3r/components/src/rules-engine/placeholder-action-handler.ts
@@ -17,23 +17,24 @@ import {
   DynamicContentService,
 } from '@o3r/dynamic-content';
 import {
-  LocalizationService,
-} from '@o3r/localization';
-import {
   LoggerService,
 } from '@o3r/logger';
 import {
   combineLatest,
   distinctUntilChanged,
   firstValueFrom,
+  from,
   map,
+  Observable,
   of,
   startWith,
   Subject,
+  switchMap,
   withLatestFrom,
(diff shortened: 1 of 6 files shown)


matthieu-crouzet (approved):

packages/@o3r/components/src/rules-engine/placeholder-action-handler.ts
```diff
@@ -54,6 +55,15 @@ import {
 export class PlaceholderRulesEngineActionHandler implements RulesEngineActionHandler<ActionUpdatePlaceholderBlock> {
   private readonly logger = inject(LoggerService);
   private readonly injector = inject(Injector);
+  // Prefer `@o3r/transloco`, fall back to `@o3r/localization`; both are optional peer dependencies.
+  // TODO: use `injectAsync` to resolve the optional `LocalizationService` token once migrated to ng22
+  // TODO: remove the dynamic-import fallback and inject the `LocalizationService` from the single supported translation package in v16
+  private readonly translationService: Promise<PlaceholderLocalizationService | undefined> = import('@o3r/transloco')
```
kpanot, line 61: Can you add a `todo` to migrate to `asyncInject` when migrating to `ng22`?

mrednic-1A: It is added, no ?

kpanot: Indeed, not sure why I missed it :S

kpanot (approved):

PR #4286, opened 2026-06-17
fix: add readonly on message type ama-mfe

mrednic-1A:
## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@ama-mfe/ng-utils/src/managers/interfaces.ts
+++ b/packages/@ama-mfe/ng-utils/src/managers/interfaces.ts
@@ -33,6 +33,7 @@ export interface MessageConsumer<T extends VersionedMessage = VersionedMessage> 
 
   /** The message type which will be handled */
   type: T['type'];
+
   /** The map of functions which will be executed for each supported version */
   supportedVersions: {
     /**
@@ -52,7 +53,7 @@ export interface MessageConsumer<T extends VersionedMessage = VersionedMessage> 
  */
 export interface MessageProducer<T extends VersionedMessage = VersionedMessage> {
   /** The types of the produced message */
-  types: T['type'] | T['type'][];
+  types: T['type'] | readonly T['type'][];
 
   /**
    * The producer should handle an error in case that it has produced a message which cannot be consumed by other peers


packages/@ama-mfe/ng-utils/src/managers/interfaces.ts
```diff
@@ -52,7 +50,7 @@ export interface MessageConsumer<T extends VersionedMessage = VersionedMessage>
  */
 export interface MessageProducer<T extends VersionedMessage = VersionedMessage> {
   /** The types of the produced message */
-  types: T['type'] | T['type'][];
+  types: T['type'] | T['type'][] | Readonly<T['type']> | readonly T['type'][];
```
kpanot:
```suggestion
  types: T['type'] | T['type'][] | Readonly<T['type']> | Readonly<T['type'][]>;
```
?

mrednic-1A: It is the linter changing it before commit. 
my intention  is `ReadonlyArray<T['type']>`.
Should I bypass the linter ?

packages/@ama-mfe/ng-utils/src/managers/interfaces.ts
```diff
@@ -23,16 +23,14 @@ export interface MessageVersions<T extends VersionedMessage> {
 /** Description for the base message consumer */
 export interface BasicMessageConsumer<T extends VersionedMessage = VersionedMessage> {
   /** Contains the message type */
-  type: T['type'];
+  type: T['type'] | Readonly<T['type']>;
```
fpaul-1A: is it not redundant typing?
```suggestion
  type: Readonly<T['type']>;
```

packages/@ama-mfe/ng-utils/src/managers/interfaces.ts
kpanot: If it is working, let's follow the linter then.
What's about `Readonly<T['type']>` to `readonly T['type']` then?

packages/@ama-mfe/ng-utils/src/managers/interfaces.ts
mrednic-1A: Good catch — and actually Readonly<T['type']> itself is also redundant: Readonly<> is a mapped type, so applied to a primitive/string-literal like T['type'] it's a no-op (Readonly<'theme'> ≡ 'theme'). 
I reverted the change

packages/@ama-mfe/ng-utils/src/managers/interfaces.ts
mrednic-1A: Readonly has no effect on the constant, so I kept it only for the array.

fpaul-1A (approved):

kpanot (approved):

PR #4289, opened 2026-06-19
feat: ama-mfe nav exit

mrednic-1A:
## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@ama-mfe/ng-utils/package.json
+++ b/packages/@ama-mfe/ng-utils/package.json
@@ -33,7 +33,7 @@
     "postbuild": "patch-package-json-exports",
     "prepare:compile": "cp-package-json",
     "build": "yarn nx build ama-mfe-ng-utils",
-    "prepare:build:builders": "yarn cpy 'schematics/**/*.json' dist/schematics && yarn cpy 'collection.json' dist",
+    "prepare:build:builders": "yarn cpy 'schematics/**/*.json' dist/schematics && yarn cpy 'collection.json' 'migration.json' dist",
     "build:builders": "tsc -b tsconfig.builders.json --pretty && yarn generate-cjs-manifest"
   },
   "devDependencies": {
@@ -93,6 +93,9 @@
     "communication protocol"
   ],
   "schematics": "./collection.json",
+  "ng-update": {
+    "migrations": "./migration.json"
+  },
   "peerDependenciesMeta": {
     "@ama-mfe/messages": {
       "optional": true
--- a/packages/@ama-mfe/ng-utils/schematics/ng-update/v14-5/index.ts
+++ b/packages/@ama-mfe/ng-utils/schematics/ng-update/v14-5/index.ts
@@ -1,0 +1,78 @@
+import {
+  noop,
+  type Rule,
+  type SchematicContext,
+  type Tree,
+} from '@angular-devkit/schematics';
+import {
+  applyEsLintFix,
+  createOtterSchematic,
+  findFilesInTree,
+} from '@o3r/schematics';
+
+/**
+ * In-repo {@link AbstractMessageConsumer} subclasses that used to auto-start in their constructor.
+ * In v15 the constructor is removed, so the application must start them explicitly.
+ */
+const KNOWN_CONSUMERS = [
+  'NavigationConsumerService',
+  'ThemeConsumerService',
+  'HistoryConsumerService',
+  'ResizeConsumerService'
+] as const;
+
+/**
+ * For every known consumer injected in the given file, append a `.start()` call to its `inject(<Consumer>)`
+ * expression. Idempotent: an `inject(<Consumer>)` already followed by `.start(` is left untouched.
+ * @param tree Schematic tree
+ * @param filePath Path of the `app.config.ts` / `main.ts` file
+ * @param context Schematic context for logging
+ */
+const appendStartToInjectedConsumers = (tree: Tree, filePath: string, context: SchematicContext): void => {
+  const original = tree.readText(filePath);
+  let updated = original;
+  const started: string[] = [];
+
+  for (const consumer of KNOWN_CONSUMERS) {
+    // Match `inject(<Consumer>)` that is not already followed by `.start(`.
+    const injectRegex = new RegExp(`inject\\(\\s*${consumer}\\s*\\)(?!\\s*\\.start\\b)`, 'g');
+    if (injectRegex.test(updated)) {
+      updated = updated.replace(injectRegex, `inject(${consumer}).start()`);
+      started.push(consumer);
+    }
+  }
+
+  if (updated !== original) {
+    tree.overwrite(filePath, updated);
+    context.logger.info(`[ama-mfe ng-update] Added explicit ".start()" call for ${started.join(', ')} in ${filePath}.`);
+  }
+};
+
+/**
+ * Migration to v14.5.0 of `@ama-mfe/ng-utils`.
+ *
+ * The consumers no longer auto-start in their constructor. This migration scans the
+ * application configuration files (`app.config.ts` / `main.ts`) and, for every known consumer already injected there,
+ * appends an explicit `.start()` call so the consumer keeps being started.
+ */
+// eslint-disable-next-line @typescript-eslint/naming-convention -- version contained in the function name
+function updateV14_5Fn(): Rule {
+  return (tree: Tree, context: SchematicContext) => {
+    const files = findFilesInTree(
+      tree.getDir('/'),
+      (filePath) => /\b(?:app\.config|main)\.ts$/.test(filePath) && !/\bdist\b/.test(filePath)
+    );
+    for (const file of files) {
+      appendStartToInjectedConsumers(tree, file.path, context);
+    }
+    return applyEsLintFix() ?? noop();
+  };
+}
+
+/**
+ * Migration of `@ama-mfe/ng-utils` to v14.5.0.
+ *
+ * Appends explicit `.start()` calls for the in-repo consumers that no longer auto-start.
+ */
+// eslint-disable-next-line @typescript-eslint/naming-convention -- version contained in the function name
+export const updateV14_5 = createOtterSchematic(updateV14_5Fn);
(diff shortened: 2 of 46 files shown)


packages/@ama-mfe/ng-utils/src/managers/abstract-message-consumer-service.ts
```diff
@@ -34,16 +34,6 @@ export abstract class AbstractMessageConsumer<T extends VersionedMessage = Versi
   /** @inheritdoc */
   public abstract readonly supportedVersions: MessageConsumer<T>['supportedVersions'];
 
-  /**
-   * Creates an instance of AbstractMessageConsumer and optionally starts it.
-   * @param autoStart If true, the consumer will be started immediately after creation. Defaults to true.
-   */
-  constructor(autoStart = true) {
```
matthieu-crouzet, line 41: Why do we cancel this change ?

mrednic-1A: There was a misalignment between the consumer services. 
At app level (shell app) all the existing services are calling themselves the `.start()` method. 
For the new services in this PR I wanted to keep the same pattern, but I had to call the constructor with false (`super(false)`). Moreover the other services were autostarted here too (so there was a double start for the existing services already). 
In the end I opted to remove the constructor and let the user start the service whenever is needed. (no change in the app, start is still called for the existing services and it will be called for the new ones too).

packages/@ama-mfe/ng-utils/src/managers/abstract-message-consumer-service.spec.ts
```diff
@@ -31,8 +31,9 @@ class TestConsumer extends AbstractMessageConsumer<TestMessage> {
     '1.0': ((_message: RoutedMessage<TestMessage & { version: '1.0' }>) => {}) as MessageCallback<TestMessage & { version: '1.0' }>
   };
 
-  constructor(autoStart = true) {
-    super(autoStart);
+  constructor() {
```
kpanot, line 37: Following you previous comment, I think you should add a warning to indicate that the "autoStart" is deprecated, potentially be removed in next major and it should now add the `.start()` in the AppConfig.
(I think a simple schematic doing the job would be nice as well)

packages/@ama-mfe/ng-utils/src/navigation/navigation-consumer-service.ts
```diff
@@ -77,10 +77,6 @@ export class NavigationConsumerService extends AbstractMessageConsumer<Navigatio
     }
   };
 
-  constructor() {
```
kpanot, line 80: This is a breaking change (as it is not autoStarted anymore).
This change should be done on next major and/or warned with a message

mrednic-1A: The abstract service was added in this prerelease 14.5.0-prerelease.* 
The file does not exist in release/14.4.0-rc branch, I would say we are doing a breaking change in a prerelease so it is ok.

packages/@ama-mfe/ng-utils/src/managers/abstract-message-consumer-service.spec.ts
mrednic-1A: I added in the service constructor. 
I can;t do a schematic here because it has to be done when the constructor will be removed, so in a v15 branch, 
But I created an issue to track this 
https://github.com/AmadeusITGroup/otter/issues/4309

__EDIT__

Now the final version is:
-removed the constructor from the abstract service - no breaking as it was added in this prerelease
-revert the call to start in the constructor in the services extending the abstract service and deprecate this (to keep the backward compatibility - autostart)
-new consumers do not call the start in their constructor 
-schematic to add the .start() method in apps for the imported consumers (there is no harm even if there is already a call to start in the constructor)

packages/@ama-mfe/ng-utils/src/managers/abstract-message-consumer-service.ts
mrednic-1A: after discussions, I only deprecate the autostart feature in this PR and the removal will be done in v15. 
__EDIT__ 
The final version is:
-removed the constructor from the abstract service - no breaking as it was added in this prerelease
-revert the call to start in the constructor in the services extending the abstract service and deprecate this (to keep the backward compatibility - autostart)
-new consumers do not call the start in their constructor 
-schematic to add the .start() method in apps for the imported consumers (there is no harm even if there is already a call to start in the constructor)

packages/@ama-mfe/ng-utils/src/navigation/navigation-consumer-service.ts
mrednic-1A: after discussions, I only deprecate the autostart feature in this PR and the removal will be done in v15.

matthieu-crouzet: It seems strange to me to add something that is directly deprecated, I think we should fallback this change but also this kind of changes https://github.com/AmadeusITGroup/otter/commit/04b8b1276170cd35f810e5f78c92833bf820dcb9#diff-e2c560e92f7f0c515e16ecc8245e12dde3724927ff195de0351cfa34d45e10abL51-L52

mrednic-1A: Indeed. 
Now the final version is:
-removed the constructor from the abstract service - no breaking as it was added in this prerelease
-revert the call to start in the constructor in the services extending the abstract service and deprecate this (to keep the backward compatibility - autostart)
-new consumers do not call the start in their constructor 
-schematic to add the .start() method in apps for the imported consumers (there is no harm even if there is already a call to start in the constructor)

packages/@ama-mfe/ng-utils/package.json
```diff
@@ -33,7 +34,7 @@
     "postbuild": "patch-package-json-exports",
     "prepare:compile": "cp-package-json",
     "build": "yarn nx build ama-mfe-ng-utils",
-    "prepare:build:builders": "yarn cpy 'schematics/**/*.json' dist/schematics && yarn cpy 'collection.json' dist",
+    "prepare:build:builders": "yarn cpy 'schematics/**/*.json' dist/schematics && yarn cpy 'collection.json' dist && yarn cpy 'migration.json' dist",
```
kpanot:
```suggestion
    "prepare:build:builders": "yarn cpy 'schematics/**/*.json' dist/schematics && yarn cpy 'collection.json' 'migration.json' dist",
```

packages/@ama-mfe/ng-utils/package.json
```diff
@@ -17,7 +17,8 @@
     "@o3r/logger": "workspace:~",
     "@o3r/schematics": "workspace:~",
     "rxjs": "^7.8.1",
-    "type-fest": "^5.3.1"
+    "type-fest": "^5.3.1",
+    "typescript": "~5.9.2"
```
kpanot: why is it needed?

packages/@ama-mfe/ng-utils/src/managers/abstract-message-consumer-service.ts
```diff
@@ -25,6 +28,8 @@ export abstract class AbstractMessageConsumer<T extends VersionedMessage = Versi
   protected readonly consumerManagerService = inject(ConsumerManagerService);
   /** The reference used to handle the destruction of the consumer */
   protected readonly destroyRef = inject(DestroyRef);
+  /** Logger shared with subclasses. */
+  protected readonly logger = inject(LoggerService);
```
kpanot, line 32: not needed anymore

packages/@ama-mfe/ng-utils/schematics/ng-update/v14-5/index.ts
```diff
@@ -0,0 +1,78 @@
...
+    const injectRegex = new RegExp(`inject\\(\\s*${consumer}\\s*\\)(?!\\s*\\.start\\b)`, 'g');
+    if (injectRegex.test(updated)) {
+      updated = updated.replace(injectRegex, `inject(${consumer}).start()`);
+      started.push(consumer);
+    }
+  }
+
+  if (updated !== original) {
+    tree.overwrite(filePath, updated);
+    context.logger.info(`[ama-mfe ng-update] Added explicit .start() call for ${started.join(', ')} in ${filePath}.`);
```
kpanot:
```suggestion
    context.logger.info(`[ama-mfe ng-update] Added explicit ".start()" call for ${started.join(', ')} in ${filePath}.`);
```

kpanot (dismissed):

packages/@ama-mfe/ng-utils/src/managers/abstract-message-consumer-service.ts
mrednic-1A: it is used in 2 consumers so I was thinking to keep it here.

packages/@ama-mfe/ng-utils/package.json
mrednic-1A: nice catch, it was needed in a previous version of the code, not needed now, I removed it

packages/@ama-mfe/ng-utils/src/history/history-consumer-service.ts
```diff
@@ -44,5 +44,12 @@ export class HistoryConsumerService extends AbstractMessageConsumer<HistoryMessa
 
   constructor() {
     super();
+    /**
+     * Auto-starts the consumer on creation.
+     * @deprecated The constructor auto-starts the consumer for backwards compatibility. It will be removed in v15;
+     * call `inject(HistoryConsumerService).start()` from your application configuration (e.g. via
+     * `provideAppInitializer`) instead.
```
fpaul-1A: this deprecated message seems odd, it looks like the message is addressed to application owners but they will probably never see it

fpaul-1A (dismissed):

packages/@ama-mfe/ng-utils/src/history/history-consumer-service.ts
mrednic-1A: indeed, it is more for us, the devs, for the app users there is a section in the doc. 
I will rewrite it 

__EDIT__ updated

matthieu-crouzet (approved):

fpaul-1A (approved):

PR #4298, opened 2026-06-22
fix: remove InputSignal from inputs in context.ts

mrednic-1A:
## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@o3r/core/src/core/interfaces/context.ts
+++ b/packages/@o3r/core/src/core/interfaces/context.ts
@@ -1,6 +1,5 @@
 import type {
   EventEmitter,
-  InputSignal,
 } from '@angular/core';
 import type {
   FormControl,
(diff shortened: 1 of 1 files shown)


matthieu-crouzet (approved):

packages/@o3r/core/src/core/interfaces/context.ts
```diff
@@ -49,7 +48,7 @@ export interface TemplateContext<
   /** Component configuration */
   config?: Partial<N>;
   /** Component inputs context */
-  inputs: { [T in keyof S]: S[T] | InputSignal<S[T]> } & { [key: string]: any };
```
fpaul-1A, line 52: is it not a breaking change?

fpaul-1A (approved):

PR #4313, opened 2026-06-24
fix: schematic - o3r transloco as external dep for apps

mrednic-1A:
## Proposed change

schematic to add o3r/transloco as external dependency for apps build

## Related issues

*- No issue associated -*

### DIFF
--- a/packages/@o3r/components/schematics/ng-update/v14-4/index.ts
+++ b/packages/@o3r/components/schematics/ng-update/v14-4/index.ts
@@ -1,0 +1,116 @@
+import {
+  chain,
+  type Rule,
+} from '@angular-devkit/schematics';
+import {
+  createOtterSchematic,
+  getWorkspaceConfig,
+  writeAngularJson,
+} from '@o3r/schematics';
+import type {
+  PackageJson,
+} from 'type-fest';
+
+const TRANSLOCO_PACKAGE = '@o3r/transloco';
+
+/** Angular builders that perform JavaScript bundling and support the externalDependencies option */
+const BUNDLING_BUILDERS = new Set([
+  '@angular-devkit/build-angular:browser',
+  '@angular-devkit/build-angular:browser-esbuild',
+  '@angular/build:application',
+  '@angular-devkit/build-angular:server',
+  '@angular-devkit/build-angular:app-shell',
+  '@angular-devkit/build-angular:prerender',
+  '@angular-devkit/build-angular:karma'
+]);
+
+/**
+ * Builders from `@angular-builders/custom-webpack` that bundle JavaScript through a user-provided webpack config file.
+ * These do not support the `externalDependencies` option, so the equivalent fix must be applied in the webpack config
+ * referenced by their `customWebpackConfig.path` option.
+ */
+const CUSTOM_WEBPACK_BUILDERS = new Set([
+  '@angular-builders/custom-webpack:browser',
+  '@angular-builders/custom-webpack:server',
+  '@angular-builders/custom-webpack:karma'
+]);
+
+/**
+ * Add `@o3r/transloco` to externalDependencies in bundling targets of angular.json
+ * when it is not a direct dependency of the project.
+ * This is required because `@o3r/components` performs a dynamic `import('@o3r/transloco')` at runtime,
+ * and webpack/esbuild will fail at build time if the module is not installed and not marked as external.
+ * @param tree
+ * @param context
+ */
+const addTranslocoExternalDependency: Rule = (tree, context) => {
+  const workspace = getWorkspaceConfig(tree);
+  if (!workspace) {
+    context.logger.warn('No angular.json found, skipping migration.');
+    return;
+  }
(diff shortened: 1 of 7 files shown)


matthieu-crouzet (dismissed):

packages/@o3r/components/schematics/ng-update/v14-4/index.ts
```diff
@@ -0,0 +1,94 @@
...
+    for (const [targetName, target] of Object.entries(architect)) {
+      if (!BUNDLING_BUILDERS.has(target.builder)) {
+        continue;
+      }
+      if (!target.options) {
+        target.options = {};
+      }
+      const existingExternals: string[] = (target.options).externalDependencies ?? [];
+      if (!existingExternals.includes(TRANSLOCO_PACKAGE)) {
+        (target.options).externalDependencies = [...existingExternals, TRANSLOCO_PACKAGE];
```
fpaul-1A, line 93: should we remove this on the `ng add` of `@o3r/transloco`?

mrednic-1A: we can remove it.
o3r/transloco is a temporary experimental package. in v16 it will be renamed to o3r/localization, so this change will have no effect. I would say that we would need to remove it at that time too (for o3r/localization v16).

mrednic-1A: updated

packages/@o3r/transloco/schematics/ng-add/helpers/remove-transloco-external-dependency.ts
```diff
@@ -0,0 +1,71 @@
...
+ * Remove `@o3r/transloco` from the `externalDependencies` of all application bundling targets in angular.json.
+ *
+ * The `@o3r/components` `ng update` migration adds `@o3r/transloco` to `externalDependencies` when the package is not
+ * installed, so that the bundler skips its dynamic `import()`. Once `@o3r/transloco` is added to the project (via this
+ * `ng add`), it must be bundled again, so the previously added entry has to be cleaned up.
+ * Custom-webpack builders cannot be updated automatically, so a warning is logged to revert the manual `IgnorePlugin`.
+ * @param tree
+ * @param context
+ */
+export const removeTranslocoExternalDependency: Rule = (tree: Tree, context: SchematicContext) => {
```
kpanot:
```suggestion
export const removeTranslocoExternalDependency: Rule = (tree, context) => {
```

packages/@o3r/components/schematics/ng-update/v14-4/index.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+    const architect = project.architect || {};
+    for (const [targetName, target] of Object.entries(architect)) {
+      if (CUSTOM_WEBPACK_BUILDERS.has(target.builder)) {
+        const webpackConfigPath = (target.options as { customWebpackConfig?: { path?: string } } | undefined)?.customWebpackConfig?.path;
+        const webpackConfigLabel = webpackConfigPath ? `the webpack config "${webpackConfigPath}"` : 'your webpack config';
+        context.logger.warn(
+          `${projectName}/${targetName} uses the custom-webpack builder "${target.builder}", which does not support the externalDependencies option.\n`
+          + `Please add the following to ${webpackConfigLabel} manually so the bundler skips the optional ${TRANSLOCO_PACKAGE} import:\n`
+          + '  const { IgnorePlugin } = require(\'webpack\');\n'
```
kpanot, line 78: Is it overkill to use [ts-syntax-highlighter](https://www.npmjs.com/package/ts-syntax-highlighter) to display code sample in the console?

packages/@o3r/components/schematics/ng-update/v14-4/index.ts
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Add `@o3r/transloco` to externalDependencies in bundling targets of angular.json
+ * when it is not a direct dependency of the project.
+ * This is required because `@o3r/components` performs a dynamic `import('@o3r/transloco')` at runtime,
+ * and webpack/esbuild will fail at build time if the module is not installed and not marked as external.
+ * @param tree
+ * @param context
+ */
+const addTranslocoExternalDependency: Rule = (tree: Tree, context: SchematicContext) => {
```
kpanot:
```suggestion
const addTranslocoExternalDependency: Rule = (tree, context) => {
```

packages/@o3r/components/schematics/ng-update/v14-4/index.ts
mrednic-1A: I checked: the package is not a dependency anywhere in the repo and is not used in either schematic. 
I would say it is a bit overkill for this use case which is targeting only a part of the users.

fpaul-1A (approved):

matthieu-crouzet (approved):

PR #4336, opened 2026-07-01
fix: issues around migration scripts

guillaumerochelle:
## Proposed change

Changes to make migration scripts work:
- Added `--allow-dirty` option in the list of authorized ng update options
- Added `--` when npm exec is used to make sure arguments are passed to the command executed
- Modified use-register-action-handlers migration script to avoid reading files with an incorrect path inside the filter callback.

*- No issue associated -*

### DIFF
--- a/packages/@o3r/rules-engine/schematics/ng-update/v11.6/use-register-action-handlers.ts
+++ b/packages/@o3r/rules-engine/schematics/ng-update/v11.6/use-register-action-handlers.ts
@@ -11,8 +11,11 @@ import {
  * @param tree
  */
 export const useRegisterActionHandlers: Rule = (tree: Tree) => {
-  const files = findFilesInTree(tree.root, (file) => /\.actionHandlers\.(add|delete)/.test(tree.readText(file)));
+  const files = findFilesInTree(tree.root, (file) => file.endsWith('.ts'));
   files.forEach(({ content, path }) => {
+    if (!/\.actionHandlers\.(add|delete)/.test(content.toString())) {
+      return;
+    }
     tree.overwrite(
       path,
       content
--- a/packages/@o3r/schematics/src/utility/package-manager-runner.ts
+++ b/packages/@o3r/schematics/src/utility/package-manager-runner.ts
@@ -94,9 +94,11 @@ export function getPackageManagerRunner(workspaceConfig?: WorkspaceSchema | stri
 export function getPackageManagerExecutor(workspaceConfig?: WorkspaceSchema | string | null, packageName?: string): string {
   const pckManager = getPackageManager({ workspaceConfig });
   if (!packageName) {
-    return `${pckManager} exec` as SupportedPackageManagerExecutors;
+    return pckManager === 'npm' ? 'npm exec --' : `${pckManager} exec`;
   }
-  return `${pckManager} ${PACKAGE_MANAGER_WORKSPACE_MAPPING[pckManager]} ${packageName} exec`;
+  return pckManager === 'npm'
+    ? `npm ${PACKAGE_MANAGER_WORKSPACE_MAPPING[pckManager]} ${packageName} exec --`
+    : `${pckManager} ${PACKAGE_MANAGER_WORKSPACE_MAPPING[pckManager]} ${packageName} exec`;
 }
 
 /**
--- a/apps/chrome-devtools/package.json
+++ b/apps/chrome-devtools/package.json
@@ -57,7 +57,7 @@
     "archiver": "^7.0.0",
     "chokidar": "^4.0.3",
     "chrome-webstore-upload": "^3.0.0",
-    "concurrently": "^9.1.0",
+    "concurrently": "^9.2.3",
     "cpy-cli": "^5.0.0",
     "eslint": "~9.28.0",
     "eslint-import-resolver-node": "~0.3.9",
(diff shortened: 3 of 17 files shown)


packages/@o3r/rules-engine/schematics/ng-update/v11.6/use-register-action-handlers.ts
```diff
@@ -11,8 +11,11 @@ import {
  * @param tree
  */
 export const useRegisterActionHandlers: Rule = (tree: Tree) => {
-  const files = findFilesInTree(tree.root, (file) => /\.actionHandlers\.(add|delete)/.test(tree.readText(file)));
+  const files = findFilesInTree(tree.root, (file) => file.endsWith('.ts'));
```
guillaumerochelle, line 14: Filter callback is called with file param containing only the name of the file: `foo.ts` instead of `path/to/foo.ts`.
That makes `tree.readText` fail.

It is the first time a filter callback tries to read the content of the file (other usages show that filtering is done based on the filename only).

I added a filter for .ts files here and added the filtering on the actual content in the forEach loop below.

Another way to fix this is to make the argument given to the filter callback (in `findFilesInTreeRec` function) be the actual path to the file (in this case `path/to/foo.ts`) but that would have an impact on all callers of `findFilesInTree`.

packages/@o3r/rules-engine/schematics/ng-update/v11.6/use-register-action-handlers.ts
```diff
@@ -11,8 +11,11 @@ import {
...
   files.forEach(({ content, path }) => {
+    if (!/\.actionHandlers\.(add|delete)/.test(content.toString())) {
```
guillaumerochelle, line 16: If I don't filter out here, the logs show all files as touched even though they have not been changed.

packages/@o3r/schematics/src/utility/package-manager-runner.ts
```diff
@@ -94,7 +94,7 @@ export function getPackageManagerRunner(workspaceConfig?: WorkspaceSchema | stri
 export function getPackageManagerExecutor(workspaceConfig?: WorkspaceSchema | string | null, packageName?: string): string {
   const pckManager = getPackageManager({ workspaceConfig });
   if (!packageName) {
-    return `${pckManager} exec` as SupportedPackageManagerExecutors;
+    return pckManager === 'npm' ? 'npm exec --' : `${pckManager} exec`;
```
guillaumerochelle, line 97: I am not sure this is actually the way to fix it... line 99 below could also have the same issue.

.husky/commit-msg
fpaul-1A, line 1: this should be done in a separate pull-request

packages/@o3r/rules-engine/schematics/ng-update/v11.6/use-register-action-handlers.ts
fpaul-1A: looks good to me, we could also update (or pass) the list of ignored directories (3rd parameter) to not dive into `.angular` for example. 
Ideally, we could follow the ignores defined in `.gitignore` (as done by glob), but it might not be straightforward with angular schematics file tree

packages/@o3r/schematics/src/utility/package-manager-runner.ts
fpaul-1A: looks good to me, and yes, line 99 should be updated as well

matthieu-crouzet (dismissed):

.husky/commit-msg
guillaumerochelle: I thought I had to push the changes to be able to commit.
Removed husky files from this PR.

packages/@o3r/rules-engine/schematics/ng-update/v11.6/use-register-action-handlers.ts
guillaumerochelle: Added .angular to list of ignored directories by default.

packages/@o3r/schematics/src/utility/package-manager-runner.ts
```diff
@@ -94,9 +94,11 @@
...
   }
-  return `${pckManager} ${PACKAGE_MANAGER_WORKSPACE_MAPPING[pckManager]} ${packageName} exec`;
+  return pckManager === 'npm'
+    ? `npm ${PACKAGE_MANAGER_WORKSPACE_MAPPING[pckManager]} ${packageName} exec --`
+    : `${pckManager} ${PACKAGE_MANAGER_WORKSPACE_MAPPING[pckManager]} ${packageName} exec`;
```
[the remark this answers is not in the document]
fpaul-1A: not linked to this PR

packages/@o3r/schematics/src/utility/package-manager-runner.ts
```diff
@@ -94,9 +94,11 @@
 export function getPackageManagerExecutor(workspaceConfig?: WorkspaceSchema | string | null, packageName?: string): string {
   const pckManager = getPackageManager({ workspaceConfig });
   if (!packageName) {
-    return `${pckManager} exec` as SupportedPackageManagerExecutors;
+    return pckManager === 'npm' ? 'npm exec --' : `${pckManager} exec`;
   }
-  return `${pckManager} ${PACKAGE_MANAGER_WORKSPACE_MAPPING[pckManager]} ${packageName} exec`;
+  return pckManager === 'npm'
+    ? `npm ${PACKAGE_MANAGER_WORKSPACE_MAPPING[pckManager]} ${packageName} exec --`
```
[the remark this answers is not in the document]
fpaul-1A: not linked to this PR

fpaul-1A (approved):

kpanot (approved):

PR #4340, opened 2026-07-03
[draft] ci: use dev-drive for it-tests

fpaul-1A:
## Related issues

*- No issue associated -*


fpaul-1A:
clearly worse performances than before

PR #4440, opened 2026-08-19
chore: add enableInlineBuilds debug hint in yarnrc

matthieu-crouzet:
## Proposed change

Add a commented-out `enableInlineBuilds: true` option in `.yarnrc.yml` with an explanatory comment. This provides developers with a quick reference on how to enable inline build output for debugging postinstall scripts.

## Related issues

*- No issue associated -*

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -7,6 +7,10 @@ enableGlobalCache: false
 
 enableScripts: false
 
+# Enable the following line will display postinstall script output in the console, which can be useful for debugging.
+# Documentation: https://yarnpkg.com/configuration/yarnrc#enableInlineBuilds
+enableInlineBuilds: false
+
 nodeLinker: pnp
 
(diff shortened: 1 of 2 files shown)


.yarnrc.yml
```diff
@@ -7,6 +7,9 @@ enableGlobalCache: false
 
 enableScripts: false
 
+# Uncomment the following line will display postinstall script output in the console, which can be useful for debugging.
+# enableInlineBuilds: true
```
matthieu-crouzet: I didn't want to enable it by default to reduce logs when not useful (debug mode)

fpaul-1A:
```suggestion
# Enable the following line will display postinstall script output in the console, which can be useful for debugging.
enableInlineBuilds: false
```

fpaul-1A (approved):

cpaulve-1A (approved):

PR #4448, opened 2026-08-19
feat: add app(or lib) logic placement skill

mrednic-1A:
## Related issues

*- No issue associated -*

### DIFF
--- a/tools/llm/plugins/otter/README.md
+++ b/tools/llm/plugins/otter/README.md
@@ -14,6 +14,25 @@ This plugin bundles three MCP servers:
 
 All servers are launched via `npx` and require no additional local installation.
 
+## Skills
+
+| Skill | Description |
+|-------|-------------|
+| **logic-placement** | Design mode: where new logic belongs. Routes the three decisions, owns the stop-and-ask rules and the recommendation format |
+| **logic-review** | Review mode: misplaced logic in existing code — presenter leaks, `componentType` mismatches, wrong store variants, anti-patterns |
+| **logic-component-shape** | Decision tree: single component vs container/presenter, `componentStructure`, `componentType` |
+| **logic-state-layer** | Decision tree: component class vs service vs store, service scope, where API and SDK calls go |
+| **logic-store-type** | Decision tree: which NgRx store variant — entity or simple, sync or async |
+| **rxjs-vs-signals** | Choosing between Angular Signals and RxJS Observables, with bridging rules for `toSignal()` and `toObservable()` |
+| **otter-schematics** | Ensures Otter schematics are used over manual file creation when scaffolding Angular artifacts |
+| **otter-new-config** | Adding a configuration flag to a component: scope decisions, safe defaults, absence detection |
+| **otter-rules-engine** | Authoring and debugging facts, operators, actions, and rulesets in `@o3r/rules-engine` |
+| **playwright-e2e-testing** | Writing and reviewing Playwright E2E scenarios, actions, and checks |
+| **otter-mcp-setup** | Installing and configuring the `@o3r/mcp` MCP server when it is unavailable |
+
+The `logic-*` decision trees are shared knowledge, read by both **logic-placement** and **logic-review** so that a design recommendation and a review verdict can never disagree.
+They are invoked by those agent skills rather than by a question asked directly.
+
 ## Agents
 
 This plugin ships specialized subagents in a format that both **Claude Code** (CLI and VS Code extension) and **GitHub Copilot in VS Code** auto-load when the plugin is installed.
@@ -21,6 +40,8 @@ This plugin ships specialized subagents in a format that both **Claude Code** (C
 | Agent | Description |
 |-------|-------------|
 | **otter-migration** | Safely upgrades `@o3r/*`, `@ama-mfe/*`, and `@ama-sdk/*` packages across an Otter monorepo, handling Angular peer-dependency bumps when required. Plans the migration, bumps versions, resolves peer-dependency conflicts, reads migration schematics and applies the equivalent changes manually, then validates the result. |
+| **logic-design** | Decides where new application logic belongs — component shape, logic layer, store variant — and returns a recommendation with the schematic to run and what to verify. Stops to ask when the choice is a trade-off the user owns. |
+| **logic-review** | Reviews existing components, services, and stores for misplaced logic and reports findings with `file:line`, the rule broken, and the concrete fix. Reads the same decision trees as **logic-design**, so the two cannot disagree. |
 
 Invoke an agent by mentioning it (e.g. "use the otter-migration agent to upgrade to latest") or via the agent picker in Claude Code or GitHub Copilot.
 
@@ -44,7 +65,8 @@ Invoke an agent by mentioning it (e.g. "use the otter-migration agent to upgrade
   }
   ```
 
-The same [`.claude-plugin/`](../../../../.claude-plugin/marketplace.json) manifest is recognized by both tools — see VS Code's [agent plugins documentation](https://code.visualstudio.com/docs/copilot/customization/agent-plugins) for the format-detection rules.
+The same [`.claude-plugin/`](../../../../.claude-plugin/marketplace.json) manifest is recognized by both tools —
+see VS Code's [agent plugins documentation](https://code.visualstudio.com/docs/copilot/customization/agent-plugins) for the format-detection rules.
 
 ## Prerequisites
 
--- a/tools/llm/plugins/otter/agents/logic-design.agent.md
+++ b/tools/llm/plugins/otter/agents/logic-design.agent.md
@@ -1,0 +1,29 @@
+---
+name: logic-design
+description: Decide where new application logic belongs — component shape, logic layer, store variant — and return a recommendation with the schematic to run and what to verify.
+  Use the logic-review agent for existing code.
+---
+
+You are an Angular and Otter architecture advisor.
+Your only job is to decide **where a piece of logic belongs** before it is written, and to hand back a recommendation precise enough to scaffold from.
+
+You do not write feature code. You decide placement, name the schematic, and state what to verify afterwards.
+
+## How to work
+
+1. **Load the `logic-placement` skill first, and follow it.** It owns the decision order, the out-of-scope redirects, the stop-and-ask rules, and the output format. Do not answer from memory.
+2. Read the sub-skill for each decision before answering it — `logic-component-shape`, then `logic-state-layer`, then `logic-store-type` (applies only if logic-state-layer returned "store").
+3. Inspect the workspace before recommending.
+   Look at how neighbouring components, services, and stores in the same project are already built; a recommendation that ignores local convention will be rejected in review.
+4. Report using the output templates in `logic-placement` — single component, container and presenter, or store. Nothing looser.
+
+## Hard rules
+
+- **Default to the shallowest option.** Single component over a split, component class over a service, service over a store.
+  Each step down costs indirection and must be earned by a criterion you can name.
+- **Never invent a justification.** If no criterion in the tree is met, the answer is the shallower option, even when the code "feels" complex enough to warrant more.
+- **Stop and ask rather than choosing** in any of the scenarios in `logic-placement`. Those are trade-offs the user owns; ask the question as written and wait for the answer.
+- **Redirect out-of-scope questions** instead of answering them.
+  Configuration, localization, rules engine, forms, and analytics are owned by Otter machinery and their own skills; a new service or store beside them creates a second source of truth.
+- **Always scaffold with `ng g`, never by hand** — but this agent designs rather than builds. Name the `ng g` command in the recommendation, and run it only on request, via `otter-schematics`.
+- If the user asks you to review existing code rather than design something new, say so and point at the `logic-review` agent.
--- a/tools/llm/plugins/otter/skills/logic-component-shape/SKILL.md
+++ b/tools/llm/plugins/otter/skills/logic-component-shape/SKILL.md
@@ -1,0 +1,190 @@
+---
+name: logic-component-shape
+description: Decision tree for single component versus container/presenter, which componentStructure to generate, and which componentType to declare.
+  Invoked by `logic-placement` and `logic-review`, not directly.
+---
+
+# Decision tree — component shape
+
+Returns: **single component**, or **container + presenter**, plus the `componentStructure` to generate and the `componentType` to declare.
+
+**Default to a single component.** A split separates data and business-logic access from pure presentation,
+and the [container/presenter documentation](https://github.com/AmadeusITGroup/otter/blob/main/docs/components/CONTAINER_PRESENTER.md) says a component interacting with external entities
+*may* be split — not that it must be.
+
+## Will something vary?
+
+The split puts a seam between data access and presentation so one can vary without touching the other. It is worth its cost only when something on one side actually varies.
+
+```
+CRITERION A  one data source, several UI flavours
+             dense table on desktop and stacked cards on mobile; an A/B variant; or a
+             presenter integrators swap or configure at runtime — through the CMS or
+             Otter's component replacement mechanism — which is just another flavour
+             supplied from outside, and needs the display half to be its own component
+             whose inputs and outputs are the contract
+             without the seam: the fetch is duplicated per variant, or the template
+             branches on display concerns
+
+CRITERION B  one UI flavour, several data sources
+             a passenger list rendered from a search result on one screen and from a
+             saved itinerary on another
+             without the seam: the presentation is copied, and the copies drift
+
+IF   A or B is true  -> BRANCH A: container + presenter
+ELSE                 -> BRANCH B: single component
+```
+
+**Name what varies before splitting.** If you cannot say which of data source, display, or integrator-supplied implementation will differ, there is nothing for the seam to absorb.
+
+"The logic feels complex enough to separate" is **not** an additional criterion. Complexity alone is a human judgement call — see the **split without variation** stop-and-ask in `logic-placement`.
+
+## Branch A — container + presenter
+
+```
+SCHEMATIC       ng g component <name> --componentStructure=full
+VERIFY          the container injects Store, services, or SDK clients, and dispatches
+                the presenter declares inputs and outputs only
+                the presenter renders from a spec by supplying inputs alone
+```
+
+### Who owns what
+
+| Concern | Owner |
+|---|---|
+| Inject `Store`, dispatch, select | container |
+| Call a service or SDK client for shared data | container |
+| Map store or API models to view models | container |
+| Business rules, custom validators, submit handling | container |
+| Router, guards, query params | container |
+| Choose which presenter flavour to render | container |
+| Format, sort, filter data already received | presenter |
+| Local UI state (open/closed, tab, page, hover) | presenter |
+| Emit user intent upward | presenter |
+| Styling, animation, breakpoints | presenter |
+
+Data down through inputs, events up through outputs.
+
+**The leak is often indirect.** A presenter that injects no store and makes no HTTP client can still reach both through a service.
+Judge by what the service *does*, not by what the presenter imports: if it selects from the store, calls an API, or triggers a side effect beyond the component, it belongs in the container.
+A presenter may inject presentation helpers only — formatting, breakpoints.
+
+## Branch B — single component
+
+```
+SCHEMATIC       ng g component <name> --componentStructure=simple
+VERIFY          no -cont or -pres suffix anywhere in file, class, selector, or template
+                no container that would only forward its inputs
+```
+
+Shapes where nothing can vary:
+
+- **Small presentational elements** — a badge, an icon, a price label, a spinner. They take inputs, render, and emit; there is nothing for a container to own.
+- **Self-contained screens with one data source** — fetch once, render the result. Splitting becomes defensible when a second flavour or a CMS requirement appears; on day one it buys nothing.
+
+Splitting anyway produces a container that only forwards inputs, which is the failure mode to avoid. The documentation applies the same rule to the folder layout:
+
+> A component with only a presenter and no container should not have a presenter folder
+
+> A functional area containing only one block should not have a sub-folder for it
+
+A single component is not a dead end: `ng g component --componentStructure=presenter` adds the display half later to what already exists.
+
+## Which componentStructure to generate
+
+| Decision | `componentStructure` |
+|---|---|
+| Single component | `simple` |
+| Container + presenter pair | `full` |
+| Container only, presenter already exists | `container` |
+| Presenter only, added to an existing container | `presenter` |
+
+Use `otter-schematics` to discover the schematic and its current options.
+It also produces the naming and folder layout, so neither needs deriving by hand —
+see the [container/presenter documentation](https://github.com/AmadeusITGroup/otter/blob/main/docs/components/CONTAINER_PRESENTER.md) for the `*-cont` / `*-pres` conventions.
+
+## componentType as a placement cross-check
+
+Every Otter component carries `@O3rComponent({ componentType })`, and the value must agree with the container/presenter role assigned above.
+
+| Value | Meaning | Role |
+|---|---|---|
+| `Page` | Displays an application route | container |
+| `Block` | Handles a functional area | container |
+| `ExposedComponent` | Needs to be exposed in the CMS | presenter |
+| `Component` | Neither exposed nor fitting the others | presenter, or a single unsplit component |
+
+A mismatch reliably means logic sits in the wrong place: a `Page` that injects nothing and only holds display state is a presenter; a `Component` that dispatches store actions is a container.
+Either the type or the logic must move.
+
+Choosing between `Block` and `ExposedComponent` is a CMS-exposure decision, not a placement one — see the **CMS replacement undecided** stop-and-ask in `logic-placement`.
+
+## Example — the seam, both halves
+
+```typescript
+@O3rComponent({ componentType: 'Page' })
+@Component({
+  selector: 'my-app-passengers-cont',
+  imports: [PassengersPres],
+  templateUrl: './passengers-cont.html',
+  encapsulation: ViewEncapsulation.None,
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class PassengersCont implements OnInit {
+  private readonly store = inject(Store);
+  private readonly passengerApi = inject(PassengerApi);
+
+  /** Passengers to display, selected from the store */
+  public readonly passengers = toSignal(this.store.pipe(select(selectAllPassengers)), { initialValue: [] });
+
+  public ngOnInit() {
+    this.store.dispatch(setPassengersEntitiesFromApi({
+      call: this.passengerApi.getPassengers(),
+      requestId: 'passengers-initial-load'
+    }));
+  }
+
+  /** Forwarded from the presenter's output */
+  public onPassengerSelected(passengerId: string) {
+    this.store.dispatch(selectPassenger({ passengerId }));
+  }
+}
+
+@O3rComponent({ componentType: 'ExposedComponent' })
+@Component({
+  selector: 'my-app-passengers-pres',
+  templateUrl: './passengers-pres.html',
+  styleUrls: ['./passengers-pres.scss'],
+  encapsulation: ViewEncapsulation.None,
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class PassengersPres {
+  /** Passengers to render */
+  public readonly passengers = input.required<Passenger[]>();
+
+  /** Emitted when the user picks a passenger */
+  public readonly passengerSelected = output<string>();
+
+  /** Local UI state: nothing outside this component needs it */
+  public readonly sortDirection = signal<'asc' | 'desc'>('asc');
+
+  /** Display ordering derived from inputs plus local state */
+  public readonly sortedPassengers = computed(() => {
+    const factor = this.sortDirection() === 'asc' ? 1 : -1;
+    return [...this.passengers()].toSorted((a, b) => factor * a.lastName.localeCompare(b.lastName));
+  });
+}
+```
+
+The presenter renders from a spec by supplying `passengers` alone — that is the test for a clean split.
+Additional details:
+
+- The container bridges the store observable to a signal at the seam, so it can pass `passengers` into the presenter's required signal input.
+  Why the bridge needs `initialValue` and why `| async` would not fit are in `rxjs-vs-signals`.
+- Not every Otter decorator has a signal equivalent.
+  `@Localization` on a `translations` property still requires the decorator-based `@Input()`, so check which form the decorator you need supports rather than generalizing to signals.
+
+## Documentation
+
+- [Container / Presenter](https://github.com/AmadeusITGroup/otter/blob/main/docs/components/CONTAINER_PRESENTER.md) — naming conventions, folder layout, structural guidelines
+- [Component introduction](https://github.com/AmadeusITGroup/otter/blob/main/docs/components/INTRODUCTION.md) — the `@O3rComponent` decorator and component types
--- a/tools/llm/plugins/otter/skills/logic-placement/SKILL.md
+++ b/tools/llm/plugins/otter/skills/logic-placement/SKILL.md
@@ -1,0 +1,159 @@
+---
+name: logic-placement
+description: Decide where new logic belongs — component shape, logic layer, and store variant — when designing a component, service, or store in an Angular/Otter workspace.
+  Use `logic-review` to assess existing code.
+---
+
+# Logic placement — design mode
+
+Decide where a piece of logic belongs **before** it is written.
+This skill holds the order of the decisions, the rules for halting, and the output format.
+The criteria for each decision live in a sub-skill: read it before answering, do not answer from memory.
+
+To assess code that already exists, use `logic-review`.
+
+## Out of scope
+
+Stop and redirect when the question is really about:
+
+| Question | Handled by |
+|---|---|
+| A configuration flag, feature toggle, or CMS-editable value | `otter-new-config` |
+| Behaviour that should be runtime-conditional rather than coded | `otter-rules-engine` |
+| Which reactive primitive to express the logic with | `rxjs-vs-signals` |
+| How to run the generator once the shape is decided | `otter-schematics` |
+| Reimplementing localization, forms, or analytics | the corresponding Otter package, not a new service or store |
+
+Otter machinery already owns state in those areas, so adding a service or store beside it creates a second source of truth.
+Deciding **which half of a split wires that machinery up** is still in scope — the container owns the data and business-logic access, the presenter only displays, per `logic-component-shape`.
+
+## Decision order
+
+| # | Decision | Read | Returns |
+|---|---|---|---|
+| 1 | Component shape | `logic-component-shape` | single component, or container + presenter |
+| 2 | Logic layer | `logic-state-layer` | component class, service, or store |
+| 3 | Store implementation | `logic-store-type` | one of the four `ng g store*` variants |
+
+The store implementation decision runs **only** if the logic layer decision returned "store".
+Never start there: most state that feels store-shaped is service-shaped, and most components that feel split-shaped are not.
+
+## When to settle each decision
+
+| Decision | Settle it | Cost of changing later |
+|---|---|---|
+| Component shape | Before `ng g component` | Low — rerun the schematic, or add the display half later with `--componentStructure=presenter` |
+| Logic layer | Before scaffolding | Low — extract or inline before anything consumes it |
+| Store implementation | Before `ng g store*` | **High** — reshapes state, selectors, and every consumer |
+
+Settle shape and layer first, then scaffold immediately to lock the structure in. A store type raised for the first time in PR review is already expensive; it belongs in the design discussion.
+
+## Stop and ask
+
+These situations are trade-offs the user owns. **Stop, ask, and wait** — do not pick for them. Other skills cite these by name, so keep the names stable when adding or removing one.
+
+### Local status tracking
+
+```
+TRIGGER  state is local to one screen, but the UI must show pending or failure
+ASK      "This state is local to this screen but needs pending/failure tracking.
+          A. Async store — you get AsyncStoreItem, but the state is promoted to application scope
+          B. Local signals — the scope stays right, but you hand-code isLoading/hasError
+          Which fits?"
+THEN     A -> the store implementation decision, via logic-store-type
+         B -> component class, no store
+```
+
+### Shared state lifetime
+
+```
+TRIGGER  state is shared, but it is unclear whether it should outlive the screen
+ASK      "When this screen closes, should this state:
+          A. die with the screen
+          B. survive until logout
+          C. survive a page reload"
+THEN     A -> service in the container's providers: []
+         B -> service with providedIn: 'root'
+         C -> store, persisted through @o3r/store-sync
+```
+
+### Store type migration
+
+```
+TRIGGER  a requirement needs entity<->simple or sync<->async to change on an existing store
+ACTION   STOP. Do not fold the migration into the current change.
+SAY      "This needs <current> to become <target>, which reshapes the state and every
+          consumer. That belongs in its own change — confirm before starting."
+```
+
+### Split without variation
+
+```
+TRIGGER  a container/presenter split is suggested, but none of the variation
+         criteria in logic-component-shape is met
+ASK      "I cannot see what would vary between container and presenter here.
+          What varies today — not in theory?"
+THEN     the user names it -> split
+         the user cannot   -> single component
+```
+
+### CMS replacement undecided
+
+```
+TRIGGER  whether integrators configure or replace the component is unknown, and it
+         determines both the split and the componentType value
+ASK      "Will integrators configure or replace this component through the CMS?"
+THEN     yes -> split, presenter as ExposedComponent
+         no  -> single component, or Block for a container
+```
+
+## Output format
+
+Report every recommendation in one of these three shapes, and nothing looser.
+
+### Single component
+
+```
+RECOMMENDATION  single component <name>
+  componentType <Page | Block | Component>
+  logic layer   <component class | service | store>
+  because       <the criteria that were and were not met>
+NEXT STEP       ng g component <name> --componentStructure=simple
+VERIFY          no -cont or -pres suffix
+                logic sits in the class, a service, or a store — never in template bindings
+```
+
+### Container and presenter
+
+```
+RECOMMENDATION  <name>-cont + <name>-pres
+  container     componentType <Page | Block>, owns <data source / state / dispatch>
+  presenter     componentType <ExposedComponent | Component>, owns display
+  because       <which variation criterion applies, and what varies>
+NEXT STEP       ng g component <name> --componentStructure=full
+VERIFY          the presenter renders from inputs alone: no store, no SDK client, no router
+                the container carries no .scss and no display formatting
+```
+
+### Store
+
+```
+RECOMMENDATION  <entity | simple> + <sync | async> store for <name>
+  because       <which store criterion is met: cross-feature, persistence, async, machinery>
+  shape         <keyed collection | single value>, id property <name> (entity only)
+NEXT STEP       ng g store-<entity|simple>-<sync|async>
+                then write provide<Name>Store, call provideEffects (async), extend the root state
+VERIFY          the container dispatches; presenters never import @ngrx/store
+                remote calls sit in an effect, never in a reducer
+                no hand-rolled isLoading/hasError beside AsyncStoreItem
+```
+
+Run the NEXT STEP through `otter-schematics`.
+If the schematic fails or does not exist, do not hand-write the files from memory — that skill covers reading the schematic's own templates and reproducing them.
+
+## Related skills
+
+- `logic-review` — the same decision trees applied to code that already exists
+- `logic-component-shape`, `logic-state-layer`, `logic-store-type` — the three trees
+- `rxjs-vs-signals` — which reactive primitive to express the chosen layer with
+- `otter-schematics` — discovering and running the generator named in NEXT STEP
--- a/tools/llm/plugins/otter/skills/logic-review/SKILL.md
+++ b/tools/llm/plugins/otter/skills/logic-review/SKILL.md
@@ -1,0 +1,118 @@
+---
+name: logic-review
+description: Review existing components, services, or stores for misplaced logic — presenter leaks, componentType mismatches, wrong store variants.
+  Use `logic-placement` when designing something new.
+---
+
+# Logic review — review mode
+
+Assess code that already exists against the same trees the design side uses, so a review can never contradict a design recommendation.
+
+To decide where something new should go, use `logic-placement`.
+
+## Procedure
+
+Read in this order, and read the tree before judging its dimension — do not judge from memory.
+
+| # | Look at | Read | Judge |
+|---|---|---|---|
+| 1 | Component classes, decorators, templates | `logic-component-shape` | is the split, or its absence, justified; does `componentType` match the role |
+| 2 | What each class injects and holds | `logic-state-layer` | is the layer the shallowest one that works; are API calls in the right place |
+| 3 | Store files, if any | `logic-store-type` | does the variant match the data |
+
+Rules for the whole pass:
+
+- **Judge a split by what varies, not by whether it exists.** A missing split is a finding only when one of the variation criteria is met.
+  An existing split whose container only forwards inputs is a finding regardless.
+- **Do not recommend a store type migration inside the change under review.** See the **store type migration** stop-and-ask in `logic-placement`.
+
+## Check — presenter purity
+
+The test: **can the presenter be rendered from a spec by supplying inputs only?** If it needs a store or an HTTP mock, logic has leaked.
+
+Follow the injections one level deeper than the imports.
+A presenter that injects no store and calls no HTTP client can still reach both through a service, which hides the dependency without removing it.
+
+```
+FOR each service a presenter injects:
+  does it select from the store, call an API, or trigger a side effect
+  beyond the component?
+    yes -> FINDING: the dependency belongs in the container
+    no  -> allowed, if it is a presentation helper such as formatting or breakpoints
+```
+
+## Check — componentType against the actual role
+
+A mismatch between `@O3rComponent({ componentType })` and what the class does is a reliable signal that logic sits in the wrong place.
+Either the type or the logic must move — decide which from the responsibility table in `logic-component-shape`.
+
+| Symptom | Reading |
+|---|---|
+| `Page` or `Block` that injects nothing and only holds display state | it is really a presenter |
+| `Component` or `ExposedComponent` that dispatches store actions | it is really a container |
+| `-pres` suffix on a class that fetches its own data | the suffix or the fetch is wrong |
+
+## Anti-patterns
+
+### Components
+
+| Symptom | Why it is wrong | Fix |
+|---|---|---|
+| Presenter injecting `Store`, dispatching, or fetching — including indirectly through a service that does | the presenter can no longer be rendered from inputs, so it cannot be reused or CMS-replaced | move the dependency to the container, pass the result down as an input |
+| Container with a `.scss` file, or doing display formatting | display concerns on the data side of the seam, so the presenter is no longer swappable | move styling and formatting into the presenter |
+| Container that only forwards its inputs | indirection with no seam to absorb variation | collapse it back into a single component |
+| Business logic in template bindings | components run `ChangeDetectionStrategy.OnPush`; the expression re-evaluates and the intent is untestable | derive in `computed()`, or in the container |
+| `componentType` disagreeing with the class's role | CMS metadata extraction reflects the wrong shape | see the componentType check above |
+| Naming drift between file, class, selector, template, and spec | the `-cont` / `-pres` convention is what makes the roles legible | realign all five, or regenerate |
+
+### Stores and services
+
+| Symptom | Why it is wrong | Fix |
+|---|---|---|
+| A store created to pass data between two components | six files and a serializer for a guarantee nobody needed | a service — see the shared-or-stateless step in `logic-state-layer` |
+| API call inside a reducer | reducers must be pure; the call is untestable and re-runs on replay | move it into an effect |
+| `.subscribe()` in a container followed by a dispatch | the subscription is unmanaged and the sequencing is implicit | move it into an effect |
+| Hand-rolled `isLoading` or `hasError` beside an async store | duplicates `AsyncStoreItem`, and the copies drift | read the status off the state |
+| Simple store holding records updated one at a time | no per-item status, and updates rewrite the whole model | entity store. An array of primitives is fine in a simple store |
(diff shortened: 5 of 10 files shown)


tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,251 @@
...
+
+Do not skip to step 3. Most state that feels store-shaped is service-shaped, and most components that feel split-shaped are not.
+
+## When to Use
+
+- Creating a component, service, or store and unsure which of them the logic belongs in
+- Adding an API or SDK call and deciding whether the result needs a store
+- Introducing state that more than one component reads
+- Choosing among `entity-async`, `entity-sync`, `simple-async`, `simple-sync`
+- Reviewing a change for misplaced logic: a presenter that fetches, a store that should be a service
```
matthieu-crouzet: Is this section really useful, as the decision to trigger the skill or not is the one in the frontmatter description

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,251 @@
...
+
+# Where application logic belongs
+
+Three decisions, in order. Each one narrows the next.
+
+1. **Component shape** - one component, or a container/presenter split
+2. **Logic layer** - component class, injectable service, or NgRx store
+3. **Store implementation** - only if step 2 said store: entity or simple, sync or async
+
+Do not skip to step 3. Most state that feels store-shaped is service-shaped, and most components that feel split-shaped are not.
```
matthieu-crouzet: instead of "step 3", I think we should prefer "store implementation step"
it's more future proof in case we add more steps later
(you may have others similar references)

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: It is redundant indeed

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: yes, updated in all references

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: updated

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,251 @@
...
+| Inject `Store`, dispatch, select | yes | no |
+| Call a service or SDK client for shared data | yes | no |
+| Map store/API models to view models | yes | no |
+| Business rules, custom validators, submit handling | yes | no |
+| Router, guards, query params | yes | no |
+| Choose which presenter flavour to render | yes | no |
+| Format, sort, filter data already received | no | yes |
+| Local UI state (open/closed, tab, page, hover) | no | yes |
+| Emit user intent upward | no | yes |
+| Styling, animation, breakpoints | no | yes |
```
matthieu-crouzet: There is never a line with `yes | yes` or `no | no` could it happens
or should it be `presenter`, `container` or `both`

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+
+Data down through inputs, events up through outputs. The test: can the presenter be rendered from a spec by supplying inputs only? If it needs a store or an HTTP mock, logic has leaked.
+
+Once the shape is decided, scaffold with the schematics rather than hand-rolling files. This decision maps directly onto the component schematic's `componentStructure` option:
```
matthieu-crouzet: Question not related to this PR, shouldn't we convert schematics into skills ?

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+
+Behaviour or state shared between components that needs no serialization, no time-travel, no persistence, no async request tracking.
+
+Also the right home for anything stateless and reusable regardless of sharing: a wrapper over a browser API, a pure computation, orchestration across several API calls, a cross-cutting concern like logging or analytics forwarding.
+
+```typescript
+// Shared between several components, but never persisted and never fetched.
+// A store here would add six files for no benefit.
+@Injectable({ providedIn: 'root' })
+export class InPageNavService {
```
matthieu-crouzet: Maybe for the example keep it with signal to not bring confusion between rxjs and signal

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+- it needs **async lifecycle tracking**: pending, failure, cancellation, concurrent request IDs
+- **Otter machinery reads from the store**: configuration, localization, rules engine, dynamic content, forms, analytics all do
+
+Cost of a store: state, actions, reducer, selectors, sync serializer, plus an effect if async, plus specs. That is the price of the guarantees above. Two sibling components exchanging a value do not need it.
+
+Two Otter conventions:
+
+- **No facade layer.** Containers inject `Store` and dispatch directly. Otter ships no facade schematic and the framework's own stores use none. Do not introduce one.
+- **Component-scoped services before root ones.** State shared only between a container and its own subcomponents belongs in a service listed in that container's `providers: []`, not `providedIn: 'root'`. A root singleton outlives the screen, which is a store-like property you probably did not intend.
+- **`@ngrx/store` belongs to containers only.** It should never be imported by a `*-pres.ts`.
```
matthieu-crouzet:
```suggestion
- **`@ngrx/store` belongs to containers only.** It should never be imported by a `presenter`.
```

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+### Where API and SDK calls go
+
+The `ng g store` schematic states the rule when it prompts for a store type:
+
+> The async store is designed to interact with an api, and handles the asynchronous call via effects.
+
+| Situation | Placement |
+|---|---|
+| Result shared, cached, or persisted | Async store. Container dispatches `set<Store>FromApi` / `set<Store>EntitiesFromApi` with the SDK promise in the payload; the effect runs it |
+| Result local to one screen, discarded on leave | Inject the SDK client, hold the result in signals |
```
matthieu-crouzet:
```suggestion
| Result local to one screen, discarded on leave | Inject the SDK client, hold the result in signals or observables |
```

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+
+Two Otter conventions:
+
+- **No facade layer.** Containers inject `Store` and dispatch directly. Otter ships no facade schematic and the framework's own stores use none. Do not introduce one.
+- **Component-scoped services before root ones.** State shared only between a container and its own subcomponents belongs in a service listed in that container's `providers: []`, not `providedIn: 'root'`. A root singleton outlives the screen, which is a store-like property you probably did not intend.
+- **`@ngrx/store` belongs to containers only.** It should never be imported by a `*-pres.ts`.
+
+### Where API and SDK calls go
+
+The `ng g store` schematic states the rule when it prompts for a store type:
```
matthieu-crouzet: I don't understand why this line is here

tools/llm/plugins/otter/skills/logic-placement/references/component-architecture.md
```diff
@@ -0,0 +1,289 @@
+# Component architecture reference
+
+Detail behind the component shape decision in `SKILL.md`. Based on the Otter [container/presenter](https://github.com/AmadeusITGroup/otter/blob/main/docs/components/CONTAINER_PRESENTER.md) and [component introduction](https://github.com/AmadeusITGroup/otter/blob/main/docs/components/INTRODUCTION.md) documentation.
+
+## Naming conventions
+
```
cpaulve-1A: It lacks the logic behind a component which needs to be split in container / presenter and small elements which do not require this kind of logic. It appears in the example but there are no explanation about it

tools/llm/plugins/otter/skills/logic-placement/references/component-architecture.md
```diff
@@ -0,0 +1,289 @@
+# Component architecture reference
```
matthieu-crouzet: Is this file useful can't we only reference https://github.com/AmadeusITGroup/otter/blob/main/docs/components/CONTAINER_PRESENTER.md and https://github.com/AmadeusITGroup/otter/blob/main/docs/components/INTRODUCTION.md

tools/llm/plugins/otter/skills/logic-placement/references/component-architecture.md
```diff
@@ -0,0 +1,289 @@
...
+export class ComponentName {}
+```
+
+`ComponentType = 'Block' | 'Page' | 'ExposedComponent' | 'Component'`.
+
+| Value | Meaning | Typical role |
+|---|---|---|
+| `Page` | Displays an application route | Container |
+| `Block` | Handles a functional area | Container |
+| `ExposedComponent` | Needs to be exposed in the CMS | Presenter |
```
cpaulve-1A: Do you think we may give more context so the MCP may require the user to provide more details on how to choose (if the agent needs to code the feature) ?

tools/llm/plugins/otter/skills/logic-placement/references/store-types.md
```diff
@@ -0,0 +1,226 @@
+# Store implementation reference
```
matthieu-crouzet: It seems this file treat a lot of information already handle in the skill
and is explaining the generated file useful to help llm to choose the logic placement ?

tools/llm/plugins/otter/skills/logic-placement/references/component-architecture.md
```diff
@@ -0,0 +1,289 @@
...
+    } finally {
+      this.isLoading.set(false);
+    }
+  }
+}
+```
+
+Acceptable because the data never leaves this screen. The moment a second feature needs the same list, or it must survive navigation, this moves to an async store and the local `isLoading`/`hasErrors` flags are replaced by `AsyncStoreItem` status.
+
+## Signals and observables
```
cpaulve-1A: Should this be mixed in the same file? It is no longer a placement logic here

tools/llm/plugins/otter/skills/logic-placement/references/component-architecture.md
```diff
@@ -0,0 +1,289 @@
...
+`ComponentType = 'Block' | 'Page' | 'ExposedComponent' | 'Component'`.
+
+| Value | Meaning | Typical role |
+|---|---|---|
+| `Page` | Displays an application route | Container |
+| `Block` | Handles a functional area | Container |
+| `ExposedComponent` | Needs to be exposed in the CMS | Presenter |
+| `Component` | Neither exposed nor fitting the others | Presenter, or a single unsplit component |
+
+The type should agree with what the class does. A `Page` that injects nothing and only holds display state, or a `Component` that dispatches store actions, signals misplaced logic.
```
cpaulve-1A: I am not sure I understand what this sentence is for. Are these bad patterns?

tools/llm/plugins/otter/skills/logic-placement/references/component-architecture.md
cpaulve-1A: Same comment for all this section (state recommendation etc.)

tools/llm/plugins/otter/skills/logic-placement/references/component-architecture.md
```diff
@@ -0,0 +1,289 @@
...
+```typescript
+export class MyComponentPres implements Translatable<MyComponentPresTranslation> {
+  /** Localization of the component */
+  @Input()
+  @Localization('./my-component-pres-localization.json')
+  public translations: MyComponentPresTranslation = translations;
+}
+```
+
+`@Localization` is a property decorator: it redefines the property with a getter/setter so the incoming value can be merged over the defaults with `deepFill`. A signal input is a readonly signal that cannot be assigned, so the two are incompatible. The `localization-to-component` schematic generates this decorator form, so it is the intended shape rather than legacy code.
```
cpaulve-1A: " so it is the intended shape rather than legacy code." 
What do you mean?

tools/llm/plugins/otter/skills/logic-placement/references/component-architecture.md
```diff
@@ -0,0 +1,289 @@
...
+
+Note the asymmetry with configuration, which *has* a signal-based path (`configSignal` with `@O3rConfig`). Localization has no signal equivalent, so do not generalize "prefer signals" from config to translations.
```
cpaulve-1A: This is a strange way to put it.
"All the Otter decorators may not yet have a signal equivalent. Do not generalize the "prefer signals" strategy and make sure to use an existing form."

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+| Map store/API models to view models | yes | no |
+| Business rules, custom validators, submit handling | yes | no |
+| Router, guards, query params | yes | no |
+| Choose which presenter flavour to render | yes | no |
+| Format, sort, filter data already received | no | yes |
+| Local UI state (open/closed, tab, page, hover) | no | yes |
+| Emit user intent upward | no | yes |
+| Styling, animation, breakpoints | no | yes |
+
+Data down through inputs, events up through outputs. The test: can the presenter be rendered from a spec by supplying inputs only? If it needs a store or an HTTP mock, logic has leaked.
```
cpaulve-1A: Beware that the store usage or HTTP calls may not be directly visible in the presenter but hidden behind a service. We should ensure the presenter does not call this kind of services.

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+  public readonly links$ = this.links.asObservable();
+
+  /** Register the set of links discovered in the page */
+  public initialize(links: NavLink[]) {
+    this.links.next(links);
+  }
+}
+```
+
+**Prefer a service whenever sharing is the only requirement.** This is the step most often skipped.
```
cpaulve-1A: This is the step most often skipped.
=> What do you mean?

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+Reach for a store when state is genuinely application-scoped and at least one of these holds:
+
+- **multiple unrelated features** read or write it
+- it must **survive navigation** or be **persisted and rehydrated** through `@o3r/store-sync`
+- it needs **async lifecycle tracking**: pending, failure, cancellation, concurrent request IDs
+- **Otter machinery reads from the store**: configuration, localization, rules engine, dynamic content, forms, analytics all do
+
+Cost of a store: state, actions, reducer, selectors, sync serializer, plus an effect if async, plus specs. That is the price of the guarantees above. Two sibling components exchanging a value do not need it.
+
+Two Otter conventions:
```
cpaulve-1A:
```suggestion
### Important Otter conventions:
```

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+
+- **No facade layer.** Containers inject `Store` and dispatch directly. Otter ships no facade schematic and the framework's own stores use none. Do not introduce one.
+- **Component-scoped services before root ones.** State shared only between a container and its own subcomponents belongs in a service listed in that container's `providers: []`, not `providedIn: 'root'`. A root singleton outlives the screen, which is a store-like property you probably did not intend.
+- **`@ngrx/store` belongs to containers only.** It should never be imported by a `*-pres.ts`.
+
+### Where API and SDK calls go
+
+The `ng g store` schematic states the rule when it prompts for a store type:
+
+> The async store is designed to interact with an api, and handles the asynchronous call via effects.
```
cpaulve-1A: Would that be a specific case where the user requires a persistent state? Do we recommend a store to hold all the api results? Or is it only the pattern to follow if the user wants to preserve the state on a refresh ?

cpaulve-1A: Okay, I just read the next line. I agree with Matthieu, the two lines above bring confusion, the description of the async store should be described after. You could leave a reference to the doc after it has been mentionned

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+
+Never call an API from a reducer. Never `subscribe` in a container and then dispatch a plain `set` - that is what the effect is for. On the `set*FromApi` path the generated effect also gives you cancellation of superseded requests; the `update`/`upsert` effects deliberately do not cancel, since concurrent per-item updates should not interrupt each other.
+
+## Store implementation: which of the four variants
+
+Two independent questions - collection shape and data source - combine into the four `ng g store*` variants (`store-entity-async`, `store-entity-sync`, `store-simple-async`, `store-simple-sync`; plain `ng g store` prompts for the type, and `ng g store-action` adds an action to an existing store).
+
+The schematic's own framing:
+
+> The entity store contains a collection of items in the state, while the simple one contains only one. The async store is designed to interact with an api, and handles the asynchronous call via effects.
```
cpaulve-1A:
```suggestion
> The entity store contains a collection of independent items in the state, while the simple one contains only one. The async store is designed to interact with an api, and handles the asynchronous call via effects.
```

I mean we should not confuse a list of request which could be updated separately (pending state etc.) and a list of string (destination code for example) which cannot be treated as different entities.

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+
+### Collection shape: entity or simple
+
+**Entity** - a keyed collection. State extends `EntityState<Model>` from `@ngrx/entity`, so you get an adapter, normalized `ids`/`entities`, and generated `selectAll`/`selectEntities`/`selectIds`/`selectTotal`. Choose it when items are addressable by ID, added and removed independently, or looked up individually. Requires an ID property (`modelIdPropName`).
+
+**Simple** - one object. State holds a single `model`. Choose it for singletons: current user session, active search criteria, feature toggles, one configuration blob.
+
+Deciding question: *is the payload a collection of items?* Yes → entity, even if today's UI only renders the list end to end. Entity is the default for collections; a collection in a simple store is the wrong shape, and reshaping later means touching state, selectors, and every consumer. Reserve simple stores for a single object, or for a fixed-size tuple that is always read and written as one unit.
```
cpaulve-1A: even if today's UI only renders the list end to end
=> what do you mean?

cpaulve-1A: Also, do we really want to push for entity stores when we deal with an array of non complex types?

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,243 @@
+---
+name: logic-placement
+description: This skill should be used when the user asks "where should this logic go", "should this be a service or a store", "which store type should I use", "should I split this into container and presenter", "add an API call to a component", "share state between components", or when authoring or reviewing any component, service, or store in an Otter workspace. Guides where application logic belongs across component class, service, and store, and the choice between entity, simple, sync, and async store implementations.
```
cpaulve-1A: Should this skill be split? It seems like a one big skills to answer all design questions

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+- Container with a `.scss` file, or doing display formatting
+- A container that only forwards inputs - collapse it back into one component
+- Business logic in template bindings. Otter components use `ChangeDetectionStrategy.OnPush`; derive in `computed()` or in the container
+- `componentType` disagreeing with the class's actual role
+- Naming drift: file, class, selector, template, and spec must all carry the same `-cont` / `-pres` suffix
+
+**Stores and services**
+
+- A store created to pass data between two components. Use a service
+- A `*.facade.ts` wrapper around the store. Otter has no such pattern
```
cpaulve-1A: That is rather specific, why this pattern specifically?

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: I would say yes. BTW the migration agent I have created is reading the schematics files and applies the changes itself without running the schematic. So indeed it would be easier with a skill instead of a schematic :D 
so I vote for it :D

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: removed the yes/no and kept only where the use-case should be implemented

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: done

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: applied

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: applied

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: mistake, removed

tools/llm/plugins/otter/skills/logic-placement/references/component-architecture.md
mrednic-1A: updated

tools/llm/plugins/otter/skills/logic-placement/references/store-types.md
mrednic-1A: I removed the file, now we have all centralized in one skill file

tools/llm/plugins/otter/skills/logic-placement/references/component-architecture.md
mrednic-1A: added a section dedicated ### When to split

tools/llm/plugins/otter/skills/logic-placement/references/component-architecture.md
mrednic-1A: I added more context

tools/llm/plugins/otter/skills/logic-placement/references/component-architecture.md
mrednic-1A: I deleted the whole file. Signals/observables guidance belongs to rxjs-vs-signals and is now delegated there entirely.

tools/llm/plugins/otter/skills/logic-placement/references/component-architecture.md
mrednic-1A: moved and rephrased

tools/llm/plugins/otter/skills/logic-placement/references/component-architecture.md
mrednic-1A: I removed this part, it came from a previous question of mine to the ai.

tools/llm/plugins/otter/skills/logic-placement/references/component-architecture.md
mrednic-1A: yes, rephrased

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: yes, updated

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: updated to be something like 'sharing on its own is the weakest reason to introduce a store, yet it's the most commonly given one — state is needed in two places, so it goes to the store. A service covers that case at a fraction of the cost, so rule it out explicitly before continuing.'

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: restructured

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: applied

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: nope, this was an error, updated

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: I would say no. if we split we'll have several skills answering the same design question 'where the logic should stay'.

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: this comes from ngrx, and I removed it as it was not studied in otter. nice catch

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: https://github.com/AmadeusITGroup/otter/issues/4452

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,405 @@
...
+# Where application logic belongs
+
+Work through these decisions in order. Each one narrows the next.
+
+1. **Component shape** - one component, or a container/presenter split
+2. **Logic layer** - component class, injectable service, or NgRx store
+3. **Store implementation** - only if the logic layer decision said store: entity or simple, sync or async
+
+Do not jump straight to the store implementation decision. Most state that feels store-shaped is service-shaped, and most components that feel split-shaped are not.
+
```
cpaulve-1A, line 16:
```suggestion
## When to Apply This Skill

| Decision | Timing | Cost of Change |
|---|---|---|
| Component shape (single vs split) | **Before** `ng g component` | Easy (rerun schematic or refactor) |
| Logic layer (component/service/store) | **Before** scaffolding | Easy (move/extract before generate) |
| Store type (entity/simple, sync/async) | **Before** `ng g store*` | **Expensive** (reshapes state, breaks consumers) |

**Guideline**: Make component shape and logic layer decisions first. Once decided, scaffold *immediately* to lock in structure. Store type decisions should happen in design phase; changing them in PR review is late.
```

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,405 @@
...
+- API call inside a reducer
+- `.subscribe()` in a container followed by a dispatch, instead of an effect
+- Hand-rolled loading/error flags shadowing `AsyncStoreItem`
+- A simple store holding records that are updated one at a time. Those belong in an entity store (an array of primitives is fine in a simple store)
+- A sync store whose data actually comes from an API
+- Hand-written store files. Use `ng g store*` so naming, adapter, selectors, and sync serializer stay conventional
+
+## Stop Rules
+
+Halt and ask the user when:
```
cpaulve-1A: To make it more AI friendly, add clear instructions:
```
## Stop & Ask: Trade-Off Decisions

When you encounter these exact scenarios, STOP and ask the user.
Do NOT choose alone.

### Scenario 1: Local State Needs Pending/Failure

TRIGGER: 
  state is screen-local AND UI needs to track pending/failure status

ACTION:
  STOP. Ask user:
  
  "This state is local to one screen but needs pending/failure tracking.
  
  Option A: AsyncStore
    - Gives AsyncStoreItem (pending/failure tracking)
    - Con: Promotes to application scope
  
  Option B: Local signals + manual tracking
    - Keeps screen scope
    - Con: You hand-code isLoading/hasError
  
  Which fits your requirements?"
  
CONTINUE: Based on their answer, recommend:
  - "Option A" → recommend: entity-async or simple-async store
  - "Option B" → recommend: local signals with computed state

---

### Scenario 2: Sharing Between Components — Which Scope?

TRIGGER:
  State must be shared between components AND unclear if it survives navigation

ACTION:
  STOP. Ask user:
  
  "This state is shared, but unclear about scope:
  
  When this screen closes, should this state:
  
  A) Die with the screen (component-scoped service)
  B) Survive until logout (root service)  
  C) Survive reload (store with persistence)
  
  Which is it?"

CONTINUE: Based on their answer, recommend:
  - "A" → component-scoped service in container's providers
  - "B" → root service (providedIn: 'root')
  - "C" → NgRx store with store-sync

---

### Scenario 3: Store Type Migration

TRIGGER:
  Reviewing code where store type (entity↔simple or sync↔async) needs to change

ACTION:
  STOP. Do NOT recommend the change in this PR.
  
  Message to user:
  "This store needs to change from [current type] to [new type].
  
  This reshapes state structure and breaks all consumers.
  
  Action: Create a separate PR for the migration. 
  Confirm with team before starting."

CONTINUE: Do not recommend the type change in the current PR.

---

### Scenario 4: CMS Replacement Uncertain

TRIGGER:
  Component appears to need a split (container/presenter) based on complexity,
  but no clear data-source or UI-flavour variation, AND unclear if CMS replacement is needed

ACTION:
  STOP. Ask user:
  
  "You're suggesting a split, but I don't see:
  - Multiple UI designs
  - Multiple data sources
  - An obvious CMS replacement need
  
  Will integrators configure or replace this component via CMS?"

CONTINUE: Based on answer:
  - "Yes" → recommend split (required for CMS replacement)
  - "No" → recommend single component (don't split yet)

---

### Scenario 5: No Variation But Split Suggested

TRIGGER:
  A component is suggested for split but:
  - No multiple UI flavours
  - No multiple data sources
  - No CMS replacement need
  - No clear future variation

ACTION:
  STOP. Ask user:
  
  "I don't see what will vary between container and presenter here.
  
  Can you name the specific thing that varies (today, not theoretically)?"

CONTINUE: Based on answer:
  - User can name it → proceed with split recommendation
  - User cannot name it → recommend single component instead
```

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,405 @@
...
+
+Do not jump straight to the store implementation decision. Most state that feels store-shaped is service-shaped, and most components that feel split-shaped are not.
+
+## Component shape: one component or container/presenter
+
+The intent, from the [container/presenter documentation](https://github.com/AmadeusITGroup/otter/blob/main/docs/components/CONTAINER_PRESENTER.md):
+
+> From a UI perspective it is a good practice to separate access of data/business logic from pure presentation.
+
+And the division of labour, from the [fixtures documentation](https://github.com/AmadeusITGroup/otter/blob/main/docs/components/FIXTURES.md):
```
matthieu-crouzet: not sure to understand why we speak about fixture in this skill

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,405 @@
...
+
+The common thread: **name what varies before splitting.** If you cannot say which of data source, display, or integrator-supplied implementation will differ, there is nothing for the seam to absorb.
+
+"The logic feels complex enough to separate" is not a fourth criterion. Complexity alone is a human judgement call, so leave that split to a reviewer who asks for it rather than inferring it.
+
+### When a component does not need splitting
+
+If nothing varies, generate a single component. Two shapes come up often where nothing can:
+
+- **Small presentational elements** - a badge, an icon, a price label, a spinner. They take inputs, render, and emit. There is nothing for a container to own. Otter keeps purely presentational, app-wide elements of this kind under `elements/` rather than `components/`.
```
matthieu-crouzet: Do we still recommend to have an `elements` folder ?

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,405 @@
+---
+name: logic-placement
+description: This skill should be used when the user asks "where should this logic go", "should this be a service or a store", "which store type should I use", "should I split this into container and presenter", "add an API call to a component", "share state between components", or when authoring or reviewing any component, service, or store in an Otter workspace. Guides where application logic belongs across component class, service, and store, and the choice between entity, simple, sync, and async store implementations.
```
cpaulve-1A: Suggestion to avoid going through this skill if user ask a question on an irrelevant topic:


```suggestion
description: This skill should be used when the user asks "where should this logic go", "should this be a service or a store", "which store type should I use", "should I split this into container and presenter", "add an API call to a component", "share state between components", or when authoring or reviewing any component, service, or store in an Otter workspace. Guides where application logic belongs across component class, service, and store, and the choice between entity, simple, sync, and async store implementations.

**Out of scope**: Configuration, localization, rules engine, forms, analytics.
Otter's machinery (@o3r/configuration, @o3r/localization, etc.) handles those.
If that's your question, consult the appropriate Otter machinery skill instead.

```

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,405 @@
...
+- [ ] Split justified by flavours, reuse, or CMS replacement - not by habit
+- [ ] Presenter renderable from inputs alone: no store, no SDK client, no router
+- [ ] Container holds no styling and no display formatting
+- [ ] Store justified by cross-feature sharing, persistence, async tracking, or Otter machinery
+- [ ] Store type matches the data: collection vs singleton, remote-fed vs locally controlled
+- [ ] API calls sit in an effect or a service
+- [ ] Files generated with the Otter schematics, naming conventions intact
+- [ ] JSDoc on exported symbols and public members; narrowest possible property types
+- [ ] The project's lint and unit-test tasks pass
+
```
cpaulve-1A, line 127: You should provide guidance to the AI on how to output the data:
exemple:
```markdown
## Output Templates for AI Agent

When recommending, use this format:

### Template: Recommend Single Component

"Based on the code, this should be a single component.

RECOMMENDATION:
  ✓ Component: [component-name]
  ✓ Role: componentType: 'Component' (or 'Block' if in a feature area)
  ✓ Logic layer: [component class / service / store]
  ✓ Scope: [local / shared]

NEXT STEP:
  Run: ng g component [name] --componentStructure=simple

VERIFY:
  - No split suffixes (-cont, -pres) in filenames
  - No @O3rComponent decorator needed unless Page/Block/ExposedComponent
  - Logic is in component class, service, or store (not template)"

---

### Template: Recommend Container/Presenter

"Based on the code, this should be split.

RECOMMENDATION:
  ✓ Container: [container-name-cont]
  ✓ Presenter: [presenter-name-pres]
  ✓ Container role: [Page or Block]
  ✓ Presenter role: ExposedComponent
  ✓ Logic layer: [where data/state lives]

NEXT STEP:
  Run: ng g component [name] --componentStructure=full

VERIFY:
  - Container injects Store/service, dispatches, selects state
  - Presenter receives inputs only, emits events
  - Presenter can be tested from a spec"

---

### Template: Recommend Store

"Based on the code, this needs a store.

RECOMMENDATION:
  ✓ Store type: [entity|simple] + [sync|async]
  ✓ Reason: [meets criterion A/B/C/D]
  ✓ Container: Injects Store, dispatches [*FromApi] actions
  ✓ Effect: Calls [API/SDK], triggers success/failure

NEXT STEP:
  Run: ng g store-[entity|simple]-[sync|async]

VERIFY:
  - State shape matches data structure
  - Container dispatches, not components
  - API calls in effects, not reducers"
```

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: updated

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: implemented

tools/llm/plugins/otter/README.md
```diff
@@ -14,13 +14,33 @@ This plugin bundles three MCP servers:
...
+| **logic-store-type** | Decision tree: which NgRx store variant — entity or simple, sync or async |
+| **rxjs-vs-signals** | Choosing between Angular Signals and RxJS Observables, with bridging rules for `toSignal()` and `toObservable()` |
+| **otter-schematics** | Ensures Otter schematics are used over manual file creation when scaffolding Angular artifacts |
+| **otter-new-config** | Adding a configuration flag to a component: scope decisions, safe defaults, absence detection |
+| **otter-rules-engine** | Authoring and debugging facts, operators, actions, and rulesets in `@o3r/rules-engine` |
+| **playwright-e2e-testing** | Writing and reviewing Playwright E2E scenarios, actions, and checks |
+| **otter-mcp-setup** | Installing and configuring the `@o3r/mcp` MCP server when it is unavailable |
+
+The three `logic-*` decision trees are shared knowledge, read by both **logic-placement** and **logic-review** so that a design recommendation and a review verdict can never disagree. They are invoked by those two entry skills rather than by a question asked directly.
```
matthieu-crouzet:
```suggestion
The `logic-*` decision trees are shared knowledge, read by both **logic-placement** and **logic-review** so that a design recommendation and a review verdict can never disagree. They are invoked by those agent skills rather than by a question asked directly.
```

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: it is removed. there was a mistake

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: I removed it from the skill. 
now it has references only in the old doc

tools/llm/plugins/otter/skills/logic-component-shape/SKILL.md
```diff
@@ -0,0 +1,194 @@
...
+    const factor = this.sortDirection() === 'asc' ? 1 : -1;
+    return [...this.passengers()].toSorted((a, b) => factor * a.lastName.localeCompare(b.lastName));
+  });
+}
+```
+
+The presenter renders from a spec by supplying `passengers` alone — that is the test for a clean split. Two details:
+
+- The store selection stays an observable and is bridged with `toSignal` at the component boundary. `initialValue` is what makes it assignable to a required signal input; `passengers$ | async` would widen the type to `Passenger[] | null` and fail under `strictTemplates`. See `rxjs-vs-signals`.
+- Not every Otter decorator has a signal equivalent. `@Localization` on a `translations` property still requires the decorator-based `@Input()`, so check which form the decorator you need supports rather than generalizing to signals.
```
matthieu-crouzet: Shouldn't we open an issue to support this use case ?

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,153 @@
+---
+name: logic-placement
+description: This skill should be used when the user asks "where should this logic go", "should this be a service or a store", "which store type should I use", "should I split this into container and presenter", "add an API call to a component", "share state between components", or when designing a new component, service, or store in an Angular or Otter workspace. Routes each decision to its dedicated decision tree and returns a placement recommendation. Out of scope, each covered by its own skill — configuration flags, localization, rules engine, forms, analytics, and the choice between signals and observables. Use `logic-review` instead to assess code that already exists.
```
matthieu-crouzet: Description seems a bit too long, we need to think that frontmatter will be loaded in the context, so we need to keep them shorter as possible 


```suggestion
description: Decide where new logic belongs — component shape, logic layer, and store variant — when designing a component, service, or
  store in an Angular/Otter workspace. Use `logic-review` to assess existing code.
```

maybe we should review the others descriptions as well

tools/llm/plugins/otter/skills/logic-component-shape/SKILL.md
mrednic-1A: yes, https://github.com/AmadeusITGroup/otter/issues/4453

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: updated all descriptions

tools/llm/plugins/otter/agents/logic-design.agent.md
```diff
@@ -0,0 +1,25 @@
...
+2. Read the sub-skill for each decision before answering it — `logic-component-shape`, then `logic-state-layer`, then `logic-store-type`. The third runs only if the second returned "store".
+3. Inspect the workspace before recommending. Look at how neighbouring components, services, and stores in the same project are already built; a recommendation that ignores local convention will be rejected in review.
+4. Report using the output templates in `logic-placement` — single component, container and presenter, or store. Nothing looser.
+
+## Hard rules
+
+- **Default to the shallowest option.** Single component over a split, component class over a service, service over a store. Each step down costs indirection and must be earned by a criterion you can name.
+- **Never invent a justification.** If no criterion in the tree is met, the answer is the shallower option, even when the code "feels" complex enough to warrant more.
+- **Stop and ask rather than choosing** in any of the five scenarios in `logic-placement`. Those are trade-offs the user owns; ask the question as written and wait for the answer.
```
matthieu-crouzet:
```suggestion
- **Stop and ask rather than choosing** in any of the scenarios in `logic-placement`. Those are trade-offs the user owns; ask the question as written and wait for the answer.
```

tools/llm/plugins/otter/agents/logic-design.agent.md
```diff
@@ -0,0 +1,25 @@
...
+---
+
+You are an Angular and Otter architecture advisor. Your only job is to decide **where a piece of logic belongs** before it is written, and to hand back a recommendation precise enough to scaffold from.
+
+You do not write feature code. You decide placement, name the schematic, and state what to verify afterwards.
+
+## How to work
+
+1. **Load the `logic-placement` skill first, and follow it.** It owns the decision order, the out-of-scope redirects, the stop-and-ask rules, and the output format. Do not answer from memory.
+2. Read the sub-skill for each decision before answering it — `logic-component-shape`, then `logic-state-layer`, then `logic-store-type`. The third runs only if the second returned "store".
```
matthieu-crouzet:
```suggestion
2. Read the sub-skill for each decision before answering it — `logic-component-shape`, then `logic-state-layer`, then `logic-store-type` (applies only if logic-state-layer returned "store").
```

tools/llm/plugins/otter/agents/logic-design.agent.md
```diff
@@ -0,0 +1,25 @@
...
+4. Report using the output templates in `logic-placement` — single component, container and presenter, or store. Nothing looser.
+
+## Hard rules
+
+- **Default to the shallowest option.** Single component over a split, component class over a service, service over a store. Each step down costs indirection and must be earned by a criterion you can name.
+- **Never invent a justification.** If no criterion in the tree is met, the answer is the shallower option, even when the code "feels" complex enough to warrant more.
+- **Stop and ask rather than choosing** in any of the five scenarios in `logic-placement`. Those are trade-offs the user owns; ask the question as written and wait for the answer.
+- **Redirect out-of-scope questions** instead of answering them. Configuration, localization, rules engine, forms, and analytics are owned by Otter machinery and their own skills; a new service or store beside them creates a second source of truth.
+- **Do not scaffold unless asked.** Name the `ng g` command; run it only on request, and then via `otter-schematics`.
```
matthieu-crouzet: What if the MCP server is not up and running ?

tools/llm/plugins/otter/skills/logic-component-shape/SKILL.md
```diff
@@ -0,0 +1,195 @@
...
+             dense table on desktop and stacked cards on mobile; an A/B variant
+             without the seam: the fetch is duplicated per variant, or the template
+             branches on display concerns
+
+CRITERION B  one UI flavour, several data sources
+             a passenger list rendered from a search result on one screen and from a
+             saved itinerary on another
+             without the seam: the presentation is copied, and the copies drift
+
+CRITERION C  the presenter must be CMS-replaceable or configurable
```
matthieu-crouzet: isn't it the same as criterion A ?

tools/llm/plugins/otter/skills/logic-component-shape/SKILL.md
```diff
@@ -0,0 +1,195 @@
...
+```
+
+**Name what varies before splitting.** If you cannot say which of data source, display, or integrator-supplied implementation will differ, there is nothing for the seam to absorb.
+
+"The logic feels complex enough to separate" is **not** an additional criterion. Complexity alone is a human judgement call — see the **split without variation** stop-and-ask in `logic-placement`.
+
+## Branch A — container + presenter
+
+```
+RECOMMENDATION  <name>-cont + <name>-pres
```
matthieu-crouzet:
```suggestion
```

tools/llm/plugins/otter/skills/logic-component-shape/SKILL.md
```diff
@@ -0,0 +1,195 @@
...
+### Forms across the seam
+
+A form is the case where the seam is easiest to get wrong, because both halves hold part of it. Otter's form documentation calls them **parent component** and **input component**; those map onto container and presenter respectively.
+
+- **Custom validators** — business rules, or rules spanning several controls — are declared in the **container** and passed to the presenter through an input. The presenter applies them to the form but does not own them.
+- **Primitive validators** — simple, configurable, single-control — are declared in the **presenter**.
+- **Submit logic always sits in the container.** The submit button may render in the presenter or at page level; either way the presenter only notifies that submit was fired, and the container runs the business logic and emits the outcome.
+- The presenter implements `ControlValueAccessor` and `Validator`, so value, status, and errors propagate up to the container as they would from a native input.
+
+That is the same rule as everywhere else in this table: business meaning in the container, the mechanics of display and input in the presenter.
```
matthieu-crouzet: Maybe we should not consider form in this skill, as form control recommendation from angular changed, we may rework that before

tools/llm/plugins/otter/skills/logic-component-shape/SKILL.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Branch B — single component
+
+```
+RECOMMENDATION  single component <name>
```
matthieu-crouzet:
```suggestion
```

tools/llm/plugins/otter/skills/logic-component-shape/SKILL.md
matthieu-crouzet: or maybe we should ask the user if it's planned to support another UI flavor at some point

tools/llm/plugins/otter/skills/logic-component-shape/SKILL.md
```diff
@@ -0,0 +1,195 @@
...
+SCHEMATIC       ng g component <name> --componentStructure=simple
+VERIFY          no -cont or -pres suffix anywhere in file, class, selector, or template
+                no container that would only forward its inputs
+```
+
+Two shapes where nothing can vary:
```
matthieu-crouzet:
```suggestion
Shapes where nothing can vary:
```

tools/llm/plugins/otter/skills/logic-component-shape/SKILL.md
```diff
@@ -0,0 +1,195 @@
...
+| Decision | `componentStructure` |
+|---|---|
+| Single component | `simple` |
+| Container + presenter pair | `full` |
+| Container only, presenter already exists | `container` |
+| Presenter only, added to an existing container | `presenter` |
+
+Use `otter-schematics` to discover the schematic and its current options. It also produces the naming and folder layout, so neither needs deriving by hand — see the [container/presenter documentation](https://github.com/AmadeusITGroup/otter/blob/main/docs/components/CONTAINER_PRESENTER.md) for the `*-cont` / `*-pres` conventions.
+
+## Which componentType to declare
```
matthieu-crouzet: Is it linked to this skill ?
there we are not speaking about logic placement no ?

tools/llm/plugins/otter/skills/logic-component-shape/SKILL.md
```diff
@@ -0,0 +1,195 @@
...
+
+  /** Display ordering derived from inputs plus local state */
+  public readonly sortedPassengers = computed(() => {
+    const factor = this.sortDirection() === 'asc' ? 1 : -1;
+    return [...this.passengers()].toSorted((a, b) => factor * a.lastName.localeCompare(b.lastName));
+  });
+}
+```
+
+The presenter renders from a spec by supplying `passengers` alone — that is the test for a clean split. Two details:
```
matthieu-crouzet:
```suggestion
The presenter renders from a spec by supplying `passengers` alone — that is the test for a clean split.
Additionnal details:
```

tools/llm/plugins/otter/skills/logic-component-shape/SKILL.md
```diff
@@ -0,0 +1,195 @@
...
+
+- The store selection stays an observable and is bridged with `toSignal` at the component boundary. `initialValue` is what makes it assignable to a required signal input; `passengers$ | async` would widen the type to `Passenger[] | null` and fail under `strictTemplates`. See `rxjs-vs-signals`.
```
matthieu-crouzet: Shouldn't this be directly in `rxjs-vs-signals` instead of here ?

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
```diff
@@ -0,0 +1,154 @@
...
+  store in an Angular/Otter workspace. Use `logic-review` to assess existing code.
+---
+
+# Logic placement — design mode
+
+Decide where a piece of logic belongs **before** it is written. This skill holds the order of the decisions, the rules for halting, and the output format. The criteria for each decision live in a sub-skill: read it before answering, do not answer from memory.
+
+To assess code that already exists, use `logic-review`.
+
+## Out of scope
```
matthieu-crouzet, line 15: Is it really useful ?
because the skill should not be triggered in these cases as not related to the frontmatter description no ?

tools/llm/plugins/otter/skills/logic-review/SKILL.md
```diff
@@ -0,0 +1,114 @@
...
+
+Read in this order, and read the tree before judging its dimension — do not judge from memory.
+
+| # | Look at | Read | Judge |
+|---|---|---|---|
+| 1 | Component classes, decorators, templates | `logic-component-shape` | is the split, or its absence, justified; does `componentType` match the role |
+| 2 | What each class injects and holds | `logic-state-layer` | is the layer the shallowest one that works; are API calls in the right place |
+| 3 | Store files, if any | `logic-store-type` | does the variant match the data |
+
+Two rules for the whole pass:
```
matthieu-crouzet:
```suggestion
Rules for the whole pass:
```

tools/llm/plugins/otter/skills/logic-review/SKILL.md
```diff
@@ -0,0 +1,114 @@
...
+
+- **Judge a split by what varies, not by whether it exists.** A missing split is a finding only when one of the three variation criteria is met. An existing split whose container only forwards inputs is a finding regardless.
```
matthieu-crouzet:
```suggestion
- **Judge a split by what varies, not by whether it exists.** A missing split is a finding only when one of the variation criteria is met. An existing split whose container only forwards inputs is a finding regardless.
```

tools/llm/plugins/otter/skills/logic-review/SKILL.md
```diff
@@ -0,0 +1,114 @@
...
+
+```
+FOR each service a presenter injects:
+  does it select from the store, call an API, or trigger a side effect
+  beyond the component?
+    yes -> FINDING: the dependency belongs in the container
+    no  -> allowed, if it is a presentation helper such as formatting or breakpoints
+```
+
+## Check — componentType against the actual role
```
matthieu-crouzet, line 44: same comment for me it's not linked

tools/llm/plugins/otter/skills/otter-schematics/SKILL.md
```diff
@@ -18,8 +18,21 @@ If a schematic exists for the artifact type, use it. Manual file creation is onl
 
 Use the `@o3r/mcp` MCP server to discover available schematics for the project and their options. If the MCP server is not available, invoke the `otter-mcp-setup` skill to guide installation, then re-run this skill.
 
+Preview before committing to a run: `ng g <schematic> --dry-run` lists the files that would be created or modified, and `--help` lists the current options.
+
+## When a Schematic Cannot Be Run
```
matthieu-crouzet:
```suggestion
## When a schematic cannot be run
```

tools/llm/plugins/otter/agents/logic-design.agent.md
cpaulve-1A: Is that true? Don't we recommend to use ng g as much as possible?

tools/llm/plugins/otter/skills/logic-component-shape/SKILL.md
mrednic-1A: I merged both criteria in one, as the logic placement is the same

tools/llm/plugins/otter/skills/logic-component-shape/SKILL.md
mrednic-1A: yes, removed

tools/llm/plugins/otter/skills/logic-component-shape/SKILL.md
mrednic-1A: yes, this is removed/rephrased

tools/llm/plugins/otter/skills/logic-component-shape/SKILL.md
mrednic-1A: yes, rephrased and linked directly to the other skill

tools/llm/plugins/otter/skills/logic-placement/SKILL.md
mrednic-1A: It's not redundant with the frontmatter gating, because the two do different jobs:

The frontmatter description decides whether the skill is entered, 
the Out of scope table does runtime routing once inside — mid-decision.

tools/llm/plugins/otter/skills/logic-review/SKILL.md
mrednic-1A: here in the review we are checking if the component type matches the logic placement.

tools/llm/plugins/otter/agents/logic-design.agent.md
mrednic-1A: yes we do, I updated the rephrasing

cpaulve-1A (approved):

matthieu-crouzet (approved):

PR #4496, opened 2026-09-15
feat: add renovate preset for generator deps

mrednic-1A:
## Related issues

*- No issue associated -*


migration-guides/15.2.md
```diff
@@ -0,0 +1,56 @@
...
+  "packageRules": [
+    {
+      "matchDepTypes": ["generatorDependencies"],
+      "groupName": "Otter generator dependencies"
+    }
+  ]
+}
+```
+
+### 2. You restrict `enabledManagers`
```
sdo-1A, line 38: I'm a bit confused, if we recommend our users to add `custom.regex` to their renovate config shouldn't we add the same to our own config since we also use `enabledManagers`?

tools/renovate/base.json
```diff
@@ -14,7 +14,8 @@
     "github>AmadeusITGroup/otter//tools/renovate/group/angular",
     "github>AmadeusITGroup/otter//tools/renovate/group/typescript",
     "github>AmadeusITGroup/otter//tools/renovate/managers/github-node-version.json5",
-    "github>AmadeusITGroup/otter//tools/renovate/managers/mcp-servers.json5"
+    "github>AmadeusITGroup/otter//tools/renovate/managers/mcp-servers.json5",
+    "github>AmadeusITGroup/otter//tools/renovate/managers/otter-generator-dependencies.json5"
```
kpanot, line 18: Do you have `generatorDependencies` in package.json outside from this repository?
If so, not sure we want to keep it as recommendation (regarding the different inconveniences we have with it).
If not, we may not want this manager to be exposed in `base` which is used by other repositories.

migration-guides/15.2.agent.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Task 1 — Deduplicate the `generatorDependencies` custom manager
+
+The Otter `base` preset now ships a custom regex manager that updates the `generatorDependencies` block of `package.json`.
+It is defined in [`otter-generator-dependencies.json5`](https://github.com/AmadeusITGroup/otter/blob/main/tools/renovate/managers/otter-generator-dependencies.json5).
+A repository-local manager doing the same job now produces duplicate extractions, and therefore duplicate branches and pull requests per bump.
+
+### Detect
+
+Renovate configuration can live in any of these locations, so check all of them:
```
kpanot, line 17: Is it not enough to say that it is Renovate configuration for the LLM to look for it at all the places it can be?

tools/renovate/base.json
mrednic-1A: Agree with the fact that we want to keep it outside of recommendation. 
I can discard this PR. We will use it as base for discussions related to the new migration way.

mrednic-1A:
After discussions we do not want to recommend the generated dependencies. So I close this PR.

migration-guides/15.2.md
mrednic-1A: we already have it, but with an older name and we did not migrate it.
https://github.com/AmadeusITGroup/otter/blob/7b4863ba635bb6278be5000ea1d460b93be5ac16/.renovaterc.json5#L19

